PR #38, opened 2016-02-02
edof change

### LINKED ISSUE #37: Petition to remove the first row in EDOF
This is a petition to remove the unnecessary first row of the _edof_ matrix. It is never used and makes life harder than it needs to be.

Advantages:
- Everywhere the _edof_ matrix is used, the first row is removed
- Simplifies coding since, you can write `edof(:,i)` instead of `edof(2:end,i)`
- The code looks nicer, i.e `edof(:,i)` looks much better than `edof(2:end,i)`

Disadvantages:
- None (except some rewriting of code

Please vote [here!](http://strawpoll.me/6706409)

> Nature is pleased with simplicity. And nature is no dummy
― Isaac New

KristofferC:
Refactor code to remove the element index row.

Fixes #37 

Also enable precompilation for faster load times.


KristofferC:
cc @fredrikekre

fredrikekre:
Excellent!

PR #39, opened 2016-02-03
added fe_values

KristofferC:
There is currently no good user friendly way to use JuAFEM to write your own elements from the available quadrature rules and shape functions. This is a step towards that.

This PR introduces an object called FEValues (stolen from https://dealii.org/developer/doxygen/deal.II/classFEValues.html)

It is initialized by calling `FEValues(function_space, quadrature_rule)`.

A function space is named `FunctionSpaceName{Order, ReferenceShape}`.

The currently available function spaces are:

* `Lagrange{1, JuaFEM.Square}`
* `Lagrange{1, JuaFEM.Triangle}`
* `Lagrange{1, JuaFEM.Triangle}`
* `Lagrange{2, JuaFEM.Triangle}`
* `Lagrange{1, JuaFEM.Cube}`
* `Serendipity{2, JuaFEM.Square}`

A function space is created by, for example: `func_space = Lagrange{1, JuaFEM.Square}()`.

The only quadrature rule are gauss quadrature on triangles, squares and cubes.
They are created by for example `quad_rule = get_gaussrule(JuAFEM.Square(), 2)` for an order 2 quadrature on the unit square.

The `FEValues` object are then created by `fe_values = FEValues(func_space, quad_rule)`.
Upon creation, `FEValues` caches the values of the shape functions and derivatives in the gauss points.
The points of `FEValues` is that you can call `reinit!(fev, x)` where `x` is the coordinate matrix for an element. This will update the global shape function derivatives, jacobians, weights etc.
The different accessor functions are seen in https://github.com/KristofferC/JuAFEM.jl/blob/8d92261b6ad64296ead84e6d30e950321eede3f2/src/utilities/fe_values.jl#L44

You then repeatedly call `reinit!(fev, x)` for all the elements in your mesh that shares the function space + quadrature rule and the values needed are updated in an efficient manner.

For a complete use case of `FEValues` see the test at: https://github.com/KristofferC/JuAFEM.jl/blob/8d92261b6ad64296ead84e6d30e950321eede3f2/test/test_fevalues.jl which assembles linear elasticity on a first order square using `FEValues`.


KristofferC:
cc @fredrikekre

PR #45, opened 2016-02-18
WIP: started on function interpolations

KristofferC:
This is a work in progress to simplify taking operators on vector / scalar fields defined on the nodes in a finite element.
Currently, the functions that return second order tensors return them in matrix form but I should add so they can be returned in Voigt form as well.

The way these functions work is basically how you write them in math with a sum over all base functions etc etc.
The nice thing is that they tend to be significantly quicker than the "FEM Basics" way where you create a B-matrix or N-matrix and matrix multiply.
For a 8 node quadraterial the `function_vector_symmetric_gradient` is about 3x faster than creating the B-matrix and doing the B * u matrix multiplication

TODO:

- [ ] Look a bit on the performance by changing the loop order
- [ ] Test the functions that return a scalar
- [ ] Maybe create an example that uses this.

This PR also changes the function name `shape_derivative` to `shape_gradient` but the old name is kept for a while.

### DIFF
--- a/doc/api/fe_values.rst
+++ b/doc/api/fe_values.rst
@@ -1,0 +1,121 @@
+.. currentmodule:: JuAFEM
+
+.. _api-fe_values:
+
+FE Values
+---------
+
+To assist with creating new finite elements ``JuAFEM`` has the concept of an object called
+``FEValues``. The idea for this was taken from deal.ii_.
+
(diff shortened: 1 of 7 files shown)


KristofferC:
Så, nu finns det en dokumentationsbit om FE values iaf.

PR #48, opened 2016-03-01
parameterize quadrature rule on dim

KristofferC:
This changes to that quadrature rule and FEValues are parametrized on `dim`.

@fredrikekre Du kan behöva ändra https://github.com/fredrikekre/pgd/blob/067cbe047f9c45dbe2fedfe740ec13e300c94c13/src/PGDmodule.jl#L11 till `JuAFEM.FEValues{dim_comp, T functionspace_comp}` om du uppdaterar efter denna merge


fredrikekre:
:+1:

PR #49, opened 2016-03-01
make function space parameterized on dim

KristofferC:
This continues the work to basically write the same code and only change a dim parameter to run it in different dimensions.

- FunctionSpace is now parameterized on dim.
- Line, Cube are now replaced with simply Square.
- Function spaces needs to be given dimension in order to be created., ie `FunctionSpace{dim, shape, order}`.
- Quadrature rules needs to be given dimension in order to be created, ie `make_quadule(dim, shape, order)`.

See modified tests for how to update code.

To fix:

- [x] Make order of "order" and "dim" consistent.
- [x] Fix comments in function space

### DIFF
--- a/src/JuAFEM.jl
+++ b/src/JuAFEM.jl
@@ -35,7 +35,7 @@ export FEValues, reinit!, shape_value, shape_gradient, detJdV, get_quadrule, get
                  function_vector_gradient!, function_vector_divergence, function_vector_symmetric_gradient!
 export n_dim, n_basefunctions
 export Lagrange, Serendipity
-export get_gaussrule
+export get_gaussrule, Dim
 
 include("materials/hooke.jl")
 include("utilities/utilities.jl")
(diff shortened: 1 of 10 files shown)


KristofferC:
cc @fredrikekre

KristofferC:
This is work for the future to throw out all the Vector and Matrices and replace them with real tensor like objects.

KristofferC:
Updates. It is now always `dim, shape, order` for both function space and quadrules.

KristofferC:
Added a commit that moves the error checking in functin spaces to the abstract type so it can be written once and for all. https://github.com/KristofferC/JuAFEM.jl/commit/ebe63f562762a2bc195c331b49ea72c46e09b7e6

KristofferC:
Behövde ändra hur man skapar kvadratureregler nu för att behålla "type stability".

En kvadraturregel skapas nu som: `get_gaussrule(Dim{1}, Square(), 2)`

för en dimension 1, linje order 2.

Notera att "1:an" i `Dim` ska vara en känd för kompilatorn för att få effektiv kod. Dvs den ska antingen vara hårdkodad som ett tal eller se ur som:

```jl
function foo{dim}(...)
    get_gaussrule(Dim{dim}, Square(), 2)
end
```

men det ska inte se ut såhär:

```jl
function foo(dim::Int)
    get_gaussrule(Dim{dim}, Square(), 2)
end
```

KristofferC:
Nu är det fika..

KristofferC:
Mergar in det nu, får du några problem så kom och fråga bara.

PR #50, opened 2016-03-03
Faster version of vtk_grid

fredrikekre:
Added a faster version of `vtk_grid` which can be used if the topology is known (which is quite common)

Test with a mesh with 100x100 elements:
```jl
julia> @time vtk_grid(edof, Coord, Dof, 4, "vtk_outfile")
  1.652264 seconds (334.70 k allocations: 12.642 MB)

julia> @time vtk_grid(top, Coord, "vtk_outfile")
  0.034052 seconds (79.24 k allocations: 4.011 MB)
```

200x200 elements:
```jl
julia> @time vtk_grid(edof, Coord, Dof, 4, "vtk_outfile")
 25.855713 seconds (1.35 M allocations: 50.621 MB, 0.05% gc time)

julia> @time vtk_grid(top, Coord, "vtk_outfile")
  0.153537 seconds (319.34 k allocations: 15.901 MB)
```

1000x1000 elements (this definitely crashes with the original version):
```jl
julia> @time vtk_grid(top, Coord, "vtk_outfile")
  4.168526 seconds (8.00 M allocations: 376.221 MB, 11.23% gc time)
```

### DIFF
--- a/src/utilities/VTK.jl
+++ b/src/utilities/VTK.jl
@@ -35,4 +35,38 @@ function vtk_grid(Edof, Coord, Dof, nen, filename::AbstractString)
     return vtk
 end
 
+"""
+    vtk_grid(top, Coord, filename::AbstractString) -> vtkgrid
+Creates an unstructured VTK grid. `nen` is the number of nodes per element
 
+Faster version, can be used if the topology is known, i.e. the nodes for each element
+"""
+function vtk_grid(top, Coord, filename::AbstractString)
+
+    nele = size(top, 2)
+    nen = size(top,1)
+    nnodes = size(Coord, 2)
+    ndim = size(Coord, 1)
+
+    # TODO: This is a bad way of figuring out the eltype
+    if nen == 3 && ndim == 2
+        cell =  VTKCellType.VTK_TRIANGLE
+    elseif nen == 4 && ndim == 2
+        cell = VTKCellType.VTK_QUAD
+    elseif nen == 4 && ndim == 2
+        cell = VTKCellType.VTK_HEXAHEDRON
+    end
+
+    points = Coord
+
+    if ndim == 2
+        points = [points; zeros(nnodes)']
+    elseif ndim == 1
+        points = [points; zeros(nnodes)'; zeros(nnodes)']
+    end
+
+    cells = MeshCell[MeshCell(cell, top[:,i]) for i = 1:nele]
+
+    vtk = vtk_grid(filename, points, cells)
+    return vtk
+end
--- a/src/utilities/VTK.jl
+++ b/src/utilities/VTK.jl
@@ -36,15 +36,15 @@ function vtk_grid(Edof, Coord, Dof, nen, filename::AbstractString)
 end
 
 """
-    vtk_grid(top, Coord, filename::AbstractString) -> vtkgrid
+    vtk_grid(topology, Coord, filename::AbstractString) -> vtkgrid
 Creates an unstructured VTK grid. `nen` is the number of nodes per element
 
 Faster version, can be used if the topology is known, i.e. the nodes for each element
 """
-function vtk_grid(top, Coord, filename::AbstractString)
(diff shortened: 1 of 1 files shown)


fredrikekre:
#racetopullrequest50

src/utilities/VTK.jl
```diff
@@ -35,4 +35,38 @@ function vtk_grid(Edof, Coord, Dof, nen, filename::AbstractString)
     return vtk
 end
 
+"""
+    vtk_grid(top, Coord, filename::AbstractString) -> vtkgrid
+Creates an unstructured VTK grid. `nen` is the number of nodes per element
+
+Faster version, can be used if the topology is known, i.e. the nodes for each element
+"""
+function vtk_grid(top, Coord, filename::AbstractString)
```
KristofferC: Kan du byta namn på den här till `topology` samt säga att det ska vara en `::Matrix{Int}`.

fredrikekre:
Was thinking about adding meshing for 1D and 3D aswell, and maybe add a type called JuAMesh or something which can be parametrized on the dimension, following the convention for the function spaces, i.e. a "square" mesh in 1, 2 and 3 dimensions.

KristofferC:
I also thought about something like that. It is a bit annoying to deal with all `Edof, Ex, Ey, Coord, Dof` etc so packing them in a type is definitely something that is needed.

It would also be cool to use `ContMechTensors` (when it is done)` to store for example `Coord` as a Vector of first order tensors.

fredrikekre:
Yep I can put together a proposed mesh type with meshfunctions, and then enhance it with the use of `ContMechTensors` later on.

KristofferC:
That would be great

fredrikekre:
Not sure, either it could be just a geometrical object with elements, nodes and coordinates, but then there should be a function that set up a desired number of dofs in each node and generate edof.

Or it could be like now, that you specify the number of dofs when creating the mesh and get back everything.

But technically the dofs doesnt belong to the mesh, idk...

KristofferC:
My "ideal" situation would be: The mesh is a geometrical concept. It represents nodes, the coordinates of the nodes and how these are connected to elements. Basically coordinates + topology.

Then there is a function that takes a mesh and a collection of field dimensions, like `(3, 1, 1)` which would represent a vector field and two scalar field and from that constructs the Edof.

For example if the field representation would be (3,1) which represent 3 displacements and the pressure a coord would have `x, y, z, p` as dofs and an edof for a triangle would look like:

`x, y, z, x, y, z, x, y, z, p, p ,p` so that each field is close together.

fredrikekre:
Yes, that makes sense.

Something like:

```jl
mymesh = JuAMesh(startcoordinates,endcoordinates,dim,nElements,...)

edof = create_edof(mymesh,(3,1,1))
```

maybe?

KristofferC:
Yes something like that. Although I don't think we should return a "naked" edof, it should probably be wrapped in some type as well so you can do `get_dofs(dof_object, i)` to get the dofs of the i:th element.

fredrikekre:
Yes, and a similar function to get the dofs for a specified field, which can be used for adding the data to the vtk-file. Like:

```jl
displacements = u[get_dofs(dof_object,1)]
pressure = u[get_dofs(dof_object,2)]
vtk_point_data(vtkfile, displacements, "displacement")
vtk_point_data(vtkfile, pressure, "pressure")
```
if `u` is the solution vector, field 1 is displacement and 2 the pressure. But then it is needed to somehow define if its for an element or a field the dofs should be extracted for, else the functions look the same.

KristofferC:
The function could also look like `get_field(dof_object, 1, u)` which would give the displacements.

fredrikekre:
Yea, thats much better!

KristofferC:
Good reading to avoid unnecessary commits, like "fix comment": http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

fredrikekre:
I am a git noob! :)

PR #51, opened 2016-03-05
Use ContMechTensors whith JuAFEM

KristofferC:
This is a WIP to using Tensors wherever something in JuAFEM actually is a tensor instead of just using a Julia Vector. 

I added a full example on how this would look for a large deformation problem.

Things to do before merging:

- [x] Finish https://github.com/KristofferC/ContMechTensors.jl/pull/1.
- [x] Make the normal CALFEM elements work with this
- [x] Do some benchmarking to see that nothing got much slower
- [x] Fix symmetric gradient
- [x] Documentation
- [x] Benchmark on julia 0.4 (as expected, it is garbage, 20x slower)

Also, tuples on julia 0.4 seems quite slow so it is possible we wait with merging this or just decide that it is worth starting using julia 0.5.


KristofferC:
Ok, juafem now works with the stack allocated version of ContMechTensors. Still some stuff missing in that package though, like inverses of 4th order tensors.

KristofferC:
Benchmarked a bit and 100 000 quad elements take about 0.5 seconds to assemble which seems to be a bit faster than before.

fredrikekre:
Looks good!

JimBrouzoulis:
So we obtain higher preformance while still gaining benefits in terms of tensor abstraction (instead of having to resort to Voigt vectors and matrices)?

KristofferC:
That's the plan ;)

KristofferC:
This is unfortunately garbage slow on 0.4 because https://github.com/JuliaLang/julia/issues/15277 and https://github.com/JuliaLang/julia/issues/15274 is not fixed on 0.4..

KristofferC:
I think https://github.com/KristofferC/JuAFEM.jl/blob/7c9661bc9cbb884182d1f63c3fc03023ea2bee4f/examples/Yeoh_large_deformations.ipynb is a nice example of the new stuff. The internal element force function feels quite clean

KristofferC:
So this is pretty much done. I want to add some more examples and fix up some stuff at https://github.com/KristofferC/ContMechTensors.jl but nothing big.

However, since the performance is so bad on julia 0.4 I won't merge this. Anyone who wants to use it can just `Pkg.checkout("JuAFEM", "kc/dim_param")` or just use git to check out the branch.

fredrikekre:
Sweet!

I started a bit on some mesh and dof functions. Should I use 1st order tensors for the coordinates then? For the `reinit!` function I saw that you use

```jl
function reinit!{dim, T}(fe_v::FEValues{dim}, x::Vector{Vec{dim, T}})
    ...
end
```

so I guess `x` is a vector of first order tensors, with one tensor per node? So that

```jl
x[i] = [x_i, y_i, z_i]
```

is the coordinates of node `i` ? Just making sure I have understand it correctly!

KristofferC:
Go for it :). Yes like you say, one first order tensor per node. I need to add more documentation for the tensor stuff. The simplest way to create a first order tensor should be for example in 2D `Tensor{1, 2}((1.0,2.0))`. 

If you get something going we can iron out the details together.

KristofferC:
Added an example for tangent stiffness with tensors: https://github.com/KristofferC/JuAFEM.jl/blob/kc/dim_param/examples/stiffness_example.ipynb

KristofferC:
I think we should merge this. It is only us using this at the moment and it is awkward to have it in a PR.

KristofferC:
Julia 0.5 shouldnt be that far away as well. Maybe two months..?

fredrikekre:
Okay, maybe that's another issue then, that only we are using it? ;) We need to spread the word!

KristofferC:
Hehe, need better documentation and examples for it to be usable by other people though.

KristofferC:
Ok, lets do this.

KristofferC:
You didnt see my bug in https://github.com/KristofferC/JuAFEM.jl/blob/master/examples/Yeoh_large_deformations.ipynb?

This is wrong:
```jl
 F = function_vector_gradient(fe_values, q_point, x_vec)
```

This is right:
```jl
H = function_vector_gradient(fe_values, q_point, x_vec)
F = H + one(H)
```

I hope..

kimauth:
> Okay, maybe that's another issue then, that only we are using it? ;) We need to spread the word!

Almost a decade later people take flights and book hotel rooms to come and hear about this! 🎉 

DRJ01344

PR #52, opened 2016-03-07
WIP: Meshing functionality in JuAFEM

fredrikekre:
Added some types with functions for some mesh stuff in JuAFEM.

Cannot manage to create a vector of ContMechTensors though. For instance [here](https://gist.github.com/fredrikekre/04dba1aaf6295c29ddd7), where the node coordinates are gathered in a matrix. Is it easier to gather them in a vector of vectors and then convert to tensors?


KristofferC:
One way to create the tensors in the example mesh: `coords_tens = reinterpret(Vec{2, Float64}, coords, (div(length(coords),2),))`

src/utilities/fe_dofs.jl
```diff
@@ -0,0 +1,66 @@
+type FEField
+    name::AbstractString # Descriptive string
```
KristofferC: Try not to have abstract fields in a type, change to `UTF8String`.

src/utilities/fe_dofs.jl
```diff
@@ -0,0 +1,66 @@
...
+    dim::Int # Number of dimensions for the field
+end
+
+type FEDofs
```
KristofferC: Maybe the FEDofs should have a `FEMesh` field. I am thinking that `FEDofs` is pretty useless without it's corresponding mesh so if you want to use `FEDofs` you will also always have to carry around `FEMesh`? If that is the case then maybe just putting the mesh inside the dof object makes sense.

fredrikekre: Okay yea, then you could only use the FEDofs object then, and only use the mesh object to set up the FEDofs?

KristofferC: Yep!

fredrikekre:
Okay, so I updated the [test code](https://gist.github.com/fredrikekre/04dba1aaf6295c29ddd7) but get the following error, is there something wrong with my typedefinition?

```jl
ERROR: LoadError: MethodError: `convert` has no method matching convert(::Type{FEMesh{dim,T<:Real}}, ::Array{ContMechTensors.Tensor{1,2,Float64,2},1}, ::Array{Int64,2}, ::Array{Array{T,1},1})
This may have arisen from a call to the constructor FEMesh{dim,T<:Real}(...),
since type constructors fall back to convert methods.
```

KristofferC:
By saying `Vector[[1,2,3],[3,6,9],[9,8,7],[7,4,1]]` the `Vector` part explicitly says that this should be a container that can hold any type of `Vector`s. Since you want it to be a container that holds only `Vector{Int}` you should say that like:  `Vector{Int}[[1,2,3],[3,6,9],[9,8,7],[7,4,1]]`

fredrikekre:
Okay, cool!

Will restructure the types so that only the FEDofs object is used, and add the actual mesh generating functions and commit later.

KristofferC:
In the closest candidate list from the error message:

image

you can see that the first two fields have white text which means that they were ok but the last one is red which means that you didn't match the type you specified.

fredrikekre:
:+1:

fredrikekre:
Restructured the types and rewrote some functions, still missing the mesh generating functions. I have them from previous projects, just need to rewrite them to fit this format.

See the [example](https://gist.github.com/fredrikekre/04dba1aaf6295c29ddd7) for how it works right now for a small example mesh.

KristofferC:
The performance hit is irrelevant since it is only done once per element and the element function does a lot more work than the cost of looking up a few values in a vector.

KristofferC:
Good random reading about FE software: 

How deal.ii builds up their classes: https://www.dealii.org/8.3.0/doxygen/deal.II/index.html

How deal.ii handles vector valued problems: https://www.dealii.org/developer/doxygen/deal.II/group__vector__valued.html

deal.ii is has in my view a pretty much optimal design for FE software for researchers which is why I like taking their things :P

KristofferC:
What they do differently from us is that they always loop over degrees of freedoms when they want to assemble for example internal forces. Since we only have one base function per node, what we are saying is that x,y,z have identical base functions.

src/utilities/fe_mesh.jl
```diff
@@ -0,0 +1,162 @@
...
+end
+
+type FEMesh{dim, T <: Real}
+    coords::Vector{Vec{dim,T}} # Node coordinates
+    topology::Matrix{Int} # Element nodes
+    boundary::Vector{Vector{}} # Boundary nodes
+end
+
+
+function FEMesh(dim)
```
fredrikekre: I kinda want to use a function space as an input variable to the meshing, since that contains information about shape, dim and order, which then can be used to generate the mesh.

KristofferC: In my view is a function space a completely unrelated to a mesh. You can have mesh for video game models for example. I think it is fine to just give the dimension and shape etc as separate input arguments.

fredrikekre: Okay, what about the order? Or should the mesh not contain the nodes on the middle of the boundary? Only the 4 corners for a 2D Square then?

fredrikekre: So when we add a field, we define that the field should be quadratic somehow and then calculated middle nodes?

src/utilities/fe_mesh.jl
```diff
@@ -0,0 +1,162 @@
+# Typ definitions
+type FEBoundary # Tänkte typ parametrisera detta på nåt sätt så att man har FEboundary i 2D, 1D, och 0D
```
fredrikekre: Kan man göra det typ:
```jl
type FEBoundary{T}
    topology::Array{Int}
end
```

and then for a 2D square mesh generate 4 `FEBoundary{1}` for the 1D line boundaries and 4 `FEBoundary{0}` for the 0D corners?

src/utilities/fe_mesh.jl
KristofferC: In my opinion yes. When we ask for the coordinates for an element we just compute the coordinates needed on the fly.

src/utilities/fe_mesh.jl
KristofferC: Yes it should funka.

```jl
julia> type FEBoundary{T}
           topology::Array{Int}
       end

julia> FEBoundary{1}([1.0,2.0])
FEBoundary{1}([1,2])
```

src/utilities/fe_mesh.jl
fredrikekre: But these new calculated nodes should be stored somehow at least, so we don't have to calculate the coordinates all the time. (in each iteration or whatever). I guess that if you add a quadratic field they will be calculated and stored.

src/utilities/fe_mesh.jl
fredrikekre: So then I would want to gather all these boundaries into the `FEMesh` field `boundary` as a vector of vectors, so that the first vector contains all the 1D boundaries and the second all the 0D boundaries.

Example:

```jl
fe_m = get2Dmesh(...)
fe_m.boundary[1][1] # This gives the first 1D boundary
fe_m.boundary[1][2] # This gives the second 1D boundary
fe_m.boundary[0][1] # This gives the first 0D boundary
```

But how do I define that in the `FEMesh` type?

```jl
type FEMesh
    ...
    boundary::Vector{Vector{FEBoundary}}
end
```

But with the definition of `FEBoundary` with

```jl
type FEBoundary{T}
    topology::Array{Int}
end
```
That means that just `FEBoundary` is abstract, right? So then I don't want it like above in the type definition of `FEMesh`? :confused:

src/utilities/fe_mesh.jl
KristofferC: I don't really care of we calculate them again. It is a few simple operations. Comparing running reinit on an element it should be completely negligible. The cost of recompiling them could even be smaller than fetching them from memory.

src/utilities/fe_dofs.jl
```diff
@@ -0,0 +1,116 @@
+type FEField
```
KristofferC: Maybe a FEField should have a function space. In that way you could make for example a pressure field with linear base functions and a displacement field with quadratic basefunctions.

src/utilities/fe_mesh.jl
```diff
@@ -0,0 +1,186 @@
...
+    # return FEMesh(coords,topology,boundary)
+
+end
+
+
+#############
+# Accessors #
+#############
+
+# Should these even exist? Atleast I should add the same accessors for the FEDofs aswell so that contains all information needed
```
KristofferC: Yes I think these should exist. You don't have to add them to `FEDofs`, instead I think we just give a method to `FEDofs` that returns the mesh and then we can use these functions on the mesh object.

src/utilities/fe_dofs.jl
fredrikekre: I think that would be possible to get around using two different `FEValues` for the two functions though, since that will be needed anyway, right? But it makes sense that a `FEField` has a function space over which it is defined.

src/utilities/fe_mesh.jl
fredrikekre: But as it is now, the `FEMesh` does not know which fields it has. So getting the node coordinates to `reinit!` a function space which has quadratic approximation will be more awkward.

fredrikekre:
Are you in :school: today?

KristofferC:
Nope,  on train to Gbg. There tomorrow though.

KristofferC:
This starts to look nice by the way. Would be nice with some tests and for bonus points, an example notebook that uses this

KristofferC:
I will play around with this a bit and might make a PR towards this branch later.

KristofferC:
An example how the meshing could be done with what I talked about earlier today: https://gist.github.com/KristofferC/42f20f903716e3209c63

Also exports to VTK.

fredrikekre:
Looks good, but are there enough symbols to keep track of the 12 edges for a cube? :trollface:

KristofferC:
That was just some quick hack to more easily build the triangles ;)

fredrikekre:
So the idea is to incorporate the dofs in the function space? So that `Lagrange{2, Triangle, 1}` knows it should have dofs in each node, and `Lagrange{2, Triangle, 2}` knows it should have dofs in each node as well as one on each edge? Maybe this can be combined with the` reference_coordinates` you added for each function space?

KristofferC:
Yes, here is an example https://gist.github.com/KristofferC/b5b29ed1689013054c89

It adds in the end a function that create dofs on edges, nodes etc. As an example I used quadratic and linear function spaces to simulate the Taylor Hood element.

KristofferC:
Now of course the dofs will have a bad numbering but after all the dofs are created we can just go element by element, function space by function space and renumber them.

KristofferC:
One idea could be that the mesh has a function space that describes the geometry. The geometry is then interpolated using that function space. The fields are interpolated with whatever function space they have.

KristofferC:
We could store the dofs for each edge, node etc in a separate https://github.com/mbauman/RaggedArrays.jl.

This would make the dofs lie continuously in memory instead of having to store it as a `Vector{Vector{Int}}`. which is garbage for memory.

KristofferC:
Can you rebase this on `kc/dim_param`?

Fetch my stuff from `kc/dim_param`, checkout this branch and do a `git rebase kc/dim_param`. Should be no merge conflicts. That way the stuff here is put on top of that branch.

fredrikekre:
I have already merged all from `kc/dim_param`
When I did rebase all my mesh generating functions disappeared...

fredrikekre:
I actually based this branch on `kc/dim_param` from the start too

KristofferC:
Yes but I have pushed new stuff to `kc/dim_param`. In order for this branch to get the new stuff you rebase it on top of the other branch. Then it "plays" your commits on top of the last commit of `kc/dim_param`,

fredrikekre:
Okay, after solving like 15 conflicts, is this right? :)

KristofferC:
Perfect :+1: 

I'm surprised that it was that many conflicts..  You haven't really touched the files I have changed. Oh well. At least you learned how to rebase...

fredrikekre:
All the conflicts where in `fe_mesh.jl`, don't know why it caused so much trouble...

KristofferC:
That file doesn't even exist on my branch.. Maybe you had conflicts with yourself :P

fredrikekre:
Something like that :confused:

JimBrouzoulis:
> One idea could be that the mesh has a function space that describes the geometry. The geometry is then interpolated using that function space. The fields are interpolated with whatever function space they have.

Yes, this gives rise to what is classically called sub- and superparametric elements. When geometry and fields have the same interpolation, we call them isoparametric elements - but you know this already :) 

Another note: the mesh could also consist of different elements with different geometry interpolations. So where do the geometry description lie right now, with the elements or the mesh?
Maybe it should lie at the mesh level and then the elements have to adapt to whatever geometry interpolation that is currently used?

KristofferC:
Can this be closed?

fredrikekre:
I am keeping it open just to take bits and pieces from the discussion/WIP code for the grid/dofhandler, but yea, this branch can not be used.

KristofferC:
Ok, np.

fredrikekre:
ref #125

PR #58, opened 2016-04-03
clean up quadrature creation

KristofferC:
This changes so that instead of `get_gaussrule...` we create them with `GaussQuadrature{Dim[d}, Shape, Order}`.

I also removed the caching of quadrature rules because it is pretty much useless here. I cached quadrature rules so that the CALFEM elements didnt have to recreate them in every element so I will move the caching there instead.


src/quadrature_tables/gaussquad_tri_table.jl
```diff
@@ -0,0 +1,79 @@
+# Dunavant, D. A. (1985), High degree efficient symmetrical Gaussian quadrature
+# rules for the triangle. Int. J. Numer. Meth. Engng., 21: 1129–1148. doi:
+# 10.1002/nme.1620210612
+function _get_gauss_tridata(n::Int)
+    if (n == 1)
+    xw=[0.33333333333333 0.33333333333333 1.00000000000000 / 2.0];
```
fredrikekre: There were incorrect weights before?  All the` / 2.0` are new :)

fredrikekre: Actually I noticed this before as well, and I think I checked that we got the correct result, did we divide by 2 somewhere else before? Instead of in this table I mean.

src/quadrature.jl
```diff
@@ -12,136 +14,41 @@ end
...
-    data = _get_gauss_tridata(order)
-    n_points = size(data,1)
-    weights = Array(Float64, n_points)
-    points = Array(Vec{2, Float64}, n_points)
-
-    for p in 1:size(data, 1)
-        points[p] = Vec{2, Float64}((data[p, 1], data[p, 2]))
-    end
-
-    weights = 0.5 * data[:, 3]
```
fredrikekre: Okay, here I guess :D

KristofferC: Yep :)

KristofferC: But I still found a bug now... I will add tests for this. Tests shouldnt pass with bugs :P

KristofferC:
Added a test that at least the weights are correct.. However, we should check that the points are correct as well by evaluating a polynomial of the order that gets exactly evaluated.

KristofferC:
Added some tests

KristofferC:
+11% code coverage :P nice

KristofferC:
I might change the syntax for creating quadrature rules again.. `QuadratureRule{:legendre, DIm{d}, Shape, Order}` or something like that. Then it is easy to add constructors for i.e Gauss-Lobotto quadrature.

PR #59, opened 2016-04-03
remove coordxtr and topologyxtr

KristofferC:
These are too much tied to the CALFEM api and can just be used from CALFEM.jl if needed.

Also removed the VTK version that doesn't take the topology. Just create it beforehand if you dont have it instead of creating it in every VTK export call.


KristofferC:
Best way to increase code coverage, delete code.

PR #65, opened 2016-04-18
[WIP] Some convenient accessor functions for FEValues

fredrikekre:
Now we can write:

``` jl
for q_point in 1:n_quadpoints(fe_values)
    ....
end
```

instead of 

``` jl
for q_point in 1:length(points(get_quadrule(fe_values)))
    ...
end
```

in the element function.

### DIFF
--- a/src/fe_values.jl
+++ b/src/fe_values.jl
@@ -115,6 +115,19 @@ The quadrature rule for the `FEValues` type.
 get_quadrule(fe_v::FEValues) = fe_v.quad_rule
 
 """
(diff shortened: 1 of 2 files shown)


KristofferC:
Nice. I agree that this are good to have. Would of course be nice with some tests so it doesn't get accidentally broken and even better if you can figure out how to add the new methods to the docs. Ask me if you can't figure it out.

fredrikekre:
I will look into it sometime, just leave this PR open for now.

fredrikekre:
included in #88

PR #70, opened 2016-09-21
New version of vtk_grid

fredrikekre:
Thinking about removing everything above [this](https://github.com/fredrikekre/JuAFEM.jl/blob/caa497b172708c122b5c5e4f9220db0dd051f480/src/VTK.jl#L47), or should we leave it there? I think its more neat if we use the tensors and specify the celltype explicitly instead of trying to guessing.

~~Will fail until the new `WriteVTK` version is tagged.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - release
+  - 0.5
   - nightly
 notifications:
   email: false
--- a/src/VTK.jl
+++ b/src/VTK.jl
@@ -1,5 +1,10 @@
-function get_cell_type(nen, ndim)
-    # TODO: This is a bad way of figuring out the eltype
+# Deprecated function
+function vtk_grid{T}(topology::Matrix{Int}, coord::Matrix{T}, filename::AbstractString)
+    Base.depwarn("vtk_grid(topology::Matrix{Int}, coord::Matrix, filename::AbstractString) is deprecated, use vtk_grid{dim,T}(filename::AbstractString, coords::Vector{Vec{dim,T}}, topology::Matrix{Int}, celltype::VTKCellTypes.VTKCellType) instead", :vtk_grid)
+    nen = size(topology,1)
+    nnodes = size(coord, 2)
+    ndim = size(coord, 1)
+
     if nen == 3 && ndim == 2
         cell =  VTKCellTypes.VTK_TRIANGLE
     elseif nen == 4 && ndim == 2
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -4,7 +4,8 @@ os:
   - linux
   - osx
 julia:
-  - release
+  - 0.4
```
fredrikekre: Any reason to still support 0.4?

src/VTK.jl
```diff
@@ -1,40 +1,39 @@
-function get_cell_type(nen, ndim)
-    # TODO: This is a bad way of figuring out the eltype
+# Deprecated function
+function vtk_grid{T}(topology::Matrix{Int}, coord::Matrix{T}, filename::AbstractString)
```
fredrikekre: Remove this completely?

.travis.yml
KristofferC: Nope

test/REQUIRE
```diff
@@ -1,2 +1,4 @@
 BaseTestNext
-ForwardDiff
\ No newline at end of file
+ForwardDiff
+WriteVTK 0.4.0
```
KristofferC: No need to put stuff here that is required by the package itself.

test/test_VTK.jl
```diff
@@ -0,0 +1,71 @@
+import Compat.is_linux
+import Compat: readstring
+if is_linux() # sha1sum only for linux
```
KristofferC: Could use SHA.jl for support on all OS.

src/VTK.jl
KristofferC: Makes sense. It is quite nice to not have to specify but if we get a proper mesh structure then it would know. Can remove for now I guess.

src/VTK.jl
```diff
@@ -43,48 +42,44 @@ Creates an unstructured VTK grid from the element topology and coordinates.
...
-
-    nele = size(topology, 2)
-    nen = size(topology,1)
-    nnodes = size(coord, 2)
-    ndim = size(coord, 1)
-    if ndim > 3
-        throw(ArgumentError("dimension > 3, maybe you transposed the input coord matrix"))
+function vtk_grid{dim,T}(filename::AbstractString, coords::Vector{Vec{dim,T}}, topology::Matrix{Int}, celltype::VTKCellTypes.VTKCellType)
+
+    Nel = size(topology,2)
```
KristofferC: Why capital letter?

fredrikekre: idk..

fredrikekre:
Forgot to change to 0.5 and master in appveyor, is the config there correct now?

KristofferC:
Think so.

fredrikekre:
Should we tag a last working 0.4 version before merging? (Assuming this is good to go?)

KristofferC:
I did that now.

KristofferC:
Nice stuff :)

PR #71, opened 2016-09-24
Separate geometric interpolation space from function interpolation space

fredrikekre:
This PR separates the geometric interpolation from the function interpolation in `FEValues`. This makes it easier to extend to sub/super-parametric elements. It works quite nice, since the only extra thing that needs to be stored is the derivative of a (potential) other function space. Defaults to iso-parametric, so it doesn't break anything.

This also adds a new function space, `Lagrange{2,RefCube,2}` which were missing.

Some things we should decide before merging:

- [x] ~~What should the two function spaces be called in the `FEValues` type?~~
- `FunctionSpace (FS)` and `GeometricSpace (GS)`
- [x] ~~What symbol should be used for the geometric functions? Cf. `N` for the other function space. Currently I use `M`.~~
- `M`
- [x] ~~What should we call the `get` functions to get the function spaces? Previously `get_functionspace(fev::FEValues)` made more sense. I changed to `get_func_functionspace(fev::FEValues)` and `get_geom_functionspace(fev::FEValues)` to get the function space used for approximating the function and the space to interpolate the geometry respectively. It is kind of ugly, but the are mostly used internally anyways. Lets name them based on the decision of the first point above.~~
- `get_functionspace()` and `get_geometricspace()`


fredrikekre:
I found one source where they call the geometric interpolation functions: `shape functions` and the functions used to interpolate the approximated field `interpolation functions`.

src/fe_values.jl
```diff
@@ -38,25 +38,33 @@ immutable FEValues{dim, T <: Real, FS <: FunctionSpace}
     detJdV::Vector{T}
     quad_rule::QuadratureRule{dim, T}
     function_space::FS
+    dMdξ::Vector{Vector{Vec{dim, T}}}
+    geom_space::FS
```
KristofferC: If this is FS then function_space and geom_space has to be the same.

src/function_spaces.jl
```diff
@@ -186,11 +181,80 @@ function reference_coordinates(fs::Lagrange{2, RefCube, 1})
...
 
+##################################
+# Lagrange dim 2 RefCube order 2 #
+##################################
+n_basefunctions(::Lagrange{2, RefCube, 2}) = 8
 
-###################################
-# Lagrange dim 2 RefTetrahedron order 1 #
-###################################
+function value!(fs::Lagrange{2, RefCube, 2}, N::Vector, ξ::Vec{2})
```
KristofferC: Isn't this what I call serendipity below? A la grange element would have 9 nodes?

src/fe_values.jl
```diff
@@ -38,25 +38,33 @@ immutable FEValues{dim, T <: Real, FS <: FunctionSpace}
...
 end
 
-FEValues{dim, FS <: FunctionSpace}(quad_rule::QuadratureRule{dim}, func_space::FS) = FEValues(Float64, quad_rule, func_space)
+FEValues{dim, FS <: FunctionSpace}(quad_rule::QuadratureRule{dim}, func_space::FS, geom_space::FS=func_space) = FEValues(Float64, quad_rule, func_space, geom_space)
```
fredrikekre: Okay, is it enough to have like:
`FEValues{dim, FS1 <: FunctionsSpace, FS2 <: FunctionSpace}(...)`

src/function_spaces.jl
fredrikekre: Oh.... Will update later

fredrikekre: I got confused from the VTK specification I think, where the `QuadraticQuad` element is 8 nodes.

src/fe_values.jl
KristofferC: yes

fredrikekre:
What about the names? :)

fredrikekre:
YOLO.

KristofferC:
Pff, deletar sin kommentar och grejjer,

PR #72, opened 2016-09-27
check if detJdV > 0

### LINKED ISSUE #16: Warn if det(J) <= 0

fredrikekre:
Fix #16 

Should maybe be an error and not just a warning. But then our tests fail since we just random nodes to reinit the fe_values.

### DIFF
--- a/src/fe_values.jl
+++ b/src/fe_values.jl
@@ -105,7 +105,9 @@ function reinit!{dim, T}(fe_v::FEValues{dim}, x::Vector{Vec{dim, T}})
         for j in 1:n_func_basefuncs
             fe_v.dNdx[i][j] = Jinv ⋅ fe_v.dNdξ[i][j]
         end
-        fe_v.detJdV[i] = det(fev_J) * w
+        detJ = det(fev_J)
+        detJ <= 0.0 && throw(ArgumentError("detJ is not positive: detJ = $(detJ)"))
+        fe_v.detJdV[i] = detJ * w
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/fe_values.jl
```diff
@@ -105,7 +105,8 @@ function reinit!{dim, T}(fe_v::FEValues{dim}, x::Vector{Vec{dim, T}})
         for j in 1:n_func_basefuncs
             fe_v.dNdx[i][j] = Jinv ⋅ fe_v.dNdξ[i][j]
         end
-        fe_v.detJdV[i] = det(fev_J) * w
+        detJdV = det(fev_J) * w
+        detJdV > 0.0 ? fe_v.detJdV[i] = detJdV : warn("detJdV is not positive: detJdV = $(detJdV)")
```
KristofferC: Gauss quadrature weights are not necessarily positive. Just do:
```jl
detJ = det(fev_J)
detJ < 0 && warn("blabla")
fe_v.detJdV[i] = detJ * w
```
?

fredrikekre:
Should it be an `ArgumentError`? I guess the `x`-vectors giving a negative `detJ` is a wrong argument.

KristofferC:
Arg error is fine in my opinion.

fredrikekre:
Hmm... The fail on AppVeyor actually says `detJ < 0.0` for one case... Thought my minor perturbation of the the reference coords should be ok.

KristofferC:
0.5 is not so minor.

fredrikekre:
Wait what? Oh, I thought I changed it to 0.1

PR #73, opened 2016-10-02
Introducing FEBoundaryValues

### LINKED ISSUE #66: Computing things on faces of elements
There is no easy way right now to compute stuff needed for, e.g. surface traction. dealii has a `FEFaceValues` https://dealii.org/8.4.0/doxygen/deal.II/classFEFaceValues.html which has a `reinit!` method that takes a cell and a face number.

Not sure how this is done best.

fredrikekre:
This PR introduces the `FEFaceValues` type which handles boundary integration. Fix #66. It is initialized with a `FunctionSpace` and a `QuadratureRule` eg:

```jl
julia> fs = Lagrange{2,RefCube,1}();

julia> QR = QuadratureRule(Dim{1},RefCube(),3);

julia> fefaceval = FEFaceValues(QR,fs);
```

Note the dimensions: the input `QuadratureRule` has one dimension lower than the function space, since the boundary integral is carried out in a lower dimension.

The FEFaceValues object is then reinitialized for each boundary in the mesh, using reinit!:

```jl
julia> x = [0.0 0.0;
            5.0 3.0;
            6.0 10.0;
            1.0 9.0]';

julia> X = reinterpret(Vec{2,Float64},x,(4,));

julia> reinit!(fefaceval,X,1)
```

The only difference from `FEValues` is the last input, an integer indicating which boundary of the element that should be initialized.

Now, all the same functionality can be used to query for information on the boundary:

Values of the test function:
```jl
julia> shape_value(fefaceval,1)
4-element Array{Float64,1}:
 0.887298
 0.112702
 0.0     
 0.0 
```

Getting the value of the function:
```jl
julia> u = rand(4);

julia> function_scalar_value(fefaceval,1,u)
0.5713983986197158
```

See [here](https://gist.github.com/fredrikekre/7f8195f7413573fcc69c9a9bceb06b31) for some more stuff.

WIP since only one `FunctionSpace` is implemented so far and some other things that should be taken care of, like names etc...

TODO:
- [x] ~~Implement all the function spaces for boundary integration~~
  3D RefTetrahedron
- [x] Tests and TESTS!
- [x] ~~Documentation~~ Lets make a new issue for the docs and redo it for all of JuAFEM
- [x] Name it `FEFaceValues`?
- [x] Better name for the union of `FEValues` and `FEFaceValues`? `FEValuesOrFEFaceValues` is kinda awkward...

```jl
    FECellValues <: AbstractFEValues
    FEBoundaryValues <: AbstractFEValues
```


src/JuAFEM.jl
```diff
@@ -41,7 +42,10 @@ immutable Dim{T} end
 include("function_spaces.jl")
 include("quadrature.jl")
 include("fe_values.jl")
+include("fe_face_values.jl")
+include("commons_feval_feface.jl")
```
fredrikekre: Good idea to split this in separate file?

src/boundary_integrals.jl
```diff
@@ -0,0 +1,30 @@
+#########################
```
fredrikekre: Good way to set up the boundary quad rule?

src/commons_feval_feface.jl
```diff
@@ -0,0 +1,293 @@
+# Common methods for both FEValues and FEFaceValues
+typealias FEValuesOrFEFaceValues{dim,T,FS,GS} Union{FEValues{dim,T,FS,GS}, FEFaceValues{dim,T,FS,GS}}
```
fredrikekre: This union is kinda awkward...

src/commons_feval_feface.jl
```diff
@@ -0,0 +1,293 @@
...
+**Details:**
+
+This value is typically used when one integrates a function on a finite element as
+
+``\\int\\limits_\\Omega f(\\mathbf{x}) d \\Omega \\approx \\sum\\limits_{q = 1}^{n_q} f(\\mathbf{x}_q) \\det(J(\\mathbf{x})) w_q``
+``\\int\\limits_\\Gamma f(\\mathbf{x}) d \\Gamma \\approx \\sum\\limits_{q = 1}^{n_q} f(\\mathbf{x}_q) \\det(J(\\mathbf{x})) w_q``
+
+"""
+@inline detJdV(fe_v::FEValues, q_point::Int) = fe_v.detJdV[q_point]
+@inline detJdS(fe_fv::FEFaceValues, q_point::Int) = fe_fv.detJdS[fe_fv.current_boundary[]][q_point]
```
fredrikekre: `detJdS` for `FEFaceValues`? Or keep using `detJdV`?

KristofferC: Keep

src/commons_feval_feface.jl
KristofferC: Why not make an abstract type?

src/boundary_integrals.jl
```diff
@@ -0,0 +1,124 @@
...
+    push!(boundary_quad_rule, QuadratureRule(w,reinterpret(Vec{3, T},new_points,(length(w),))))
+    # Boundary 6
+    new_points = [t; s; ones(t)] # ξ = t, η = s, ζ = 1
+    push!(boundary_quad_rule, QuadratureRule(w,reinterpret(Vec{3, T},new_points,(length(w),))))
+
+    return boundary_quad_rule
+end
+
+function detJ_boundary(::Lagrange{3, RefCube, 1}, J::Tensor{2, 3}, boundary::Int)
+    boundary == 1 && return norm(J[:,1] × J[:,2])
```
KristofferC: TODO: Optimize or fix in contmechtensors.

fredrikekre: Adding a method `getindex(S::Tensor{2,dim}, :, j::Int)` that returns a `Vec{2,dim}`?

src/commons_feval_feface.jl
fredrikekre: Name suggestion?
I wanna name it `FEValues` and rename the current `FEValues` to `FECellValues` and `FEFaceValues` to `FEBoundaryValues` or something

src/boundary_integrals.jl
KristofferC: Yeah.

fredrikekre: see KristofferC/ContMechTensors.jl#35

fredrikekre: They should probably be transposed btw. `J[1,:]` etc.

src/function_spaces.jl
```diff
@@ -87,8 +103,8 @@ function derivative!{T}(fs::Lagrange{1, RefCube, 1}, dN::Vector{Vec{1, T}}, ξ::
 end
 
 function reference_coordinates(fs::Lagrange{1, RefCube, 1})
-    return (Vec{1, Float64}((-1.0,)),
-            Vec{1, Float64}(( 1.0,)))
+    return [Vec{1, Float64}((-1.0,)),
```
fredrikekre: They are only used in tests, and to use `reinit!` we need them in a vector.

fredrikekre: https://github.com/KristofferC/JuAFEM.jl/pull/73/files#diff-5f674da623f3bdd1df72ab1f1e44a8beR18

fredrikekre: But maybe that should be changed? so that we reinit with a tuple?

fredrikekre:
https://codecov.io/gh/KristofferC/JuAFEM.jl/pull/73

src/function_spaces.jl
KristofferC: If they are only used in tests could we move them to the test files then. Vector is fine.

src/commons_feval_feface.jl
KristofferC: AbstractFEValues? It is not important since it is not user facing. Your names are also good. Deprecate fevalues then?

src/function_spaces.jl
fredrikekre: Sure, there are a number of other functions which are only used for testing. We can extract them all to a `test_utils.jl` file or something.

src/commons_feval_feface.jl
fredrikekre: Like this?
```jl
FECellValues <: AbstractFEValues
FEBoundaryValues <: AbstractFEValues
```

src/function_spaces.jl
KristofferC: Yeah.

KristofferC:
Grönt är skönt :)

fredrikekre:
Maybe skip the `FE`?
```jl
CellValues <: AbstractValues
BoundaryValues <: AbstractValues
```

Because then it also makes sense to include the Applied Mechanics Core Values in `JuAFEM`, like

```jl
CoreValues <: AbstractValues
```

since they should influence our daily work. Or maybe even like this, since the `CoreValues` are pretty abstract already:

```jl
CellValues <: CoreValues
BoundaryValues <: CoreValues
```

KristofferC:
Maybe all our types should be `<: CoreValues`. For maximum.. value.

src/commons_feval_feface.jl
KristofferC: Yes

src/fe_face_values.jl
```diff
@@ -0,0 +1,82 @@
...
+FEFaceValues{dim_qr, FS <: FunctionSpace, GS <: FunctionSpace}(quad_rule::QuadratureRule{dim_qr}, func_space::FS, geom_space::GS=func_space) = FEFaceValues(Float64, quad_rule, func_space, geom_space)
+
+function FEFaceValues{dim_qr, T, FS <: FunctionSpace, GS <: FunctionSpace}(::Type{T}, quad_rule::QuadratureRule{dim_qr}, func_space::FS, geom_space::GS=func_space)
+    @assert n_dim(func_space) == n_dim(geom_space)
+    @assert ref_shape(func_space) == ref_shape(geom_space)
+    n_qpoints = length(weights(quad_rule))
+    dim = dim_qr + 1
+
+    boundary_quad_rule = create_boundary_quad_rule(func_space,quad_rule)
+    n_bounds = length(boundary_quad_rule)
```
KristofferC: `n_boundaries`

fredrikekre: name conflict:
https://github.com/KristofferC/JuAFEM.jl/pull/73/files/2a594fb5887eff7eb2f3fa95fba2976cc1a676b3#diff-1b8244793f5c4fc18ff71bab2986f30fR64

src/JuAFEM.jl
```diff
@@ -14,10 +14,11 @@ import WriteVTK: vtk_grid
 
 export start_assemble, assemble, end_assemble
 
-export FEValues, reinit!, shape_value, shape_gradient, shape_divergence, detJdV, get_quadrule, get_functionspace,
+export FEValues, reinit!, shape_value, shape_gradient, shape_divergence, detJdV, get_quadrule, get_functionspace, get_geometricspace,
                  function_scalar_value, function_vector_value, function_scalar_gradient,
                  function_vector_gradient, function_vector_divergence, function_vector_symmetric_gradient
-export n_dim, n_basefunctions
+export FEFaceValues
+export FunctionSpace, n_dim, ref_shape, fs_order, n_basefunctions
```
fredrikekre: name suggestion? its just that the [old version](https://github.com/KristofferC/JuAFEM.jl/blob/master/src/function_spaces.jl#L29) didn't work

```jl
julia> JuAFEM.order
order (generic function with 0 methods)
```

src/boundary_integrals.jl
```diff
@@ -0,0 +1,150 @@
...
+##################
+function create_boundary_quad_rule{T}(::FunctionSpace{2, RefCube}, quad_rule::QuadratureRule{1, T})
+    w = weights(quad_rule)
+    p = reinterpret(T,points(quad_rule),(1,length(w)))
+    t = p[1,:]'
+    boundary_quad_rule = QuadratureRule{2, T}[]
+
+    # Boundary 1
+    new_points = [t; -ones(t)] # ξ = t, η = -1
+    push!(boundary_quad_rule, QuadratureRule(w,reinterpret(Vec{2, T},new_points,(length(w),))))
```
fredrikekre: For the `QuadratureRule` we need them in a vector anyway. As I see it the option is to do:

```jl
# Boundary 1
new_points = Vec{2,T}[]
for i in 1:length(w)
    push!(new_points, Vec{2,T}((t[i], -one(T))))
end
push!(boundary_quad_rule, QuadratureRule(w, new_points))
# Boundary 2
...
```

or perhaps
```jl
# Boundary 1
new_points = Array(Vec{2,T}, length(w))
for i in 1:length(w)
    new_points[i] = Vec{2,T}((t[i], -one(T)))
end
push!(boundary_quad_rule, QuadratureRule(w, new_points))
# Boundary 2
...
```
Don't know whats the most efficient. Could probably use @ inbounds aswell

src/function_spaces.jl
```diff
@@ -52,11 +52,27 @@ end
...
+n_boundaries{dim}(::FunctionSpace{dim, RefCube}) = 2*dim
+n_boundaries(::FunctionSpace{2, RefTetrahedron}) = 3
+n_boundaries(::FunctionSpace{3, RefTetrahedron}) = 4
+
 ############
 # Lagrange #
 ############
 type Lagrange{dim, shape, order} <: FunctionSpace{dim, shape, order} end
 
+fs_lower_dim{dim,shape,order}(::Lagrange{dim,shape,order}) = Lagrange{dim-1,shape,order}()
```
fredrikekre: I guess this is also quite bad naming, same as `fs_order`. But these are only used internally I think.

src/fe_face_values.jl
KristofferC: Ah

src/function_spaces.jl
KristofferC: fs -> funcspace maybe?

src/boundary_integrals.jl
fredrikekre: Maybe something like this is more cleaner:
```jl
# Boundary 1
new_points = [Vec{2,T}((t[i], -one(T))) for i in 1:length(w)]
push!(boundary_quad_rule, QuadratureRule(w, new_points))
# Boundary 2
...
```

src/function_spaces.jl
fredrikekre: `funcspace_lower_dim`, `funcspace_lower_order`, `funcspace_order` ?
and also `funcspace_ref_shape` and `funcspace_n_dim`?

They are mostly used internally so maybe write out the full name?
`functionspace_lower_dim`, `functionspace_lower_order`, `functionspace_order`, `functionspace_ref_shape` and `functionspace_n_dim`

KristofferC: Yeah, sounds good to me.

fredrikekre:
Test passes locally, what happened with Travis and AppVeyor?

fredrikekre:
Something like this?
```jl
function get_side(boundary_nodes::Vector{Int}, cell_nodes::Vector{Int}, fs::Lagrange{2, RefCube, 1})
    boundary_nodes == cell_nodes[[1,2]] && return 1
    boundary_nodes == cell_nodes[[2,3]] && return 2
    boundary_nodes == cell_nodes[[3,4]] && return 3
    boundary_nodes == cell_nodes[[4,1]] && return 4
    throw(ArgumentError("can't find side in cell"))
end
```

KristofferC:
Something like this perhaps:

```jl

using JuAFEM
import Base.Cartesian.@ntuple

facelist(::Lagrange{2, RefCube, 1}) = ((1,2), (2,3), (3,4), (1,4))

function get_facenumber(face_nodes, cell_nodes, fs::JuAFEM.FunctionSpace)
    # @assert length(face_nodes) == n_facenodes # Implement something like this?
    @assert length(cell_nodes) == n_basefunctions(fs)

    tmp = zeros(face_nodes)
    for i in 1:length(face_nodes)
        tmp[i] = findfirst(j -> j == face_nodes[i], cell_nodes)
    end

    if 0 in tmp
        throw(ArgumentError("at least one face node: $face_nodes not in cell nodes: $cell_nodes"))
    end
    sort!(tmp)
    face_nodes_sorted = ntuple(i -> tmp[i], Val{2})
    for (i, face) in enumerate(facelist(fs))
        face_nodes_sorted == face && return i
    end

    throw(ArgumentError("invalid node numbers for face"))
end
```

KristofferC:
`    face_nodes_sorted = ntuple(i -> tmp[i], Val{2})` the 2 here to be fixed. Either `length(face_nodes)` or better something that is determined from the functionspace.

fredrikekre:
Was also thinking about adding something like the `facelist` and `n_boundary_nodes(fs)`.

I think that would be helpful for the mesh-stuff later as well

fredrikekre:
You better let me push the button on this one. Kappa.

fredrikekre:
`Close and comment? ... what a button.

fredrikekre:
Some final comments?

KristofferC:
Push that button :) Nice job!

PR #83, opened 2016-10-18
Remove loop in assemble

fredrikekre:
Some artifact from calfem perhaps?

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -1,7 +1,7 @@
-type Assembler
+type Assembler{T}
     I::Vector{Int}
     J::Vector{Int}
-    V::Vector{Float64}
+    V::Vector{T}
 end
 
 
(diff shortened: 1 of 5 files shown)


fredrikekre:
Btw, should we make `assemble` -> `assemble!` ?

fredrikekre:
I think it would be better "flow" if the order of the input was

```jl
assemble(a::Assembler, Ke::Matrix, edof::Vector)
```

i.e. "assemble into `a`, the matrix `Ke`, on position `edof`".

src/assembler.jl
```diff
@@ -1,7 +1,7 @@
-type Assembler
+type Assembler{T}
```
fredrikekre: `immutable`?

KristofferC: Yes.

KristofferC:
Order was a left over from calfem. Makes sense to change and call it `assemble!`

fredrikekre:
Oh look! The review is outdated, nice.

PR #84, opened 2016-10-18
Grid class

fredrikekre:
This currently only implements the `Grid` class, i.e. only the geometric object. Not sure if it makes sense yet to separate the `DoFHandler` stuff into another PR, might include it here.

Replaces #52 but keeping that open since the start of the mesh stuff there and the discussion will be used here.

Example with a small grid:

```jl
# Nodes
n1 = Node(Vec{2}((0.0,0.0)))
n2 = Node(Vec{2}((0.5,0.0)))
n3 = Node(Vec{2}((0.5,1.0)))
n4 = Node(Vec{2}((0.0,1.0)))
nodes = [n1, n2, n3, n4]

# Cells
c1 = Triangle(1, (1,2,3))
c2 = Triangle(2, (1,3,4))
celler = [c1,c2]

# Grid
grid = Grid(celler, nodes)
```

It is now convenient to loop through the cells like:
```jl
for cell in cells(grid)
    x = cell_coordinates(grid, cell)
    reinit!(cellvalues, x)
    ...
end
```

TODO:
- [x] Grid vs Mesh?
- [x] How to gather boundary info?
- [x] Tests
- [x] Decide what other functionalities that are needed
- [x] Implement some more `grid_generators` (can be done separately)
- [x] Docs (heh)


src/grid.jl
```diff
@@ -0,0 +1,90 @@
...
+# Grid utilities
+export cells, nodes, cell_coordinates, n_cells, n_nodes
+
+#########################
+# Main types for meshes #
+#########################
+"""
+A `Node` is a point in space.
+"""
+immutable Node{dim, T}
```
fredrikekre: Maybe have an id for the `Node` type aswell? Could maybe help with the implementation of `DoFHandler`.

src/grid.jl
```diff
@@ -0,0 +1,90 @@
...
+    nodes::NTuple{N, Int}
+end
+
+"""
+A `Grid` is a collection of `Cells` and `Node`s which covers the computational domain.
+"""
+immutable Grid{dim, N, T <: Real}
+    cells::Vector{Cell{dim, N}}
+    nodes::Vector{Node{dim, T}}
+    # cell_sets # Dict? with symbol/name => Vector of id's
```
fredrikekre: Ideas about this boundary stuff?

src/grid.jl
```diff
@@ -0,0 +1,90 @@
...
+    # node_sets # Dict?
+    # interpolation::FunctionSpace
+end
+
+# Helper functions
+(::Type{Cell{dim}}){dim,N}(id::Int, nodes::NTuple{N}) = Cell{dim,N}(id, nodes)
```
KristofferC: Put where the Cell is defined

src/grid.jl
KristofferC: `Dict{String, Vector{Int}}` for cell sets and node sets seems good. Perhaps worth to have a boundary sets like `Dict{String, Vector{Tuple{Int,Int}}` where first int is cell and second is edge number.

fredrikekre: You mean one set exclusive for boundaries?

KristofferC: yes

fredrikekre: Is it necessary with cell sets and node sets then?

src/grid.jl
```diff
@@ -0,0 +1,90 @@
...
+"""
+immutable Node{dim, T}
+    x::Vec{dim, T}
+end
+
+"""
+A `Cell` is a sub-domain defined by a collection of `Node`s as vertices.
+"""
+immutable Cell{dim, N}
+    id::Int
```
fredrikekre: This is bad...
```jl
c1 = Triangle(1, (1,2,3))
c2 = Triangle(2, (1,3,4))
c = [c2, c1]
```
It is really convenient to assume that node 1 is at place 1 in the array. And if they are in order, then it is unnecessary to also store the id....

KristofferC: You mean **cell** 1?

I'm just thinking what datatype `cell` should be when looping over `for cell in cells(mesh)`? Perhaps a `CellIndex` where
```
immutable CellIndex
    idx::Int
end
```

so we can dispatch on the `cell` compared to for example a `BoundaryIndex` or something...

KristofferC: The difference is that we can dispatch on `CellIndex`. So instead of `cell_coordinates` we can just have a `coordinates` that takes a `CellIndex` and one that takes a `BoundaryIndex` etc. Just an idea.

fredrikekre: Yea thats nice, although, the way `FEBoundaryValues` got implemented we need all coordinates of the cell for the boundary too (for `reinit!`), so for the case of `coordinates`, the implementation would be the same

KristofferC: Yeah. Just feels like by giving it a separate type we leave up for potential expansions in the future.

fredrikekre: Like the above comment? Or did you mean that `cells(grid)` should return a vector with `CellIndex`?

fredrikekre:
Maybe we should introduce something like `CellBoundary`. That way the difference between a boundary of the domain and the boundary of a cell is clearer. Would mimic the behaviour of domain and `Cell`, i.e. a `Cell` is a part of the domain, and `CellBoundary` is part of the boundary.

KristofferC:
And a `CellBoundary` is basically a wrapped `Tuple{Int,Int}`? Yeah makes sense. I don't see why you need a dict of cellboundaries though. Should be enough with a Vector of them and then if you want to split them up into parts you do that with sets, just like for cells.

fredrikekre:
Yes, or perhaps a wrapped Tuple{CellIndex, Int}. Where the int is for the side.

KristofferC:
Yes. Not sure it needs to be a cellindex here. Doesnt matter much though.

fredrikekre:
We can probably just have

```jl
immutable CellBoundary
    idx::Tuple{Int, Int}
end

typealias CellBoundaryIndex CellBoundary
```

KristofferC:
Makes sense.

fredrikekre:
For the DoFHandler I think it would be nice with some kind of `apply`/`map` function, that describes what happens when a `FunctionSpace` is applied to a `Cell`.

e.g.

```jl
apply(Triangle, Lagrange{2, RefCube, 1}) # returns info that there is one dof per node
apply(Triangle, Lagrange{2, RefCube, 2}) # returns info tha tthere is one dof per node and one dof per face.
```

Don't really know how the info should be returned but...

KristofferC:
By the way, is it possible to separate the mesh PR from the dofhandler PR or is it better that they are developed together?

fredrikekre:
They can be separate. This PR is all grid stuff. In the worst case we need to rewrtite some grid codes to better fit the dofhandler in the next PR.

KristofferC:
So what is needed now here is

* Sets: Element set, Node set, Boundary set.
* VTK export. Can you from the cell uniquely determine the corresponding vtk shape?

That should be all?

fredrikekre:
Its all done, let me push it up!

> Can you from the cell uniquely determine the corresponding vtk shape?
Yes, as it is right now. Will work as long as we dont have any Cells with same dim and same number of nodes; but corresponds to different elements. In that case we have to change to

```
abstract Cell{dim, N}

immutable Triangle <: Cell{2, 3}
    nodes::NTuple{3, Int}
end

immutable QuadraticTriangle <: Cell{2, 6}
    nodes::NTuple{6, Int}
end

...
```

But then we have to name each element type explicit.

fredrikekre:
- Are there any more functionality that are needed?
- Do we want some basic grid generator functions in this package? Was thinking about writing a separate package for that, but maybe it fits in here.
- Need to write some tests.

KristofferC:
It could be convenient to have some simple generators. You had already written a few? But for fancy stuff it is easier to just write a converter from another format I think.

KristofferC:
I think this is good now when it comes to functionality. Can always add more stuff later.

fredrikekre:
Will finish the generators then, implemented some in #52 which I will convert to this format.

fredrikekre:
What about names for getter functions. Right now there are alot of names which are likely to be used in a program, like `n_nodes`, `cells` etc. Should we add get to such functions?

```jl
nodes => get_nodes
cells => get_cells
n_nodes => get_n_nodes
# Existing functions
points => get_points
weights => get_weights
```

KristofferC:
Julia uses `getindex`, `getfield` for example. `getnodes` looks good to me? Don't know if the underscore is needed or not.

fredrikekre:
Like above?

KristofferC:
Looks good to me :)

fredrikekre:
Will add tests, then this is done I think.

fredrikekre:
They can be separate. This PR only contains grid stuff so far.



Sent from my Samsung device


-------- Original message --------

fredrikekre:
Test ideas?

fredrikekre:
Simple as this:

```jl
julia> grid = generate_grid(Tetrahedron, (5,5,5));

julia> vtkfile = vtk_grid("tetra", grid);

julia> vtk_save(vtkfile);
```
tetra

KristofferC:
Nice!

src/grid.jl
```diff
@@ -93,6 +93,15 @@ end
 @inline getnodeset(grid::Grid, set::Symbol) = grid.nodes[grid.nodesets[set]]
 @inline getcellboundaryset(grid::Grid, set::Symbol) = grid.cellboundaries[grid.cellboundarysets[set]]
 
+function addcellset!(grid::Grid, name::Symbol, cellid::Vector{Int})
+    haskey(grid.cellsets, name) && warn("There already exists a cellset with the name: $name")
```
KristofferC: Error here instead, in my opinion.

fredrikekre: Okay.

Btw, why does it show as an outdated diff?

fredrikekre:
Let's go.

KristofferC:
Sweet!

Issue #87, opened 2016-10-27
Change FEBoundaryValues


fredrikekre:
For a 3D case, the `FEBoundaryValues` should be a 2D space, embedded in a 3D space. Like dealii has a `dim` and a `spacedim`.

fredrikekre:
Skip this?

KristofferC:
For now maybe yes.

fredrikekre:
I don't think this is needed, the current solution is pretty nice.

PR #88, opened 2016-10-28
add FEVectorCellValues

KristofferC:
@fredrikekre


src/JuAFEM.jl
```diff
@@ -14,8 +14,10 @@ import WriteVTK: vtk_grid
 
 export start_assemble, assemble!, end_assemble
 
-export FECellValues, reinit!, shape_value, shape_gradient, shape_divergence, getdetJdV, getquadrule, getfunctionspace, getgeometricspace,
-                     function_value, function_gradient, function_symmetric_gradient, function_divergence, spatial_coordinate
+export FECellValues, FEVectorCellValues
+
+export reinit!, shape_value, shape_gradient, shape_divergence, getdetJdV, getquadrule, getfunctionspace, getgeometricspace,
```
fredrikekre: add `shape_symmetric_gradient` ?

src/fe_vector_cell_values.jl
```diff
@@ -0,0 +1,55 @@
...
+        derivative!(func_space, dNdξ_temp, ξ)
+        basefunc_count = 1
+        for basefunc in 1:getnbasefunctions(func_space)
+            for comp in 1:dim
+                N_comp = zeros(T, dim)
+                N_comp[comp] = N_temp[basefunc]
+                N[i][basefunc_count] = Vec{dim, T}((N_comp...))
+                
+                dN_comp = zeros(T, dim, dim)
+                dN_comp[:, comp] = dNdξ_temp[basefunc]
```
fredrikekre: shouldn't this be `dN_comp[comp,:] = dNdξ_temp[basefunc]` instead?

KristofferC: No, I did that first but it turns out wrong. I think it is because of the transposed jacobian :P

fredrikekre: Then we should transpose the Jacobian!

src/JuAFEM.jl
KristofferC: Added and exported

fredrikekre: And it should only be defined for `FEVectorCellValues`, since `symmetric()` only works for second order tensors.

fredrikekre:
Should it be named `FECellScalarValues` and `FECellVectorValues`? Since it is not "vectorcells" but rather "vectorvalues"?

fredrikekre:
If we remove them: Then it could just as well be in a matrix as well, right?

KristofferC:
See #89

KristofferC:
Updated a bit. Gonna go for a while now :)

fredrikekre:
```jl
@deprecate FECellValues FEScalarCellValues
```

Or should we name it like my above comment?

KristofferC:
Like you said. I made the change for your suggestion locally but gotta be social for the rest of the day...

fredrikekre:
Btw, if we send in a vector of `Vec`s to get the function value and function gradient, wouldn't that basically mean we have to reinterpret to vector of `Vec`s and then reinterpret back inside the function?

KristofferC:
Not really, just do like: https://github.com/KristofferC/JuAFEM.jl/pull/88/files#diff-c2eaa020dc34bec1e323f380b27fc1d3R212 ?

fredrikekre:
Yea, but typically we have the values in a vector. And we have to reinterpret to send it in I mean.

KristofferC:
Don't see how anything has changed from before? Yes, the element internal forces routine gets a vector of values but then I always chunk it up into Vector of Vecs or Vector of T depending on the field.

fredrikekre:
Something is wrong I think:

```jl
julia> fs = Lagrange{2, RefCube, 1}();

julia> fev = FECellVectorValues(qr, fs);

julia> qr = QuadratureRule{2, RefCube}(2); fs = Lagrange{2, RefCube, 1}();

julia> fs = Lagrange{2, RefCube, 1}();

julia> fev = FECellVectorValues(qr, fs);

julia> x = reference_coordinates(fs);

julia> reinit!(fev, x)

julia> shape_value(fev, 1, 1) # First shape function
2-element ContMechTensors.Tensor{1,2,Float64,2}:
 0.622008
 0.0     

julia> shape_gradient(fev, 1, 1) # First gradient
2×2 ContMechTensors.Tensor{2,2,Float64,4}:
 -0.394338  0.0
 -0.394338  0.0

```

The gradient of the first shape function should be:
```jl
2×2 ContMechTensors.Tensor{2,2,Float64,4}:
 -0.394338  -0.394338
  0.0        0.0
```

fredrikekre:
The failing tests suggests the same, the gradient is transposed, see the test logs.

KristofferC:
Yeah, just pushed what I had on laptop.

fredrikekre:
I think [this](https://github.com/KristofferC/JuAFEM.jl/pull/88/files#diff-c8255a1936df95b1753569cd8e4d91b5R44) should actually be `dN_comp[comp, :] = dNdξ_temp[basefunc]` and also in [`reinit!`](https://github.com/KristofferC/JuAFEM.jl/blob/cab43626099ae7cc5f9c1d5abac3d34ed7931ab4/src/fe_cell_values.jl#L111) I think it should be:

```jl
for j in 1:n_func_basefuncs
            fe_cv.dNdx[j,i] = Jinv ⋅ fe_cv.dNdξ[j, i]'
end
```
with the transpose at the end. Hence why I made it a no-op to transpose a vector in the scalar case. Thus it would be better to change it to 

```jl
for j in 1:n_func_basefuncs
            fe_cv.dNdx[j,i] = fe_cv.dNdξ[j, i] ⋅ Jinv
end
```

and change the calculation of `Jinv` to actually be `Jinv` and not its transpose.

KristofferC:
https://github.com/KristofferC/JuAFEM.jl/pull/88/commits/b10865303df70688fd31512af867aa28102b6fbb#diff-41bd23b93bca8688dda6c880b4b3143cL107

KristofferC:
Should pass now.

src/JuAFEM.jl
```diff
@@ -14,8 +14,10 @@ import WriteVTK: vtk_grid
 
 export start_assemble, assemble!, end_assemble
 
-export FECellValues, reinit!, shape_value, shape_gradient, shape_divergence, getdetJdV, getquadrule, getfunctionspace, getgeometricspace,
-                     function_value, function_gradient, function_symmetric_gradient, function_divergence, spatial_coordinate
+export FECellValues, FECellVectorValues
+
+export reinit!, shape_value, shape_gradient, symmetric_shape_gradient, shape_divergence, getdetJdV, getquadrule, getfunctionspace, getgeometricspace,
```
fredrikekre: `symmetric_shape_gradient => shape_symmetric_gradient`

KristofferC:
Maybe you can add `getnbasefunctions` for BoundaryValues?

fredrikekre:
Can do that in a separate PR and also define FEBoundaryVectorValues in that PR.

fredrikekre:
Just deprecate `FECellValues => FECellScalarValues` and its good to go?

src/fe_cell_vector_values.jl
```diff
@@ -11,6 +11,7 @@ immutable FECellVectorValues{dim, T <: Real, FS <: FunctionSpace, GS <: Function
 end
 
 FECellVectorValues{dim, FS <: FunctionSpace, GS <: FunctionSpace}(quad_rule::QuadratureRule{dim}, func_space::FS, geom_space::GS=func_space) = FECellVectorValues(Float64, quad_rule, func_space, geom_space)
+getnbasefunctions(fe_cvv::FECellVectorValues) = getnbasefunctions(fe_cvv.function_space) * dim
```
fredrikekre: need to parametrize on `dim to` make this work.

KristofferC: ups, will add tests but have to go lecture now :)

fredrikekre:
I'm surprised tests passed for boundary values since we transposed the jacobian, this should be needed to change too: https://github.com/KristofferC/JuAFEM.jl/blob/kc/fe_vector_cells/src/boundary_integrals.jl#L111

src/JuAFEM.jl
```diff
@@ -32,14 +34,16 @@ immutable RefTetrahedron <: AbstractRefShape end
 immutable RefCube <: AbstractRefShape end
 
 """
-Abstract type which has `FECellValues` and `FEBoundaryValues` as subtypes
+Abstract type which has `FECellScalarValues` and `FEBoundaryValues` as subtypes
 """
 abstract AbstractFEValues{dim, T, FS, GS}
+abstract AbstractFECellScalarValues{dim, T, FS, GS} <: AbstractFEValues{dim, T, FS, GS}
```
fredrikekre: Should this be `AbstractFECellValues`? `FECellScalarValues` it the concrete type.

I thought the hierarchy should be:

```jl
abstract AbstractFEValues
abstract AbstractFECellValues <: AbstractFEValues
immutable FECellScalarValues <: AbstractFECellValues end
immutable FECellVectorValues <: AbstractFECellValues end

abstract AbstractFEBoundaryValues <: AbstractFEValues
immutable FEBoundaryScalarValues <: AbstractFEBoundaryValues end
immutable FEBoundaryVectorValues <: AbstractFEBoundaryValues end

```

src/commons_abstract_fevalues.jl
```diff
@@ -1,4 +1,4 @@
-# Common methods for both FECellValues and FEBoundaryValues
+# Common methods for both FECellScalarValues and FEBoundaryValues
```
fredrikekre: see above, isnt this commone methods for all values now? eg `Common methods for AbstractFEValues`

src/commons_abstract_fevalues.jl
```diff
@@ -14,10 +14,18 @@ The quadrature rule for the `AbstractFEValues` type.
 * `::QuadratureRule`: the quadrature rule.
 
 """
-getquadrule(fe_cv::FECellValues) = fe_cv.quad_rule
+getquadrule(fe_cv::FECellScalarValues) = fe_cv.quad_rule
```
fredrikekre: with my suggested hierarchy this could be `getquadrule(fe_cv::AbstractFECellValues) = fe_cv.quad_rule`

src/commons_abstract_fevalues.jl
```diff
@@ -71,7 +79,7 @@ This value is typically used when one integrates a function on a finite element
 ``\\int\\limits_\\Gamma f(\\mathbf{x}) d \\Gamma \\approx \\sum\\limits_{q = 1}^{n_q} f(\\mathbf{x}_q) \\det(J(\\mathbf{x})) w_q``
 
 """
-@inline getdetJdV(fe_cv::FECellValues, q_point::Int) = fe_cv.detJdV[q_point]
+@inline getdetJdV(fe_cv::AbstractFECellScalarValues, q_point::Int) = fe_cv.detJdV[q_point]
```
fredrikekre: this should be `AbstractFECellValues` then?

src/commons_abstract_fevalues.jl
```diff
@@ -82,25 +90,26 @@ Computes the value of the shape function
 Gets the values of the shape function for a given quadrature point and base_func
 
 """
-@inline shape_value(fe_cv::FECellValues, q_point::Int, base_func::Int) = fe_cv.N[base_func, q_point]
+@inline shape_value(fe_cv::AbstractFECellScalarValues, q_point::Int, base_func::Int) = fe_cv.N[base_func, q_point]
```
fredrikekre: same

src/commons_abstract_fevalues.jl
```diff
@@ -82,25 +90,26 @@ Computes the value of the shape function
...
 @inline shape_value(fe_bv::FEBoundaryValues, q_point::Int, base_func::Int) = fe_bv.N[base_func, q_point, fe_bv.current_boundary[]]
 
-@inline geometric_value(fe_cv::FECellValues, q_point::Int, base_func::Int) = fe_cv.M[base_func, q_point]
+@inline geometric_value(fe_cv::AbstractFECellScalarValues, q_point::Int, base_func::Int) = fe_cv.M[base_func, q_point]
```
fredrikekre: same

src/commons_abstract_fevalues.jl
```diff
@@ -82,25 +90,26 @@ Computes the value of the shape function
...
 @inline geometric_value(fe_bv::FEBoundaryValues, q_point::Int, base_func::Int) = fe_bv.M[base_func, q_point, fe_bv.current_boundary[]]
 
 
+
 """
 Get the gradient of the shape functions for a given quadrature point and base function
 """
-@inline shape_gradient(fe_cv::FECellValues, q_point::Int, base_func::Int)     = fe_cv.dNdx[base_func, q_point]
+@inline shape_gradient(fe_cv::AbstractFECellScalarValues, q_point::Int, base_func::Int) = fe_cv.dNdx[base_func, q_point]
```
fredrikekre: same

src/commons_abstract_fevalues.jl
```diff
@@ -82,25 +90,26 @@ Computes the value of the shape function
...
 @inline shape_gradient(fe_bv::FEBoundaryValues, q_point::Int, base_func::Int) = fe_bv.dNdx[base_func, q_point, fe_bv.current_boundary[]]
-
+@inline shape_symmetric_gradient(fe_cv::FECellVectorValues, q_point::Int, base_func::Int) = symmetric(shape_gradient(fe_cv, q_point, base_func))
 const shape_derivative = shape_gradient
 
 """
 Get the divergence of the shape functions for a given quadrature point and base function
 """
-@inline shape_divergence(fe_cv::FECellValues, q_point::Int, base_func::Int) = sum(fe_cv.dNdx[base_func, q_point])
+@inline shape_divergence(fe_cv::AbstractFECellScalarValues, q_point::Int, base_func::Int) = sum(fe_cv.dNdx[base_func, q_point])
```
fredrikekre: same

src/commons_abstract_fevalues.jl
```diff
@@ -128,7 +137,7 @@ where ``u_i`` are the value of ``u`` in the nodes. For a vector valued function
 ``\\mathbf{u}(\\mathbf{x}) = \\sum\\limits_{i = 1}^n N_i (\\mathbf{x}) \\mathbf{u}_i`` where ``\\mathbf{u}_i`` are the
 nodal values of ``\\mathbf{u}``.
 """
-@inline function function_value{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{T})
+function function_value{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{T})
```
fredrikekre: this could be defined for the union of `FECellScalarValues` and `FEBoundaryScalarValues` later on.

src/commons_abstract_fevalues.jl
```diff
@@ -138,7 +147,7 @@ nodal values of ``\\mathbf{u}``.
     return s
 end
 
-@inline function function_value{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_value{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: this could be defined for the union of FECellVectorValues and FEBoundaryVectorValues later on.

fredrikekre: Wait, when will this method even be called?

src/commons_abstract_fevalues.jl
```diff
@@ -174,7 +197,7 @@ For a vector valued function the gradient is computed as
 ``\\mathbf{\\nabla} \\mathbf{u}(\\mathbf{x}) = \\sum\\limits_{i = 1}^n \\mathbf{\\nabla} N_i (\\mathbf{x}) \\otimes \\mathbf{u}_i``
 where ``\\mathbf{u}_i`` are the nodal values of ``\\mathbf{u}``.
 """
-@inline function function_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{T})
+function function_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{T})
```
fredrikekre: union of both scalar values

src/commons_abstract_fevalues.jl
```diff
@@ -184,7 +207,7 @@ where ``\\mathbf{u}_i`` are the nodal values of ``\\mathbf{u}``.
     return grad
 end
 
-@inline function function_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: union of the vector values

fredrikekre: same, this method will never be used?

src/commons_abstract_fevalues.jl
```diff
@@ -194,6 +217,21 @@ end
...
 
+function function_gradient{dim, T}(fe_v::FECellVectorValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+    n_base_funcs = getnbasefunctions(getfunctionspace(fe_v))
+    @assert length(u) == n_base_funcs
+    grad = zero(Tensor{2, dim, T})
+    basefunc_count = 1
+    @inbounds for i in 1:n_base_funcs
+        for j in 1:dim
+            grad += u[i][j] * shape_gradient(fe_v, q_point, basefunc_count)
+            basefunc_count += 1
```
fredrikekre: called just `basefunc` elsewhere

src/commons_abstract_fevalues.jl
```diff
@@ -219,7 +257,7 @@ The symmetric gradient of a scalar function is computed as
 
 where ``\\mathbf{u}_i`` are the nodal values of the function.
 """
-@inline function function_symmetric_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_symmetric_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: Can be defined only for the union of vector values.

src/commons_abstract_fevalues.jl
```diff
@@ -248,7 +286,7 @@ The divergence of a vector valued functions in the quadrature point ``\\mathbf{x
 
 where ``\\mathbf{u}_i`` are the nodal values of the function.
 """
-@inline function function_divergence{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_divergence{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: union of vector values

fredrikekre: maybe should be deleted? same as above

src/fe_cell_scalar_values.jl
```diff
@@ -71,18 +72,18 @@ function FECellValues{dim, T, FS <: FunctionSpace, GS <: FunctionSpace, shape <:
...
     detJdV = zeros(T, n_qpoints)
 
-    FECellValues(N, dNdx, dNdξ, detJdV, quad_rule, func_space, M, dMdξ, geom_space)
+    FECellScalarValues(N, dNdx, dNdξ, detJdV, quad_rule, func_space, M, dMdξ, geom_space)
 end
 
 
 """
-Updates the `FECellValues` object for an element.
+Updates the `FECellScalarValues` object for an element.
```
fredrikekre: maybe: "Updades the `AbstractFEValues` object for an element" since this method is used for them all?

src/fe_cell_scalar_values.jl
```diff
@@ -94,20 +95,21 @@ Updates the `FECellValues` object for an element.
 
 
 """
-function reinit!{dim, T}(fe_cv::FECellValues{dim}, x::Vector{Vec{dim, T}})
+function reinit!{dim, T}(fe_cv::AbstractFECellScalarValues{dim}, x::Vector{Vec{dim, T}})
```
fredrikekre: AbstractFECellValues?

src/fe_cell_vector_values.jl
```diff
@@ -0,0 +1,56 @@
+immutable FECellVectorValues{dim, T <: Real, FS <: FunctionSpace, GS <: FunctionSpace, shape <: AbstractRefShape, M} <: AbstractFECellScalarValues{dim, T, FS, GS}
```
fredrikekre: SHouldnt this subtype `AbstractFECellValues`? weird that the vector values subtypes the scalar values...

fredrikekre:
I suggest this:

```jl
abstract AbstractFEValues
abstract AbstractFECellValues <: AbstractFEValues
immutable FECellScalarValues <: AbstractFECellValues end
immutable FECellVectorValues <: AbstractFECellValues end

abstract AbstractFEBoundaryValues <: AbstractFEValues
immutable FEBoundaryScalarValues <: AbstractFEBoundaryValues end
immutable FEBoundaryVectorValues <: AbstractFEBoundaryValues end

typealias AbstractFEScalarValues Union{AbstractFECellScalarValues, AbstractFEBoundaryScalarValues}
typealias AbstractFEVectorValues Union{AbstractFECellVectorValues, AbstractFEBoundaryVectorValues}
```

KristofferC:
Some search and replace errors.

KristofferC:
Can we just drop the `abstract`.. It gets soo verbose.

fredrikekre:
Sure, why not drop `FE` as well while we are at it?

KristofferC:
Sure

fredrikekre:
```jl
abstract Values
abstract CellValues <: Values
immutable CellScalarValues <: CellValues end
immutable CellVectorValues <: CellValues end

abstract BoundaryValues <: Values
immutable BoundaryScalarValues <: BoundaryValues end
immutable BoundaryVectorValues <: BoundaryValues end

typealias ScalarValues Union{CellScalarValues, BoundaryScalarValues}
typealias VectorValues Union{CellVectorValues, BoundaryVectorValues}
```

so much cleaner.

src/JuAFEM.jl
```diff
@@ -32,16 +34,23 @@ immutable RefTetrahedron <: AbstractRefShape end
 immutable RefCube <: AbstractRefShape end
 
 """
-Abstract type which has `FECellValues` and `FEBoundaryValues` as subtypes
+Abstract type which has `CellScalarValues` and `BoundaryScalarValues` as subtypes
```
fredrikekre: ... which has `CellValues` and `BoundaryValues` as ...

src/boundary_scalar_values.jl
```diff
@@ -1,18 +1,18 @@
 """
-An `FEBoundaryValues` object facilitates the process of evaluating values shape functions, gradients of shape functions,
+An `BoundaryScalarValues` object facilitates the process of evaluating values shape functions, gradients of shape functions,
```
fredrikekre: *A* `BoundaryScalarValues` object ...
Kappa

src/common_values.jl
```diff
@@ -1,62 +1,110 @@
...
-The quadrature rule for the `AbstractFEValues` type.
+Updates the `CellValues` object for an element.
 
-    getquadrule(fe_v::AbstractFEValues)
+    reinit!{dim, T}(cv::CellValues{dim}, x::Vector{Vec{dim, T}})
 
 ** Arguments **
 
-* `fe_v`: the `AbstractFEValues` object
+* `cv`: the `CellScalarValues` object
```
fredrikekre: CellValues

KristofferC:
By the way, it seems that Abaqus defines surfaces as element + side so we are not totally doing things wrong...

src/common_values.jl
```diff
@@ -219,19 +297,19 @@ The symmetric gradient of a scalar function is computed as
 
 where ``\\mathbf{u}_i`` are the nodal values of the function.
 """
-@inline function function_symmetric_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_symmetric_gradient{dim, T}(fe_v::Values{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: only for VectorValues

src/common_values.jl
```diff
@@ -248,7 +326,7 @@ The divergence of a vector valued functions in the quadrature point ``\\mathbf{x
 
 where ``\\mathbf{u}_i`` are the nodal values of the function.
 """
-@inline function function_divergence{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
+function function_divergence{dim, T}(fe_v::ScalarValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
```
fredrikekre: this method should be deleted?

fredrikekre: no defined with `u::Vector{T}` right?

fredrikekre: Can you even take divergence for a scalar function?

fredrikekre:
Maybe keep it like it is then instead of embedding as suggested in #87 ?

src/common_values.jl
KristofferC: This is only for vector valued *functions* (`u`)

src/common_values.jl
fredrikekre: Basically allow for the previous behaviour? Do we do that on all the functions then?

KristofferC: Yes allow for the previous behavior. I don't want all my code to break and it is technically better performance even if you have to do some algebra to get the corresponding expression from the weak form. :P

src/common_values.jl
fredrikekre: Alright!

fredrikekre:
Btw, would be interesting with some benchmarking problems.

src/JuAFEM.jl
```diff
@@ -32,16 +34,23 @@ immutable RefTetrahedron <: AbstractRefShape end
...
-include("fe_boundary_values.jl")
-include("commons_abstract_fevalues.jl")
+include("cell_scalar_values.jl")
+include("cell_vector_values.jl")
+include("boundary_scalar_values.jl")
+
+typealias ScalarValues{dim, T, FS, GS} Union{CellScalarValues{dim, T, FS, GS}, BoundaryScalarValues{dim, T, FS, GS}}
+typealias VectorValues{dim, T, FS, GS} Union{CellVectorValues{dim, T, FS, GS} #=,BoundaryVectorValues{dim, T, FS, GS}=# } # TODO
+
+include("common_values.jl")
```
fredrikekre: why not `core_values.jl`

KristofferC: haha ok

KristofferC:
https://github.com/JuliaCI/PkgBenchmark.jl

Would be simple to benchmark reinits for different function spaces and the different  function evaluators. I have to focus on getting my curve today though :P 

OK to merge this if CI passes?

fredrikekre:
Go! Will fix the boundary stuff sometime.

PR #89, opened 2016-10-28
make stuff compact

KristofferC:
Changes all the nested vectors to pack their data compactly. Easier to read and better performance imo.

@fredrikekre

### DIFF
--- a/src/commons_abstract_fevalues.jl
+++ b/src/commons_abstract_fevalues.jl
@@ -72,43 +72,36 @@ This value is typically used when one integrates a function on a finite element 
 
 """
 @inline getdetJdV(fe_cv::FECellValues, q_point::Int) = fe_cv.detJdV[q_point]
-@inline getdetJdV(fe_bv::FEBoundaryValues, q_point::Int) = fe_bv.detJdV[fe_bv.current_boundary[]][q_point]
+@inline getdetJdV(fe_bv::FEBoundaryValues, q_point::Int) = fe_bv.detJdV[q_point, fe_bv.current_boundary[]]
 
 """
 Computes the value of the shape function
 
-    shape_value(fe_v::AbstractFEValues, quadrature_point::Int, [base_function::Int])
+    shape_value(fe_v::AbstractFEValues, quadrature_point::Int, base_function::Int)
 
 Gets the values of the shape function for a given quadrature point and base_func
 
 """
-@inline shape_value(fe_cv::FECellValues, q_point::Int) = fe_cv.N[q_point]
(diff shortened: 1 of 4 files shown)


fredrikekre:
Out of curiosity; why the change from `Vector` to `AbstractVector` on all the places?

KristofferC:
Because I pass in a subarray when creating fevalues which is not a vector.

src/commons_abstract_fevalues.jl
```diff
@@ -186,21 +177,19 @@ where ``\\mathbf{u}_i`` are the nodal values of ``\\mathbf{u}``.
...
 
 @inline function function_gradient{dim, T}(fe_v::AbstractFEValues{dim}, q_point::Int, u::Vector{Vec{dim, T}})
     n_base_funcs = getnbasefunctions(getfunctionspace(fe_v))
     @assert length(u) == n_base_funcs
-    dN = shape_gradient(fe_v, q_point)
     grad = zero(Tensor{2, dim, T})
-    for i in 1:n_base_funcs
-        grad += u[i] ⊗ dN[i]
+    @inbounds for i in 1:n_base_funcs
+        grad += u[i] ⊗ shape_gradient(fe_v, q_point, i)
```
fredrikekre: This method won't work I think. Atleast the \otimes should be a regular *
the input `u` should also now be a regular vector of values, not a vector of `Vec`'s.
We can probably just combined the two `function_gradient` methods now.

src/commons_abstract_fevalues.jl
```diff
@@ -150,9 +142,8 @@ end
     n_base_funcs = getnbasefunctions(getfunctionspace(fe_v))
     @assert length(u) == n_base_funcs
     vec = zero(Vec{dim, T})
-    N = shape_value(fe_v, q_point)
     @inbounds for i in 1:n_base_funcs
-        vec += N[i] * u[i]
+        vec += shape_value(fe_v, q_point, i) * u[i]
```
fredrikekre: same here, the `u` needs to be a regular vector.

src/commons_abstract_fevalues.jl
KristofferC: Are these comments regarding #88?

KristofferC: FWIW, I still feel it makes sense to send in a vector of `Vec`s.

fredrikekre: Well, we still have to rewrite the code.

KristofferC: yes

PR #107, opened 2016-11-09
FunctionSpace -> Interpolation

fredrikekre:
Search and replace...

What do you think about this?


KristofferC:
Good with me, maybe fix in notebook too?

fredrikekre:
Go?

PR #100, opened 2016-11-02
[WIP] Grid fixes

fredrikekre:
Lets push updates to the grid stuff on to this branch and merge it all together.

PR #108, opened 2016-11-14
Cherries from #100

fredrikekre:


KristofferC:
Maybe we should turn off nightly testing for a while?

src/grid.jl
```diff
@@ -150,8 +150,36 @@ function addnodeset!(grid::Grid, name::String, f::Function)
...
+
+"""
+@inline function getcoordinates!{dim, T, N}(x::Vector{Vec{dim, T}}, grid::Grid{dim, N, T}, cell::Int)
+    @assert length(x) == N
+    @inbounds for i in 1:N
+        x[i] = grid.nodes[grid.cells[cell].nodes[i]].x
+    end
+end
+@inline getcoordinates!{dim, T, N}(x::Vector{Vec{dim, T}}, grid::Grid{dim, N, T}, cell::CellIndex) = getcoordinates!(x, grid, cell.idx)
+@inline getcoordinates!{dim, T, N}(x::Vector{Vec{dim, T}}, grid::Grid{dim, N, T}, boundary::CellBoundaryIndex) = getcoordinates!(x, grid, boundary.idx[1])
```
KristofferC: This will give all coordinates in the cell that the boundary is attached to. To me it makes more sense to only give the ones that exists on the boundary?

fredrikekre: Well, the problem is that we need them all to `reinit!` the boundary values.

KristofferC: Yeah ok. In the long run it would probably be better to `reinit!` with a `Grid + CellBoundaryIndex` (or just an index in case we save more in it).

fredrikekre: Yes, we should rewrite the boundary stuff...

KristofferC: Don't really need to rewrite it though? Just pass a "higher level" object instead of the raw coordinates.

fredrikekre: Yes, I am trying some stuff now for regular cells, can probably map that to boundary stuff too later.

fredrikekre:
Coverage is only run for nightly though.
Perhaps we can skip that for a while as well, since these functionalities are not really tested anyway hehe

KristofferC:
yeah

fredrikekre:
I need to rebase on this branch now. Merge?

fredrikekre:
fdklafjasdklfjdaklös

fredrikekre:
Docs are not run in the testscript like on `ContMechTensors`, maybe update that some time too

KristofferC:
yeah, create issue maybe?

PR #109, opened 2016-11-14
grid generator fix

fredrikekre:
Think this fixes the problem with tetrahedron boundaries. @KristofferC can you try this with your notebook when possible?


fredrikekre:
Rebased. Did this fix the problem btw?

fredrikekre:
Yolo

PR #111, opened 2016-12-06
relax matrix in assembler

fredrikekre:
```jl
julia> ke = rand(SMatrix{4,4,Float64});

julia> typeof(ke) <: Matrix{Float64}
false

julia> typeof(ke) <: AbstractMatrix{Float64}
true
```

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -33,7 +33,7 @@ end
 
 Assembles the element matrix `Ke` into `a`.
 """
-function assemble!{T}(a::Assembler{T}, Ke::Matrix{T}, edof::Vector{Int})
+function assemble!{T}(a::Assembler{T}, Ke::AbstractMatrix{T}, edof::AbstractVector{Int})
     n_dofs = length(edof)
     append!(a.V, Ke[:])
     for dof1 in 1:n_dofs, dof2 in 1:n_dofs


KristofferC:
`edof` could probably be an abstractvector as well.

fredrikekre:
Derp, can not index into a `SMatrix` anyway, so pretty useless to have `ke` as an `SMatrix`.

KristofferC:
You can use an `MMatrix`.

fredrikekre:
Oh, nice!

Seems like it is defined for SMatrix too, just not working?

```jl
julia> ke = zero(SMatrix{4,4,Float64});

julia> @which ke[1,1] = 1.0
setindex!{SM<:StaticArrays.StaticArray{T,2}}(m::SM, val, i1::Integer, i2::Integer) at /home/fredrik/.julia/v0.5/StaticArrays/src/indexing.jl:226
```

KristofferC:
`SMatrix` is stored on the stack so you can basically only have `setindex` without the `!` that returns a new object.

fredrikekre:
Yea, but there is a `setindex!` defined in the code is what I meant.

fredrikekre:
Btw, `StaticArrays` implements indexing with tuples, which is pretty nice! I always create a vector from the tuple and index with that before :)

KristofferC:
Yes, but it might be removed because it should probably be in base. But you can also index with an `SVector` to get another `SVector`.

fredrikekre:
> Not sure it will matters for performance though.

Seems to be a bit slower, at least if you use a predefined matrix

fredrikekre:
AV is on vacation

PR #112, opened 2016-12-14
make assember for residual

fredrikekre:
Such symmetry!

screenshot from 2016-12-14 17-12-23

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -51,3 +51,14 @@ assembled values.
 function end_assemble(a::Assembler)
     return sparse(a.I, a.J, a.V)
 end
+
+"""
+    assemble!(g, ge, edof)
+
+Assembles the element residual `ge` into the global residual vector `g`.
+"""
+function assemble!{T}(g::AbstractVector{T}, ge::AbstractVector{T}, edof::AbstractVector{Int})
+    for i in 1:length(edof)
+        g[edof[i]] += ge[i]
+    end
+end
--- a/docs/src/lib/utility_functions.md
+++ b/docs/src/lib/utility_functions.md
@@ -59,6 +59,14 @@ getboundarynumber
 getcurrentboundary
 ```
 
+## Assembling
+
+```@docs
+start_assemble
+assemble!
+end_assemble
+```
+
 
 ## VTK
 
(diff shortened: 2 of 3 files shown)


src/assembler.jl
```diff
@@ -51,3 +51,12 @@ assembled values.
...
     return sparse(a.I, a.J, a.V)
 end
+
+"""
+    assemble!(g, ge, edof)
+
+Assembles the element residual `ge` into the global residual vector `g`.
+"""
+@inline function assemble!{T}(g::AbstractVector{T}, ge::AbstractVector{T}, edof::AbstractVector{Int})
+    g[edof] += ge
```
KristofferC: Bad performance from `+=` due to the extra array created. Use a loop :)

fredrikekre: 25 times faster for a small test

src/assembler.jl
```diff
@@ -33,10 +33,10 @@ end
 
 Assembles the element matrix `Ke` into `a`.
 """
-function assemble!{T}(a::Assembler{T}, Ke::AbstractMatrix{T}, edof::AbstractVector{Int})
+@inline function assemble!{T}(a::Assembler{T}, Ke::AbstractMatrix{T}, edof::AbstractVector{Int})
     n_dofs = length(edof)
     append!(a.V, Ke[:])
-    for dof1 in 1:n_dofs, dof2 in 1:n_dofs
+    @inbounds for dof1 in 1:n_dofs, dof2 in 1:n_dofs
```
KristofferC: This is not safe if user puts a bad `edof`. Should also not be inlined

src/assembler.jl
```diff
@@ -51,3 +51,14 @@ assembled values.
...
     return sparse(a.I, a.J, a.V)
 end
+
+"""
+    assemble!(g, ge, edof)
+
+Assembles the element residual `ge` into the global residual vector `g`.
+"""
+@inline function assemble!{T}(g::AbstractVector{T}, ge::AbstractVector{T}, edof::AbstractVector{Int})
+    @inbounds for i in 1:length(ge)
```
fredrikekre: Do we need to `@assert length(ge) == length(edof) < length(g)` ?
We don't do such a check for the stiffness.

KristofferC: asserts are typically used for purely internal errors (like getting into an "impossible" state), while throwing is done for user errors.

fredrikekre:
What was the test failure here? Didn't seem related though

PR #113, opened 2016-12-15
use > rather than <=, check before inverting

fredrikekre:
1.5% faster Kappa


KristofferC:
But if it is zero it is still bad?

fredrikekre:
`detJ > 0.0 || throw(ArgumentError("detJ is not positive: detJ = $(detJ)"))`

KristofferC:
Oh OK.

PR #115, opened 2016-12-19
[RFC] Better way too loop over grid cells

fredrikekre:
First take on a better cell iterator interface.

This only provide a way to loop over the grid, without coupling with `CellValues`.

(keeping it in a separate file until ready to be put in `grid.jl` I guess)

### DIFF
--- a/src/cell_iterator.jl
+++ b/src/cell_iterator.jl
@@ -2,6 +2,33 @@ using Base: RefValue
 
 export getid
 
+"""
+A `CellIterator` is returned when looping over the cells in a grid. The `CellIterator`
+contains information about the cell which can be queried from the object.
+
+**Example:**
+
+```julia
+for cell in grid                 # cell is now a CellIterator
+
+    id = getid(cell)             # get the cell number
+    coord = getcoordinates(cell) # get the coordinates
+    nodes = getnodes(cell)       # get the node numbers
+
+end
+```
+The `CellIterator` can also be used directly to [`reinit!`](@ref) the [`CellValues`](@ref):
+
+```julia
+for cell in grid
+
+    reinit!(cv, cell)
+
+    # do stuff
+
+end
+```
+"""
 immutable CellIterator{dim, T}
     cellid::RefValue{Int}
     nodes::Vector{Int}
@@ -9,16 +36,12 @@ immutable CellIterator{dim, T}
 end
 
 @inline function Base.start{dim, N, T}(grid::Grid{dim, N, T})
-    # N is number of nodes per cell in the grid
     nodes = zeros(Int, N)
     coords = zeros(Vec{dim, T}, N)
     return CellIterator(RefValue{Int}(0), nodes, coords)
 end
 
 @inline function Base.next{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T})
-    # this is a bit awkward, we need to basically reinit the state and also send it out
-    # that means that we send out the object twice, but it works I guess.
-    # the second output state is only used internally and is hidden from the user
     ci.cellid[] += 1
     nodeids = grid.cells[ci.cellid[]].nodes
     @inbounds for i in 1:N
@@ -27,7 +50,7 @@ end
         ci.coords[i] = grid.nodes[nodeid].x
     end
 
-    return ci, ci # the fact that we return ci twice here is allocating memory?!
+    return ci, ci
 end
 
 @inline Base.done{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T}) = ci.cellid[] >= getncells(grid)
--- a/src/cell_iterator.jl
+++ b/src/cell_iterator.jl
@@ -12,14 +12,14 @@ end
     # N is number of nodes per cell in the grid
     nodes = zeros(Int, N)
     coords = zeros(Vec{dim, T}, N)
-    return CellIterator(RefValue{Int}(1), nodes, coords)
+    return CellIterator(RefValue{Int}(0), nodes, coords)
 end
 
 @inline function Base.next{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T})
     # this is a bit awkward, we need to basically reinit the state and also send it out
     # that means that we send out the object twice, but it works I guess.
     # the second output state is only used internally and is hidden from the user
-    ci.cellid[] += 0
+    ci.cellid[] += 1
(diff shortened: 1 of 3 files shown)


src/cell_iterator.jl
```diff
@@ -0,0 +1,32 @@
...
+end
+
+function Base.start{dim, N, T}(grid::Grid{dim, N, T})
+    # N is number of nodes per cell in the grid
+    nodes = zeros(Int, N)
+    coords = zeros(Vec{dim, T}, N)
+    return CellIterator(Ref(0), nodes, coords)
+end
+
+function Base.next{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T})
```
fredrikekre: for some reason `next` seem to allocate memory...

fredrikekre:
The alternative approach would be to create a new `CellIterator` in each `next`-call, which would allow state to be just an integer. But that should be less efficient, right?

fredrikekre:
Why is it so much slower to return the same object twice?

returning just one:

```jl
julia> @benchmark next($grid, $ci)
BenchmarkTools.Trial: 
  memory estimate:  0.00 bytes
  allocs estimate:  0
  --------------
  minimum time:     13.282 ns (0.00% GC)
  median time:      13.442 ns (0.00% GC)
  mean time:        14.544 ns (0.00% GC)
  maximum time:     84.151 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     998
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

returning two:

```jl
julia> @benchmark next($grid, $ci)
BenchmarkTools.Trial: 
  memory estimate:  32.00 bytes
  allocs estimate:  1
  --------------
  minimum time:     19.894 ns (0.00% GC)
  median time:      20.283 ns (0.00% GC)
  mean time:        23.095 ns (5.91% GC)
  maximum time:     1.246 μs (96.98% GC)
  --------------
  samples:          10000
  evals/sample:     996
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

fredrikekre:
Before latest commit:

```jl
julia> @benchmark next($grid, $ci)
BenchmarkTools.Trial: 
  memory estimate:  2.03 kb
  allocs estimate:  41
  --------------
  minimum time:     1.683 μs (0.00% GC)
  median time:      1.762 μs (0.00% GC)
  mean time:        2.031 μs (5.89% GC)
  maximum time:     147.508 μs (97.36% GC)
  --------------
  samples:          10000
  evals/sample:     10
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

fredrikekre:
```jl
julia> foo(x) = x;

julia> bar(x) = (x, x);

julia> x = rand(4);

julia> @code_native foo(x)
	.text
Filename: REPL[1]
	pushq	%rbp
	movq	%rsp, %rbp
Source line: 1
	movq	%rdi, %rax
	popq	%rbp
	retq
	nopl	(%rax)

julia> @code_native bar(x)
	.text
Filename: REPL[2]
	pushq	%rbp
	movq	%rsp, %rbp
	pushq	%r14
	pushq	%rbx
	subq	$32, %rsp
	movq	%rdi, %r14
	movq	%fs:0, %rbx
	addq	$-2672, %rbx            # imm = 0xFFFFFFFFFFFFF590
	vxorps	%xmm0, %xmm0, %xmm0
	vmovups	%xmm0, -32(%rbp)
	movq	$4, -48(%rbp)
	movq	(%rbx), %rax
	movq	%rax, -40(%rbp)
	leaq	-48(%rbp), %rax
	movq	%rax, (%rbx)
Source line: 1
	movabsq	$jl_gc_pool_alloc, %rax
	movl	$1456, %esi             # imm = 0x5B0
	movl	$32, %edx
	movq	%rbx, %rdi
	callq	*%rax
	movabsq	$140325128561296, %rcx  # imm = 0x7F9FFD70E290
	movq	%rcx, -8(%rax)
	movq	%rax, -32(%rbp)
	movq	%r14, -24(%rbp)
	movq	%r14, (%rax)
	movq	%r14, 8(%rax)
	movq	-40(%rbp), %rcx
	movq	%rcx, (%rbx)
	addq	$32, %rsp
	popq	%rbx
	popq	%r14
	popq	%rbp
	retq
	nopw	%cs:(%rax,%rax)

julia> using BenchmarkTools

julia> @benchmark foo($x)
BenchmarkTools.Trial: 
  memory estimate:  0.00 bytes
  allocs estimate:  0
  --------------
  minimum time:     1.815 ns (0.00% GC)
  median time:      1.891 ns (0.00% GC)
  mean time:        1.890 ns (0.00% GC)
  maximum time:     42.746 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark bar($x)
BenchmarkTools.Trial: 
  memory estimate:  32.00 bytes
  allocs estimate:  1
  --------------
  minimum time:     7.399 ns (0.00% GC)
  median time:      10.367 ns (0.00% GC)
  mean time:        11.983 ns (10.84% GC)
  maximum time:     1.050 μs (97.74% GC)
  --------------
  samples:          10000
  evals/sample:     999
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

KristofferC:
You need to allocate memory for the tuple when it contains objects that are allocated on the heap (like arrays).

fredrikekre:
But its is the same array?

fredrikekre:
eg.
```jl
julia> x = rand(4);

julia> bar(x) = (x,x);

julia> a, b = bar(x)
([0.637593,0.308879,0.653693,0.639757],[0.637593,0.308879,0.653693,0.639757])

julia> a[1] = 1
1

julia> b
4-element Array{Float64,1}:
 1.0     
 0.308879
 0.653693
 0.639757

julia> pointer(a)
Ptr{Float64} @0x00007fe1bad04d00

julia> pointer(b)
Ptr{Float64} @0x00007fe1bad04d00
```

fredrikekre:
Anyhow, now we can do this:

```jl
julia> qr = QuadratureRule{3, RefCube}(2);

julia> ip = Lagrange{3, RefCube, 1}();

julia> cv = CellScalarValues(qr, ip);

julia> grid = generate_grid(Hexahedron, (3,3,3));

julia> for cell in grid
           reinit!(cv, cell)
           # do stuff
       end
```

KristofferC:
Yea but the tuple itself needs to be allocated.

PR #118, opened 2017-01-25
Relax to abstractvector

fredrikekre:
Since these methods do not modify the input vectors it should be safe to input views.

eg:

```julia
u_element = view(u_global, edof)

∇u = function_gradient(cv, qp, u_element)
```


fredrikekre:
Using `view`s might not be the best thing performance wise. But no reason to restrict the methods to `Vector`s so I'm merging 😄

PR #119, opened 2017-02-14
swap to Tensors.jl

KristofferC:


appveyor.yml
```diff
@@ -28,7 +28,7 @@ build_script:
 # Need to convert from shallow to complete for Pkg.clone to work
   - IF EXIST .git\shallow (git fetch --unshallow)
   - C:\projects\julia\bin\julia -e "versioninfo();
-      Pkg.clone(\"https://github.com/KristofferC/ContMechTensors.jl\");
+      Pkg.clone(\"https://github.com/KristofferC/Tensors.jl\");
```
fredrikekre: Could this be in `REQUIRE` instead? Since it is registered??

KristofferC: Yeah should work

PR #121, opened 2017-02-16
rework celliterator

fredrikekre:
This is much cleaner. Usage is now:

```julia
grid = generate_grid(Triangle, (2, 2));

ci = CellIterator(grid);

for i in 1:getncells(grid)
    reinit!(ci, grid, i) # reinit the celliterator
    reinit!(cv, ci)      # reinit the cellvalues via the celliterator
    # do stuff
end
```

Edit: Take 3:
```julia
grid = generate_grid(Triangle, (2, 2));

for cell in CellIterator(grid)
    reinit!(cv, cell)      # reinit the cellvalues via the celliterator
    # do stuff
end
```

it can also be done manually like this:

```julia
grid = generate_grid(Triangle, (2, 2));

ci = CellIterator(grid)

for i in 1:getncells(grid)
    reinit!(ci, i)         # reinit the "iterator"
    reinit!(cv, cell)      # reinit the cellvalues via the celliterator
    # do stuff
end
```

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -32,10 +32,15 @@ function CellIterator{dim, N, T}(grid::Grid{dim, N, T})
     return CellIterator(grid, nodes, coords)
 end
 
+# iterator interface
 Base.start(::CellIterator) = 1
 Base.next{dim, N, T}(ci::CellIterator{dim, N, T}, i) = (reinit!(ci, i), i+1)
 Base.done(ci::CellIterator, i) = i > getncells(ci.grid)
 Base.length(ci::CellIterator) = getncells(ci.grid)
+
+Base.iteratorsize{dim, N, T}(::Type{CellIterator{dim, N, T}}) = Base.HasLength()   # this is default in Base
+Base.iteratoreltype{dim, N, T}(::Type{CellIterator{dim, N, T}}) = Base.HasEltype() # this is default in Base
+Base.eltype{dim, N, T}(::Type{CellIterator{dim, N, T}}) = CellIterator{dim, N, T}
 
 # utility
 @inline getnodes(ci::CellIterator) = ci.nodes
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -8,39 +8,43 @@ A `CellIterator` contains information about the cell which can be queried from t
 **Example:**
 
 ```julia
-ci = CellIterator(grid)
-
-for i in 1:getncells(grid)
-    reinit!(grid, ci, i)          # update the CellIterator for cell i
+for cell in CellIterator(grid)
     coords = getcoordinates(cell) # get the coordinates
     nodes = getnodes(cell)        # get the node numbers
 
-    reinit!(cv, ci)               # reinit the FE-base with a CellIterator
+    reinit!(cv, cell)             # reinit! the FE-base with a CellIterator
 end
 ```
 """
-immutable CellIterator{dim, T}
+immutable CellIterator{dim, N, T}
+    grid::Grid{dim, N, T}
     nodes::Vector{Int}
     coords::Vector{Vec{dim, T}}
 end
 
-@inline function CellIterator{dim, N, T}(grid::Grid{dim, N, T})
+function CellIterator{dim, N, T}(grid::Grid{dim, N, T})
     nodes = zeros(Int, N)
     coords = zeros(Vec{dim, T}, N)
-    return CellIterator(nodes, coords)
+    return CellIterator(grid, nodes, coords)
 end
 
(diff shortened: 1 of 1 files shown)


KristofferC:
It is a bit weird to not iterate over a `CellIterator`? Should be called something else?

fredrikekre:
Hmm... Its called that in deal.ii I think. 
https://www.dealii.org/8.4.1/doxygen/deal.II/classFEValuesBase_1_1CellIterator.html

and fenics:

KristofferC:
Yes but don't you loop over it then? `for cell : CellIterator..`?

KristofferC:
Like now, you dont need `next` `start` etc. So not even following the iterator interface

fredrikekre:
ah, right...
Suggestions?

fredrikekre:
`CellValues` kappa

fredrikekre:
`GridValues`?

KristofferC:
`FECellValues` + `GeometryCellValues`.. :P

fredrikekre:
Perhaps this instead:

```julia
for cell in CellIterator(grid)
    # do stuff
end
```
Then we can keep the name? What usage is nicest?

KristofferC:
Yeah, I like that

src/iterators.jl
```diff
@@ -8,39 +8,43 @@ A `CellIterator` contains information about the cell which can be queried from t
...
 
-@inline function reinit!{dim, N, T}(ci::CellIterator{dim, T}, grid::Grid{dim, N, T}, i::Int)
-    nodeids = grid.cells[i].nodes
-    @inbounds for i in 1:N
-        nodeid = nodeids[i]
-        ci.nodes[i] = nodeid
-        ci.coords[i] = grid.nodes[nodeid].x
+@inline Base.start(::CellIterator) = 1
+
+@inline function Base.next{dim, N, T}(ci::CellIterator{dim, N, T}, i)
```
KristofferC: No point inlining

fredrikekre:
updated top post

KristofferC:
Much gusto. I like being explicit of what to loop over.

fredrikekre:
It is a bit weird to return the full thing though. It is like returning the entire array when looping over the elements 😛

KristofferC:
Should it be `reinit!(cv, cell)` in top post?

src/iterators.jl
```diff
@@ -1,59 +1,54 @@
...
-        ci.coords[i] = grid.nodes[nodeid].x
-    end
-    return ci, ci
-end
-
-@inline Base.done{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T}) = ci.cellid[] >= getncells(grid)
+Base.start(::CellIterator) = 1
+Base.next{dim, N, T}(ci::CellIterator{dim, N, T}, i) = (reinit!(ci, i), i)
+Base.done(ci::CellIterator, i) = i > getncells(ci.grid)
+Base.lenth(ci::CellIterator) = getncells(ci.grid)
```
fredrikekre: `lenth` 4Head

KristofferC:
Can also add https://github.com/KristofferC/Tokenize.jl/blob/master/src/lexer.jl#L57 functions

src/iterators.jl
```diff
@@ -1,59 +1,54 @@
...
-        nodeid = nodeids[i]
-        ci.nodes[i] = nodeid
-        ci.coords[i] = grid.nodes[nodeid].x
-    end
-    return ci, ci
-end
-
-@inline Base.done{dim, N, T}(grid::Grid{dim, N, T}, ci::CellIterator{dim, T}) = ci.cellid[] >= getncells(grid)
+Base.start(::CellIterator) = 1
+Base.next{dim, N, T}(ci::CellIterator{dim, N, T}, i) = (reinit!(ci, i), i+1)
```
fredrikekre: Will `reinit!` inline here? add `@inline` here? https://github.com/KristofferC/JuAFEM.jl/pull/121/files#diff-a5af3864a37bfecc3c1b632d479b21b4R44

fredrikekre:
What are they used for? `collect` and friends?

KristofferC:
Yeah, not sure how valuable they are here but still

fredrikekre:
Didn't effect benchmarking for regular looping, but added anyway

fredrikekre:
mac build not starting, but we are not testing this currently anyways so...

fredrikekre:
I like this implementation. It is very easy to extend the `CellIterator` with more fields in the future, if we want to. And add boolean flags to set which fields should be updated.

Eg like this:
```julia
ci = CellIterator(grid, nodes = false)

for c in ci
    # do stuff
end
```
if the nodes should not be updated.

PR #124, opened 2017-03-06
inbounds assembling

fredrikekre:
This should be safe, but doesn't give much of an improvement.

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -4,7 +4,6 @@ immutable Assembler{T}
     V::Vector{T}
 end
 
-
 function Assembler(N)
     I = Int[]
     J = Int[]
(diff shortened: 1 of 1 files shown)


src/assembler.jl
```diff
@@ -58,7 +58,8 @@ end
 Assembles the element residual `ge` into the global residual vector `g`.
 """
 function assemble!{T}(g::AbstractVector{T}, ge::AbstractVector{T}, edof::AbstractVector{Int})
-    for i in 1:length(edof)
+    @boundscheck checkbounds(g, edof)
```
fredrikekre: IIUC this `checkbounds` call will be skipped if this function is called with `@inbounds assemble!(g, ge, edof)` ?

src/assembler.jl
```diff
@@ -35,10 +34,12 @@ Assembles the element matrix `Ke` into `a`.
 """
 function assemble!{T}(a::Assembler{T}, Ke::AbstractMatrix{T}, edof::AbstractVector{Int})
     n_dofs = length(edof)
```
KristofferC: No boundscheck here?

fredrikekre: what is there to check?

fredrikekre: we just append stuff.

fredrikekre: We could do something like:
`@assert checksquare(Ke) == length(edof)`
But if this doesn't hold, it will error in `sparse(I, J, V)` anyway. Thoughts?

KristofferC: Leave it as is. I said that yesterday but forgot to "finish review"

PR #125, opened 2017-03-20
Lots of stuff...

KristofferC:


fredrikekre:
Nice, possible to extract renaming commits to another PR and merge that?

KristofferC:
Yes

KristofferC:
Sorry for the mess... I will try to fix it up later. The interesting changes right now are the following:

* Fix so looping over faces is more natural
* Add a `DofHandler` type
* Add a `DirichletBoundaryCondition` type.
* Add a `create_sparsity_pattern` function that creates the sparsity pattern from a dofhandler.
* Add a new assembler type that takes a sparse matrix and efficiently assemble into it.
* Create a function `apply!` that modifies the stiffness matrix and optionally the force vector so that `K \ f` solves the system including boundary conditions.

Everything is showed in the committed `cantilevel.ipynb` notebook.

fredrikekre:
(we should make a new release before merging this)

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+    push!(dbcs.bcs, DirichletBoundaryCondition(f, Set(nodes), field, components, idxoffset))
+
+end
+
+# Updates the DBC's to the current time `time`
+function update!(dbcs::DirichletBoundaryConditions, time::Float64 = 0.0)
+    @assert dbcs.closed[]
+    bc_offset = 0
+    for dbc in dbcs.bcs
+        # Function barrier
```
fredrikekre: ELI5

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+
+function add!(dbcs::DirichletBoundaryConditions, field::Symbol,
+                          nodes::Union{Set{Int}, Vector{Int}}, f::Function, component::Int=1)
+    add!(dbcs, field, nodes, f, [component])
+end
+
+# Adds a boundary condition
+function add!(dbcs::DirichletBoundaryConditions, field::Symbol,
+                          nodes::Union{Set{Int}, Vector{Int}}, f::Function, components::Vector{Int})
+    @assert field in dbcs.dh.field_names || error("Missing: $field")
```
fredrikekre: the `@assert` should be removed?

src/Dofs/DofHandler.jl
```diff
@@ -0,0 +1,250 @@
...
+julia> a = rand(ndofs(dh))
+
+julia> vtkfile = vtk_grid(grid)
+
+julia> vtk_point_data(vtkfile
+```
+
+"""
+
+# TODO: Make this immutable?
```
fredrikekre: Any reason it's not?

src/Dofs/DofHandler.jl
```diff
@@ -0,0 +1,250 @@
...
+type DofHandler{dim, N, T, M}
+    dofs_nodes::Matrix{Int}
```
fredrikekre: I guess this is the reason this is not immutable?
We could just make this Vector so it can be resized? IIRC they did this in fenics, with a corresponding offset vector. That can also handle different elements, which this cant

src/Dofs/DofHandler.jl
```diff
@@ -0,0 +1,250 @@
...
+
+# Creates a sparsity pattern from the dofs in a dofhandler.
+# Returns a sparse matrix with the correct pattern.
+function create_sparsity_pattern(dh::DofHandler)
+    grid = dh.grid
+    n = ndofs_per_cell(dh)
+    N = n^2 * getncells(dh.grid)
+    I = Int[]
+    J = Int[]
+    sizehint!(I, N)
```
fredrikekre: We could just allocate the full vectors here instead of just sizehinting? or is it faster to push?

src/Grid/grid.jl
```diff
@@ -23,96 +23,111 @@ immutable Node{dim, T}
...
+Face() = Face(false)
+=#
 
 """
 A `Cell` is a sub-domain defined by a collection of `Node`s as it's vertices.
 """
-immutable Cell{dim, N}
+immutable Cell{dim, N, M}
     nodes::NTuple{N, Int}
+    onboundary::NTuple{M, Bool}
```
fredrikekre: Why is this not N? Isn't it easier to define if the node is on the boundary instead of the face?

KristofferC: We could say what nodes are on the boundary and then `onboundary(cell, node)` checks if all the nodes are on the boundary.

src/Dofs/DofHandler.jl
KristofferC: I think the speed should be exactly the same but allocating up front is probably cleaner.

src/Dofs/DofHandler.jl
KristofferC: Or just not have it at all and use `node_dofs`? In Fenics, I think they use it only for the mesh?

KristofferC: But node dofs doesn't even make sense because dofs can be on other places than nodes...

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+    DirichletBoundaryConditions
+
+A dirichlet boundary conditions is a a boundary where a dof is fixed to take a certain value.
+The struct `DirichletBoundaryConditions` represents a collection of such boundary conditions.
+
+It is created from a `DofHandler`
+
+```jldoctest dbc
```
KristofferC: No, documentation is not complete, just some stuff to make reviewing easier :P

examples/hyperelasticity.jl
```diff
@@ -0,0 +1,216 @@
+module HyperElasticity
```
KristofferC: Yes, this can be removed

src/Dofs/DirichletBoundaryConditions.jl
KristofferC: Yes

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+# Adds a boundary condition
+function add!(dbcs::DirichletBoundaryConditions, field::Symbol,
+                          nodes::Union{Set{Int}, Vector{Int}}, f::Function, components::Vector{Int})
+    @assert field in dbcs.dh.field_names || error("Missing: $field")
+    for component in components
+        @assert 0 < component <= ndim(dbcs.dh, field)
+    end
+
+    if length(nodes) == 0
+        warn("Added Dirichlet BC to node set containing 9 nodes")
```
KristofferC: Eeh, it should be a 0 here...

KristofferC: It is to warn if you apply a boundary condition to a node set with 0 nodes

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+    components::Vector{Int}
+    idxoffset::Int
+end
+
+
+immutable DirichletBoundaryConditions{DH <: DofHandler}
+    bcs::Vector{DirichletBoundaryCondition}
+    dofs::Vector{Int}
+    free_dofs::Vector{Int}
+    values::Vector{Float64}
```
KristofferC: Yeah, makes sense

src/Dofs/DirichletBoundaryConditions.jl
KristofferC: If we have to do dynamic dispatch on `dbc.f` every time we call it, things will be slow. Therefore do dynamic dispatch once and call out to a function which will specialize on every `f`.

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,237 @@
...
+        for dbc in dbcs.bcs
+            println(io, "    ", "Field: ", dbc.field, " ", "Components: ", dbc.components)
+        end
+    end
+end
+
+isclosed(dbcs::DirichletBoundaryConditions) = dbcs.closed[]
+dirichlet_dofs(dbcs::DirichletBoundaryConditions) = dbcs.dofs
+free_dofs(dbcs::DirichletBoundaryConditions) = dbcs.free_dofs
+function close!(dbcs::DirichletBoundaryConditions)
```
KristofferC: I thought `close!` like "the door is closed so you can't add anything more now. The fact that a bunch of things are happening under the scene doesn't really concern a user.

fredrikekre: Sure

src/Dofs/DirichletBoundaryConditions.jl
fredrikekre: Okay, cool

test/test_utils.jl
```diff
@@ -99,17 +143,18 @@ function rotmat(dim, θ=π/6)
         R = Tensor{2,2}((cos(θ), sin(θ), -sin(θ), cos(θ)))
         return R
     else
-        u = Vec{3}((1.0, 2.0, 3.0)); u *= norm(u)
+        u = Vec{3}((1.0, 2.0, 3.0)); u /= norm(u)
```
KristofferC: Tack för den minan ;)

src/FEValues/face_integrals.jl
```diff
@@ -0,0 +1,166 @@
...
+    # Face 4
+    new_points = [Vec{2, T}((-one(T), p[i][1])) for i in 1:n_points] # ξ = -1, η = t
+    push!(face_quad_rule, QuadratureRule{2, shape, T}(w, new_points))
+
+    return face_quad_rule
+end
+
+function weighted_normal{T}(J::Tensor{2, 2}, ::FaceValues{2, T, RefCube}, face::Int)
+    @inbounds begin
+        face == 1 && return Vec{2}(( J[2,1], -J[1,1]))
```
KristofferC: By the way, this stuff computes the normal now, so you can use `getnormal(facevalue, q_point)`.

KristofferC: Also, this was previously type unstable because the checks where only

```
face = 1 && return ...
face = 2 && return  ...
...
```
and then there were no "base case". Adding an `error` in the end fixes it.

src/FEValues/face_values.jl
```diff
@@ -0,0 +1,274 @@
...
+immutable FaceScalarValues{dim, T <: Real, refshape <: AbstractRefShape} <: FaceValues{dim, T, refshape}
+    N::Array{T, 3}
+    dNdx::Array{Vec{dim, T}, 3}
+    dNdξ::Array{Vec{dim, T}, 3}
+    detJdV::Matrix{T}
+    normals::Vector{Vec{dim, T}}
+    M::Array{T, 3}
+    dMdξ::Array{Vec{dim, T}, 3}
+    qr_weights::Vector{T}
+    current_face::ScalarWrapper{Int}
```
KristofferC: Actually, `Ref` is not an leaf type so using it created type instabilities. Created a new (mutable) type that just wraps a value.

src/FEValues/face_integrals.jl
fredrikekre: Sweet!
I wonder how this will work out if we would have quadratic geometric interpolation. I am pretty sure we would need a completely other strategy?

src/Grid/grid.jl
```diff
@@ -0,0 +1,242 @@
...
+onboundary(f::Face) = f.onboundary
+Face() = Face(false)
+=#
+
+"""
+A `Cell` is a sub-domain defined by a collection of `Node`s as it's vertices.
+"""
+immutable Cell{dim, N, M}
+    nodes::NTuple{N, Int}
+    onboundary::NTuple{M, Bool}
```
KristofferC: It feels weird to store this here. Why not in a separate vector in the `grid`? We already have the cell id in the cell iterator so it is easy to retrieve anyway.

src/Grid/grid.jl
```diff
@@ -0,0 +1,242 @@
...
+end
+
+"""
+A `Grid` is a collection of `Cells` and `Node`s which covers the computational domain, together with Sets of cells, nodes and faces.
+"""
+type Grid{dim, N, T <: Real, M}
+    cells::Vector{Cell{dim, N, M}}
+    nodes::Vector{Node{dim, T}}
+    # Sets
+    cellsets::Dict{String, Set{Int}}
```
KristofferC: Changed to sets here.

src/Grid/grid.jl
```diff
@@ -0,0 +1,242 @@
...
+
+@inline getfaceset(grid::Grid, set::String) = grid.facesets[set]
+@inline getfacesets(grid::Grid) = grid.facesets
+
+n_faces_per_cell(grid::Grid) = nfaces(eltype(grid.cells))
+getfacelist(grid::Grid) = getfacelist(eltype(grid.cells))
+
+# Transformations
+
+function transform!(g::Grid, f::Function)
```
KristofferC: Added this thing to transform all the coordinates in a grid,

src/Grid/grid_generators.jl
```diff
@@ -0,0 +1,386 @@
...
+
+    _apply_onboundary!(QuadraticLine, cells, boundary)
+
+    # Cell face sets
+    facesets = Dict("left"  => Set{Tuple{Int, Int}}([boundary[1]]),
+                    "right" => Set{Tuple{Int, Int}}([boundary[2]]))
+    return Grid(cells, nodes, facesets=facesets)
+end
+
+function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
```
KristofferC: It is not possible to give all four corners in the 2d grid generators (as in the COok membrane notebook). Should probably extend to 3d as well.

src/assembler.jl
```diff
@@ -64,3 +64,122 @@ function assemble!{T}(g::AbstractVector{T}, ge::AbstractVector{T}, edof::Abstrac
...
+end
+
+
+function start_assemble(K::SparseMatrixCSC, f::Vector=Float64[])
+    fill!(K.nzval, 0.0)
+    AssemblerSparsityPattern(K, f, Int[], eltype(K)[])
+end
+
+assemble!(A::AssemblerSparsityPattern, Ke::AbstractMatrix, dofs::AbstractVector{Int}) = assemble!(A, eltype(Ke)[], Ke, dofs)
+function assemble!(A::AssemblerSparsityPattern, fe::AbstractVector, Ke::AbstractMatrix, dofs::AbstractVector{Int})
```
KristofferC: This should have time complexity `O(nd * nz) + O(nd * log(nd))` where nd is the number of dofs and nz is the number of non zeros per column. The first term is the insertion and the second is the sorting.

Just inserting by binary search would have complexity `O(nd^2 * log(nz))` because we insert `nd^2` values into sorted lists with `nz` values. For large `nz` the former method should be slower. We could benchmark what the threshold is and switch depending on the lengths.

KristofferC: Ignoring the sorting part, we have that the ratio is `O(nd * nz) / O(nd^2 * log(nz))` = `O(nz / (nd * log(nz)))` so presumable, the threshold should be given in terms of `nz / (nd * log(nz))`

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -0,0 +1,246 @@
...
+
+function add!(dbcs::DirichletBoundaryConditions, field::Symbol,
+                          nodes::Union{Set{Int}, Vector{Int}}, f::Function, component::Int=1)
+    add!(dbcs, field, nodes, f, [component])
+end
+
+# Adds a boundary condition
+function add!(dbcs::DirichletBoundaryConditions, field::Symbol,
+                          nodes::Union{Set{Int}, Vector{Int}}, f::Function, components::Vector{Int})
+    field in dbcs.dh.field_names || error("field $field did not exist in the dof handler, exisitng fields are $(dh.field_names)")
```
fredrikekre: existing

src/FEValues/cell_values.jl
```diff
@@ -87,29 +85,27 @@ function CellScalarValues{dim, T, FIP <: Interpolation, GIP <: Interpolation, sh
...
     N::Matrix{Vec{dim, T}}
     dNdx::Matrix{Tensor{2, dim, T, M}}
     dNdξ::Matrix{Tensor{2, dim, T, M}}
     detJdV::Vector{T}
-    quad_rule::QuadratureRule{dim, shape, T}
-    func_interpol::FIP
     M::Matrix{T}
     dMdξ::Matrix{Vec{dim, T}}
-    geom_interpol::GIP
+    qr_weights::Vector{T}
```
fredrikekre: what are these used for?

src/FEValues/cell_values.jl
```diff
@@ -87,29 +85,27 @@ function CellScalarValues{dim, T, FIP <: Interpolation, GIP <: Interpolation, sh
...
 end
 
-CellVectorValues{dim, FIP <: Interpolation, GIP <: Interpolation}(quad_rule::QuadratureRule{dim}, func_interpol::FIP, geom_interpol::GIP=func_interpol) =
+function CellVectorValues{dim}(quad_rule::QuadratureRule{dim}, func_interpol::Interpolation, geom_interpol::Interpolation=func_interpol)
```
fredrikekre: might as well remove `dim` parameter here too

examples/cantilever.ipynb
```diff
@@ -18,9 +24,20 @@
...
+    "deletable": true,
+    "editable": true
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "\u001b[1m\u001b[34mINFO: Recompiling stale cache file /home/kristoffer/.julia/lib/v0.5/JuAFEM.ji for module JuAFEM.\n",
```
fredrikekre: :+1:

KristofferC: Argh!

KristofferC:
I dunno why 0.5 is so slow for hyper elasticity...

KristofferC:
Everything is like super slow on 0.5... Stiffness example takes 1.6s on 0.6 and 45s on 0.5...

fredrikekre:
It's probably https://github.com/KristofferC/Tensors.jl/pull/19

KristofferC:
Ah.. yeah.

KristofferC:
Maybe we should revert it... and just add it to a 0.6 only release later? It makes things pretty unusable on 0.5

fredrikekre:
It's a non-trivial revert at this point. Perhaps just checkout an old commit of `Tensors` when testing on 0.5?

fredrikekre:
```
---------------------------------------------------------------------------

                                    Time                   Allocations      

                            ----------------------   -----------------------

      Tot / % measured:          2373s / 99.9%           68.9GiB / 99.9%    

 Section            ncalls     time   %tot     avg     alloc   %tot      avg

 ---------------------------------------------------------------------------

 assemble                6    2354s  99.3%    392s   64.5GiB  93.7%  10.8GiB

   inner assemble    30.0k    2349s  99.0%  78.3ms   64.3GiB  93.4%  2.20MiB

 linear solve            5    17.7s  0.75%   3.54s   4.33GiB  6.28%   886MiB

 export                  1   40.9ms  0.00%  40.9ms   3.49MiB  0.00%  3.49MiB

 ---------------------------------------------------------------------------
```

such performance 40 min assemble

KristofferC:
I think it is because inlining is disabled on 0.5?

fredrikekre:
How come?

KristofferC:
Because on 0.5 `Pkg.test` disables inlining because otherwise coverage got bad.

KristofferC:
https://github.com/JuliaLang/julia/blob/release-0.5/base/pkg/entry.jl#L715

fredrikekre:
Oh okay. Perhaps just run the notebooks on 0.6 then, consider them doctests which we also only run on 0.6?

KristofferC:
Yes, that sounds ok!

fredrikekre:
Easier to fix small things later than to do more on this branch imo

KristofferC:
Yes, I agree

PR #133, opened 2017-04-05
misc updates

fredrikekre:
- cache celldofs in the iterator
- show method for grid and fevalues to increase quality of life in notebook/repl: This

```julia
julia> grid = generate_grid(Hexahedron, (20, 20, 20))
JuAFEM.Grid{3,8,Float64,6} with 8000 Hexahedron cells and 9261 nodes
```
instead of this:
```jl
julia> grid = generate_grid(Hexahedron, (20, 20, 20))
JuAFEM.Grid{3,8,Float64,6}(JuAFEM.Cell{3,8,6}[JuAFEM.Cell{3,8,6}((1, 2, 23, 22, 442, 443,
464, 463)), JuAFEM.Cell{3,8,6}((2, 3, 24, 23, 443, 444, 465, 464)), JuAFEM.Cell{3,8,6}((3,
4, 25, 24, 444, 445, 466, 465)), JuAFEM.Cell{3,8,6}((4, 5, 26, 25, 445, 446, 467, 466)),
...
```


src/FEValues/common_values.jl
```diff
@@ -344,3 +344,7 @@ function spatial_coordinate{dim, T}(fe_v::Values{dim}, q_point::Int, x::Abstract
     end
     return vec
 end
+
+function Base.show(io::IO, fe_v::Values)
+    println(io, "$(typeof(fe_v)) with $(getnbasefunctions(fe_v)) base functions")
```
fredrikekre: dont really know whats interesting to show here...

src/iterators.jl
```diff
@@ -71,6 +73,10 @@ function reinit!{dim, N}(ci::CellIterator{dim, N}, i::Int)
         ci.nodes[j] = nodeid
         ci.coords[j] = ci.grid.nodes[nodeid].x
     end
+    if isdefined(ci, :dh) # update celldofs
+        resize!(ci.celldofs, ndofs_per_cell(ci.dh))
```
fredrikekre: perhaps we can put it to the correct size from the beginning?

src/Dofs/DirichletBoundaryConditions.jl
```diff
@@ -1,9 +1,9 @@
-export DirichletBoundaryConditions, update!, apply!, apply_zero!, add!
+export DirichletBoundaryConditions, update!, apply!, apply_zero!, add!, free_dofs
 
 """
     DirichletBoundaryConditions
 
-A dirichlet boundary conditions is a a boundary where a dof is fixed to take a certain value.
+A Dirichlet boundary condition is a boundary where the function is fixed to take a certain value.
```
KristofferC: the solution?

src/FEValues/common_values.jl
KristofferC: perhaps the reference shape?

src/iterators.jl
KristofferC: Yes, why not just `zeros(Int, ndofs_per_cell(dh))` in the constructor?

fredrikekre: that works as long as we have the same dofs in all elements :)

src/FEValues/common_values.jl
fredrikekre: sure

fredrikekre: Ideally we should show the interpolation but thats gone in fevalues now. Any way to retrieve it?

src/iterators.jl
KristofferC: I guess... Do we have any thoughts to not have that. Even with different interpolation of fields there would still be the same number of dofs per element.

src/FEValues/common_values.jl
KristofferC: store it in a string or something..

fredrikekre: reference shape is actually shown from the type:
```julia
julia> CellScalarValues(qr, ip)
JuAFEM.CellScalarValues{2,Float64,JuAFEM.RefCube} with 4 base functions
```

fredrikekre: Something like

```julia
julia> CellScalarValues(qr, ip)
JuAFEM.CellScalarValues{2,Float64,JuAFEM.RefCube} with 4 shape functions and 4 quadrature points
```

src/iterators.jl
KristofferC: Yeah.. but everything we have already breaks for that :P

examples/hyperelasticity.ipynb
```diff
@@ -427,6 +425,7 @@
   }
  ],
  "metadata": {
+  "anaconda-cloud": {},
```
fredrikekre: Okej.

fredrikekre:
good to go?

PR #135, opened 2017-04-07
use AD to compute shape derivatives

KristofferC:


src/JuAFEM.jl
```diff
@@ -3,6 +3,7 @@ __precompile__()
 module JuAFEM
 using FastGaussQuadrature
 using Compat
+using ForwardDiff
```
fredrikekre: Why can we not use Tensors only here?

src/interpolations.jl
```diff
@@ -115,7 +128,7 @@ getnbasefunctions(::Lagrange{1, RefCube, 1}) = 2
 getnfacenodes(::Lagrange{1, RefCube, 1}) = 1
 getfacelist(::Type{Lagrange{1, RefCube, 1}}) = ((1,),(2,))
 
-function value!(ip::Lagrange{1, RefCube, 1}, N::AbstractVector, ξ::Vec{1})
+function value!(ip::Lagrange{1, RefCube, 1}, N::AbstractVector, ξ::AbstractVector)
```
fredrikekre: is the quadpoints not Vec anymore?

KristofferC: They are but not when ForwardDiff calls this function

fredrikekre: Right. If we called `value!` for each independent shape function we could have used AD from `Tensors`, right?

PR #136, opened 2017-04-07
use GaussQuadrature instead of FastGaussQuadrature

KristofferC:


fredrikekre:
More like `SlowGaussQuadrature`<sup>kappa<sup>

PR #137, opened 2017-04-07
Notebook cleanup

fredrikekre:
strip all metadata from the notebooks using a git filter.

To enable, run `git config --local include.path ../.gitconfig` in the base directory

also use the new symmetric assembling in the cantilever notebook


examples/cantilever.ipynb
```diff
@@ -281,8 +215,8 @@
     "                δu = shape_value(cellvalues, q_point, i)\n",
     "                fe[i] += (δu ⋅ b) * dΩ\n",
     "                ɛC = ɛ[i] ⊡ C\n",
-    "                for j in 1:n_basefuncs\n",
-    "                    Ke[i, j] += (ɛC ⊡ ɛ[j]) * dΩ\n",
+    "                for j in i:n_basefuncs # assemble only upper half\n",
+    "                    Ke.data[i, j] += (ɛC ⊡ ɛ[j]) * dΩ # faster to assemble into the parent matrix\n",
```
fredrikekre: it allocates some memory wrapping `Ke` in `Symmetric` when its sent into `assemble`, so doing this..

KristofferC: Can't you wrap it once and for all in the beginning? Or will it be slower to locally assemble into the symmetric wrapper?

fredrikekre: I do wrap it once now. The only thing with `.data` here is to avoid the `uplo` check in the `Symmetric` `setindex!` so I guess that’s no cost

KristofferC: Ok, sounds good.

examples/cantilever.ipynb
```diff
@@ -216,7 +216,7 @@
     "                fe[i] += (δu ⋅ b) * dΩ\n",
     "                ɛC = ɛ[i] ⊡ C\n",
     "                for j in i:n_basefuncs # assemble only upper half\n",
-    "                    Ke.data[i, j] += (ɛC ⊡ ɛ[j]) * dΩ # faster to assemble into the parent matrix\n",
```
KristofferC: Did it make a difference in timing?

fredrikekre: dunno will check later

PR #139, opened 2017-04-07
always add diagonal elements

fredrikekre:
think this is safer...


fredrikekre:
oh.... the ci skipped last commit in #137 actually fails... `setindex!` can only operate on the diagonal as of 0.6....

KristofferC:
Well thats stupid...

KristofferC:
https://github.com/JuliaLang/julia/pull/19228/files#r110451034

PR #141, opened 2017-04-09
use Tensors.jl AD instead

fredrikekre:
will probably conflict alot with other open PR's, but this is kind of neat IMO


KristofferC:
Cool, I like it.

PR #144, opened 2017-04-25
link to the example notebooks from the docs

fredrikekre:


fredrikekre:
`nbviewer` is very nice. I like the button `View as Code` which just gives a script of the notebook!

http://nbviewer.jupyter.org/github/KristofferC/JuAFEM.jl/blob/master/examples/heat_square.ipynb

KristofferC:
Cool

PR #151, opened 2017-08-17
export cellid

fredrikekre:
Now I can remove 
```julia
@eval JuAFEM begin
    export cellid
end
```
from my code. Kappa.


fredrikekre:
StaticArrays screwing up nightly.

PR #152, opened 2017-09-07
[WIP] "Linear Assembler"

fredrikekre:
Assemble into `K.nzval` directly.

Missing tests, benchmarks, API, etc etc...

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -267,8 +267,9 @@ function _create_linear_sparsity_pattern(dh::DofHandler, sym::Bool)
     return sparse(I, J, W, maximum(I), maximum(J)), dict
 end
 
-function assemble!(a::LinearAssembler, ke::AbstractMatrix, edof::AbstractVector)
-    for dofj in edof, dofi in edof
-        a.K.nzval[a.d[(dofi, dofj)]] += ke[dofi, dofj]
+function assemble!(a::LinearAssembler, ke::AbstractMatrix, dofs::AbstractVector)
+    _ndofs = length(dofs)
+    @inbounds for dofj in 1:_ndofs, 1:_ndofs
+        a.K.nzval[a.d[(dofs[dofi], dofs[dofj])]] += ke[dofi, dofj]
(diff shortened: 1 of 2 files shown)


src/assembler.jl
```diff
@@ -209,3 +209,66 @@ function InsertionSort!(A, order, ii=1, jj=length(A))
...
+        push!(I, d)
+        push!(J, d)
+        push!(V, (d,d))
+    end
+    K = sparse(I, J, V, maximum(I), maximum(J), (x,y) -> x)
+
+    # loop again, and construct the map from (dofi, dofj)
+    # to a linear index into K.nzval
+    dict = Dict{Tuple{Int,Int},Int}()
+    for element_id in 1:ncells
```
KristofferC: Why a dict and not a matrix?

fredrikekre: Hmm... Interesting. A SparseMatrix that gives the index to another one then? :)

fredrikekre: But then this has no advantage over simply `K[i,j] += ...`?

KristofferC: You don't have to binary search?

fredrikekre: You don't think a hash lookup in a dict is faster than finding the correct index in a SparseMatrixCSC?

src/assembler.jl
```diff
@@ -209,3 +209,67 @@ function InsertionSort!(A, order, ii=1, jj=length(A))
...
+        celldofs!(global_dofs, dh, element_id)
+        @inbounds for j in 1:n, i in 1:n
+            dofi = global_dofs[i]
+            dofj = global_dofs[j]
+            sym && (dofi > dofj && continue)
+            dict[(dofi, dofj)] = findfirst(K.nzval, (dofi, dofj))
```
fredrikekre: This line takes forever...

fredrikekre:
This is garbage, right?

KristofferC:
Think so :P 🗑

PR #159, opened 2017-11-17
new dofhandler

### LINKED ISSUE #153: DofHandler does't account for higher order basis functions
In the following setting
```
dim = 2
grid = generate_grid(Quadrilateral, (1, 1))
cellvalues = CellScalarValues(qr, ip)
```
results in
```
JuAFEM.CellScalarValues{2,Float64,JuAFEM.RefCube} with 9 shape functions and 4 quadrature points
```
so i would expect the DofHandler to produce a 9x9 sparsity pattern, but it doesn't have the information for recognizing that.

fredrikekre:
fix #153 

TODO:
- [x] ~Order on faces in 3D~ leave for the future
- [x] update DirichletBoundaryConditions
- [x] Tests
- [x] Documentation


KristofferC:
Cool stuff.

Though, I wonder if we have to (for performance reasons) have the loop over the different field outermost and put the rest inside a function as a barrier. The different fields have different types so won't we need to shield the inner stuff from that?

fredrikekre:
> Though, I wonder if we have to (for performance reasons) have the loop over the different field outermost and put the rest inside a function as a barrier. The different fields have different types so won't we need to shield the inner stuff from that?

Yes, I thought so too, should be simple enough to just lift that loop out

KristofferC:
Ref https://www.dealii.org/8.5.0/doxygen/deal.II/namespaceDoFRenumbering.html

src/Dofs/DofHandler2.jl
```diff
@@ -0,0 +1,225 @@
...
+    return (a, b, c)
+end
+
+# close the DofHandler and distribute all the dofs
+function close!(dh::DofHandler{dim}) where {dim}
+    @assert !isclosed(dh)
+
+    # `vertexdict` keeps track of the visited vertices. We store the global vertex
+    # number and the first dof we added to that vertex.
+    vertexdict = ((Dict{Int,Int}() for _ in 1:nfields(dh))...)
```
fredrikekre: I guess these should really be vectors since the length of these tuples will be value dependent

src/Dofs/DofHandler2.jl
```diff
@@ -0,0 +1,225 @@
...
+    # celldofs are never shared between different cells so there is no need
+    # for a `celldict` to keep track of which cells we have added dofs too.
+
+    nextdof = 1 # next free dof to distribute
+    push!(dh.cell_dofs_offset, 1) # dofs for the first cell start at 1
+
+    # loop over all the cells in the grid
+    for (ci, cell) in enumerate(getcells(dh.grid))
+        @debug println("cell #$ci")
+        for (fi, interpolation) in enumerate(dh.field_interpolations)
```
fredrikekre: Will put a function barrier just after this `for`

src/Dofs/DofHandler2.jl
KristofferC: Yes

src/Dofs/DofHandler2.jl
```diff
@@ -0,0 +1,225 @@
+
+# struct Field
+#     name::Symbol
+#     interpolation::Interpolation
+#     dim::Int
+# end
+
+mutable struct DofHandler{dim,N,T,M}
```
lijas: What happend to `dofs_nodes::Matrix{Int}` (aka DOF in calfem-world)

KristofferC:
We probably need to come up with something for exporting the fields to VTK.

src/Dofs/DofHandler2.jl
KristofferC: That thing doesn't actually make much sense. Nothing says that all nodes need a dof, nothing says that there are no dofs at locations with no nodes, and nothing says that all nodes have equal number of dofs.

Dofs are something that is defined on the element level. When exporting data for visualization, it is common to give the field values in the nodes, but that is a post processing step.

src/Dofs/DofHandler2.jl
```diff
@@ -0,0 +1,245 @@
...
+    # and the interpolations inside without using a function barrier
+    #
+    # dofinfo =
+    # | nvertexdofs(field#1) | nvertexdofs(field#2) | ...
+    # | nedgedofs(field#1)   | nedgedofs(field#2)   | ...
+    # | nfacedofs(field#1)   | nfacedofs(field#2)   | ...
+    # | ncelldofs(field#1)   | ncelldofs(field#2)   | ...
+
+    dofinfo = fill(0, 4, nfields(dh))
+    for (i, interpolation) in enumerate(dh.field_interpolations)
```
fredrikekre: `interpolation` here is the only non-inferred variable that shows up for a `@code_warntype close!(dh)` call. Which should be OK (?).

fredrikekre:
> We probably need to come up with something for exporting the fields to VTK.

I think it is OK to only export the value in the grid nodes (which should be easy to implement). We can always add more fancy exporting later.

src/Dofs/DofHandler2.jl
KristofferC: Yes, that should be fine.

fredrikekre:
To distribute displacement and pressure dofs an a mesh with 20 000 elements: `0.023613 seconds (132 allocations: 8.302 MiB)`

src/Dofs/DofHandler2.jl
lijas: I thought dofs was defined on nodes, not on the elements, since we get the results in the nodes

lijas: I thought dofs was defined on nodes, not on the elements, since we get the results in the nodes

KristofferC:
I guess a good metric to compare it against is the cost of solving the linear system.

src/Dofs/DirichletBoundaryConditions2.jl
```diff
@@ -0,0 +1,342 @@
...
+```
+
+Dirichlet boundary conditions are added to certain components of a field for a specific nodes of the grid.
+A function is also given that should be of the form `(x,t) -> v` where `x` is the coordinate of the node, `t` is a
+time parameter and `v` should be of the same length as the number of components the bc is applied to:
+
+```jldoctest
+julia> addnodeset!(grid, "clamped", x -> norm(x[1]) ≈ 0.0);
+
+julia> nodes = grid.nodesets["clamped"]
```
KristofferC: Should update / delete this at some point

src/Dofs/DirichletBoundaryConditions2.jl
```diff
@@ -0,0 +1,342 @@
...
+
+isclosed(dbcs::DirichletBoundaryConditions) = dbcs.closed[]
+free_dofs(dbcs::DirichletBoundaryConditions) = dbcs.free_dofs
+prescribed_dofs(dbcs::DirichletBoundaryConditions) = dbcs.prescribed_dofs
+
+function close!(dbcs::DirichletBoundaryConditions)
+    fdofs = setdiff(1:ndofs(dbcs.dh), dbcs.prescribed_dofs)
+    copy!!(dbcs.free_dofs, fdofs)
+    # TODO: I *think* it is allowed to sort!(dbcs.prescribed_dofs) here,
+    # then we should be able to utilize that in apply! and friends
```
fredrikekre: Any reason this would not work?

KristofferC:
This is sick! Huge milestone.

src/Dofs/DirichletBoundaryConditions2.jl
```diff
@@ -0,0 +1,342 @@
...
+    for face in faces(func_interpol)
+        dofcoords = Vec{dim,T}[]
+        for facedof in face
+            push!(dofcoords, interpolation_coords[facedof])
+        end
+        qrf = QuadratureRule{dim,refshape,T}(fill(T(NaN), length(dofcoords)), dofcoords) # weights will not be used
+        push!(qrs, qrf)
+    end
+
+    # TODO: The rest is just to fill up the FaceScalarValues, can be removed
```
fredrikekre: We can have a separate struct for this maybe, all that we need is `M`

test/test_facevalues.jl
```diff
@@ -59,11 +59,12 @@ for (func_interpol, quad_rule) in  (
             for i in 1:getnquadpoints(fv)
                 vol += getdetJdV(fv,i)
             end
-            x_face = coords_on_faces(xs, func_interpol)[face]
+            # x_face = coords_on_faces(xs, func_interpol)[face]
```
fredrikekre: rm

src/Dofs/DirichletBoundaryConditions2.jl
KristofferC: It should work, although the gain is probably dwarfed by the amount of work that's done in `apply!`

src/Dofs/DirichletBoundaryConditions2.jl
KristofferC: Yeah, seems reasonable when so much is just garbage to actually create the struct.

src/Dofs/DirichletBoundaryConditions2.jl
fredrikekre: Right, but since we sort there anyway we might as well just do it here instead.

src/Dofs/DirichletBoundaryConditions2.jl
fredrikekre: `BoundaryConditionValues` incoming.

examples/hyperelasticity.ipynb
```diff
@@ -341,7 +343,8 @@
     }
    ],
    "source": [
-    "Base.Test.@test norm(u) ≈ 4.870833706518008\n",
+    "# Base.Test.@test norm(u) ≈ 4.870833706518008\n",
+    "Base.Test.@test norm(u) ≈ 4.856506220915417 # HEHE\n",
```
fredrikekre: This makes me a bit uneasy. Worth investigating more I think.

fredrikekre:
yolo

PR #162, opened 2017-11-22
Add the option of selecting cells having any point satisfying a condition, not just all points

mohamed82008:
So that not all points have to satisfy the condition function, can also select cells with any point having some feature.

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -120,7 +120,7 @@ function addcellset!(grid::Grid, name::String, cellid::Union{Set{Int}, Vector{In
     grid
 end
 
-function addcellset!(grid::Grid, name::String, f::Function; all = true)
+function addcellset!(grid::Grid, name::String, f::Function; all::Bool = true)
     cells = Set{Int}()
     for (i, cell) in enumerate(getcells(grid))
         pass = all
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -120,15 +120,16 @@ function addcellset!(grid::Grid, name::String, cellid::Union{Set{Int}, Vector{In
     grid
 end
 
-function addcellset!(grid::Grid, name::String, f::Function)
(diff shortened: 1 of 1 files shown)


src/Grid/grid.jl
```diff
@@ -120,15 +120,16 @@ function addcellset!(grid::Grid, name::String, cellid::Union{Set{Int}, Vector{In
     grid
 end
 
-function addcellset!(grid::Grid, name::String, f::Function)
+function addcellset!(grid::Grid, name::String, f::Function; all = true)
```
fredrikekre: `all::Bool=true`

src/Grid/grid.jl
```diff
@@ -120,15 +120,16 @@ function addcellset!(grid::Grid, name::String, cellid::Union{Set{Int}, Vector{In
...
     cells = Set{Int}()
     for (i, cell) in enumerate(getcells(grid))
-        all_true = true
+        pass = all
         for node_idx in cell.nodes
             node = grid.nodes[node_idx]
-            !f(node.x) && (all_true = false; break)
+            all && !f(node.x) && (pass = false; break)
```
fredrikekre: Perhaps store `f(node.x)` in a variable instead of calling it twice?

mohamed82008: Sure will do it tomorrow, thanks.

mohamed82008: But won't it be called only once per iteration? Because `all && ...` and `!all && ...` are like if statements.

mohamed82008: Actually storing `f(node.x)` is still useful to avoid calling it multiple times for different cells sharing the same node.

fredrikekre: Right, this should be free anyway so doesn't really matter.

fredrikekre:
I included this commit in #174, thanks!

PR #163, opened 2017-11-22
improve perf for apply!

KristofferC:
Found out during discussion with @PetrKryslUCSD that applying the boundary conditions is slow. It is still slower than I like but significantly faster. Perhaps `zero_out_rows!` can be improved.

Edit: Wups, need to break if i goes OOB.


PetrKryslUCSD:
If I'm reading this right, the new functions are specialized to a particular storage  pattern? (CCS?)

KristofferC:
Yes, we use CSC.

PR #170, opened 2017-12-08
Moved field-variables in DofHandler to struct Field

lijas:


src/Dofs/DofHandler.jl
```diff
@@ -42,29 +40,30 @@ end
...
+    end
+    error("did not find field $field_name")
 end
 
 # Calculate the offset to the first local dof of a field
 function field_offset(dh::DofHandler, field_name::Symbol)
     offset = 0
     for i in 1:find_field(dh, field_name)-1
-        offset += getnbasefunctions(dh.field_interpolations[i]) * dh.field_dims[i]
+        offset += getnbasefunctions(f.interpolation) * dh.fields[i].dim
```
fredrikekre: Test failure here: `f not defined`

lijas: Fixed

src/Dofs/DofHandler.jl
```diff
@@ -126,23 +126,23 @@ function close!(dh::DofHandler{dim}) where {dim}
     # loop over all the cells, and distribute dofs for all the fields
     for (ci, cell) in enumerate(getcells(dh.grid))
         @debug println("cell #$ci")
-        for fi in 1:nfields(dh)
-            interpolation_info = interpolation_infos[fi]
-            @debug println("  field: $(dh.field_names[fi])")
+        for (i,fi) in enumerate(dh.fields)
```
fredrikekre: Hmmm, I think this is why I didn't use a `Field` struct before, this loop will now be type unstable because all fields in `dh.fields` are not of the same type...

lijas: Maybe it´s ok since `close` is only called in the initialization of the problem? An other possibility is to add `dim` and `name` of the field to `InterpolationInfo` so we don´t need to loop over `dh.fields` at all.

src/interpolations.jl
```diff
@@ -84,9 +84,13 @@ struct InterpolationInfo
     nedgedofs::Int
     nfacedofs::Int
     ncelldofs::Int
-    InterpolationInfo(interpolation::Interpolation) =
+    #Not really interpolation info, but needed when 
+    #distributing dofs
+    name::Symbol
+    dim::Int
+    InterpolationInfo(interpolation::Interpolation, name::Symbol, dim::Int) =
```
lijas: Maybe change name from InterpolationInfo to FieldInfo?

lijas: Maybe change name from InterpolationInfo to FieldInfo?

lijas:
Old PR / Replaced by other PRs

PR #171, opened 2017-12-09
loop over cellset using CellIterator

lijas:
Loop over a cell set, for example

`CellIterator(dh, getcellset(grid, "mycellset"))`

wip


lijas:
Implemented in other PRs

PR #179, opened 2018-03-18
rename DirichletBoundaryConditions to ConstraintHandler

fredrikekre:
Matches `DofHandler` more nicely, and is more general if we extend with other type of constraints.

### DIFF
--- a/docs/src/examples/heat_equation.jl
+++ b/docs/src/examples/heat_equation.jl
@@ -99,10 +99,10 @@ K = create_sparsity_pattern(dh)
 # ```
 
 # ### Boundary conditions
-# In JuAFEM boundary Dirichlet boundary conditions are handled by
-# `DirichletBoundaryConditions`.
-dbcs = DirichletBoundaryConditions(dh)
-# Next we need to add contraints to `dbcs`. For this problem we define
+# In JuAFEM constraints like Dirichlet boundary conditions
+# are handled by a `ConstraintHandler`.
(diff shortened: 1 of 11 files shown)


KristofferC:
Sounds good!

lijas:
Dbcs and other constraints are treated quite differently, so maybe it will be difficult to implement all constraint in a single ConstraintHandler?

fredrikekre:
The current situation is just a bit awkward IMO:
```julia
dbcs = DirichletBoundaryConditions(dh)
dbc = DirichletBoundaryCondition(:u, faceset, (x, t) -> 0)
add!(dbcs, dbc)
```
Perhaps we can go back to the pre-#159 style, but rename `DirichletBoundaryConditions` to `Dirichlet`. That would look like:
```julia
dbc = Dirichlet(dh)
add!(dbc, :u, faceset, (x, t) -> 0)    <- We don't create a DirichletBoundaryCondition here, only internally
```

fredrikekre:
But something like
```julia
ch = ConstraintHandler(dh)
dbc = Dirichlet(:u, faceset, (x, t) -> 0)
add!(ch, dbc)
```
Looks very nice...

fredrikekre:
Interesting failure only Windows64-bit:
```
 !! failed to run `@setup` block.
LoadError("string", 1, ErrorException("syntax: invalid escape sequence")) [src\examples\heat_equation.md]
```

PR #180, opened 2018-03-18
rename DirichletBoundaryCondition to Dirichlet 

fredrikekre:


src/Dofs/ConstraintHandler.jl
```diff
@@ -1,64 +1,53 @@
...
-julia> addnodeset!(grid, "clamped", x -> norm(x[1]) ≈ 0.0);
-
-julia> nodes = grid.nodesets["clamped"]
-
-julia> push!(dbc, :temperature, nodes, (x,t) -> t * [x[2], 0.0, 0.0], [1, 2, 3])
+    Dirichlet(u, ∂Ω, f)
+    Dirichlet(u, ∂Ω, f, component)
+
+Create a Dirichlet boundary condition on `u` on the `∂Ω` part of
+the boundary. `f` is a function that takes two arguments, `x` and `t`
```
KristofferC: Actually, I guess `f` could be just a function of `x` and optionally you could close over a `t` if you wanted. This means `update!` wouldn't take a time either, you would just update `t`. Can do this later though, if we want.

src/Dofs/ConstraintHandler.jl
```diff
@@ -1,64 +1,53 @@
...
-julia> update!(dbc, t)
-```
-
-The boundary conditions can be applied to a vector:
-
-```jldoctest
-julia> u = zeros(ndofs(dh))
-
-julia> apply!(u, dbc)
+```julia
```
KristofferC: Add a way of specifying multiple components with a list?

PR #181, opened 2018-03-18
remove the old style assembler

fredrikekre:


fredrikekre:
Hmm, we use this for assembling just the residual e.g. `assemble!(g::Vector{Float64}, ge::::Vector{Float64}, edof::Vector{Int})`. Worth keeping that method perhaps? It's essentially just a wrapper around `g[edof] += ge` but...

fredrikekre:
Ah, it was used internally too, added it back.

fredrikekre:
I think this can be useful in some situations so we can keep this. Perhaps we can rename it though, and use `Assembler` as an abstract type?

PR #182, opened 2018-03-18
[RFC] replace start_assemble with methods of abstract type Assembler

fredrikekre:


samuelpowell:
I prefer this API, are you likely to proceed with this change?

fredrikekre:
Sure, why not. I think there was a reason I didn't follow through, but can't remember why now....

PR #183, opened 2018-03-19
clean up testfiles

fredrikekre:
Cleans up the `.vtu` files that we create when running examples.


KristofferC:
Should do `mktempdir() do tmp; cd(tmp) do` to clean up the files afterwards.

fredrikekre:
> Should `do mktempdir() do tmp; cd(tmp) do` to clean up the files afterwards.

Yea, just thought about that.

PR #185, opened 2018-03-31
colon of steprange vs ternary op fix

louisponet:


KristofferC:
This means we have no test / example with an element with edge dofs? Guess we should add that :)

fredrikekre:
Thanks!

> This means we have no test / example with an element with edge dofs? Guess we should add that :)

Yes, this code is sparsely tested, will submit a PR with better testing soon™.

PR #186, opened 2018-04-01
add an example computing the potential and using AD to get gradient and hessian

KristofferC:
Quite a lot slower, hm.

### DIFF
--- a/examples/hyperelasticity.ipynb
+++ b/examples/hyperelasticity.ipynb
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


examples/hyperelasticity.ipynb
termi-official, line 1: Is there any advantage here over the current implementation?

termi-official:
This is essentially the same as https://ferrite-fem.github.io/Ferrite.jl/stable/gallery/landau/ , but for elasticity, right? Should we close this then?

PR #187, opened 2018-04-06
wip add dof range to compute range for a field

KristofferC:


src/Dofs/DofHandler.jl
```diff
@@ -59,6 +59,13 @@ function field_offset(dh::DofHandler, field_name::Symbol)
     return offset
 end
 
+function dof_range(dh::DofHandler, field_name::Symbol)
+    f = find_field(dh, field_name)
+    offset = field_offset(dh, field_name)
+    n_field_dofs = getnbasefunctions(dh.field_interpolations[f]) * dh.fieldn_dims[f]
```
fredrikekre: `fieldn_dims` -> `field_dims`

src/Dofs/DofHandler.jl
```diff
@@ -59,6 +59,13 @@ function field_offset(dh::DofHandler, field_name::Symbol)
...
+    return offset:n_field_dofs
```
fredrikekre: `(offset+1):n_field_dofs`

fredrikekre:
Is this better called `field_range` (like the branch name) rather than `dof_range`?

KristofferC:
You get the dof range for a given field. Seems good to me?

PR #190, opened 2018-04-11
Fix issue 189

mohamed82008:
https://github.com/KristofferC/JuAFEM.jl/issues/189


mohamed82008:
I should probably combine the 2 commits.

fredrikekre:
> I should probably combine the 2 commits.

They can be squashed when merging. Thanks for the fix!

mohamed82008:
Ok I will leave it to you then!

PR #200, opened 2018-04-20
use a build stage for doc build

fredrikekre:


KristofferC:
"ERROR: LoadError: LoadError: something went wrong when evaluating code"

:)

fredrikekre:
#227

PR #202, opened 2018-04-24
simple renumber

fredrikekre:
also test and make `dof_range` inferable.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -324,6 +324,26 @@ function _create_sparsity_pattern(dh::DofHandler, sym::Bool)
     return K
 end
 
+# dof renumbering
+"""
+    renumber!(dh::DofHandler, perm)
+
+Renumber the degrees of freedom in the DofHandler according to the
+permuation `perm`.
+
+!!! warning
+    Remember to do renumbering *before* adding boundary conditions,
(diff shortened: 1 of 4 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -59,7 +59,7 @@ end
 function field_offset(dh::DofHandler, field_name::Symbol)
     offset = 0
     for i in 1:find_field(dh, field_name)-1
-        offset += getnbasefunctions(dh.field_interpolations[i]) * dh.field_dims[i]
+        offset += getnbasefunctions(dh.field_interpolations[i])::Int * dh.field_dims[i]
```
fredrikekre: Even if `dh.field_interpolations[i]` here I find this weird -- all `getnbasefunctions` return `Int`... Would it be better to add a fallback `getnbasefunctions` that errors?

src/Dofs/DofHandler.jl
```diff
@@ -324,6 +324,39 @@ function _create_sparsity_pattern(dh::DofHandler, sym::Bool)
     return K
 end
 
+# dof renumbering
```
KristofferC: Can remove these comments?

KristofferC: Or is it for later?

PR #205, opened 2018-05-02
add Literate as test/REQUIRE and go back to default test script

fredrikekre:


fredrikekre:
Why did having Literate cloned prevent test/REQUIRE dependencies to be removed when checking out Documenter master?
https://github.com/KristofferC/JuAFEM.jl/blob/af763996748ea45b076f8cfbab1219d351758901/docs/make.jl#L1

PR #211, opened 2018-05-11
Fix tetrahedron grid generator

### LINKED ISSUE #210: generate_grid(Tetrahedron) results in bad grid
As far as I can tell the current generate_grid() for Tetrahedral elements in 3D results in discontinuous "hat" functions.

To see this, notice that for every hexahedron the line (1,0,0) to (1,1,1) passes through a face and separates two elements. But for the adjacent hexahedron, this line is within an element.  See also http://www.baumanneduard.ch/Splitting%20a%20cube%20in%20tetrahedras2.htm (which is cited in generate_grid which does a rotated version of #13): if you look at the image of #13, you'll see that the lines on adjacent cubes don't fit together.

haampie:
Fixes #210. Split cubes into 6 tetrahedrons that match faces when stacked in a larger cube.

I've found that most time is spent in calling `boundaries_to_sparse` by the way.

```
@benchmark generate_grid(Tetrahedron, (64, 64, 64)) # *without* constructing boundary_matrix
# median time:      17.135 ms (11.75% GC)

@benchmark generate_grid(Tetrahedron, (64, 64, 64)) # *with* constructing boundary_matrix
# median time:      80.319 ms (71.48% GC)
```

edit: tests failed previously because `det(J) < 0` in the change of variables -- I've changed it to use `abs(det(J))` now, but I'm sure there is some reason why `det(J) > 0` was enforced previously, so I'm willing to undo that commit if necessary and fix the problem elsewhere. Also, the notebook values are obviously not the same anymore, so they're updated as well. Someone should manually check however that the tet's make sense now!


haampie:
Hm, Github seems to have trouble with the order of the commits after rebase :P, but `fix notebook tests ...` should be the last one.

haampie:
Seems good to go now after the tests are green!

I could still add a test to verify that the faces of two adjacent tets match and that the facesets are correct etc.

KristofferC:
Looks good to me superficially but let's give Fredrik a chance to look over it.

haampie:
A slightly cleaner version is here: https://github.com/haampie/JuAFEM.jl/blob/feature-generate-tets-without-loops/src/Grid/grid_generators.jl#L322-L374 I could update this pr with it.

fredrikekre:
Thanks!

Issue #167, opened 2017-11-28
Export of super parametric elements


fredrikekre:
I don't *think* it works :trollface:.

fredrikekre:
```julia
julia> grid = generate_grid(QuadraticTriangle, (5, 5));

julia> dh = DofHandler(grid);

julia> push!(dh, :u, 1, Lagrange{2,RefTetrahedron,1}());

julia> close!(dh);

julia> u = collect(1:ndofs(dh));

julia> vtk = vtk_grid("test", grid);

julia> vtk_point_data(vtk, dh, u);
ERROR: BoundsError: attempt to access 3-element Array{Int64,1} at index [4]
Stacktrace:
```

KristofferC:
Always just export it as linear?

fredrikekre:
Yes, or if both fields are quadratic I guess we can export that

PR #212, opened 2018-05-14
make it possible to add bcs on nodesets

### LINKED ISSUE #166: Support adding boundary conditions on specific nodes
After #159 we add boundary conditions on facesets. It is sometimes needed to lock individual nodes, to hinder e.g. rotation. We should have a way to do this.

fredrikekre:
Alternative to #194 

This reuses the same struct, as I suggested in #194 https://github.com/KristofferC/JuAFEM.jl/pull/194/files#r185399881. I find this implementation a bit simpler.

cc @mohamed82008

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -28,16 +28,16 @@ which applies the condition via `apply!`.
 """
 struct Dirichlet # <: Constraint
     f::Function # f(x,t) -> value
-    faces::Set{Tuple{Int,Int}}
+    faces::Union{Set{Int},Set{Tuple{Int,Int}}}
     field_name::Symbol
     components::Vector{Int} # components of the field
     local_face_dofs::Vector{Int}
     local_face_dofs_offset::Vector{Int}
 end
-function Dirichlet(field_name::Symbol, faces::Set{Tuple{Int,Int}}, f::Function, component::Int=1)
+function Dirichlet(field_name::Symbol, faces::Union{Set{Int},Set{Tuple{Int,Int}}}, f::Function, component::Int=1)
     Dirichlet(field_name, faces, f, [component])
 end
-function Dirichlet(field_name::Symbol, faces::Set{Tuple{Int,Int}}, f::Function, components::Vector{Int})
+function Dirichlet(field_name::Symbol, faces::Union{Set{Int},Set{Tuple{Int,Int}}}, f::Function, components::Vector{Int})
     unique(components) == components || error("components not unique: $components")
     # issorted(components) || error("components not sorted: $components")
     return Dirichlet(f, faces, field_name, components, Int[], Int[])
@@ -104,7 +104,7 @@ function dbc_check(ch::ConstraintHandler, dbc::Dirichlet)
         0 < component <= ndim(ch.dh, dbc.field_name) || error("component $component is not within the range of field $field which has $(ndim(ch.dh, field)) dimensions")
     end
     if length(dbc.faces) == 0
-        warn("added Dirichlet Boundary Condition to face set containing 0 faces")
+        warn("added Dirichlet Boundary Condition to set containing 0 entities")
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -154,6 +158,34 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, interpolation::Interpolati
...
 
+function add_node_dbc!(ch::ConstraintHandler, dbc::Dirichlet, interpolation::Interpolation, field_dim::Int, offset::Int)
+    if interpolation !== default_interpolation(getcelltype(ch.dh.grid))
+        warn("adding constraint to nodeset is not recommended for sub/super-parametric approximations.")
+    end
+    constrained_dofs = Int[]
+    _celldofs = fill(0, ndofs_per_cell(ch.dh))
+    added_nodes = Set{Int}() # keep track of the nodes we have added to
+    for (cellidx, cell) in enumerate(ch.dh.grid.cells)
+        idx = findfirst(x -> in(x, dbc.faces) && !in(x, added_nodes), cell.nodes)
```
mohamed82008: The time complexity of this part is O(ncells * length(dbc.faces)), it is basically a masked nested loop. There is also one hidden assumption if I understand correctly which is that every cell can at most result in a single node being found/added. So when 2 boundary nodes belong to 1 and only 1 cell, 1 of them will be missed. Think of 2 nodes on the edge of a corner quadratic hexahedron. I know that quadratic hexahedrons are not supported but there is probably a better way to do this?

mohamed82008: I think `find` instead of `findfirst` will solve the second problem.

fredrikekre: > There is also one hidden assumption if I understand correctly which is that every cell can at most result in a single node being found/added. 

Yea, good catch, will fix this.

TBH I don't think the time for this will matter at all compared to other things, but I can do some benchmarks.

fredrikekre:
> I think `node_dofs` should be added to make it possible to query the dofs of a node at constant time

I don't like this, this does not generalize at all and only "works" for iso-parametric elements. Other than adding bcs, why do you need to query which dofs happen to be located at a given node?

mohamed82008:
>only "works" for iso-parametric elements.

I don't see what you mean here. I believe that JuAFEM assumes the order of function interpolation >= the order of geometric interpolation == the order of geometric cells. Given this assumption, and given that the additional support of points for interpolations of order `o+1` are always given indices starting from `n+1`, where `n` is the number of support points of the interpolation of order `o`, it should be entirely possible to extract the dofs of nodes for non-isoparametric elements, which is what my PR did. 

>Other than adding bcs, why do you need to query which dofs happen to be located at a given node? 

Currently no reason but in the future I imagine it would be useful for post-processing and visualization, exporting the displacement field for example to a vtk file. I think it makes sense to be able to query information about nodes without going through a linear time search for every node.

fredrikekre:
>  I believe that JuAFEM assumes the order of field interpolation >= the order of geometric interpolation == the order of geometric cells.

This is not true:
```julia
grid = generate_grid(QuadraticQuadrilateral, (1, 1))
dh = DofHandler(grid)
push!(dh, :u, 1, Lagrange{2,RefCube,1}())
close!(dh)
```
with
```julia
julia> ndofs(dh)
4

julia> getnnodes(grid)
9
```
meaning that all nodes does not have dofs.

> > Other than adding bcs, why do you need to query which dofs happen to be located at a given node?
> 
> Currently no reason but in the future I imagine it would be useful for post-processing and visualization, exporting the displacement field for example to a vtk file.

We already do this, and I think it is pretty fast, no?

mohamed82008:
Hmm, ok that was a misconception I had then. I tried to find the part of the code that made me think like that but couldn't find except the `function WriteVTK.vtk_point_data(vtkfile, dh::DofHandler, u::Vector)` in https://github.com/KristofferC/JuAFEM.jl/blob/fb3cc17a57f54ac89a137d80fa85328e082dda0b/src/Dofs/DofHandler.jl#L350 which understandably makes the above assumption and can be easily generalized. May I ask why are you completely against `node_dofs`, can't we have `NaN`s where nodes have no dofs. The `NaN` can also enable a more precise error message when trying to add a dbc to a node with no dof. 

Would it make it any better if the `node_dofs` is optionally populated when constructing the `DofHandler` or when adding node dbcs, but left empty by default? Your implementation can be the default, but when a user wants to have constant time complexity for adding a single node dbc, he/she can ask for the `node_dofs` to be computed in advance, and if computed it can be used every time after. I can find and use the `node_dofs` myself outside of JuAFEM, if I need them for anything other than adding BCs, so I think that's the main relevant use case for this discussion.

fredrikekre:
Yea, that function is bugged, see #167

> May I ask why are you completely against node_dofs



can't we have NaNs where nodes have no dofs. The NaN can also enable a more precise error message when trying to add a dbc to a node with no dof.
> 
> Would it make it any better if the node_dofs is optionally populated when constructing the DofHandler or when adding node dbcs, but left empty by default? Your implementation can be the default, but when a user wants to have constant time complexity for adding a single node dbc, he/she can ask for the node_dofs to be computed in advance, and if computed it can be used every time after.

mohamed82008:
Even if you are not storing the node dofs permanently, I think if you find them temporarily, with NaNs where no dofs are there, you can reduce the time complexity from O(ncells * length(dbc.faces)) to O(ncells + length(dbc.faces)), just avoid the nested loop. The space complexity will increase though from O(length(dbc.faces)) to O(ncells), so it's a compromise.

fredrikekre:
I did some timing and the time to search for the nodes are negligible compared to e.g. creating the DofHandler and assembling.

mohamed82008:
If you are not in a hurry, I will make another PR when I can to make this into linear time and memory complexity. Creating the dofhandler and assembling are all linear time so if you have enough nodes in the dbc, it will be the dominant cost.

PR #213, opened 2018-05-14
call update! with arbitrary arguments, fix  #207

### LINKED ISSUE #207: Having to send a time into the function for constraints is a design failure.

fredrikekre:
Actually non-breaking unless you rely on the default value of `t = 0.0`.


fredrikekre:
Perhaps only propagate kwargs?

KristofferC:
So all functions need to support exactly the same arguments / kwargs?

fredrikekre:
You can always define your functions as `(x, y, z, args...; kwargs...)`.

PR #215, opened 2018-06-01
Linear time and memory addition of node dbcs

mohamed82008:
The PR modifies https://github.com/KristofferC/JuAFEM.jl/pull/212 to make it of linear time complexity, but it has linear memory complexity ~~now, as opposed to the constant memory your PR takes~~.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -158,24 +158,40 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpo
     if interpolation !== default_interpolation(getcelltype(ch.dh.grid))
         warn("adding constraint to nodeset is not recommended for sub/super-parametric approximations.")
     end
-    constrained_dofs = Int[]
+
+    ncomps = length(dbc.components)
+    nnodes = getnnodes(ch.dh.grid)
+    interpol_points = ndofs_per_cell(ch.dh)
     _celldofs = fill(0, ndofs_per_cell(ch.dh))
-    added_nodes = Set{Int}() # keep track of the nodes we have added to
+    node_dofs = zeros(Int, ncomps, nnodes)
+    visited = BitVector(nnodes)
+    visited .= false
     for (cellidx, cell) in enumerate(ch.dh.grid.cells)
-        idx = findfirst(x -> in(x, bcnodes) && !in(x, added_nodes), cell.nodes)
-        while idx !== 0
-            push!(added_nodes, cell.nodes[idx]) # add to visited nodes
-            celldofs!(_celldofs, ch.dh, cellidx) # update the dofs for this cell
-            noderange = (offset + (idx-1)*field_dim + 1):(offset + idx*field_dim) # the dofs in this node
-            for d in 1:field_dim
-                if d ∈ dbc.components # skip unless this component should be constrained
-                    push!(constrained_dofs, _celldofs[noderange[d]])
-                    @debug println("adding dof $(_celldofs[noderange[d]]) to dbc")
+        celldofs!(_celldofs, ch.dh, cellidx) # update the dofs for this cell
+        for idx in 1:min(interpol_points, length(cell.nodes))
+            node = cell.nodes[idx]
+            if !visited[node]
+                noderange = (offset + (idx-1)*field_dim + 1):(offset + idx*field_dim) # the dofs in this node
+                for (i,c) in enumerate(dbc.components)
+                    node_dofs[i,node] = _celldofs[noderange[c]]
+                    @debug println("adding dof $(_celldofs[noderange[c]]) to node_dofs")
                 end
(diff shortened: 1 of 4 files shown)


mohamed82008:
This PR is ready, so please review it when you can.

KristofferC:
Would be interesting to see a timing before and after.

mohamed82008:
For a cantilever beam of width 10000 and height 500, i.e. 5M quads, the time of the `add` function went down from 0.69 sec to 0.22 sec on my machine.

src/Dofs/ConstraintHandler.jl
```diff
@@ -154,6 +154,52 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, interpolation::Interpolati
...
+        warn("adding constraint to nodeset is not recommended for sub/super-parametric approximations.")
+    end
+
+    ncomps = length(dbc.components)
+    nnodes = getnnodes(ch.dh.grid)
+    interpol_points = ndofs_per_cell(ch.dh)
+    _celldofs = fill(0, ndofs_per_cell(ch.dh))
+    node_dofs = zeros(Int, ncomps, nnodes)
+    visited = BitVector(nnodes)
+    visited .= false
```
fredrikekre: `falses(nnodes)` can replace those two lines.

src/Dofs/ConstraintHandler.jl
```diff
@@ -154,6 +154,52 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, interpolation::Interpolati
...
+            end
+        end
+    end
+
+    constrained_dofs = Int[]
+    sizehint!(constrained_dofs, ncomps*length(bcnodes))
+    sizehint!(dbc.local_face_dofs, length(bcnodes))
+    for node in bcnodes
+        if !visited[node]
+            throw("Unable to add a Dirichlet boundary condition to node $node as there are no degrees of freedom on this node.")
```
fredrikekre: `throw` -> `error`

src/Dofs/ConstraintHandler.jl
```diff
@@ -154,6 +154,52 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, interpolation::Interpolati
...
+    ncomps = length(dbc.components)
+    nnodes = getnnodes(ch.dh.grid)
+    interpol_points = ndofs_per_cell(ch.dh)
+    _celldofs = fill(0, ndofs_per_cell(ch.dh))
+    node_dofs = zeros(Int, ncomps, nnodes)
+    visited = BitVector(nnodes)
+    visited .= false
+    for (cellidx, cell) in enumerate(ch.dh.grid.cells)
+        celldofs!(_celldofs, ch.dh, cellidx) # update the dofs for this cell
+        for idx in 1:min(interpol_points, length(cell.nodes))
```
fredrikekre: Why the `min`? Don't we always want to loop over all the nodes in the cell?

src/Dofs/ConstraintHandler.jl
mohamed82008: Cool!

src/Dofs/ConstraintHandler.jl
mohamed82008: Sure.

src/Dofs/ConstraintHandler.jl
mohamed82008: When `interpol_points < length(cell.nodes)`, this will an avoid out-of-bounds error in `_celldofs[noderange[c]]`. And when `interpol_points > length(cell.nodes)`, it will just do the right thing.

src/Dofs/ConstraintHandler.jl
fredrikekre: We should throw an exception not a String :)

mohamed82008: True, but this works:
```julia
julia> throw("eee")
ERROR: "eee"
```

src/Dofs/ConstraintHandler.jl
fredrikekre: So this is for a case with e.g. quadratic geometry and linear function interpolation?

mohamed82008: Yes.

mohamed82008: So any node `i` without dofs will just have 0 in its `node_dofs[:,i]`.

fredrikekre: But `interpol_points` is just the number of dofs in the cell, not the "number of nodes with dofs", which is what you want here, right?

mohamed82008: Oh yes that's a mistake, oops!

mohamed82008: `getnbasefunctions(interpolation)` should do the right thing.

mohamed82008:
Any doubts about this?

PR #216, opened 2018-06-03
Experimental inp file importer

mohamed82008:
This PR extends #215 to add an experimental inp file importer. This PR requires a way to add nodal Dirichlet boundary conditions. The importer currently only supports the following features:

1. Homogeneous meshes
2. Single elastic material objects
3. Pressure on faces
4. Concentrated loads on nodes
5. Loads on faces
6. Zero and non-zero Dirichlet boundary conditions on nodes (and faces using FreeCAD which is written as node dbcs)
7. Cell sets
8. Node sets

It is easily extensible though. I am calling it experimental because its feature support is very limited and perhaps more tests won't hurt. Please let me know your feedback, the PR is ready.

### DIFF
--- a/src/Import/INP/FeatureExtractors/extract_cells.jl
+++ b/src/Import/INP/FeatureExtractors/extract_cells.jl
@@ -1,0 +1,47 @@
+function extract_cells(file, ::Type{TI}=Int) where TI
+    line = readline(file)
+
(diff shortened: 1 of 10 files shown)


fredrikekre:
This functionality now exist in https://github.com/Ferrite-FEM/FerriteMeshParser.jl

PR #219, opened 2018-06-18
Femtoclean

KristofferC:


dkarrasch:
The checks seemed to fail because, at the time, Tensors.jl had not been upgraded to Julia 0.7. Can we give this another try now that Tensors.jl is working on Julia 0.7?

fredrikekre:
Yes, I'll try to update this today

fredrikekre:
https://github.com/KristofferC/JuAFEM.jl/pull/222

PR #220, opened 2018-07-05
implement function_curl and shape_curl

fredrikekre:
I like how simple this is.

https://stackoverflow.com/questions/51125430/solve-poisson-equation-for-3d-mesh


src/FEValues/common_values.jl
```diff
@@ -83,6 +83,11 @@ quadrature point `q_point`.
 @inline shape_divergence(bv::FaceVectorValues, q_point::Int, base_func::Int) = trace(bv.dNdx[base_func, q_point, bv.current_face[]])
 
 
+function shape_curl(cv::VectorValues, q_point::Int, base_func::Int)
+    ∇v = shape_gradient(cv, q_point, base_func)
+    return Vec{3}((∇v[3,2] - ∇v[2,3], ∇v[1,3] - ∇v[3,1], ∇v[2,1] - ∇v[1,2]))
```
KristofferC: Perhaps lift this out in a function to not have the same code copy pasted three times?

PR #229, opened 2018-08-28
Landau

louisponet:
This is an example of how a Landau model can be solved performantly using JuAFEM, Optim, and ForwardDiff. 

For some reason there is 1 allocation that happens per element when calculating the gradient, namely in the gradient! calculation. I can't seem the find the reason why, I have almost exactly the same code running a more complex model where there are 0 allocations. For the life of me I can't find it, send help!

Let me know what you think!


dkarrasch:
I was impressed, and tried to run it. Seems like `alphatest` is not defined, and `using SparseArrays` in the beginning is required.

louisponet:
Woops, my bad, forgot to run everything from a new repl. Should work now!

KristofferC:
This is great to have. Thanks a lot of taking the time to do this!

docs/src/examples/landau.jl
```diff
@@ -0,0 +1,265 @@
...
+function JuAFEM.vtk_save(path, model, dofs=model.dofs)
+    vtkfile = vtk_grid(path, model.dofhandler)
+    vtk_point_data(vtkfile, model.dofhandler, dofs)
+    vtk_save(vtkfile)
+end
+
+# ## Assembly
+# This macro defines most of the assembly step, since the structure is the same for
+# the energy, gradient and Hessian calculations.
+macro assemble!(innerbody)
```
KristofferC: Out of curiosity, why couldn't this just be a function?

louisponet: Oh hmm no reason. I didn't think about just passing in an innerbody "function" to the assembly function. I'll do an experiment there shouldn't be a difference, maybe it makes the example a bit more accessible?

louisponet:
On a sidenote, how would one change this example to work with second order Lagrange interpolation? I can't seem to figure that out

docs/src/examples/landau.jl
KristofferC: But do you even have to pass in a function, can't you just call this function from within the assembly function?

KristofferC: Yeah, you are right, you probably need to pass a function into `assemble!` to do the "inner" part. We can leave as is, I just have a thing for trying to not use macros for as long as possible heh.

louisponet: Oh yea, I'll clean it up a bit later, I also noticed that this `CellCache` is actually not necessary using the function as a field of the `ThreadCache` also works. Before I had it since I thought for performance I could not store functions as fields, but had to do something like `(x::CellCache)(...)` as the function call for the gradient and hessian.

KristofferC:
Could we run this as one of the tests (by adding it to https://github.com/KristofferC/JuAFEM.jl/blob/master/test/test_examples.jl).

KristofferC:
I updated this and will merge for now, fixups can happen later.

PR #232, opened 2018-09-22
improve threading performance example

KristofferC:
For four threads improves performance by ~20% but for 16 threads improves performance by 3x.

This is likely because I am now allocating the scratch space on each separate thread which removes false sharing.


fredrikekre:
Subsumed by #1068.

PR #233, opened 2018-09-24
remove boundscheck on f when inbounds assembling

KristofferC:


fredrikekre:
KrylovMethods seems capped

KristofferC:
Should probably swap to IterativeSolvers

KristofferC:
Passes on 0.7 so perhaps good to go?

fredrikekre:
Did this actually change something? We have https://github.com/KristofferC/JuAFEM.jl/blob/6b340e6744cec28530e30f83b2fc907980524034/src/assembler.jl#L62-L63 already

KristofferC:
Yes but I don't think the call from where I made the change make it applies the inbounds. At least I saw allocations before this change but I can double check.

PR #243, opened 2019-06-02
Added an example implementation of a von Mises plasticity model. 

JimBrouzoulis:
Note: Code has not been optimized for speed (...yet).


KristofferC:
Pushed some small updates (updating Manifest, some fixes to performance etc).

Performance before / after

```
  7.405634 seconds (162.86 M allocations: 5.050 GiB, 22.96% gc time)
 
  0.576641 seconds (287.54 k allocations: 156.701 MiB, 5.59% gc time)
```

Something seems a bit odd with the rendered docs:

Screenshot 2019-06-03 at 08 26 30

Also the "continued" mode that puts comments as "free text"  doesn't look supergood here. Is it possible to disable that?
@fredrikekre.

docs/src/examples/plasticity.jl
```diff
@@ -0,0 +1,410 @@
...
+            end
+
+            apply_zero!(K, r, dbcs)
+            Δu = Symmetric(K) \ r
+            u -= Δu
+        end
+
+        # Update all the material states after we have reached equilibrium
+        for cell_states in states
+            foreach(update_state!, cell_states)
```
lijas: Maybe it is possible to make use of a `previous_states` variable which could be used for the equilibrium iterations. Then we could get rid of  `update_state!` and the `temp_` variables in `MaterialState`, but we would need to copy the state-vector each iteration...

KristofferC: Yes, the alternative would be to not store `temp_` stuff in the material state but instead have two arrays with states. The advantage is that the material state can then be made immutable.

lijas: But the state is modified in `compute_stress_tangent`, so it still has to be mutable?
Or should `compute_stress_tangent` return the `σ`, `D`` and a `new_state`?

fredrikekre: Just `new_state` would be sufficient, and you extract `σ` and `D` from the new state.

KristofferC: Yes, in that case you return `state, D`. I don't think `D` is needed to be stored.

JimBrouzoulis: So, would this necessarily lead to a speedup, or is this more a question of design?
Modifying a mutable struct vs. creating new immutable structs inside the material routine.

KristofferC: It would probably be a bit higher performance since in julia, immutable types can be stored inline in arrays (no pointer indirection).

I think the current way is fine though, we can always refine it later if we want to.

JimBrouzoulis:
> Something seems a bit odd with the rendered docs:
> 
> Screenshot 2019-06-03 at 08 26 30

I noticed that on my local build I don't get the last block (code rendered as markdown).

fredrikekre:
Works for me locally too, lets see what happens on deploy. Thank you!

KristofferC:
Seems like it is there in the deployed docs as well: http://kristofferc.github.io/JuAFEM.jl/dev/examples/generated/plasticity/.

fredrikekre:
Right, I missed that and forgot to fix it.

PR #244, opened 2019-06-04
Added example with large elasto-plastic deformations

lijas:
An example with large deformation plasticity. I basically copied the hyperelasticity problem but added an elasto-plastic material model.

Don't know if this example is needed since there already exists an example with plasticity. 

Do you want the example to be in a ipynb-format?
Can someone review the code to see where efficiency improvements can be made?

(I have no way in checking if the results is correct, but it looks reasonable in paraview.)

### DIFF
--- a/examples/elasto_plastic.jl
+++ b/examples/elasto_plastic.jl
@@ -1,0 +1,300 @@
+using JuAFEM
+using Tensors
+using KrylovMethods
+using TimerOutputs
+using Statistics: mean
+import ProgressMeter
+const ∇ = Tensors.gradient
+
+struct Material{T}
+    μ::Float64		#Lamé constant
+    λ::Float64		#Lamé constant
+    τ₀::Float64		#Yield stress
+    H::Float64		#Hardening
+end
+
+struct MaterialState{T}
+	S::SymmetricTensor{2,3,T,6} #2nd PK stress
+    ϵᵖ::Float64					#Plastic strain
+    F⁻¹::Tensor{2,3,T,9}		#Plastic deformation grad
+    ν::Tensor{2,3,T,9}			
+end
+
+function MaterialState()
+    return MaterialState(
+    				zero(SymmetricTensor{2,3,Float64,6}),
+    				0.0, 
+    	            one(Tensor{2,3,Float64,9}), 
+    	            zero(Tensor{2,3,Float64,9}),)
+end
+
+function _compute_2nd_PK(mp::Material, E, state::MaterialState)
+    μ, λ, τ₀, H = mp.μ, mp.λ, mp.τ₀, mp.H
+    
+    ϵᵖ = state.ϵᵖ
+    F⁻¹ = state.F⁻¹
+    ν = state.ν
+    dγ = 0 
+
+    C = 2E + one(E)
+
+    phi, Mdev, Stilde = ϕ(C, F⁻¹, ϵᵖ, μ, λ, τ₀, H)
+
+    if phi > 0 #Plastic step
+        dγ = 0.0
+        
+        #Setup newton varibles
+        TOL = 0.00001
+        newton_error, counter = TOL + 1, 0
+
+        # Define residual for semi-emplicit backward euler (assuming ⁿ⁺¹ν = ⁿν)
+        ϕ_residual(dγ) = ϕ(C, F⁻¹ - dγ*F⁻¹⋅ν, ϵᵖ + dγ, μ, λ, τ₀, H)[1]
+
+        while newton_error > TOL; counter +=1 
+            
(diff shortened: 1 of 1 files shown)


examples/elasto_plastic.jl
```diff
@@ -0,0 +1,300 @@
...
+
+        while newton_error > TOL; counter +=1 
+            
+            res = ϕ_residual(dγ)
+
+            #Not able to take Automatic derivative of ϕ_residual for some reaseon
+            #Use numerical derivative for simplicity
+
+            #jac = ForwardDiff.derivative(ϕ_residual, dγ)
+            jac = (ϕ_residual(dγ + 0.00001) - ϕ_residual(dγ))/(0.00001)
```
lijas: I am not able to use ` ForwardDiff.derivative` here because the `_compute_2nd_PK` is also callen with `Tensors.gradient`. Any way to fix this?

examples/elasto_plastic.jl
```diff
@@ -0,0 +1,300 @@
...
+    yield_func = sqrt(3/2)*norm(Mdev) - (τ₀ + H*ϵᵖ)
+    
+    return yield_func, Mdev, Stilde
+end
+
+function constitutive_driver(mp::Material, E, state::MaterialState)
+
+	#Is it possible to combine the computation for S and ∂S∂E (using auto diff?)
+    S, ϵᵖ, ν, F⁻¹ = _compute_2nd_PK(mp, E, state)
+    ∂S∂E, _ = ∇(E -> _compute_2nd_PK(mp, E, state)[1], E, :all)
```
lijas: I am not able to get `∂S∂E` and `S` at the same time, since `_compute_2nd_PK` exports multiple values

lijas:
Do you know if there is a "solution" for the two issues i commented on above?

examples/elasto_plastic.jl
```diff
@@ -0,0 +1,300 @@
+using JuAFEM
+using Tensors
+using KrylovMethods
+using TimerOutputs
+using Statistics: mean
+import ProgressMeter
+const ∇ = Tensors.gradient
+
+struct Material{T}
+    μ::Float64		#Lamé constant
```
KristofferC: Remove tabs and maybe add a space after `#`?

examples/elasto_plastic.jl
```diff
@@ -0,0 +1,300 @@
...
+    λ::Float64		#Lamé constant
+    τ₀::Float64		#Yield stress
+    H::Float64		#Hardening
+end
+
+struct MaterialState{T}
```
KristofferC: Same with tabs here.

If you don't want to hardcode the dimension you could do something like

```jl
struct MaterialState{T, dim, ST <: SymmetricTensor{2, dim, T}, TT <: Tensor{2, dim, T}}
S::ST
F⁻¹::TT
...
end
```
```

File: examples/elasto_plastic.jl [hunk 5]
```diff
@@ -0,0 +1,300 @@
...
+	S::SymmetricTensor{2,3,T,6} #2nd PK stress
+    ϵᵖ::Float64					#Plastic strain
+    F⁻¹::Tensor{2,3,T,9}		#Plastic deformation grad
+    ν::Tensor{2,3,T,9}			
+end
+
+function MaterialState()
+    return MaterialState(
```
KristofferC: Tabs

examples/elasto_plastic.jl
```diff
@@ -0,0 +1,300 @@
...
+    dγ = 0 
+
+    C = 2E + one(E)
+
+    phi, Mdev, Stilde = ϕ(C, F⁻¹, ϵᵖ, μ, λ, τ₀, H)
+
+    if phi > 0 #Plastic step
+        dγ = 0.0
+        
+        #Setup newton varibles
```
KristofferC:
```suggestion
        #Setup newton variables
```

examples/elasto_plastic.jl
KristofferC: Yes. You shouldn't use AD on an iterative solver, that will be crazy expensive. Instead, exploit that `dS/dE = inv(-dr/dS) * dr/dE` evaluated at the solved state where `r` is the residual function.

lijas: I have looked at this for a while now, but I dont quite understand. My residual is `ϕ_residual` (it is a scalar), but it does not depend on SPK. how should I take `dr/dS`?

lijas: I realised that  I should calculate `dF/dC` from the residual and use that to calculate `S`

termi-official:
Any plans on reviving this for the code gallery (or tutorials?)?

PR #246, opened 2019-06-22
Support for mixed celltypes

JimBrouzoulis:
This PR is to discuss:

* an example implementation of having mixed celltypes 
* other changes required in the code

Example script using the MixedGrid
https://gist.github.com/JimBrouzoulis/98e5b8e50895ea322da354eb0ab197ba


src/Dofs/MixedDofHandler.jl
```diff
@@ -0,0 +1,254 @@
+
+mutable struct FieldHandler
+    field_names::Vector{Symbol}
```
lijas: Maybe add a `struct Field` with`field_name`, `field_dims` and `and field_interpolations`

src/Dofs/MixedDofHandler.jl
```diff
@@ -0,0 +1,254 @@
...
+    field_dims::Vector{Int}
+    field_interpolations::Vector{JuAFEM.Interpolation}
+    quad_rules::Vector{JuAFEM.QuadratureRule}
```
lijas: I dont think QuadratureRule should be stored here. User should handle this on their own?

src/Dofs/MixedDofHandler.jl
JimBrouzoulis: I have done so now. The FieldHandler stores a number of Fields.

src/Dofs/MixedDofHandler.jl
JimBrouzoulis: Yes, I agree. I don't know why I added it there.

fredrikekre:
:tada: 

I think we can probably combine the two types in the future, but let's try it out separately first.

KristofferC:
Woot!

KristofferC:
Need an example now :)

JimBrouzoulis:
> Need an example now :)

A small test that can be reworked into an example. This can then be used to benchmark and find bottlenecks in the current implementation.
https://github.com/KristofferC/JuAFEM.jl/blob/07f6d065865cacc891909261b2bd243ae80b1ef2/test/test_mixeddofhandler.jl#L295

PR #247, opened 2019-06-24
Add VertexIndex, EdgeIndex and FaceIndex and update ConstraintHandler

lijas:
Currently it is only possible to add constraints on faces and node-ids. This PR makes it possible add constraints on edges and vertices. It was possible to only need one `add(::ConstraintHandler, d::Dirichlet)` -function regardless if one adds constraints on vertices ,edges or faces by utilizing the  `vertices`, `edges` and `faces` functions.

I also removed the possibility to add constraints on nodes since one can now put them on vertices. One could add a `node2vertex()` function 

This is not a completely working PR. One more change in `_update(::ConstraintHandler)` function is needed regarding `BCValues`


lijas:
Replaced by other PR

PR #249, opened 2019-06-28
Delete stiffness_example.ipynb

KristofferC:
This file is kinda pointless and was mostly just as a way for me to benchmark using cellvalues vs using normal "B-matrices" to create the stiffness matrix.


JimBrouzoulis:
Does serve as an example on how one can go about using a more traditional `B-matrix` approach if one likes. But a small code-snippet in the docs would probably be enough.

KristofferC:
Yeah, I guess it might be good to have for people that want to use JuAFEM just for some small stuff like computing the shape value.

I think what is good about JuAFEM is that you can sort of chose the amount you want to use it. It should be a system of composable tools and you don't need to use everything for it to be useful.

JimBrouzoulis:
I totally agree.
________________________________

KristofferC:
Has been moved to `docs/old_examples`.

PR #261, opened 2019-11-13
add helmholtz example to the docs

jw3126:


docs/src/literate/helmholtz.jl
```diff
@@ -0,0 +1,169 @@
...
+
+vtkfile = vtk_grid("helmholtz", dh)
+vtk_point_data(vtkfile, dh, u)
+vtk_save(vtkfile)
+
+using Test
+# TODO where does this number come from?
+Test.@test maximum(u) ≈ 0.05637592090022005
+# TODO would be nice to actually compute e.g. the L2 error between the analytic solution
+# (projected to the fem space) and the numeric solution.
```
jw3126: Some guidance would be appreciated. I have no idea, where this number comes from. 
Also, I would really like to compute some error norms. But it seems, that `JuAFEM` does not really have much infrastructure to do that (no struct that represents an element of a FEM space, no way of projecting a `Function` onto a FEM space). Is that correct?

fredrikekre: > Some guidance would be appreciated. I have no idea, where this number comes from.

I think that number is just the value we got, not compared to any other software. The purpose is just to make sure it does not regress.

> I would really like to compute some error norms. But it seems, that JuAFEM does not really have much infrastructure to do that (no struct that represents members of a FEM space, no way of projecting a Function onto a FEM space). Is that correct?

Yea, you have to roll your own function for that, but I think it should not be too difficult. So far the functionality in here is mostly what we have needed ourselves for our own research projects, but what you are talking about would be a great addition.

docs/src/literate/helmholtz.jl
```diff
@@ -0,0 +1,169 @@
...
+
+K, f = doassemble(cellvalues, facevalues, K, dh);
+apply!(K, f, dbcs)
+u = Symmetric(K) \ f;
+
+vtkfile = vtk_grid("helmholtz", dh)
+vtk_point_data(vtkfile, dh, u)
+vtk_save(vtkfile)
+
+using Test
```
KristofferC: This could be made source only (like I think we do in the other examples) to not have it show up in the generated documentation, but still have it get tested in CI.

docs/src/literate/helmholtz.jl
KristofferC: It is just the number that I got... not scientific but better than nothing 🤷‍♂

jw3126: I might add some helper functions to compute error norms in the future. But not in this PR.

docs/src/literate/helmholtz.jl
jw3126: Done!

docs/src/literate/helmholtz.jl
```diff
@@ -0,0 +1,170 @@
...
+vtkfile = vtk_grid("helmholtz", dh)
+vtk_point_data(vtkfile, dh, u)
+vtk_save(vtkfile)
+using Test #src
+#src this test catches unexpected changes in the result over time
+#src it does not certify that the solution is any good
+Test.@test maximum(u) ≈ 0.05637592090022005 #src
+@assert u_ana(Vec{2}((-0.5, -0.5))) >= 1 #src
+@assert u_ana(Vec{2}((0.5, -0.5))) >= 1 #src
+@assert u_ana(Vec{2}((-0.5, 0.5))) >= 1 #src
```
jw3126: It is my understanding that since we use Lagrange elements, the entires of `u` are values of the solution at node points. But the maximum of `u_ana` seems to be slightly bigger then `1`? So this would be an indicator that our solution is garbage?

KristofferC: Probably not complete garbage since the fields have at least similar shape :D. But maybe some factor wrong? Or something else. 

bild

bild

jw3126: I will check if I spot an error.

docs/src/literate/helmholtz.jl
```diff
@@ -0,0 +1,170 @@
...
+function u_ana(x::Vec{2, T}) where {T}
+    xs = (Vec{2}((-0.5,  0.5)),
+          Vec{2}((-0.5, -0.5)),
+          Vec{2}(( 0.5,  -0.5)))
+    σ = 1/8
+    s = zero(eltype(x))
+    for i in 1:3
+        s += exp(- norm(x - xs[i])^2 / σ^2)
+    end
+    return max(1e-15 * one(T), s) # Denormals, be gone
```
jw3126: Why we need to kill denormals?

docs/src/literate/helmholtz.jl
jw3126: I think I fixed the errors, see here https://github.com/KristofferC/JuAFEM.jl/pull/261/commits/cf9f78c6c6ba8226be95a500b85de3b29b966f68

jw3126: How do you visualize the `.vtu` file? I tried using paraview, but it gives some errors and just shows a grey square.

docs/src/literate/helmholtz.jl
KristofferC: They typically make computations very slow. But I am not sure why I did it like this and not just clamp to zero. We could remove it perhaps.

fredrikekre:
If you rebase on master we get preview docs from https://github.com/KristofferC/JuAFEM.jl/commit/ec8dce5d7ae43075815540c509485428390b00c0 !

jw3126:
Where can I view the online preview of the docs?

fredrikekre:
Ah, sorry I was wrong. I forgot that only works for PRs with the branch in the same repository.

fredrikekre:
Thanks!

KristofferC:
Great, thanks.

fredrikekre:
https://kristofferc.github.io/JuAFEM.jl/dev/examples/helmholtz.html

docs/src/literate/helmholtz.jl
mforets: FWIW, i had the same problem and paraview worked well only after switching the linux session to open under the "xorg" mode instead of the default "wayland". I'm on a linux machine with Fedora 31.

JimBrouzoulis:
A comment. 
The weak form contains the Dirichlet condition (so it is enforced weakly) but the implementation enforces it strongly using `Dirichlet`

So, the weak form should only be:

    \int \nabla δu \cdot \nabla u d\Omega
    + \int δu \cdot u d\Omega
    - \int δu \cdot f d\Omega
    + \int δu \cdot (n \cdot \nabla u - g_2) d\Gamma_2

And then supplemented with the equation `u = g_1 \quad x \in \Gamma_1`

PR #268, opened 2020-01-27
Mixed grid updates

JimBrouzoulis:
Some changes in the code to allow for non-contiguous ordering of cell types.


src/Dofs/MixedDofHandler.jl
```diff
@@ -4,30 +4,63 @@ mutable struct FieldHandler
...
 
+
+struct CellDofs
+    dofs::Vector{Int}
+    offset::Vector{Int}
+    length::Vector{Int}
+end
+
+import Base.getindex
+function getindex(eldofs::CellDofs, el::Int64)
```
KristofferC:
```suggestion
function Base.getindex(eldofs::CellDofs, el::Int64)
```

src/Dofs/MixedDofHandler.jl
```diff
@@ -4,30 +4,63 @@ mutable struct FieldHandler
...
 end
 
+
+struct CellDofs
+    dofs::Vector{Int}
+    offset::Vector{Int}
+    length::Vector{Int}
+end
+
+import Base.getindex
```
KristofferC:
```suggestion
```

src/Dofs/MixedDofHandler.jl
```diff
@@ -4,30 +4,63 @@ mutable struct FieldHandler
...
+function getindex(eldofs::CellDofs, el::Int64)
+     return eldofs.dofs[eldofs.offset[el]:eldofs.offset[el]+eldofs.length[el]-1]
```
KristofferC: Hoist `eldofs.offset[el]` into a variable to not lookup twice?

src/Dofs/MixedDofHandler.jl
```diff
@@ -4,30 +4,63 @@ mutable struct FieldHandler
...
 
 function MixedDofHandler(grid::MixedGrid)
-    MixedDofHandler(FieldHandler[], Int[], Int[], JuAFEM.ScalarWrapper(false), grid)
+    MixedDofHandler(FieldHandler[], CellDofs([],[],[]), JuAFEM.ScalarWrapper(false), grid)
 end
 
 getfieldnames(fh::FieldHandler) = [field.name for field in fh.fields]
 getfielddims(fh::FieldHandler) = [field.dim for field in fh.fields]
 getfieldinterpolations(fh::FieldHandler) = [field.interpolation for field in fh.fields]
+import JuAFEM.ndofs_per_cell, JuAFEM.celldofs, JuAFEM.isclosed, JuAFEM.ndofs
```
KristofferC: No need to import these, right?

src/Dofs/MixedDofHandler.jl
```diff
@@ -4,30 +4,63 @@ mutable struct FieldHandler
...
+ndofs_per_cell(dh::MixedDofHandler, cell::Int64) = dh.cell_dofs.length[cell]
+ndofs(dh::MixedDofHandler) = maximum(dh.cell_dofs.dofs)
```
KristofferC:
```suggestion
ndofs(dh::MixedDofHandler) = length(dh.cell_dofs.dofs)
```

If we are assuming that the celldofs start from 1 and are continous?

src/Dofs/MixedDofHandler.jl
```diff
@@ -139,9 +192,14 @@ function _close!(dh::MixedDofHandler{dim}, cellnumbers, field_names, field_dims,
 
         end
         # after done creating dofs for the cell, push them to the global list
-        push!(dh.cell_dofs, cell_dofs...)
+        push!(dh.cell_dofs.dofs, cell_dofs...)
+        dh.cell_dofs.length[ci] = length(cell_dofs)
+        #push!(dh.cell_dofs, cell_dofs...)
         # push! the first index of the next cell to the offset vector
-        push!(dh.cell_dofs_offset, length(dh.cell_dofs)+1)
+        #dh.cell_dofs_offset[ci] = nextdof
```
KristofferC: Should commented lines be deleted?

src/Dofs/MixedDofHandler.jl
JimBrouzoulis: Yup, no need for them.

src/Dofs/MixedDofHandler.jl
JimBrouzoulis: They are not continuous, simply an array with all the dofs for each cell after one another.
But I can't see `ndofs` being called frequently. Only in some initializations like for the sparse structure.

src/Dofs/MixedDofHandler.jl
JimBrouzoulis: Yes

src/Dofs/MixedDofHandler.jl
KristofferC: Yeah, I mixed it up with `ndofs_per_cell` -.-.

Still, I wonder if it wouldn't be better to cache it on `close!` in `dh.ndofs` or something.

JimBrouzoulis: Sure, we can do that. Saving an integer may be acceptable use of memory in this case 😄

PR #269, opened 2020-02-05
Juafem pullrequest

natschil:
Thisi is https://github.com/KristofferC/JuAFEM.jl/pull/228 but without the erroneous commits at the end. 

Add a function for generating regular P2 triangulations with Quadratic Tetrahedral elements I've briefly tested the code and it seems to work. Somebody should go  over the code that creates the boundary facesets, as I don't understand  it and just copied the relevant sections from  generate_grid(Tetrahedron, ....)

### DIFF
--- a/src/Grid/grid_generators.jl
+++ b/src/Grid/grid_generators.jl
@@ -429,19 +429,21 @@ function generate_grid(::Type{QuadraticTetrahedron}, cells_per_dim::NTuple{3,Int
     # left = (1, 4, 5, 8), right = (2, 3, 6, 7)
     # front = (1, 2, 5, 6), back = (3, 4, 7, 8)
     # bottom = (1, 2, 3, 4), top = (5, 6, 7, 8)
+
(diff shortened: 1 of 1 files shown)


KristofferC:
Thanks for the PR. We don't really have a good way to export a visualization of facesets but as a workaround I created dirichlet boundary conditions on all faces and exported that:

```
grid = generate_grid(QuadraticTetrahedron, (5,5,5))
dh = DofHandler(grid)
push!(dh, :u, 3) 
close!(dh)
dbcs = ConstraintHandler(dh)
foreach(x -> add!(dbcs, Dirichlet(:u, getfaceset(grid, x), (x,t) -> [0.0, 0.0, 0.0], [1, 2, 3])),  keys(grid.facesets))
close!(dbcs)

vtk = vtk_grid("quadtetra", grid)
vtk_point_data(vtk, dbcs)
vtk_save(vtk) 
```

I then get something that looks like:

image

Maybe this means that some faces are not part of the boundary?

fredrikekre:
Maybe you can just add `addfaceset!` etc to create the sets.

PR #279, opened 2020-03-15
Changes to MixedDofHandler

lijas:
This makes changes to the MixedDofHandler so it will work the same as regular DofHandler for one celltype. This makes it possible to later merge MixedDofHandler/MixedGrid with DofHandler/Grid, which I think is much cleaner than having two versions that basically do the same thing.

I dont think there will any loss in speed becuase of the way dofhandler stores celldofs, cellcoords and collnodes now.


docs/src/literate/heat_equation.jl
```diff
@@ -127,20 +127,25 @@ function doassemble(cellvalues::CellScalarValues{dim}, K::SparseMatrixCSC, dh::D
...
     assembler = start_assemble(K, f)
 
+    celldofs = zeros(Int, ndofs_per_cell(dh))
+    cellcoords = zeros(Vec{dim,T}, JuAFEM.nnodes_per_cell(dh))
+
     # It is now time to loop over all the cells in our grid. We do this by iterating
     # over a `CellIterator`. The iterator caches some useful things for us, for example
     # the nodal coordinates for the cell, and the local degrees of freedom.
     #+
-    @inbounds for cell in CellIterator(dh)
```
KristofferC: Does `CellIterator` not work for `MixedDofHandler`?

lijas: Not at the moment, but could quite easily fix that so it works for both dofhandler and mixeddofhandler.

lijas:
Updates:

- Remove MixedGrid 
- Updated CellIterator to work with MixedDofHandler.
- I also updated DofHandler, even if the plan is to remove it. The reason for this is to be able to make speed comparisons later maybe...
- Created AbstractCell because i need to be able to dispatch on Bezier-cells which has the same type s as a QuadraticQuadrilateral

src/iterators.jl
```diff
@@ -23,36 +23,41 @@ for cell in CellIterator(grid)
...
+struct CellIterator{dim,C,T}
     flags::UpdateFlags
-    grid::Grid{dim,N,T,M}
+    grid::Grid{dim,C,T}
     current_cellid::ScalarWrapper{Int}
     nodes::Vector{Int}
     coords::Vector{Vec{dim,T}}
-    dh::DofHandler{dim,N,T,M}
+    cellset::Vector{Int}
+    dh::Union{DofHandler{dim,C,T}, MixedDofHandler{dim,C,T}} #Futre: remove DofHandler and rename MixedDofHandler->DofHandler
```
KristofferC: Specialize on this instead?

```
CellIterator{dim, C, T, DH<:AbstractDofHandler}
`` `

File: src/Grid/MixedGrid.jl [hunk 3]
```diff
@@ -1,5 +1,5 @@
 
-mutable struct MixedGrid{dim,C,T<:Real} <: JuAFEM.AbstractGrid
+#=mutable struct Grid{dim,C,T<:Real} <: JuAFEM.AbstractGrid
```
fredrikekre: Delete this file

src/Dofs/ConstraintHandler.jl
```diff
@@ -54,13 +54,14 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     free_dofs::Vector{Int}
     values::Vector{T}
     dofmapping::Dict{Int,Int} # global dof -> index into dofs and values
+    bcvalues::Vector{BCValues{T}}
```
lijas: Store bcvalues in ConstraintHandler instead of getting it from DofHandler

src/Dofs/ConstraintHandler.jl
```diff
@@ -386,68 +393,27 @@ function meandiag(K::AbstractMatrix)
 end
 
 
-
-# TODO Remove - almost identical to the previous
-function _update!(values::Vector{Float64}, f::Function, faces::Set{Tuple{Int,Int}}, field::Symbol, local_face_dofs::Vector{Int}, local_face_dofs_offset::Vector{Int},
```
lijas: Delete this function anduse the normal _update! function for both MixedDofHandler and DofHandler

lijas:
All tests work for me and I think it is possible to merge now.

KristofferC:
Are the changes to the heat equation example still supposed to be in here?

src/Dofs/ConstraintHandler.jl
fredrikekre: So each new bc will have it's own bcvalues? Isn't it more a property of the field/interpolation?

docs/src/literate/heat_equation.jl
```diff
@@ -166,7 +174,7 @@ function doassemble(cellvalues::CellScalarValues{dim}, K::SparseMatrixCSC, dh::D
         # The last step in the element loop is to assemble `Ke` and `fe`
         # into the global `K` and `f` with `assemble!`.
         #+
-        assemble!(assembler, celldofs(cell), fe, Ke)
+        assemble!(assembler, _celldofs, fe, Ke)
```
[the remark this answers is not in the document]
lijas: I can revert the entire heat_equation example

src/Dofs/ConstraintHandler.jl
lijas: MixedDofHandler stores bcvalues in the fieldHandler, and DofHandler stores bcvalues itself, so in able to be able to have the same functionality for both Dohandlers, I must save the bcvalues in ConstraintHandler (I think)

lijas: Maybe it is better to store it in `Dirichlet`

PR #281, opened 2020-04-29
Add Line2, Line3

lijas:


lijas:
I realized that this will probably not work for higher order lines, because the faces() function for `Lagrange{1,RefCube,order}` and for `Cell{2,N,1} (Line2)` are defined differently (in 1d the the faces are defined as the end points `((1,),(2,))`, but in 2d the face is defined as the `face = ((1,2),)` ).

It works for the linear case though...

PR #283, opened 2020-04-29
Make it possible to have differ qr-rule in different directions

lijas:


lijas:
Should I create a `merge_quadrules(qr1,qr2,qr2...)` function instead?

kimauth:
I am actually using different quadrature rules in different directions. Are you using this code? I think it is something pretty specific though for which one should understand how quadrature rules work. So not sure if we should explicitly support it.

fredrikekre:
#513

PR #284, opened 2020-05-04
fix celliterator again...

lijas:
Should really create test for this so I don't need to fix it again and again... :P


fredrikekre:
Why was this wrong? I get `ERROR: UndefRefError: access to undefined reference` now.

PR #285, opened 2020-05-04
Fix CellIterator constructor

kimauth:
After @lijas CellIterator fixes, I also have something to fix here. :)


fredrikekre:
So it begins....

PR #287, opened 2020-05-09
apply_rhs!

koehlerson:
@kimauth encouraged me to do my first PR, so let's go 

adds the requested function of: https://github.com/KristofferC/JuAFEM.jl/issues/280

### DIFF
--- a/src/L2_projection.jl
+++ b/src/L2_projection.jl
@@ -1,11 +1,11 @@
 
 abstract type AbstractProjector end
 
-struct L2Projector <: AbstractProjector
-    fe_values::CellValues
+struct L2Projector{CV<:CellValues} <: AbstractProjector
+    fe_values::CV
     M_cholesky #::SuiteSparse.CHOLMOD.Factor{Float64}
     dh::MixedDofHandler
-    set::Vector{Integer}
+    set::Vector{Int}
     node2dof_map::Dict{Int64, Array{Int64,N} where N}
 end
 


fredrikekre:
Do you still need to keep a copy of the stiffness matrix around for use with `apply_rhs!` then? Cause I believe you can't use the modified one since it is, well, modified?

koehlerson:
If `A` is the stiffness matrix, then with this approach there is only one `A` and one copy of A `copyA`, before there were as much copies as time steps. So what you can do with this:

- assemble

- apply bc to `A` 

- copy `A` into `copyA` 

- start time loop and within each loop iteration apply_rhs!(copyA,b,ch)

and before it was:

- assemble

- start loop and within each loop iteration copy A,  apply!(copyA,b,ch)

KristofferC:
We only need to keep around the columns from `A` that we have constraints on though right? I wonder if there should be a function `extract_constrained_columns(A)` that gives you a new sparse matrix of size `ndofs x nconstraints` and you use that one to pass into `apply_rhs!`?

koehlerson:
Ye, I tried without scaling the meandiagonal, since we talked about this also via slack, but in my tests it was definitely needed. So it's `m` and `nodfs x nconstraints` what is actually needed instead of the full copy. However, nothing came into my mind, how this could properly designed such that a user don't have to pass `m` on their own.

KristofferC:
Yeah, you would need to pass it explicitly, something like

```
struct RHSData{T}
    m::T
    constrained_columns::SparseMatrixCSC{Int, T}
end

function get_rhs_data(ch::ConstraintHandler, A)
    m = meandiag(A)
    Aa = A[:, ch.prescribed_dofs]
    return RHSData(m, Aa)
end

function apply_rhs!(rhs, data::RHSData)
    # your function here using data.m and data.constrained_columns
end
```

So a bit more code in order to not have to save a duplicate of the original spare matrix. Do you think it is worth it?

koehlerson:
Definitely! I'll take care of it :)

koehlerson:
Sorry for the delay, I had some compulsory things to do. I added the struct and function to initialize the struct. It's tested with the same old script of mine as before: https://github.com/koehlerson/Catalyst/blob/master/scripts/unsteady-multiscale-convection-diffusion-SUPG.jl

PR #294, opened 2020-08-17
Add support for adding constraints to nodes when using MixedDofHandler

kimauth:
This introduces an additional _add! function that allows to add Dirichlet BCs to nodesets for the MixedDofHandler. I turned out to need this, so maybe somebody else also appreciates it. :)

A few comments:

- I bypass _check_cellset_dirichlet: Having to filter a nodeset by which nodes belong to the cellset of the FieldHandler beforehand doesn't seem reasonable to me, considering that the new _add! just skips all nodes that either don't belong to the FieldHandler or don't have dofs in the constrained field.

- Probably it would be possible to use the same _add! for the regular DofHandler and the MixedDofHandler, by just handing it a cellset of 1:getncells(ch.dh.grid) by default (but I was afraid to break something...).

- I think checking for a default_interpolation(getcelltype(ch.dh.grid)) doesn't make much sense for a mixed grid. Though I don't really understand whats the problem about adding node constraints to sub/superparametric approximations, so if I could get an explanation about that, I am happy to come up with a check that makes more sense for mixed grids.


kimauth:
Replaced by #346

PR #297, opened 2020-10-15
WriteVTK: pass through compress keyword argument, fixes #293.

### LINKED ISSUE #293: Allow kwargs for overloaded JuAFEM vtk_grid constructor
It seems there is a memory leak when using vtk_grid with compress=true, which is the default option, see also: https://github.com/jipolanco/WriteVTK.jl/issues/43 . I had the same issues even without using PVD, but the suggested workaround (compress=false) has worked fine for me, as the only disadvantage the size of the output files increases of course. However, in the overloaded constructor of the JuaFEM WriteVTK.vtk_grid, no kwargs are supported. A quick workaround I did was:

```
function WriteVTK.vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}; compress=false) where {dim,C,T}
    cls = MeshCell[]
    for cell in grid.cells
        celltype = JuAFEM.cell_to_vtkcell(typeof(cell))
        push!(cls, MeshCell(celltype, collect(cell.nodes)))
    end
    coords = reshape(reinterpret(T, getnodes(grid)), (dim, getnnodes(grid)))
    if(compress) # MB: added compress option
	return vtk_grid(filename, coords, cls; compress=true)
    else
	return vtk_grid(filename, coords, cls; compress=false) 
    end
end 
```

It would be nice to include this in future versions.

fredrikekre:


src/Export/VTK.jl
```diff
@@ -17,14 +17,14 @@ cell_to_vtkcell(::Type{QuadraticTetrahedron}) = VTKCellTypes.VTK_QUADRATIC_TETRA
...
-function WriteVTK.vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}) where {dim,C,T}
+function WriteVTK.vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}; compress::Bool=true) where {dim,C,T}
     cls = MeshCell[]
     for cell in grid.cells
         celltype = JuAFEM.cell_to_vtkcell(typeof(cell))
         push!(cls, MeshCell(celltype, collect(cell.nodes)))
     end
     coords = reshape(reinterpret(T, getnodes(grid)), (dim, getnnodes(grid)))
-    return vtk_grid(filename, coords, cls)
+    return vtk_grid(filename, coords, cls; compress=compress)
```
lijas, line 27: Is compress=compress needed in 1.5?

fredrikekre: No but I don't like the other syntax :)

KristofferC: Waaat.

PR #298, opened 2020-10-16
Add quadratic hexa interpolation

lijas:
Also noticed that edges was wrong order compared to VTK, so changed order of that.
https://vtk.org/doc/nightly/html/classvtkQuadraticHexahedron.html

Also, the order of the faces is wrong, which will affect vtk output if we add hexa-interpolation with facedofs.

Tested in hyperalticity.jl example, and it coverges to same result when comparing with tetrahedtron element.


termi-official:
Quadratic serendipity is fixed with #353 , right?

PR #300, opened 2020-10-22
update L2 projection

kimauth:
This PR updates the L2 projection such that it can work with mixed grids (grids with several different cell types). Therefore the internal DofHandler of the L2Projector is changed to a MixedDofHandler.
Additionally, a few problems with the L2Projection are addressed:

- An optional argument fe_values_mass is added. This enables the user to use a higher number of quadrature points for the integration of the "mass" matrix. When using (some kind of) reduced integration, this is needed as the mass matrix might become singular when not correctly integrated. However, the fe_values argument must have the same integration points that were used for computing vars. By default fe_values_mass = fe_values. (Does not change the API.)

- When only using a subset of the mesh for interpolation the values can now still be projected to the nodes. All nodes that are not part of the subset will get a NaN value (or its Tensor equivalent).

- Bugfix concerning projection of SymmetricTensors

- The project function for SymmetricTensors is adjusted to the version for general Tensors: it now has an optional argument project_to_nodes that is true by default. 

There are also some open points that I think should be addressed in the future:

-  It is possible to hand CellVectorValues to the L2Projector, but this will result in an error.
- The L2Projector takes an interpolation as an argument, but that is not the interpolation used for the L2 projection. I think this is quite confusion and not obvious to a user. Instead the interpolation from the fe_values is used.

I think it would make sense to hand a quadrature rule (and an optional quadrature rule for mass matrix integration) to the L2Projector. The the CellScalarValues object could then be created internally based on the quadrature rule and the given interpolation. This would fix both above mentioned issues, but bread the API of the L2Projector. Maybe we could introduce it as additional constructor for a while.

Finally, I think it would be worth documenting the L2Projection in the API reference. It is a useful functionality even for users that don't know how it works exactly, but these would probably need some explanation on the inputs.

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -153,6 +153,7 @@ Dofs are created in the following order: Go through each FieldHandler in the ord
 """
 function close!(dh::MixedDofHandler)
     dh, _, _, _ = __close!(dh)
+    return dh
 end
 
 function __close!(dh::MixedDofHandler{dim}) where {dim}
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -151,7 +151,7 @@ end
 Closes the dofhandler and creates degrees of freedom for each cell.
 Dofs are created in the following order: Go through each FieldHandler in the order they were added. For each field in the FieldHandler, create dofs for the cell. This means that dofs on a particular cell will be numbered according to the fields; first dofs for field 1, then field 2, etc.
 """
-function close!(dh::MixedDofHandler{dim}) where {dim}
+function close!(dh::MixedDofHandler{dim}, return_dicts=false) where {dim}
 
     @assert !JuAFEM.isclosed(dh)
     field_names = JuAFEM.getfieldnames(dh)  # all the fields in the problem
@@ -203,6 +203,9 @@ function close!(dh::MixedDofHandler{dim}) where {dim}
         push!(dh.cell_coords.length, length(cell.nodes))
     end
 
+    if return_dicts
+        return dh, vertexdicts, edgedicts, facedicts
+    end
     return dh
 end
 
@@ -234,7 +237,7 @@ function _close!(dh::MixedDofHandler{dim}, cellnumbers, field_names, field_dims,
             if ip_info.nvertexdofs > 0
                 nextdof = add_vertex_dofs(cell_dofs, cell, vertexdicts[fi], field_dims[fi], ip_info.nvertexdofs, nextdof)
             end
-                   
+
             if ip_info.nedgedofs > 0 && dim == 3 #Edges only in 3d
                 nextdof = add_edge_dofs(cell_dofs, cell, edgedicts[fi], field_dims[fi], ip_info.nedgedofs, nextdof)
             end
@@ -242,7 +245,7 @@ function _close!(dh::MixedDofHandler{dim}, cellnumbers, field_names, field_dims,
             if ip_info.nfacedofs > 0 && (ip_info.dim == dim)
                 nextdof = add_face_dofs(cell_dofs, cell, facedicts[fi], field_dims[fi], ip_info.nfacedofs, nextdof)
             end
-            
+
             if ip_info.ncelldofs > 0
                 nextdof = add_cell_dofs(cell_dofs, ci, celldicts[fi], field_dims[fi], ip_info.ncelldofs, nextdof)
             end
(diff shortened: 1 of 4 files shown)


lijas:
It looks good to me, but here are some thoughts and question.

Should we check that all the cells in the cellset has the same celltype, and give an error if it is not? (like `_check_same_celltype` in iterator.jl file). Or maybe the user should deal with it them self...

Could the interpolation be determined from the other inputs? Like `default_interpolation(grid.cells(first(cellset)))` or something? It seems like this is the most common use case, ie projecting to nodes to visualize in VTK.

Why do we have different function for symmetric tensor and tensor? They look pretty much the same?

Add type for node2dof_map in the struct.

kimauth:
That's good points!

If we check that all cells have the same celltype, we could by default use the `default_interpolation(grid.cells(first(cellset)))`, but I would let the user choose another interpolation if they want. This only becomes significant though if we do change the interface of the L2Projector (otherwise the interpolation that is given is basically meaningless).

I think the reason for the two different functions for Symmetric and regular Tensors was to return the correct type, but I agree that there is probably a better way to do this. I'll come up with a solution.

kimauth:
- Added type for node2dof_map
- Joined the project functions for regular and symmetric tensors to one function
- Added a constructor for the L2Projector based on quadrature rule and interpolation. It takes a quadrature rule and interpolation for the approximation of the field. Thus the quadrature points should be the same as the ones in which the field (given by the 'vars' argument of  `project`) was computed. It is possible to give a different quadrature rule for the computation of the "mass matrix" and a different interpolation for geometry approximation.
- The new L2Projection constructor uses the `_check_same_celltype` function from the iterator file. Of course, this results in an error message regarding the cell iterator when the cells are not the same. Should we modify this error message to be more generally applicable?

lijas:
Nice, I think you can merge this now, if not @fredrikekre has anything to remark.
...well, maybe fix the tests first :)

src/Dofs/MixedDofHandler.jl
```diff
@@ -203,6 +203,9 @@ function close!(dh::MixedDofHandler{dim}) where {dim}
         push!(dh.cell_coords.length, length(cell.nodes))
     end
 
+    if return_dicts
```
fredrikekre, line 206: This doesn't seem to be used anywhere

src/L2_projection.jl
```diff
@@ -3,24 +3,48 @@ abstract type AbstractProjector end
 
 struct L2Projector <: AbstractProjector
     fe_values::CellValues
-    M_cholesky # ::SuiteSparse.CHOLMOD.Factor{Float64}
-    dh::DofHandler
+    M_cholesky #::SuiteSparse.CHOLMOD.Factor{Float64}
+    dh::MixedDofHandler
```
fredrikekre, line 7: Does it still work with the regular DofHandler?

fredrikekre (commented):
I have not used or written this code so if it works for you feel free to merge :+1:

src/Dofs/MixedDofHandler.jl
kimauth: It is used for getting the `node2dof_map` in the constructor. That is needed for reordering the values according to the node order in `project`.

src/L2_projection.jl
kimauth: The MixedDofHandler is constructed internally, so I think there is no need for it to offer an alternative with the regular DofHandler. The old tests still work with the new version (you can do all computations with DofHandler and then still use this for interpolating your results in the end). Do you see any other reason to have another version with the regular DofHandler?

fredrikekre: :+1:

src/Dofs/MixedDofHandler.jl
fredrikekre: Ah, I thought it was a kwarg and searched for it as that. Perhaps it would be better to have
```
function close!(dh::MixedDofHandler)
    dh, _, _, _ = _close!(dh)
end
function _close!(dh::MixedDofHandler)
    # ...
    return dh, vertexdicts, edgedicts, facedicts
end
```
and you can call `_close!` directly in the projection constructor.

kimauth: I tried to comply with how this works in the DofHandler. This looks like a nice solution, but for the MixedDofHandler there already is a `_close!` function that does something else. We could still make `return_dicts` a kwarg?

fredrikekre: Make it `__close!` then

kimauth: I added this for both the MixedDofHandler and the DofHandler now (in order to be consistent). I don't think it is ever used within JuAFEM for the DofHandler though (so it also shouldn't break anything).

fredrikekre: Nice. The reason, as I am sure you know, is that it is generally bad if output types depend on input values.

fredrikekre:
They grow up so fast @KristofferC!

PR #301, opened 2020-11-09
Misc fixes for CellIterator without DofHandler

fredrikekre:


fredrikekre:
Broken by https://github.com/KristofferC/JuAFEM.jl/pull/279

lijas:
Maybe add the DofHandlers to the type-space, and have two reinit-functions depending on if the CellIterator has a `Grid` or a `DofHandler`?

```
struct CellIterator{dim,C,T,DH<AbstractDofHandler}
    ....
    dh::Union{Missing,DH}
end

function reinit!(ci::CellIterator{dim,C,T,DH}, i::Int) where DH<:AbstractDofHandler ...
function reinit!(ci::CellIterator{dim,C,T, DH}, i::Int) where DH <: Missing ...
```
That way we dont need all the `if isdefinied(dh..)`. (The code above obviously will not work, but I think you get the idea).

If you dont like the above suggenstion, your code looks good to me.

PR #302, opened 2020-11-09
Make mixeddofhandler take any grid

lijas:
In my IGA.jl package, I create a custom grid subtybing AbstractGrid. However I cant use the MixedDofHandler in that case, because it only takes the JuAFEM.Grid. This pr moves the Grid to the typespace.

I guess this does not break anything, right?

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -26,18 +26,18 @@ function Base.getindex(elvec::CellVector, el::Int)
     return elvec.values[offset:offset + elvec.length[el]-1]
  end
 
-struct MixedDofHandler{dim,C,T} <: JuAFEM.AbstractDofHandler
+struct MixedDofHandler{dim,C,T,G<:AbstractGrid} <: JuAFEM.AbstractDofHandler
     fieldhandlers::Vector{FieldHandler}
     cell_dofs::CellVector{Int}
     cell_nodes::CellVector{Int}
     cell_coords::CellVector{Vec{dim,T}}
     closed::ScalarWrapper{Bool}
-    grid::Grid{dim,C,T}
+    grid::G
     ndofs::ScalarWrapper{Int}
 end
 
 function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
-    MixedDofHandler{dim,C,T}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
+    MixedDofHandler{dim,C,T,typeof(grid)}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
 end
 
 getfieldnames(fh::FieldHandler) = [field.name for field in fh.fields]


src/Dofs/MixedDofHandler.jl
```diff
@@ -26,18 +26,18 @@ function Base.getindex(elvec::CellVector, el::Int)
...
     cell_coords::CellVector{Vec{dim,T}}
     closed::ScalarWrapper{Bool}
-    grid::Grid{dim,C,T}
+    grid::G
     ndofs::ScalarWrapper{Int}
 end
 
 function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
-    MixedDofHandler{dim,C,T}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
+    MixedDofHandler{dim,C,T,typeof(grid)}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
```
koehlerson, line 40: Just some thoughts: 
* The parametrization of `MixedDofHandler` was previously `dim,C,T` based on `Grid{dim,C,T}` and now would be doubled to `dim,C,T,Grid{dim,C,T}`
* Imo it would be better to either have `MixedDofHandler{Grid}` (don't know if possible due to cell_coords), `MixedDofHandler{dim,T,Grid}` or `MixedDofHandler{dim, C, T, Grid}`. Last one corresponds to the least invasive approach 

The last one just needs a small change in the constructor
```julia
function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T} 
    MixedDofHandler{dim,C,T,Grid}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
end
```
Another approach:
```julia
julia> struct MixedDofHandler{dim,T,G <: JuAFEM.AbstractGrid} <: JuAFEM.AbstractDofHandler
           fieldhandlers::Vector{FieldHandler}
           cell_dofs::JuAFEM.CellVector{Int}
           cell_nodes::JuAFEM.CellVector{Int}
           cell_coords::JuAFEM.CellVector{Vec{dim,T}}
           closed::JuAFEM.ScalarWrapper{Bool}
           grid::G
           ndofs::JuAFEM.ScalarWrapper{Int}
       end

julia> function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
           MixedDofHandler{dim,T,Grid}(FieldHandler[], JuAFEM.CellVector(Int[],Int[],Int[]), JuAFEM.CellVector(Int[],Int[],Int[]), JuAFEM.CellVector(Vec{dim,T}[],Int[],Int[]), JuAFEM.ScalarWrapper(false), grid, JuAFEM.ScalarWrapper(-1))
       end
MixedDofHandler

julia> grid = generate_grid(Line, (2,))
Grid{1,Cell{1,2,2},Float64} with 2 Line cells and 3 nodes

julia> mdh = MixedDofHandler(grid)
MixedDofHandler{1,Float64,Grid}(FieldHandler[], JuAFEM.CellVector{Int64}(Int64[], Int64[], Int64[]), JuAFEM.CellVector{Int64}(Int64[], Int64[], Int64[]), JuAFEM.CellVector{Tensor{1,1,Float64,1}}(Tensor{1,1,Float64,1}[], Int64[], Int64[]), JuAFEM.ScalarWrapper{Bool}(false), Grid{1,Cell{1,2,2},Float64}(Cell{1,2,2}[Cell{1,2,2}((1, 2)), Cell{1,2,2}((2, 3))], Node{1,Float64}[Node{1,Float64}([-1.0]), Node{1,Float64}([0.0]), Node{1,Float64}([1.0])], Dict{String,Set{Int64}}(), Dict{String,Set{Int64}}(), Dict{String,Set{Tuple{Int64,Int64}}}("left" => Set([(1, 1)]),"right" => Set([(2, 2)])), 
  [1, 1]  =  true
  [2, 2]  =  true), JuAFEM.ScalarWrapper{Int64}(-1))

```

koehlerson (commented):
It would be nice to have a consens on this, because I could benefit from the same changes in `DofHandler` for adaptivity

src/Dofs/MixedDofHandler.jl
lijas: Thanks for the comments. `MixedDofHandler{dim,C,T,G<:AbstractGrid}` is the only one that is not completely breaking, and I think people will want to dispatch on `dim` (and possibly `C` and `T`), so we should keep them as type parameters.

koehlerson: I think you can still dispatch on `dim, C, T`, but you would write `::MixedDofHandler{Grid{dim, C, T}}` in case you need it. The benefit would be more flexibility in general when it comes down to the mesh for any `DofHandler`.
 
Right now the `MixedDofHandler` constructor would return `MixedDofHandler{dim,C,T,Grid{dim,C,T}}` and I don't see a benefit of doubling `dim, C, T`, but maybe I'm just not aware of it. 

I cloned your branch and tested with `MixedDofHandler{dim, T, G<:AbstractGrid}` worked almost ootb with all tests passed. Only the CellIterator needs a small fix in its definition and constructor based on the new definition of `MixedDofHandler`. 

Besides this, I also tested if `MixedDofHandler{dim, C, T, Grid}` instead of `MixedDofHandler{dim, C, T, Grid{dim, C, T}}` works and this indeed works ootb. 

Just my two cents in my first ever review, so don't care too much about my opinion :smile:

lijas:
Close since another pr implemented this change...

PR #303, opened 2020-11-09
Small changes to L2projector

lijas:
In my IGA code, I want to be able to dispatch on the CellValue, type.
Also I changed `Vector{Integer}` to Vector{Int}.

### DIFF
--- a/src/L2_projection.jl
+++ b/src/L2_projection.jl
@@ -1,11 +1,11 @@
 
 abstract type AbstractProjector end
 
-struct L2Projector <: AbstractProjector
-    fe_values::CellValues
+struct L2Projector{CV} <: AbstractProjector
+    fe_values::CV
     M_cholesky #::SuiteSparse.CHOLMOD.Factor{Float64}
     dh::MixedDofHandler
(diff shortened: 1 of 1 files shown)


src/L2_projection.jl
```diff
@@ -1,11 +1,11 @@
 
 abstract type AbstractProjector end
 
-struct L2Projector <: AbstractProjector
-    fe_values::CellValues
+struct L2Projector{CV} <: AbstractProjector
```
fredrikekre, line 4:
```suggestion
struct L2Projector{CV<:CellValues} <: AbstractProjector
```

fredrikekre:
Where in the code are you intercepting this for dispatch?

lijas:
```
_assemble_L2_matrix(fe_valuesIGA.BezierValues, set, dh)
_project(vars, proj::L2Projector{<:BezierValues}, M::Integer)
```

I only have to change the X = getcoordinates(..) to something similar in my code.... Everything else is the same.
With this pr I can atleast reuse the L2Proector struct and some of the functions.

PR #304, opened 2020-11-19
apply_rhs! with rhsdata

koehlerson:
I reopen this, because my old PR was quite dirty. 

## What is new
I added a functionality to apply the Dirichlet boundary conditions solely on the rhs of the problem. This feature should benefit anyone who deals with time dependent/incremental linear problems. Instead of copying in each time step the stiffness matrix `A` we can now store once the prescribed dofs and the mean in `RHSData`. `RHSData` is constructed with `get_rhs_data`.

## How it's tested
In `test/test_apply_rhs.jl` the heat equation is tested with the new functions/struct. 

This PR is related to:
- https://github.com/KristofferC/JuAFEM.jl/pull/287
- https://github.com/KristofferC/JuAFEM.jl/issues/280
- https://discourse.julialang.org/t/understand-apply-of-juafem/35325/7


src/Dofs/ConstraintHandler.jl
```diff
@@ -59,6 +59,54 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
...
+end
+
+"""
+    get_rhs_data(ch::ConstraintHandler, A::SparseMatrixCSC) -> RHSData
+
+Returns the needed RHSData for apply_rhs!
+"""
+function get_rhs_data(ch::ConstraintHandler, A::SparseMatrixCSC)
+    m = meandiag(A)
+    Aa = A[:, ch.prescribed_dofs]
```
kimauth, line 79: Maybe call this constrained_columns for clarity?

test/test_apply_rhs.jl
```diff
@@ -0,0 +1,75 @@
...
+    
+    K, f = doassemble(cellvalues, K, dh)
+    A = create_sparsity_pattern(dh)
+    A, g = doassemble(cellvalues, A, dh)
+    rhsdata = get_rhs_data(ch, A)
+    
+    apply!(K, f, ch)
+    apply_rhs!(rhsdata, g, ch)
+    u₁ = K \ f
+    u₂ = K \ g
```
kimauth, line 70: Considering that this is the only example for usage of `apply_rhs!` so far, I think it would be nice to be clear here. Using `apply_rhs!` requires that the stiffness matrix has been modified once as well.

```suggestion
    apply!(K, f, ch)
    apply!(A, ch) #need to apply bcs to A once
    apply_rhs!(rhsdata, g, ch)
    u₁ = K \ f
    u₂ = A \ g
```

(But maybe `apply(A, ch)` could be made abundant for the user by making it happen in a joint step with generating the rhs_data? What do you think?)

src/Dofs/ConstraintHandler.jl
kimauth, line 79: Maybe call this constrained_columns for clarity?

src/Dofs/ConstraintHandler.jl
```diff
@@ -59,6 +59,54 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
...
+
+"""
+    apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)
+
+Applies the boundary condition to the rhs vector without modifying stiffness matrix
+"""
+function apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)	
+    K = data.constrained_columns
+    @assert length(f) == 0 || length(f) == size(K, 1)
+    #@boundscheck checkbounds(K, ch.prescribed_dofs, ch.prescribed_dofs)
```
kimauth, line 91: This line could be deleted.

test/test_apply_rhs.jl
```diff
@@ -0,0 +1,75 @@
...
+    
+    close!(ch)
+    update!(ch, 0.0);
+    
+    function doassemble(
+        cellvalues::CellScalarValues{dim},
+        K::SparseMatrixCSC,
+        dh::DofHandler,
+    ) where {dim}
+
```
kimauth, line 29: I would either make this `doassemble!(K, f, cellvalues, dh)` and create K and f outside or `doassemble(cellvalues, dh)` and create both K and f inside.

src/Dofs/ConstraintHandler.jl
```diff
@@ -59,6 +59,54 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
...
+function get_rhs_data(ch::ConstraintHandler, A::SparseMatrixCSC)
+    m = meandiag(A)
+    Aa = A[:, ch.prescribed_dofs]
+    return RHSData(m, Aa)
+end
+
+"""
+    apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)
+
+Applies the boundary condition to the rhs vector without modifying stiffness matrix
```
kimauth, line 86:
```suggestion
Applies the boundary condition to the right-hand-side vector without modifying the stiffness matrix.
```

kimauth (commented):
This looks really good! I only have a few minor comments/suggestions.

(Long term it could be nice to have another example that shows the usage - maybe a time dependent variation of the Heat Equation example. But I don't think there is an urgent need for it now.)

test/test_apply_rhs.jl
koehlerson: yes if people are confused by the interface we can still wrap it in `get_rhs_data`. However, I've encountered this kind of interface in various FE frameworks, so I think it's convenient to do `apply!(A,ch)` before the time loop and `apply_rhs!` right before each solve statement in the time loop

koehlerson:
Included all of your suggestions. I will do this week a time-dependent example. I can do time dependent heat without advection or with advection and SUPG stabilization. With advection and SUPG stabilization people could see how easy it is to formulate Petrov-Galerkin problems instead of Galerkin ones with JuAFEM.

kimauth:
It looks great to me now!

I think any of the examples would be nice! (Maybe make sure that it is possible to understand the apply_rhs even if one doesn't know what Petrov-Galerkin problems are.)

src/Dofs/ConstraintHandler.jl
```diff
@@ -59,6 +59,53 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
...
+    K = data.constrained_columns
+    @assert length(f) == 0 || length(f) == size(K, 1)
+    @boundscheck length(f) == 0 || checkbounds(f, ch.prescribed_dofs)
+
+	m = data.m
+    @inbounds for i in 1:length(ch.values)
+        d = ch.prescribed_dofs[i]
+        v = ch.values[i]
+        if !applyzero && v != 0
+            for j in nzrange(K, d)
```
koehlerson, line 98: @I'm actually uncertain if this line should work for arbitrary cases

koehlerson (commented):
I'm writing the transient heat example and I have doubts about this line. I'm getting a boundserror and yet the new test is working. If I change `d` to `i` the boundserror is gone + new test is still working. I included the example in the `test` folder, before merging I'll move it to the literate folder and rewrite it in literate format. 

So there are two options here, I either confused myself that much while writing the transient example, that I missed something in it which causes the boundserror (so please check the example)

Option two, the line as this is not working for the `constrained_columns` in `RHSData` because it stores less indices and so you may end up outside of `K.nzval[j]`

src/Dofs/ConstraintHandler.jl
kimauth: I think this indeed is what causes your error. `nzrange(K, col)` returns the indices for column `col`. `ch.prescribed_dofs` contains the global prescribed dofs, but `K` only contains the prescribed columns and not the full matrix. So with `d` we can index into the complete `K`, but not into the version that is stored in `RHSDATA`. As you suggest in your bugfix that needs to be indexed with `i`. 

The reason that this did not show up in your test is that all prescribed values there are 0 - thus this branch is never reached.

koehlerson: Thanks so much for your time :relaxed:  ! And good catch to wait for me to write the example, ha!

koehlerson:
- bug fixed
- literate example included 
- nightly CI still failing for some `Pkg` related reasons

tested doc building locally, looks good to me

koehlerson:
@kimauth do you think we can merge this? I think the transient heat example could need a proofread

docs/src/literate/transient_heat_equation.jl
```diff
@@ -0,0 +1,205 @@
...
+# ### Solution of the system
+# We first assemble all parts in the prior allocated `SparseMatrixCSC`.
+K, f = doassemble!(cellvalues, K, dh, Δt)
+M = doassemble!(cellvalues, M, dh)
+A = (Δt .* K) + M;
+# Now, we need to save all boundary condition related values of the unaltered system matrix `A`, which is done
+# by `get_rhs_data`. The function returns a `RHSData` struct, which contains all needed informations to apply
+# the boundary conditions solely on the right-hand-side vector of the problem.
+rhsdata = get_rhs_data(ch, A);
+# We initialize the previous time step, denoted by uₙ,  to $\mathbf{0}$.
```
kimauth, line 167: We just set the initial temperature to 0, right? Previous time step sounds like we care what happened before.

kimauth (commented):
This is a really good example to add I think. The only major point that I have about it is that I would really like an explanation in the beginning why the apply_rhs approach is so suitable for it. Then people with comparable problems can see immediately that this is what they are looking for instead of having to understand the problem set-up. Everything else is minor things and possibly a matter of taste. ;)

docs/src/literate/transient_heat_equation.jl
```diff
@@ -0,0 +1,205 @@
...
+# u(x) = 0 \quad x \in \partial \Omega_1,
+# ```
+# where $\partial \Omega_1$ denotes the left and right boundary of $\Omega$.
+#
+# Further, we define heterogeneous Dirichlet boundary conditions at the top and bottom edge $\partial \Omega_2$.
+# We choose a linearly increasing function $a(t)$ that describes the temperature at this boundary
+# ```math
+# u(x) = a(t)\cdot x \quad x \in \partial \Omega_2.
+# ```
+# The semidiscrete weak form is given by
```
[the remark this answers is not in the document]
koehlerson: https://www-m16.ma.tum.de/foswiki/pub/M16/Allgemeines/NumMethHyp14/FE-Hyperbolic-Systems.pdf
Page 14 2.1.6 . If you've already substituted the H^1 Space with a finite dimensional subspace of piecewise linear functions, then you are left with an ODE in time. But to be fair, I silently did this. So I can add to "suitable test function" something like "suitable test function. We substitute the infinite dimensional space by an appropriate finite dimensional subspace of piecewise linear functions and test against all functions of the previsouly mentioned finite dimensional space."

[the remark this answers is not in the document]
kimauth: I think it is ok, this is not a mathematical text book - I was just wondering when I read it.

fredrikekre:
Nice, I don't think we had an example of time dependent function before. It would be nice to see what it would take to use some package in the DifferentialEquations.jl ecosystem for the time integration.

koehlerson:
True, I actually thought about coding one in the upcoming days, maybe we can just change/extend this one and use some higher order scheme from DiffEq as an alternative to the used time discretization

PR #306, opened 2020-11-25
More Grid Docs, mostly related to API Reference

koehlerson:
Continued the grid docs, but this time focusing on the API reference. I tried to include everything which is relevant for subtyping `AbstractGrid`. Hopefully, it will be easier now to understand how `Grid` works, but also how one can subtype `AbstractGrid` and define a suitable interface for JuAFEM


koehlerson:
I still have to add in the utility function sections that these should be dispatched on a custom `AbstractGrid` subtype in case someone wants to use their own grid format. What do you think, should I append it below, put it at the top? I think right now I'd put it on top of the section and immediately afterwards a Documenter `@index` block listing all neccessary functions.

src/Grid/grid.jl
```diff
@@ -71,7 +82,19 @@ end
...
+    Grid{dim, C<:AbstractCell, T<:Real} <: AbstractGrid}
+
 A `Grid` is a collection of `Cells` and `Node`s which covers the computational domain, together with Sets of cells, nodes and faces.
+
+# Fields
+- `cells::Vector{C}`: stores all cells of the grid
+- `nodes::Vector{Node{dim,T}}`: stores the `dim` dimensional nodes of the grid
+- `cellsets::Dict{String,Set{CellIndex}}`: maps a `String` key to a `Set` of cell ids
+- `nodesets::Dict{String,Set{Int}}`: maps a `String` key to a `Set` of global node ids
+- `facesets::Dict{String,Set{FaceIndex}}`: maps a `String` to a `Set` of `Tuple{Int,Int} (global_cell_id, local_face_id)`
```
termi-official, line 94:
```suggestion
- `facesets::Dict{String,Set{FaceIndex}}`: maps a `String` to a `Set` of `FaceIndex`
```

termi-official, line 94: We should be more precise here on what these things are doing and not just describing the data structure.

I think we can even make a group comment here. i think something like "Helper structures to group entities by their name name." could be a start.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+    getcells(grid::AbstractGrid, v::Union{Int,Vector{Int}} 
+    getcells(grid::AbstractGrid, set::String)
+
+Returns either all `cells::Vector{C<:AbstractCell}` of a `grid` or a subset based on an `Int`, `Vector{Int}` or `String`.
+Whereas the last option tries to call a `cellset` of the `grid`.
+"""
 @inline getcells(grid::AbstractGrid) = grid.cells
 @inline getcells(grid::AbstractGrid, v::Union{Int, Vector{Int}}) = grid.cells[v]
 @inline getcells(grid::AbstractGrid, set::String) = grid.cells[collect(grid.cellsets[set])]
+"Returns and `Int` corresponding to how many cells are in the `grid`."
```
termi-official, line 138: Describes the implementation, not the interface. Maybe something like "Returns the number of cells in the grid." or just "Number of cells in the grid." or smth like this?

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
 @inline getncells(grid::AbstractGrid) = length(grid.cells)
+"Returns the celltype of the `grid`."
 @inline getcelltype(grid::AbstractGrid) = eltype(grid.cells)
 @inline getcelltype(grid::AbstractGrid, i::Int) = typeof(grid.cells[i])
 
+"""
+    getnodes(grid::AbstractGrid) 
+    getnodes(grid::AbstractGrid, v::Union{Int,Vector{Int}}
+    getnodes(grid::AbstractGrid, set::String)
```
termi-official, line 147: `set` -> `set_name` or `group_name` or similar?

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+
+Returns either all `nodes::Vector{Node{dim,T}}` of a `grid` or a subset based on an `Int`, `Vector{Int}` or `String`.
+The last option tries to call a `nodeset` of the `grid`.
```
termi-official, line 150: Do we want to be this strict in the interface? Maybe it should be more relaxed and ungrouped? E.g. for ` getnodes(grid::AbstractGrid, v::Union{Int,Vector{Int}} ` something like "Returns a node with given index or a range of nodes for a given index range."

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
 ##########################
-@inline getdim(::AbstractGrid{dim}) where {dim} = dim
+@inline getdim(grid::AbstractGrid{dim}) where {dim} = dim
+"""
+    getcells(grid::AbstractGrid) 
+    getcells(grid::AbstractGrid, v::Union{Int,Vector{Int}} 
+    getcells(grid::AbstractGrid, set::String)
+
+Returns either all `cells::Vector{C<:AbstractCell}` of a `grid` or a subset based on an `Int`, `Vector{Int}` or `String`.
+Whereas the last option tries to call a `cellset` of the `grid`.
```
termi-official, line 133: See below on `getnodes`.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+"""
 @inline getcells(grid::AbstractGrid) = grid.cells
 @inline getcells(grid::AbstractGrid, v::Union{Int, Vector{Int}}) = grid.cells[v]
 @inline getcells(grid::AbstractGrid, set::String) = grid.cells[collect(grid.cellsets[set])]
```
termi-official, line 137:
```suggestion
@inline getcells(grid::AbstractGrid, setname::String) = grid.cells[collect(getcellset(grid, setname))]
```

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+    getnodes(grid::AbstractGrid) 
+    getnodes(grid::AbstractGrid, v::Union{Int,Vector{Int}}
+    getnodes(grid::AbstractGrid, set::String)
+
+Returns either all `nodes::Vector{Node{dim,T}}` of a `grid` or a subset based on an `Int`, `Vector{Int}` or `String`.
+The last option tries to call a `nodeset` of the `grid`.
+"""
 @inline getnodes(grid::AbstractGrid) = grid.nodes
 @inline getnodes(grid::AbstractGrid, v::Union{Int, Vector{Int}}) = grid.nodes[v]
 @inline getnodes(grid::AbstractGrid, set::String) = grid.nodes[collect(grid.nodesets[set])]
```
termi-official, line 154:
```suggestion
@inline getnodes(grid::AbstractGrid, setname::String) = grid.nodes[collect(getnodeset(grid, setname))]
```

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+"returns an `Int` corresponding to how many nodes are in the `grid`"
 @inline getnnodes(grid::AbstractGrid) = length(grid.nodes)
+"returns an `Int` of how many nodes are in one `cell`"
 @inline nnodes_per_cell(grid::AbstractGrid, i::Int=1) = nnodes(grid.cells[i])
```
termi-official, line 158: Any idea what the purpose of this method is? Looks like the method name does not match its purpose.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
 
+"Accesses the cellset which is mapped to the key `set::String`"
```
termi-official, line 160: Describes the implementation, not the interface.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
 @inline getcellset(grid::AbstractGrid, set::String) = grid.cellsets[set]
+"Returns all cellsets of the `grid`"
 @inline getcellsets(grid::AbstractGrid) = grid.cellsets
 
+"Accesses the nodeset which is mapped to the key `set::String`"
```
termi-official, line 165: Describes the implementation, not the interface.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
 @inline getnodeset(grid::AbstractGrid, set::String) = grid.nodesets[set]
+"Returns all nodesets of the `grid`"
 @inline getnodesets(grid::AbstractGrid) = grid.nodesets
 
+"Accesses the faceset which is mapped to the key `set::String`"
```
termi-official, line 170: Describes the implementation, not the interface.

src/Grid/grid.jl
```diff
@@ -100,27 +123,53 @@ end
...
+
+Returns either all `nodes::Vector{Node{dim,T}}` of a `grid` or a subset based on an `Int`, `Vector{Int}` or `String`.
+The last option tries to call a `nodeset` of the `grid`.
+"""
 @inline getnodes(grid::AbstractGrid) = grid.nodes
 @inline getnodes(grid::AbstractGrid, v::Union{Int, Vector{Int}}) = grid.nodes[v]
 @inline getnodes(grid::AbstractGrid, set::String) = grid.nodes[collect(grid.nodesets[set])]
+"returns an `Int` corresponding to how many nodes are in the `grid`"
 @inline getnnodes(grid::AbstractGrid) = length(grid.nodes)
+"returns an `Int` of how many nodes are in one `cell`"
```
termi-official, line 157: Describes the implementation, not the interface.

src/Grid/grid.jl
```diff
@@ -131,7 +180,22 @@ end
 
 n_faces_per_cell(grid::Grid) = nfaces(eltype(grid.cells))
 
+"""
+    function compute_vertex_values(grid::AbstractGrid, f::Function)
+    function compute_vertex_values(grid::AbstractGrid, v::Vector{Int}, f::Function)    
+    function compute_vertex_values(grid::AbstractGrid, set::String, f::Function)
+
+Given a `grid` and some function `f`, `compute_vertex_values` computes all nodal values,
```
termi-official, line 188: Can we change the name to `compute_nodal_values` now or would this be breaking already?

src/Grid/grid.jl
```diff
@@ -189,6 +269,19 @@ function addcellset!(grid::AbstractGrid, name::String, f::Function; all::Bool=tr
     grid
 end
 
+"""
+    addfaceset!(grid::AbstractGrid, name::String, faceid::Union{Set{FaceIndex},Vector{FaceIndex}})
+    addfaceset!(grid::AbstractGrid, name::String, f::Function; all::Bool=true) 
+
+Adds a `faceset::Dict{String, Set{Tuple{Int,Int}}` to the `grid` with key `name`.
```
termi-official, line 276:
```suggestion
Adds a `faceset::Dict{String, Set{FaceIndex}` to the `grid` with key `name`.
```

src/Grid/grid.jl
```diff
@@ -164,6 +232,18 @@ end
...
+
+Adds a `cellset::Dict{String,Set{Int}}` to the `grid` with key `name`.
+Cellsets can be used to specify a boundary for `Dirichlet`, which is needed by the `ConstraintHandler`
+
+```julia
+addcellset!(grid, "left", Set((1,3))) #add cells with id 1 and 3 to cellset left
+addcellset!(grid, "right", x -> norm(x[1]) < 2.0 ) #add cell to cellset right, if x[1] of each cell's node is smaller than 2.0
+```
+"""
 function addcellset!(grid::AbstractGrid, name::String, cellid::Union{Set{Int},Vector{Int}})
```
termi-official, line 247: Maybe `CellIndex` is more explicit?

src/Grid/grid.jl
```diff
@@ -164,6 +232,18 @@ end
 _check_setname(dict, name) = haskey(dict, name) && throw(ArgumentError("there already exists a set with the name: $name"))
 _warn_emptyset(set) = length(set) == 0 && @warn("no entities added to set")
 
+"""
+    addcellset!(grid::AbstractGrid, name::String, cellid::Union{Set{Int}, Vector{Int}})
+    addcellset!(grid::AbstractGrid, name::String, f::function; all::Bool=true)
+
+Adds a `cellset::Dict{String,Set{Int}}` to the `grid` with key `name`.
+Cellsets can be used to specify a boundary for `Dirichlet`, which is needed by the `ConstraintHandler`
```
termi-official, line 240: Maybe mention MixedDofHandler here too?

src/Grid/grid.jl
```diff
@@ -164,6 +232,18 @@ end
 _check_setname(dict, name) = haskey(dict, name) && throw(ArgumentError("there already exists a set with the name: $name"))
 _warn_emptyset(set) = length(set) == 0 && @warn("no entities added to set")
 
+"""
+    addcellset!(grid::AbstractGrid, name::String, cellid::Union{Set{Int}, Vector{Int}})
+    addcellset!(grid::AbstractGrid, name::String, f::function; all::Bool=true)
+
+Adds a `cellset::Dict{String,Set{Int}}` to the `grid` with key `name`.
```
termi-official, line 239: This doc does not match the interface. Maybe split doc up into two doc strings.

termi-official (changes requested):
First big round. I think a major issue is that we do not specify what the interface does, but leak implementation details of the default grid implementation. We should focus more on what is happening conceptually.

koehlerson:
ready for review again :)

src/Grid/grid.jl
```diff
@@ -2,18 +2,30 @@
...
 getcoordinates(n::Node) = n.x
 
+
+abstract type AbstractCell{dim,N,M} end
 """
+    Cell{dim,N,M} <: AbstractCell{dim,N,M}
 A `Cell` is a sub-domain defined by a collection of `Node`s as it's vertices.
+However, a `cell` is not defined by the nodes but rather by the global node ids.
+The parameter `dim` refers here to the geometrical/ambient dimension, i.e. the dimension of the `nodes` in the grid and **not** the topological dimension of the cell.
+
```
kimauth, line 25: Perhaps add a comment about what `N` and `M` refer to?
```suggestion
A `Cell` has `N` nodes and `M` faces.

```

kimauth (approved):
Looks really good to me!

docs/src/reference/grid.md
```diff
@@ -2,4 +2,108 @@
...
+addcellset!
+addfaceset!
+addnodeset!
+```
+
+## AbstractGrid
+
+It can be very useful to use a grid type for a certain special case, e.g. mixed cell types, adaptivity, IGA, etc.
+In order to define your own `<: AbstractGrid` you need to fulfill the `AbstractGrid` interface.
+In case that certain structures are preserved from the `Ferrite.Grid` type, you don't need to dispatch on your own type, but rather rely on the fallback `AbstractGrid` dispatch.
```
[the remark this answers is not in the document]
koehlerson: I thought it's worth mentioning because of two reasons: on one hand it's the only interface documentation Ferrite.jl has and on the other hand even when I started to subtype an AbstractGrid this wasn't really clear to me, because you need to dig in the source code which functions are defined based on `AbstractGrid` and use `Grid`'s structure and which dispatch only on `Grid` even though`AbstractGrid` could have been used there as well . But I don't mind deleting it, if you want to :)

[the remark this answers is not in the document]
kimauth: Sounds like good reasoning, let's leave it as it is then. :)

PR #308, opened 2020-12-02
added apply_multiple_rhs! for applying boundary condition to a collec…

bcsj:
Added apply_multiple_rhs! for simultaneously applying boundary condition to a collection of rhs-vectors at once.

Since the matrix ``K`` is modified by ``apply!``, if you want to setup to solve ``Ku = f`` for various choices of ``f`` with the same boundary condition, it seems you would need to have a copy of ``K`` every time you want to apply the boundary condition to a new rhs. So I made an apply method which doesn't requiring any copying of ``K``.

Example:
```
# Single rhs situation
apply!(K, f, ch)
u = K \ f

# Multiple rhs situation
apply_multiple_rhs!(K, [f,g,h], ch)
u = K \ f
v = K \ g
w = K \ h
```

If this is completely superfluous and there already is a good way of dealing with this kind of situation, I haven't been able to find it at least.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -374,6 +374,52 @@ function apply!(KK::Union{SparseMatrixCSC,Symmetric}, f::AbstractVector, ch::Con
     end
 end
 
+function apply_multiple_rhs!(KK::Union{SparseMatrixCSC,Symmetric}, F::Vector{<:AbstractVector}, ch::ConstraintHandler, applyzero::Bool=false;
+                strategy::ApplyStrategy=APPLY_TRANSPOSE)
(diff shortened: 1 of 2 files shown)


KristofferC:
Is this sort of the same as https://github.com/KristofferC/JuAFEM.jl/pull/304?

bcsj:
Oh! I suppose it is kind of the same, yes.

fredrikekre:
Thanks for the contribution. Closing since you can already do this as follows:
```
rhsdata = get_rhs_data(ch, K)
apply!(K, ch)

apply_rhs!(rhsdata, f, ch)
u = K \ f

apply_rhs!(rhsdata, g, ch)
v = K \ g

apply_rhs!(rhsdata, h, ch)
w = K \ h
```

PR #316, opened 2021-01-08
WIP: Point evaluations within a JuAFEM grid

jbmuir:
This is a rough draft for a handler to interpolate fields on a JuAFEM grid onto unstructured points - useful for, e.g., the locations of sensors for inverse problems. It is a long way away from being finished but I was asked to create a pull request on the Julia discourse so that people could look at it so far

### DIFF
--- a/src/JuAFEM.jl
+++ b/src/JuAFEM.jl
@@ -5,7 +5,6 @@ using Reexport
 
 using LinearAlgebra
 using SparseArrays
-
 using Base: @propagate_inbounds
 
 include("exports.jl")
@@ -61,6 +60,9 @@ include("L2_projection.jl")
 
 # Export
 include("Export/VTK.jl")
+
+# Point Evaluation
+include(joinpath("PointEval", "PointEvalHandler.jl"))
 
 # Other
 include("deprecations.jl")
--- a/src/PointEval/PointEvalHandler.jl
+++ b/src/PointEval/PointEvalHandler.jl
@@ -1,0 +1,110 @@
+struct PointEvalHandler{dim, C, T, U}
+    dh::DofHandler{dim, C, T}
+    cells::Vector{Int}
+    cellvalues::Vector{U}
+end
+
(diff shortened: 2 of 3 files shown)


koehlerson:
Thanks @jbmuir  
Me and @kimauth  will take a look, since we are both interested in this functionality :)

kimauth:
This is a functionality I would really like to see in JuAFEM/Ferrite!

Here are a few remarks:

- Would your `is_point_inside_cell` work with higher order geometry approximations?
- The idea of a PointEvalHandler looks good to me. However, I don't think we use the functor syntax anywhere else in the package, so I'd suggest to use "regular" functions for it.
- The PointEvalHandler should get a geometry interpolation for finding its cell + the local coordinate and a function interpolation for interpolating the function value. 
- At a point I really fell over storing full CellValues for each point, because that technically also allows to compute derivatives in these points - nothing I think we should encourage to do on points to which a field has been interpolated. However, not storing full CellValues would mean we need to define another type of CellValues that only stores the shape function values. That seems to be a little overkill to me. 
- **Allow for mixed meshes** This should definitely support mixed meshes via the MixedDofHandler. Mainly I think this means that we would need to do a nearest node search for every cellset (i.e. for every FieldHandler which store the cellset). It also means that we need to allow for handing in several field_interpolations and several geometry interpolations, because cells might have different `RefShapes`. I don't see a great way how to relate the interpolations to the cellsets on which they are used yet. Maybe the order must match the order of FieldHandlers, but it seems easy to make mistakes when using this.
- The point evaluation becomes quite powerful for post-processing when combined with the L2Projection, so we should probably revisit that and make sure the outputs from it are nice to hand over the point evaluation, even if e.g. a field does not exist on all cells.

kimauth:
I've been working on this a little, its not ready yet and also missing documentation. It should be working for the `MixedDofHandler` though. In case someone else is interested, it can be found on [ka/pointevaluation](https://github.com/Ferrite-FEM/Ferrite.jl/tree/ka/pointevaluation).

Issue #315, opened 2020-12-18
Subparametric elements with MixedDofHandler and ConstraintHandler does not work:


lijas:
```
using JuAFEM

grid = generate_grid(Quadrilateral, (1,1))

ip_geom = Lagrange{2,RefCube,1}()
ip      = Lagrange{2,RefCube,2}()

field = Field(:u, ip, 2)
fh = FieldHandler([field], Set(1:getncells(grid)))

# 
dh = MixedDofHandler(grid)
push!(dh, fh)
close!(dh)

#
ch = ConstraintHandler(dh)
dbc1 = Dirichlet(:u, getfaceset(grid, "left"), (x, t) -> 0.0, 2)
add!(ch, dbc1)
close!(ch)
update!(ch, 1.0)
```

Error
```
ERROR: AssertionError: length(xh) == n_base_funcs
Stacktrace:
 [1] spatial_coordinate(::JuAFEM.BCValues{Float64}, ::Int64, ::Array{Tensor{1,2,Float64,2},1}) at /home/elias/.julia/packages/JuAFEM/6DTqj/src/FEValues/face_values.jl:253
 [2] _update!(::Array{Float64,1}, ::var"#5#6", ::Set{Tuple{Int64,Int64}}, ::Symbol, ::Array{Int64,1}, ::Array{Int64,1}, ::Array{Int64,1}, ::MixedDofHandler{2,Float64,Grid{2,Cell{2,4,4},Float64}}, ::JuAFEM.BCValues{Float64}, ::Dict{Int64,Int64}, ::Float64) at /home/elias/.julia/packages/JuAFEM/6DTqj/src/Dofs/ConstraintHandler.jl:243
 [3] update!(::ConstraintHandler{MixedDofHandler{2,Float64,Grid{2,Cell{2,4,4},Float64}},Float64}, ::Float64) at /home/elias/.julia/packages/JuAFEM/6DTqj/src/Dofs/ConstraintHandler.jl:214
 [4] top-level scope at REPL[39]:1
```

The error comes from that the FieldHandler creates the BCValues based on the field-interpolation (which in this case is quadratic), and it have no information about the grid (which as linear elements).

PR #317, opened 2021-01-21
Fix subparametric elements in MixedDofHandler

koehlerson:
Fixed https://github.com/KristofferC/JuAFEM.jl/issues/315 and adds regression tests


lijas:
Nice thanks.

In the future it could be nice to make the FieldHandler be associated with a cell-type. That way we get the geometric interpolation "automatically" (via `default_interpolation(celltype)`).
One could also then check that the added fields are compatible with the cell type (so that one does not add a triangular interpolation to a Quadrilateral for example). 

```
fh = FieldHandler(Cell{2,4,4})
push!(fh, Field(:u, Lagrange{2,RefCube,2})) #works
push!(fh, Field(:u, Lagrange{2,RefTetrahedron,2})) #error or warning
```

src/Dofs/MixedDofHandler.jl
```diff
@@ -9,9 +9,9 @@ end
 
 Construct a `FieldHandler` based on an array of `Field`s and assigns it a set of cells.
 """
-function FieldHandler(fields::Vector{Field}, cellset)
+function FieldHandler(fields::Vector{Field}, cellset, ipgeo::Vector{ip}=getproperty.(fields,:interpolation)) where ip <: Interpolation
```
kimauth, line 12: Geometry interpolation is called `geom_interpol` in most of the files, maybe be consistent with that?

src/Dofs/MixedDofHandler.jl
```diff
@@ -9,9 +9,9 @@ end
...
     # TODO for now, only accept isoparamtric mapping
```
kimauth, line 13:
```suggestion

```

kimauth (commented):
We should add a test for the dof distribution of sub/superparametric elements. (I think it should work fine though.)
Otherwise this looks good to be merged to me. :)

kimauth:
> In the future it could be nice to make the FieldHandler be associated with a cell-type. That way we get the geometric interpolation "automatically" (via `default_interpolation(celltype)`).
> One could also then check that the added fields are compatible with the cell type (so that one does not add a triangular interpolation to a Quadrilateral for example).
> 
> ```
> fh = FieldHandler(Cell{2,4,4})
> push!(fh, Field(:u, Lagrange{2,RefCube,2})) #works
> push!(fh, Field(:u, Lagrange{2,RefTetrahedron,2})) #error or warning
> ```

With [#320 ](https://github.com/KristofferC/JuAFEM.jl/pull/320) we at least check this now when the FieldHandler is added to the MixedDofHandler, so the user does get warned when using an incompatible interpolation on a cellset.

koehlerson:
Should we close this? As far as I understand it's not longer needed since `FieldHandler` got rid of the `BCValues`?

kimauth:
Does that mean #315 is resolved? Then we should close that one, too.

lijas:
I think so, but I have not tested if 315 though.

koehlerson:
Ok, so I can update the regression tests of this branch, such that #315 can be closed

PR #320, opened 2021-02-05
MixedDofHandler checks

kimauth:
Add checks when pushing a FieldHandler to the MixedDofHandler:

- _check_same_celltype for the cellset of FieldHandler
- check that RefShape of the fields we're trying to push is the same RefShape as the one of the cells we're pushing the fields to

I accidentally pushed a triangular field to a quad element recently and it did not throw any errors, but corrupted the behavior of my code quite a bit. I think it would make sense to test for the above mentioned things to avoid such problems. This will enforce all cells within a FieldHandler to have the same celltype, which I think makes sense. However, one could discuss if it would be enough to enforce the same RefShape for all fields (then e.g. linear quad and quadratic quads would be allowed together in the same FieldHandler).

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -26,7 +26,7 @@ function Base.getindex(elvec::CellVector, el::Int)
     return elvec.values[offset:offset + elvec.length[el]-1]
  end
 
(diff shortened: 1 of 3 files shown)


lijas:
LGTM.

Maybe add a test with @test_throws to actually know that it throws an error.

PR #325, opened 2021-02-23
Example Integration with DifferentialEquations.jl shown by the Bidomain model with FHN cell model

koehlerson:
Hey together,

@termi-official and me set up an example how to connect JuAFEM with DiffEq. We have some remaining discussion points:

## Remaining discussion points
-  ~~If we add further complex examples where we maybe cite different papers, it is worth to think about to include [DocumenterCitation.jl](https://github.com/ali-ramadhan/DocumenterCitations.jl) as a doc dependency~~ Done.
- The current implementation is very expensive (for an example running in CI) and therefore, we need some sort of workaround
    -  solution 1: parameterize examples in test, if executed by CI reduce element size, end time, etc.
    -  solution 2: advanced examples/integration section without CI testing them (or a subsection as soon as it's supported by Documenter.jl)
- ~~Cassette Integration, which is used by [SparsityDetection.jl](https://github.com/SciML/SparsityDetection.jl), doesn't work at all. Sometimes it returns an error that is too big for the terminal buffer size ~~ Deprecated and not necessary as we can precompute the pattern without much cost.

## TODOs 
- [ ] Correct application of BCs
- [ ] Basic integration with DiffEq's Newton-Raphson backend
- [ ] Multilevel-Newton implementation

Currently blocked by #368 which solves the first two TODOs. Turns out local solves are a concept which is not inherently supported by DiffEq.


termi-official:
Note: This is currently blocked by https://github.com/SciML/OrdinaryDiffEq.jl/issues/1570 . It can be helpful to co-design the nonlinear solver interface with this PR as an example.

KristofferC:
I rebased this and hooked it up to the figure rendering pipeline and VTKHDF, etc with some other tune-ups at https://github.com/Ferrite-FEM/Ferrite.jl/tree/examples/bidomain-differentialequationsjl_rebase. Feel free to use it if it is helpful. Figures for dark mode below:

bidomain-dark

termi-official:
I hope that I will have https://github.com/termi-official/FerriteDiffEq.jl ready soon, so users do not need to fiddle with the boundary condition integration themselves. A major thing missing is still that I need to decouple FerriteOperators from that infrastructure, so I do not force users into using that package.

PR #326, opened 2021-02-26
Mesh Interface Rework

termi-official:
Hi folks, this PR contains some sketches on how we could extend the interface to the finite element mesh (grid). The basic idea is to simply the integration of different kinds of meshes by extending the interface to the mesh. The changes here should not affect any user code directly.

TODO
* [ ] Extend faces, edges, vertices 
* [ ] What about nodes?
* [ ] Investigate interface to cells
* [ ] Example why when this is useful
* [ ] Update Docs

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -85,7 +85,7 @@ end
 
 Applies the boundary condition to the right-hand-side vector without modifying the stiffness matrix
 """
-function apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)	
+function apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)
     K = data.constrained_columns
     @assert length(f) == 0 || length(f) == size(K, 1)
     @boundscheck length(f) == 0 || checkbounds(f, ch.prescribed_dofs)
@@ -339,7 +339,7 @@ function WriteVTK.vtk_point_data(vtkfile, ch::ConstraintHandler)
             dbc.field_name != field && continue
             if eltype(dbc.faces) <: Tuple
                 for (cellidx, faceidx) in dbc.faces
-                    for facenode in faces(ch.dh.grid.cells[cellidx])[faceidx]
+                    for facenode in faces(ch.dh.grid, ch.dh.grid.cells[cellidx])[faceidx]
                         for component in dbc.components
                             data[component, facenode] = 1
                         end
@@ -451,7 +451,7 @@ function add!(ch::ConstraintHandler, fh::FieldHandler, dbc::Dirichlet)
     interpolation = getfieldinterpolations(fh)[field_idx]
     field_dim = getfielddims(fh)[field_idx]
     bcvalue = fh.bc_values[field_idx]
-    
+
     JuAFEM._add!(ch, dbc, dbc.faces, interpolation, field_dim, field_offset(fh, dbc.field_name), bcvalue)
     return ch
 end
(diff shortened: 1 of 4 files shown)


kimauth:
Thanks for giving an example of what you would like to change! Could you maybe also provide a short description of why these changes could be useful?

As far as I understand, the main purpose of adding a grid argument to `vertices` / `edges` /  `faces` would be being able to use unchanged versions of the `DofHandler` etc. with fancier grids, right?

I think I get now why this would be nice changes for using your adaptive grid. However, I have a few doubts with it:

- It seems obscure to me to have a completely unused argument in the "base functionality" of JuAFEM/Ferrite. It is difficult to understand why `vertices` / `edges` /  `faces` would have a grid argument just from reading the package's code.
- You were arguing that  `vertices` / `edges` /  `faces` is part of the interface definition when sub-typing `AbstractGrid` and that as an interface, it should be flexible towards more advanced use cases. What other use cases could we imagine apart from the adaptive mesh refinement? Would they require the same type of syntax changes? Maybe @lijas can say something about other sub-typed grids?
- When it comes to more advances grids, can the `DofHandler`/`MixedDofHandler` be trusted to work or would they likely need to be carefully reviewed anyways?

Finally, I'd like to hear @fredrikekre 's / @KristofferC 's  opinion about more interface like function definitions. :)

JimBrouzoulis:
I am also interested in hearing a bit more about why the changes in the interfaces are required/beneficial.
For example Kim's first point, I would like to see how the `grid` in `vertices(grid::AbstractGrid, ...)` is used in your context.

termi-official:
First, thanks for the interest!

Regarding Kim's 1st point, I think we agree here, but this will be more clear when I find time to give specific code examples for when exactly this becomes handy will be provided. My idea here was to provide a structured grid example and a "grid view", which I think makes the change very clear. The grid view will make more sense once we have a proper tree-based grid implementation ready, but this will take more time. This "grid view" can e.g. be useful in FAC schemes, where we have finite element subspaces defined on "intermediate refinement levels". 

Regarding Kim's 2nd and Mr Brouzoulis point - yes, you understand correctly that the main benefit is that we can use "fancier" grids. Another plus is that we can introduce some additional changes to the internals to allow full custom grids e.g a trace mesh of the original grid (for simulations requiring a trace space) or a structured grid. I think further about allowing the utilization of exernal grid implementations (e.g. in C++ SAMRAI or PUMI), which have great performance for special use-cases.
I believe this is also useful in e.g. distributed (adaptive) simulations. Here the mesh handling can become quickly complicated and there are lots of ways how to do it. So the basic idea is to allow using different, swappable "backends" for each the grid, which are optimized for their use-case. Note that most of the code can be reused (e.g. dof numbering, assembly, ...), if the abstraction layers are properly designed and if we implement the code with the [finite element operator decomposition](https://ceed.exascaleproject.org/ceed-code/) in mind. 

Now Kim's 3rd point. While I can not give a single example where the dof handlers would mess up, I also do not think we have a strong guarantee that, that it will work for every single thinkable use case. However, I am confident that we can reuse the existing dof numbering logic in at least almost all of the FEM based methods.

To also answer Mr Brouzoulis' second point: One practical example can be found in [a preliminary implementation on tree-based NC meshes](https://ceed.exascaleproject.org/ceed-code/) of two of my students. I think the relevant code changes to make the "adaptivity" work were

* Changes to faces, vertices, edges
https://github.com/Ljon4ik4/FEMRefinements.jl/blob/0fe2b71e7f7cfe2e3186c239a0ad69daceae4138/src/grid.jl#L186-L203
* Changes to close (here we just changed the calls to ``faces, vertices, edges``)
https://github.com/Ljon4ik4/FEMRefinements.jl/blob/0fe2b71e7f7cfe2e3186c239a0ad69daceae4138/src/utils.jl#L61-L113
* An example for the heat equation on an NC grid with NC level 1
https://github.com/Ljon4ik4/FEMRefinements.jl/blob/9e01c780a62b8ce4c4b378947a590c446c3b9caa/examples/HeatEquationMixed.jl

Note that this is "just" a prototype implemented by two of my students (which I think is really a great result by them for their first larger project!). Do not understand me wrong here, we can also get this one up and running without changing the interface, but this would come at the price of sacrificing efficiency in the synchroniozation logic of non-conforming entities (in this tree-based approach). 

I hope this makes some sense and helps.

PR #331, opened 2021-03-22
Improvements to L2 projection

fredrikekre:
Slight change of API.
Old:
```julia
proj = L2Projector(qr, ip, mesh)
point_vars = project(qp_values, proj)
```
New:
```julia
proj = L2Projector(ip, mesh)
point_vars = project(proj, qp_values, qr)
```

The main advantages, IMO, are that (i) the new API uses kwargs for "special stuff" (like using it for a specific cell set), (ii) better default quad rule for the lhs, and (iii) that the qr is bundled with the values in the call to project.

Also adds support for scalar fields.

Should add some more docs.

### DIFF
--- a/docs/src/literate/l2_projection.jl
+++ b/docs/src/literate/l2_projection.jl
@@ -41,10 +41,10 @@ end
 q = compute_heat_fluxes(cellvalues, dh, u);
 
 # Next, create an L2-projector using the same interpolation as was used to approximate the temperature field. On instantiation, the projector assembles the coefficient matrix `M` and computes the Cholesky factorization of it. By doing so, the projector can be reused without having to invert `M` every time.
-projector = L2Projector(cellvalues, ip, grid);
+projector = L2Projector(ip, grid);
 
 # Project the integration point values to the nodal values
-q_nodes = project(q, projector);
+q_nodes = project(projector, q, qr);
 
 
 # ## Exporting to VTK
--- a/docs/src/reference/export.md
+++ b/docs/src/reference/export.md
@@ -7,3 +7,8 @@ DocTestSetup = :(using Ferrite)
 ```@docs
 vtk_grid
 ```
+
+```@docs
+L2Projector
+project
+```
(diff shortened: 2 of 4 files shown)


lijas:
Point 3 is very nice. Before this PR i guess it was not possible to use the same `L2Projector` for different values which had different qr-rules.

fredrikekre:
Preview docs: https://ferrite-fem.github.io/Ferrite.jl/previews/PR331/reference/export/#Ferrite.L2Projector

kimauth (commented):
Looks really good to me Fredrik!

I think the `_mass_qr` still underintegrates for higher order tetrahedral interpolations.
Eg. the following leads to a `PosDefException`:
```
nodes = [Node((0.0, 0.0)),
         Node((1.0, 0.0)),
         Node((0.0, 1.0))]
cells = [Triangle((1,2,3))]
mesh = Grid(cells, nodes)

ip = Lagrange{2,RefTetrahedron,2}()

projector = L2Projector(ip, mesh)
```
but if we do 
```
projector = L2Projector(ip, mesh; qr_lhs=QuadratureRule{2,RefTetrahedron}(4))
```
instead, it works. Maybe we should at least warn about it.

lijas:
Instead of _mass_qr, maybe we could have a default_quadraturerule(::Interpolation), where the default version is the same as _mass_qr, but can  e overloaded for quadratic tetras.

fredrikekre:
Interesting, (nervously looking at my research code....). I fixed it instead of warning.

fredrikekre:
I'll merge this for now, but I thought that order `n` integration should exactly integrate polynomials of order `2n-1`. So order 3 should be enough to integrate polynomials of degree 5, and the mass matrix for quadratic functions should be of order 4. This works for the `RefCube` elements, so maybe we have chosen bad integration points for `RefTetra` in both 2 and 3 dimensions?

koehlerson:
the formula `2n-1` is definitely correct and I agree that order 3 should be sufficient to integrate a 4th order term. If the latter is true then 

image

JimBrouzoulis:
The `2n-1`-formula is not valid for triangles and tetrahedrons since these domains are not simple tensor products like squares and cubes (`dim 1 x dim 2 ...`). In OOFEM we tabulated more points in order to be able to integrate higher-order polynomials, see: 

http://mech.fsv.cvut.cz/oofem/resources/doc/oofemrefman/gaussintegrationrule_8C_source.html

koehlerson:
TIL, thanks for clarifying :+1:

PR #332, opened 2021-03-22
bugfix on CellIterator with given cellset

kimauth:


fredrikekre:
I guess this code path could use a test then?

kimauth:
I can't find any tests for the CellIterator. Is is possibly hidden within another test or do we never test it at all?

fredrikekre:
Since before this patch it errored I guess that branch (e.g. CellIterator over a specific cellset) is not tested at all.

kimauth:
Yeah, I was going to add a test for that branch after you commented. But then I couldn't find any tests for the CellIterator (not just for this branch). So I was wondering if we really never test it or if I am just missing where.

Issue #336, opened 2021-03-27
"Optimized" `function_divergence` is slow


fredrikekre:
e.g. https://github.com/Ferrite-FEM/Ferrite.jl/blob/d04709f73f1893d5e50ac9fe6e3127d4e4faabea/src/FEValues/common_values.jl#L231-L243

is 2x slower than just `tr(function_gradient(args...))`.

```julia
using Ferrite

qr = QuadratureRule{3,RefTetrahedron}(2)
linear = Lagrange{3,RefTetrahedron,1}()
quadratic = Lagrange{3,RefTetrahedron,2}()
cv = CellVectorValues(qr, quadratic, linear)
grid = generate_grid(Tetrahedron, (1,1,1))
xe = getcoordinates(grid, 1)
reinit!(cv, xe)
ue = rand(getnbasefunctions(cv))
tr_grad(cv, ue) = tr(function_gradient(cv, 1, ue))

using BenchmarkTools
@btime function_divergence($cv, 1, $ue)
@btime tr_grad($cv, $ue)
```
Gives, on my machine:
```julia
julia> @btime function_divergence($cv, 1, $ue)
  57.305 ns (0 allocations: 0 bytes)
-1.1384832740838906

julia> @btime tr_grad($cv, $ue)
  32.581 ns (0 allocations: 0 bytes)
-1.1384832740838908
```

PR #337, opened 2021-03-27
some small optimization tweaks

KristofferC:
Some small stuff I found when looking a bit into #336

### DIFF
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -1,5 +1,7 @@
 # Common methods for all `Values` objects
 
+using Base: @propagate_inbounds
+
 const ScalarValues{dim,T,shape} = Union{CellScalarValues{dim,T,shape},FaceScalarValues{dim,T,shape}}
 const VectorValues{dim,T,shape} = Union{CellVectorValues{dim,T,shape},FaceVectorValues{dim,T,shape}}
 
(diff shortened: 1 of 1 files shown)


src/FEValues/common_values.jl
```diff
@@ -228,13 +230,15 @@ function function_divergence(fe_v::ScalarValues{dim}, q_point::Int, u::AbstractV
...
-function function_divergence(fe_v::VectorValues{dim}, q_point::Int, u::AbstractVector{T}, dof_range::UnitRange = 1:length(u)) where {dim,T}
+function function_divergence(fe_v::VectorValues{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
     n_base_funcs = getn_scalarbasefunctions(fe_v)*dim
     @assert length(dof_range) == n_base_funcs
     @boundscheck checkbounds(u, dof_range)
     diverg = zero(T)
-    @inbounds for (i, j) in enumerate(dof_range)
-        grad = shape_gradient(fe_v, q_point, i)
+    basefunc_count = 1
+    @inbounds for j in dof_range
```
fredrikekre, line 239: `enumerate` is used in some other places too, but did this actually make a difference with the other changes?

fredrikekre (approved):

src/FEValues/common_values.jl
KristofferC: Not really, just made the code slightly smaller to look at.

kimauth:
I have few questions just for for the purpose of learning about it. :)

- Do we want use `@propagate_inbounds` instead of `@inline` because these functions are typically used in loops where also using `@inbounds` is a good idea?
- What is better about `eachindex(u)` compared to `1:length(u)`?

KristofferC:
1) yes, we use `@inbounds` for many of the loops where they are called.

2) `eachindex` gives for normal arrays a `OneTo` range where the compiler knows from the type that the first index is 1. When things are inlines and constant propagated it shouldn't matter but it looked like that doesn't happen here (you can look at `@code_warntype` for example).

PR #341, opened 2021-03-30
Add disassemble!(ue, u, dofs) as an optimized version of ue .= u[dofs].

fredrikekre:


src/assembler.jl
```diff
@@ -208,3 +208,23 @@ function InsertionSort!(A, order, ii=1, jj=length(A))
...
+    disassemble!(ue, u, dofs)
+
+Fill the element vector `ue` with values from the global vector `u`.
+
+This function is equivalent to `ue .= u[dofs]` but without allocating
+the intermediate array.
+"""
+@propagate_inbounds function disassemble!(ue::AbstractVector, u::AbstractVector, dofs::AbstractVector{Int})
+    @boundscheck begin
+        length(ue) == length(dofs) || throw(BoundsError())
```
KristofferC, line 222: I don't think this is worth it. It will only prevent work in the case that the code will error and in that case, the performance doesn't really matter.

src/assembler.jl
```diff
@@ -208,3 +208,23 @@ function InsertionSort!(A, order, ii=1, jj=length(A))
...
+        checkbounds(u, dofs)
+    end
```
fredrikekre, line 224:
```suggestion
    @boundscheck checkbounds(u, dofs)
```

JimBrouzoulis:
Is  `disassemble!` the best name? To me it sounds like this method would be the inverse of `assemble!` and in that case, it should actually remove elements from `u`.

fredrikekre:
I don't know. `unpack`? `extract`?

kimauth:
`extract` sounds right to me.

fredrikekre:
I kinda like the symmetry with `assemble` though...

JimBrouzoulis:
I vote for `extract`, `disassemble` implies a destructive action to me... but I won't lie sleepless over the name choice 😄 

Two follow-up questions: 
* Isn't there already something similar in Julia, simply extracting a few elements and copying them to a given array? Sounds like something one wants to do quite often.
* I guess this is more efficient (and safer!) than using a `view`? But should one expect a large difference compared to `
ue[:] = u[dofs]`?

KristofferC:
I did a MWE benchmark:

```jl
u = [1,2,3,4,5,6,7,8,9,10]
ue = [0, 0, 0, 0]
idxs = [1,3,5,8]
function assemble1!(ue, u, idxs) 
    @inbounds for (i, j) in enumerate(idxs)
        ue[i] = u[j]
    end
end
assemble2!(ue, u, idxs) = (ue .= getindex.(Ref(u), idxs); nothing)
assemble3!(ue, u, idxs) = (ue .= @view u[idxs]; nothing)
using BenchmarkTools
@btime assemble1!($ue, $u, $idxs)
@btime assemble2!($ue, $u, $idxs)
@btime assemble3!($ue, $u, $idxs)
```

```jl
julia> @btime assemble1!($ue, $u, $idxs)
  2.906 ns (0 allocations: 0 bytes)

julia> @btime assemble2!($ue, $u, $idxs)
  8.864 ns (0 allocations: 0 bytes)

julia> @btime assemble3!($ue, $u, $idxs)
  16.277 ns (0 allocations: 0 bytes)
```

fredrikekre:
Right, it would be nice if `ue .= @view u[dofs]` (or `ue .= u.[dofs]`, if that was a thing) would be as efficient, but it isn't right now.

lijas:
In some literatures i have seen the terms `gather!(ue, u, idxs)` and `scatter!(assembler, Ke, fe, idxs)`, been used. Don't know if I like them though :P

fredrikekre:
`gather` is basically a synonym to `assemble` so that seems confusing

KristofferC:
> gather is basically a synonym to assemble so that seems confusing


To me, `scatter` is more like assemble. You take something small and insert it into something big.

src/exports.jl
```diff
@@ -119,61 +89,40 @@ export
...
     UpdateFlags,
     cellid,
-    interfacedofs,
 
 # assembly
     start_assemble,
     assemble!,
-    finish_assemble,
+    end_assemble,
+    disassemble!,
```
termi-official, line 116: Maybe we can be direct here?

```suggestion
    load_local_dofs!,
```

KnutAM: Could also just recommend `map!(i -> u[i], ue, idxs)`, seems to be equivalently fast?

termi-official: I wouldn't recommend such constructs for users tbh, as this is probably not easy to read for most users. Not much performance difference either.

```
julia> @btime assemble1!($ue, $u, $idxs)
  7.688 ns (0 allocations: 0 bytes)

julia> @btime assemble4!($ue, $u, $idxs)
  8.659 ns (0 allocations: 0 bytes)
```

KnutAM: > Not much performance difference either.

Yes, my point was that this was equally good as a special function. But it is really sad that there is no `copyto!(ue, u, ids)` function, or that the view-variant is better...

KnutAM: Could provide a function that makes it fast to do `copyto!` with views (not sure if this could be upstreamed to `Base`...). Added benchmark with and without checking the keys to be comparable (no check) with `assemble1!`...

```julia
function fast_copyto!(dst::AbstractArray, src::AbstractArray)
    keys(dst) == keys(src) || copyto!(dst, src) # key check
    @inbounds for (i, v) in pairs(src)
        dst[i] = v
    end
    return dst
end

function assemble1!(ue, u, idxs) 
    @inbounds for (i, j) in enumerate(idxs)
        ue[i] = u[j]
    end
end

using BenchmarkTools, Random

for N in (10, 100, 1000, 10000)
    println("N = ", N)
    for r in (1, 2, 10, 100)
        n = max(N ÷ r, 1)
        ae = zeros(n)
        a = rand(N)
        idx = randperm(N)[1:n] 
        av = view(a, idx)
        println("n = ", n)
        @btime copyto!($ae, $av)
        @btime fast_copyto!($ae, $av)
        @btime assemble1!($ae, $a, $idx)
    end
end
```
<details> <summary> Results with key check </summary>

```julia
N = 10
n = 10
  9.218 ns (0 allocations: 0 bytes)
  6.291 ns (0 allocations: 0 bytes)
  5.875 ns (0 allocations: 0 bytes)
n = 5
  7.959 ns (0 allocations: 0 bytes)
  5.125 ns (0 allocations: 0 bytes)
  4.791 ns (0 allocations: 0 bytes)
n = 1
  6.792 ns (0 allocations: 0 bytes)
  4.208 ns (0 allocations: 0 bytes)
  2.916 ns (0 allocations: 0 bytes)
n = 1
  7.333 ns (0 allocations: 0 bytes)
  4.208 ns (0 allocations: 0 bytes)
  2.916 ns (0 allocations: 0 bytes)
N = 100
n = 100
  36.211 ns (0 allocations: 0 bytes)
  29.899 ns (0 allocations: 0 bytes)
  28.350 ns (0 allocations: 0 bytes)
n = 50
  21.480 ns (0 allocations: 0 bytes)
  17.702 ns (0 allocations: 0 bytes)
  15.949 ns (0 allocations: 0 bytes)
n = 10
  9.208 ns (0 allocations: 0 bytes)
  6.292 ns (0 allocations: 0 bytes)
  5.833 ns (0 allocations: 0 bytes)
n = 1
  7.333 ns (0 allocations: 0 bytes)
  4.334 ns (0 allocations: 0 bytes)
  2.916 ns (0 allocations: 0 bytes)
N = 1000
n = 1000
  329.813 ns (0 allocations: 0 bytes)
  259.717 ns (0 allocations: 0 bytes)
  259.707 ns (0 allocations: 0 bytes)
n = 500
  168.423 ns (0 allocations: 0 bytes)
  136.630 ns (0 allocations: 0 bytes)
  135.357 ns (0 allocations: 0 bytes)
n = 100
  36.170 ns (0 allocations: 0 bytes)
  29.899 ns (0 allocations: 0 bytes)
  28.322 ns (0 allocations: 0 bytes)
n = 10
  8.884 ns (0 allocations: 0 bytes)
  6.291 ns (0 allocations: 0 bytes)
  5.875 ns (0 allocations: 0 bytes)
N = 10000
n = 10000
  3.297 μs (0 allocations: 0 bytes)
  2.662 μs (0 allocations: 0 bytes)
  2.699 μs (0 allocations: 0 bytes)
n = 5000
  1.637 μs (0 allocations: 0 bytes)
  1.250 μs (0 allocations: 0 bytes)
  1.254 μs (0 allocations: 0 bytes)
n = 1000
  331.858 ns (0 allocations: 0 bytes)
  260.477 ns (0 allocations: 0 bytes)
  258.639 ns (0 allocations: 0 bytes)
n = 100
  36.128 ns (0 allocations: 0 bytes)
  29.857 ns (0 allocations: 0 bytes)
  28.308 ns (0 allocations: 0 bytes)
```
</details>

<details> <summary> Results without key check </summary>

```julia
N = 10
n = 10
  9.175 ns (0 allocations: 0 bytes)
  4.625 ns (0 allocations: 0 bytes)
  5.833 ns (0 allocations: 0 bytes)
n = 5
  8.000 ns (0 allocations: 0 bytes)
  3.542 ns (0 allocations: 0 bytes)
  4.791 ns (0 allocations: 0 bytes)
n = 1
  7.291 ns (0 allocations: 0 bytes)
  1.916 ns (0 allocations: 0 bytes)
  2.958 ns (0 allocations: 0 bytes)
n = 1
  7.333 ns (0 allocations: 0 bytes)
  1.917 ns (0 allocations: 0 bytes)
  2.917 ns (0 allocations: 0 bytes)
N = 100
n = 100
  36.337 ns (0 allocations: 0 bytes)
  27.051 ns (0 allocations: 0 bytes)
  28.350 ns (0 allocations: 0 bytes)
n = 50
  19.684 ns (0 allocations: 0 bytes)
  15.823 ns (0 allocations: 0 bytes)
  15.974 ns (0 allocations: 0 bytes)
n = 10
  9.208 ns (0 allocations: 0 bytes)
  4.625 ns (0 allocations: 0 bytes)
  5.875 ns (0 allocations: 0 bytes)
n = 1
  7.333 ns (0 allocations: 0 bytes)
  1.916 ns (0 allocations: 0 bytes)
  2.917 ns (0 allocations: 0 bytes)
N = 1000
n = 1000
  332.258 ns (0 allocations: 0 bytes)
  255.045 ns (0 allocations: 0 bytes)
  258.476 ns (0 allocations: 0 bytes)
n = 500
  168.534 ns (0 allocations: 0 bytes)
  133.666 ns (0 allocations: 0 bytes)
  136.143 ns (0 allocations: 0 bytes)
n = 100
  36.128 ns (0 allocations: 0 bytes)
  27.051 ns (0 allocations: 0 bytes)
  28.308 ns (0 allocations: 0 bytes)
n = 10
  9.416 ns (0 allocations: 0 bytes)
  4.625 ns (0 allocations: 0 bytes)
  5.833 ns (0 allocations: 0 bytes)
N = 10000
n = 10000
  3.297 μs (0 allocations: 0 bytes)
  2.662 μs (0 allocations: 0 bytes)
  2.699 μs (0 allocations: 0 bytes)
n = 5000
  1.650 μs (0 allocations: 0 bytes)
  1.246 μs (0 allocations: 0 bytes)
  1.258 μs (0 allocations: 0 bytes)
n = 1000
  332.226 ns (0 allocations: 0 bytes)
  255.282 ns (0 allocations: 0 bytes)
  259.633 ns (0 allocations: 0 bytes)
n = 100
  36.463 ns (0 allocations: 0 bytes)
  27.052 ns (0 allocations: 0 bytes)
  28.308 ns (0 allocations: 0 bytes)
```
</details>

PR #343, opened 2021-03-30
Allow boundary condition on edges/vertices

lijas:
(Open new PR so I can view documentation)

Summary:

Makes it possible to add bcs on edges and vertices.
- Introduce EdgeIndex and VertexIndex, similar to FaceIndex.
- Generelize BCValues to also work with edges and vertices. 
- The BCValues are now handled by the ConstraintHandler (and we can remove bcvalues-field from the DofHandler now if we want).
- Add shell tutorial in doc: https://ferrite-fem.github.io/Ferrite.jl/previews/PR343/examples/linear_shell/

I have nothing more to add to this PR, so if it looks good I think it can be merged.


kimauth:
Does this also allow to add `Dirichlet` bcs to nodesets in general?

lijas:
That was already possible right? I dont think this affects bcs on nodesets, the tests does not fail.

kimauth:
It is possible only for the `DofHandler` I think. I did a PR to add that for the `MixedDofHandler` a while ago, but I think it was never reviewed (and its not very well programmed...).

Do you think we could unify the handling of adding bcs to nodes between your suggestion and the current implementation?

lijas:
I can see if I can more or less copy your #294 in to mine and see if it works. Are you using #294 in your work now?

kimauth:
I am, but in that case I can go back to it, update it and merge #294 itself. I was more thinking if what you are doing was somewhat similar and if it would make sense to use the same approach for both.

PR #345, opened 2021-04-01
correct function_gradient docstring

kimauth:
According to the old docstring, `function_gradient` for vector valued functions does ∇N ⊗ u, while it really does u ⊗ ∇N. This makes a difference in which vector determines the columns and which one determines the rows.

### DIFF
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -98,9 +98,8 @@ for the degrees of freedom. For a scalar valued function, `u` contains scalars.
 For a vector valued function, `u` can be a vector of scalars (for use of `VectorValues`)
 or `u` can be a vector of `Vec`s (for use with ScalarValues).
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
Should probably be clarified that this is when you use scalar base functions for a vector valued problem. Also, on line 138 `function_scalar_gradient` should be `function_gradient`.

kimauth:
But the meaning of the docstring doesn't really change depending on whether it is `ScalarValues` or `VectorValues`, does it? For me the main take-away is that in the `function_gradient`  `u` spans the rows and `x` spans the columns. Or what would you document differently for vector base function?

fredrikekre:
I mean that typically you would use `VectorValues` so then the first expression is still what is happening, just that the `N_i`s are vectors (meaning that `\nabla N_i`s are tensors).

kimauth:
Modified the text a little, but I'm not so certain this actually makes it more understandable. What do you think @fredrikekre ?

PR #346, opened 2021-04-01
support adding Dirichlet BCs to nodesets

kimauth:
Allow to add Dirichlet BCs to nodesets for the `MixedDofHandler`.

Also allows to add boundary conditions to a `FieldHandler` even if not all nodes/faces are part of the corresponding cellset. In that case a warning will be given and the node/face will simply be skipped. Considering that most mesh generators give nodesets in the format of "left nodes" and not seperated according to quad/tri/etc. elements, I think this is a lot more convenient.

@lijas : Now I did mess with functions that you are touching in #343, but it is small changes.

This replaces #294.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -85,7 +85,7 @@ end
 
 Applies the boundary condition to the right-hand-side vector without modifying the stiffness matrix
 """
-function apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)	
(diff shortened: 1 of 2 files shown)


lijas:
I think this looks good, just some weird spaces/tabs in the code that made line be misaligned. 

I think it is ok to warn (rather than give an error) if an user adds constraints to a faceset/nodeset that is not included in the given FieldHandler.

kimauth:
Weird that line doesn't look misaligned at all in my editor.

KristofferC:
:tada:

fredrikekre:
> Weird that line doesn't look misaligned at all in my editor.

You probably have the tab-width set to 4.

PR #350, opened 2021-06-04
use StableRNGs for testing the sha of the exported files

KristofferC:
Prevents tests from error when Julia's RNG change.


test/test_grid_dofhandler_vtk.jl
```diff
@@ -68,8 +69,8 @@ end
         end
         close!(ch)
         update!(ch, 0.0)
-        Random.seed!(1234)
-        u = rand(ndofs(dofhandler))
+        rng = StableRNG(1234)
+        u = rand(rng, ndofs(dofhandler))
```
fredrikekre, line 73: If this is the only place rng is used perhaps just `collect(1:ndofs(dofhandler))`?

vtjnash:
This will fix https://github.com/Ferrite-FEM/Ferrite.jl/issues/351

PR #354, opened 2021-07-28
refactor vtk_point_data(vtkfile, dh, u) for access to its reordering outside of vtk export

kimauth:
I think it would be useful to have access to the reordering of the variables in nodal order outside of the vtk export. This is literally just refactoring, shouldn't change anything about how it works (apart from introducing one additional test).


KristofferC:
> (apart from introducing one additional test).

I don't see the added test.

kimauth:
Ah sorry, bad formulation. Not a test, just an added check in the `reshape_to_nodes` function ([here](https://github.com/Ferrite-FEM/Ferrite.jl/pull/354/files#diff-36df0c8049395008c0c59acc9b663b3c17f0ed2e7231b3667f1f57e1730ce78bR117).

kimauth:
I also refactored the function for the `DofHandler` and reduced code duplication between the `DofHandler` and `MixedDofHandler` functions a little on the way. 
Additionally, there indeed are two new tests now for the `reshape_to_nodes` functions. Also exported that one and added a docstring. 
Not sure if `reshape_to_nodes` is a good name or if it should be called something else. It takes a vector with dof-values and returns a matrix with the values in nodal order for a given field.

fredrikekre (approved):

PR #357, opened 2021-08-02
Add Contributing and Questions section to README

kimauth:
During the Gender Diversity BoF the point was raised that explicit information about how to reach out to package maintainers in e.g. the readme makes it a lot easier to actually reach out with questions / report issues etc., especially for newcomers. Not only valid for people from underrepresented genders of course, this should put the bar a bit lower for everyone. 
Here is a suggestion, what we could write. I took some inspiration from the Tensors.jl readme. @koehlerson suggested that some people might like Zulip better than Slack, so he opened a Zulip channel today. 

I think it would be great to incorporate such a section, please feel free to suggest changes to it if you'd like anything to be changed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ You can install Ferrite from the Pkg REPL:
 pkg> add Ferrite
 ```
 
+## Contributing and Questions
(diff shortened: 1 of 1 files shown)


KristofferC (approved):
Nice. Are the trailing backslashes needed?

kimauth:
They're for making the line breaks, I believe it could be two spaces instead.

PR #360, opened 2021-08-04
Extract dofs on face

lijas:
I think it is pretty common to want to extract the dofs on some face in order to calculate for example the reaction forces on a boundary. This PR adds a function similar to` celldofs!(dofs, dh, cellid)`, called `facedofs!(dofs, dh, FaceIndex(cellid, faceid))`.

```
grid = generate_grid(Hexahedron, (3,3,3))
dh = DofHandler(grid)
push!(dh, :u, 2)
close!(dh)

faceset = collect(getfaceset(grid, "left"))

fdofs = zeros(Int, 8) #number of dofs on a face
Ferrite.facedofs!(fdofs, dh, faceset[1])
```

This PR could also make it possible integrate for example external forces on boundaries using the face-interpolation directly, instead of using the cell-interpolation + `FaceVectorValues`  which we are currently using.

This was just a concept I wanted to try out, let me know if this is something that could be usefull, and then I can work some more on it :)

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -23,11 +23,12 @@ struct DofHandler{dim,C,T} <: AbstractDofHandler
     closed::ScalarWrapper{Bool}
     grid::Grid{dim,C,T}
     ndofs::ScalarWrapper{Int}
+    local_facedofs::CellVector{Int}
 end
 
 function DofHandler(grid::Grid)
     isconcretetype(getcelltype(grid)) || error("Grid includes different celltypes. Use MixedDofHandler instead of DofHandler")
-    DofHandler(Symbol[], Int[], Interpolation[], BCValues{Float64}[], Int[], Int[], ScalarWrapper(false), grid, Ferrite.ScalarWrapper(-1))
+    DofHandler(Symbol[], Int[], Interpolation[], BCValues{Float64}[], Int[], Int[], ScalarWrapper(false), grid, Ferrite.ScalarWrapper(-1), CellVector(Int[],Int[],Int[]))
 end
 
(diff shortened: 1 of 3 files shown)


KristofferC:
> I think it is pretty common to want to extract the dofs on some face in order to calculate for example the reaction forces on a boundary.

Isn't this done by just integrating over that boundary with a normal face integral (using `FaceValues`)?

koehlerson:
I think that's a good functionality which is also useful for other things e.g. adaptivity: https://github.com/Ljon4ik4/FEMRefinements.jl/blob/main/src/utils.jl#L273-L277
We built it back then a bit simpler but in the end we need them to construct a restriction operator.

lijas:
our code does the same think, I just cache the local dofs in the dofhandler to speed up the extraction a bit.

koehlerson:
ye exactly, your version is better, that's what I meant with "We built it back then a bit simpler" :)

kimauth:
Maybe this could be extended so that we could boundary conditions like the one on the fluid-solid interface [here](https://www.dealii.org/current/doxygen/deal.II/step_46.html)? I don't really see a straight forward way how to do this at the moment, as it requires quantities from both sides of the interface.

termi-official:
Maybe I miss something, but is something wrong in using the dicts from [__close!](https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Dofs/DofHandler.jl#L127) ?

termi-official:
It is rather simple possible to extract facedofs without the additional bookmarking : https://github.com/koehlerson/FerriteVis.jl/blob/do/discontinuous-vis/src/utils.jl#L248-L251 .

Issue #352, opened 2021-06-30
Implementing Discontinuous Lagrange Interpolations


bhaveshshrimali:
Hi,

It would be helpful if discontinuous interpolations on simplices could be supported. I am taking a look at the codebase. If someone could guide me on the relevant modules that need to be changed, I could take an attempt at implementing those. Thanks!

kimauth:
Hi Bhavesh,
you probably found the [interpolations.jl](https://github.com/Ferrite-FEM/Ferrite.jl/blob/b73c61c43cd9d799753675a98bc48165d342406c/src/interpolations.jl) file already. You could just define a new subtype of `Interpolation{dim,shape,order}`. Depending on your interpolation and use-case, you might then want to have a look at the [CellValues](https://github.com/Ferrite-FEM/Ferrite.jl/blob/b73c61c43cd9d799753675a98bc48165d342406c/src/FEValues/cell_values.jl), which store pre-computed shape function values and derivatives. But if your interpolation is similar to the existing `Lagrange` interpolations, the `CellValues` might work out of the box. (I use a custom subtype of `Interpolation` for cohesive zone modelling and it didn't need much change.)
Eventually you also want to check out the `DofHandler` or `MixedDofHandler` if you need a non-standard dof-distribution. In that case it is useful to notice that the `InterpolationInfo` is used for the dof-distribution.

koehlerson:
You can get some inspiration from https://github.com/Paulms/HDiscontinuousGalerkin.jl/, especially https://github.com/Paulms/HDiscontinuousGalerkin.jl/blob/master/src/basis.jl. Further you can check the modifications in the dofhandler and what additional information are stored in the `FESpace` which is a related entity to `CellValues`

bhaveshshrimali:
> Hi Bhavesh,
[...]

PR #363, opened 2021-08-14
Topology information in grid

koehlerson:
Quick sketch how I would construct some topology information. Currently cells with more than one corner neighbor don't work, e.g. `generate_grid(Triangle,(2,2))` doesn't work, but `generate_grid(Triangle,(2,1)` works. Quads (and hex, I think?) should work in general, same as mixed grid, except for the multiple corner neighborhood. New testset shows how it works. Open for feedback and discussions. 

If you don't like this at all, then I wouldn't pursue to get a workaround for the corner neighborhood.


JimBrouzoulis:
Nice!
Although, wouldn't it be preferable to have a lazy constructor for the topology? So you don't get the overhead of constructing it when not using it (thinking of larger meshes).

koehlerson:
That's true! I can add this, for sure.

 However, before adding it, I'm still thinking about how to handle the neighborhood information properly. The plain `Neighbor` type doesn't feel right yet, because for mixed grids you (kind of) loose the information what kind of neighbor you have. 

https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files#diff-796a5ea6f5cdd76d08a6b0a07f1bd1813fa58607624b572ef71f1cd8ffdddf8aR260-R265

The only way to notice this is by checking the symmetry of the matrix. I'm currently thinking about changing the `Vector{Tuple{Int,Int}}` in `Neighbor` to `Vector{T} where T <: GeometricEntity` and `GeometricEntity` gets `Corner`, `Edge` and `Face` subtypes, by this we could nicely dispatch on them for e.g. `reinit!`. But still, this doesn't feel 100% right, because I don't know how to handle the "no neighbor" case (currently (0,0) tuple), only thing that comes to my mind is to have a dummy subtype struct or a union in the vector of `Neighbor`

koehlerson:
I think this version is quite good. The only question for me is now how I should include the option to construct the topology for all `grid_generator` functions, should I have a arg or kwarg bool switch?

termi-official (commented):
I am leaving some commentary on the draft because @koehlerson asked explicitly. :)

While I like this draft in general I think we have to put more work into this. Especially we should think about the requirements. Different techniques in FEM require different topology informations. For example some error estimators evalute the jump in the gradients between adjacent elements, hence we need some table to associate half-faces back to the adjacent elements (if we want to do this operation efficiently). Generally we can derive the following tables in 3D grids

cell->cell, cell->face, cell->edge, cell->node, cell->vertex
face->cell, face->edge, face->node, face->vertex
edge->cell, edge->face, edge->node, edge->vertex
node->cell, node->face, node->edge
vertex->cell, vertex->face, vertex->edge

where I assume that faces are 2D entities, edges are 1D entities and vertices/nodes are 0D. Things get tricky if we have mixed dimensional grids, but we can define tables analogously over codimension relations.

Some details are found as direct comments on the code where we can follow up with some discussion to work out details.

Edit: To not cause too much confusion, some of the features above are already present in this PR. Since this is a major feature I would like to include
* Docs of the data structures and functions, especially recapping utilized definitions to avoid confusion
* Update of the grid docs
* Exemplaric usage beyond the provided unit tests. While I believe that the provided test are really nicely written and useful, I think it can be useful for users to have an example beyond that. Maybe the evaluation of jumps between elements and/or something like patch-wise energies could be something simple?
* More test coverage for fancy corner cases

koehlerson:
# Summary of the current status of this PR

## What does it change

- Grid gets another property `topology` which is in the current commit parametric for `<: AbstractTopology` maybe there are other corner cases, where one wants to plug and play the grid topology data structure 
- all `grid_generator` functions accept the keyword `build_topology::Bool` that decides whether the lazy constructor `GridTopology()` or the "true" constructor `GridTopology(cells)` is called - building topology is per default false
- a new struct `Neighbor` is included which handles the Neighborhood information of one entity of one element; esentially this collects all connected entities, e.g. one hex edge is connected to a quad face and one vertex of another element, then one `Neighbor` struct will store both information in a vector `neighbor_info::Vector{<:BoundaryIndex}` - in this particular example `neighbor_info` will be of type `Vector{BoundaryIndex}` since it stores a `VertexIndex` **and** a `FaceIndex`. 

## How does it work 

- I take a `Vector{<:AbstractCell}` and construct a table that maps the cell index to all cell nodes; then I iterate over each element and check which nodes are shared and **how many**
- `face_npoints` and `edge_npoints` then decide what kind of neighbor it is, this is for all cell types unique - at least I'm not aware of different entities that are described with the same amount of nodes for one cell type
- I heavily rely on the dispatches of `face_npoints`, `edge_npoints` together with the already defined `faces`, `edges` dispatches
- the information is stored in a `face_neighbor`, `edge_neighbor` and `vertex_neighbor` `SparseMatrixCSC` where all elements are of type `Neighbor`
- to give an example: consider the `face_neighbor` matrix, this matrix maps a cellid and a local face id to all connected entities of this face (most of the time just another face, but of course also edges are possible), same holds true for `edge_neighbor` and `vertex_neighbor` matrix

## What we can do with this

- build neighborhood patches, which can be important for some multiscale methods (included in tests)
- compute values across element boundaries, e.g. jumps  (not yet included in tests)
- compute interface stuff for e.g. FSI

## Tests

- neighborhood information tested for 
    - linear triangles
    - linear quads
    - linear hexs
    - simple mixed grid
- application test
    - build level 1 neighborhood patches

bhaveshshrimali:
I just happen to have come accross this (in its entirety). Absolutely great work @koehlerson ! :) 

I just had one question. From what I understand reading through the `README`, a potential use case would be computing jumps across element facets, right? 

I ask this because I am currently in the process of putting together the non-conforming Crouzeix-Raviart interpolation as [we briefly discussed](https://github.com/Ferrite-FEM/Ferrite.jl/issues/352#issuecomment-873088052) in #352 . As soon as I have finished the relevant dispatches, I will open an issue/discussion. :) 

CR, when used for incompressible elasticity, or stokes, much like DG, involves penalizing internal jumps across element facets. So something like `∫ᵧ1/hᵧ * [[u]] ⋅ [[v]] dΓ` where `hᵧ` is the facet diameter, and `γ` is all the internal facets.

This is sometimes beneficial as opposed to using a mixed formulation. 

Cheers, and thanks!

Bhavesh


### References

1. https://www.esaim-m2an.org/articles/m2an/pdf/2003/01/m2an0157.pdf

koehlerson:
Hey @bhaveshshrimali 

yes, you are right, one application could be computing jumps. I just included a prototype dummy jump integration along an element boundary in the test file. If you have some time, you can check it out https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files#diff-796a5ea6f5cdd76d08a6b0a07f1bd1813fa58607624b572ef71f1cd8ffdddf8aR298-R337

Although, I'm not quite sure why the absolute value approaches 10.666 in this line https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files#diff-796a5ea6f5cdd76d08a6b0a07f1bd1813fa58607624b572ef71f1cd8ffdddf8aR336 In my head, I'd compute it as 2/3 * 2 * 4 (length of the element 2/3, jump is 2 and this for 4 sides). However the other test below should be correct, the "fluxes" cancel each other out and the resulting value should be something close to 0. Note that I'm summing up the integrals for each face.

termi-official:
Hi @bhaveshshrimali , jump calculations are fine with this branch now. There is a test included which will be a basis for some follow up PRs (AMR via Kelly error estimation and a simple FSI from my side). Do you have a link to your CR element implementation? I am still not super sure what the best way to implement new element formulations is, because the dof handler and the helper functions are quite rigid atm.

bhaveshshrimali:
> Hi @bhaveshshrimali , jump calculations are fine with this branch now. There is a test included which will be a basis for some follow up PRs (AMR via Kelly error estimation and a simple FSI from my side). Do you have a link to your CR element implementation? I am still not super sure what the best way to implement new element formulations is, because the dof handler and the helper functions are quite rigid atm.



Hi @termi-official , Not yet. Will push this in a separate branch over the weekend. :)   I'll need to take a look at the dof-handler but this specific case (CR) should work ootb without much changes I think. But TBD...

Will open a discussion once I have put it together. I like Ferrite since the hacks to put together new interpolations are much more transparent than other higher-level-abstraction libraries that abstract away these details. But anyways, not to digress much on this PR, I'll push something soon. Thanks

src/Grid/grid.jl
```diff
@@ -68,12 +68,133 @@ struct VertexIndex <: BoundaryIndex
     idx::Tuple{Int,Int} # cell and side
 end
 
+struct Neighbor{T<:BoundaryIndex}
+    neighbor_info::Vector{T}
+end
```
termi-official, line 73: I think we should rename this one to something like `Neighborhood`, `GeneralNeighborhood` or something, because using a singular name is a bit misleading. No idea how to improve the data member's name.

Further I would introduce some specialized types, i.e. FaceNeighborhood, EdgeNeighborhood and VertexNeighborhood.

Since I just noted, is there some rationale behind wrapping this one in a struct? It looks a bit like direclty using the Vector could be sufficient.

src/Grid/grid.jl
```diff
@@ -84,6 +205,8 @@ mutable struct Grid{dim,C<:AbstractCell,T<:Real} <: AbstractGrid{dim}
     vertexsets::Dict{String,Set{VertexIndex}} 
     # Boundary matrix (faces per cell × cell)
     boundary_matrix::SparseMatrixCSC{Bool,Int}
+    #topology
+    topology::topologytype
```
termi-official, line 209: I think adding some abstraction to access the topology information could be helpful to not leak implementation details, i.e. something like

Get everything connected to an entity
* `full_neighborhood(::Grid, ::CellIndex, include_self=false)`
* `full_neighborhood(::Grid, ::FaceIndex, include_self=false)`
* `full_neighborhood(::Grid, ::EdgeIndex, include_self=false)`
* `full_neighborhood(::Grid, ::VertexIndex, include_self=false)`

and maybe some specialized methods for relations which can not be directly derived, e.g.

* `face_neighborhood(::Grid, ::CellIndex, include_self=false)`

I think this should be sufficient to quickly get all topological connections, but I am not sure. I think checking this is out of the scope of this PR and more future work.

src/Grid/grid.jl
```diff
@@ -68,12 +68,133 @@ struct VertexIndex <: BoundaryIndex
...
+edge_npoints(::Cell{3,N,M}) where {N,M} = 2
+
+getdim(::Cell{dim,N,M}) where {dim,N,M} = dim
+
+abstract type AbstractTopology end
+
+struct GridTopology <: AbstractTopology
+    face_neighbor::SparseMatrixCSC{Neighbor,Int}
+    corner_neighbor::SparseMatrixCSC{Neighbor,Int}
+    edge_neighbor::SparseMatrixCSC{Neighbor,Int}
```
termi-official, line 112: Names are a bit weird. I have no good alternative for now. 

Depending on the exact definitions of these tables, maybe some of the "inverse relations" are missing (e.g. vertex -> cell)

src/Grid/grid.jl
```diff
@@ -68,12 +68,133 @@ struct VertexIndex <: BoundaryIndex
...
+                #edge neighbor
+                _edge_neighbor!(V_edge,I_edge,J_edge,cellid,cell,neighbor,neighborid,neighbor_cell)
+            elseif getdim(cell) > 1 && length(neighbor) == face_npoints(cell)
+                #face neighbor
+                _face_neighbor!(V_face,I_face,J_face,cellid,cell,neighbor,neighborid,neighbor_cell)
+            else
+                continue
+            end 
+        end
+    end
```
termi-official, line 147: This construction is quite expensive ( O(#cells * #nodes), which is basically quadratic in the number of cells ).

My alternative construction would be to derive the topology as follows.

1. vertex -> cell ("invert" cell -> vertex by looping over the vertices)
1. now you can derive cell -> cell via the concatenation cell -> vertex -> cell
1. now the face neighborhood can be found by cell -> cell and comparing the unique face identifiers
1. same for the edge neighborhood

termi-official (changes requested):
PR looks great already! :) However, some more suggestions from my side has been left  
* sketch for a linear construction algorithm (instead of the current quadratic one)
* renaming suggestions
* thin abstraction layer to not leak implementation details

Further I think not all points from the previous review round has been resolved. Just double check.

Also, I think we should rename "corner" to "vertex" for consistency reasons.

src/Grid/grid.jl
koehlerson: > Since I just noted, is there some rationale behind wrapping this one in a struct? It looks a bit like direclty using the Vector could be sufficient.


The reason behind wrapping it in a struct is that I want to dispatch `Base.:+` see https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90R83-R86

We need this because if we construct `I, J, V` of each sparse matrix, `+` is called if more than one `V`entry exists at the same `I` `J` indices. Probably a `Base.:+(x::Vector{T}, y::Vector{T}) where T<:BoundaryIndex` type piracy is possible but I thought we should avoid it, if possible

koehlerson: Regarding naming, we should agree one something here before I change it in the code: maybe `EntityNeighborhood` ?

src/Grid/grid.jl
koehlerson: should be resolved

src/Grid/grid.jl
koehlerson: hm I don't know how to call them, `*_neighbor` is a matrix that maps a cell and a local id to a `*Index` where `*Index` is the exclusive neighborhood connection to the given cell and local id

termi-official: Maybe [x]_to_entity_table?

src/Grid/grid.jl
```diff
@@ -184,11 +216,15 @@ function _face_neighbor!(V_face, I_face, J_face, cellid, cell, neighbor, neighbo
...
     cell_face_id = findfirst(x->issubset(x,neighbor_face),faces(cell))
-    push!(V_face, Neighbor(face_neighbor))
+    push!(V_face, EntityNeighborhood(face_neighbor))
     push!(I_face, cellid)
     push!(J_face, cell_face_id)
 end
 
+getelement(neighbor::EntityNeighborhood{T}) where T <: BoundaryIndex = first.(neighbor.neighbor_info)
+getelement(neighbor::EntityNeighborhood{CellIndex}) = getproperty.(neighbor.neighbor_info, :idx)
+getelements(neighbors::Vector{T}) where T <: EntityNeighborhood = reduce(vcat, getelement.(neighbors))
```
termi-official, line 226: While I personally also prefer the name `element` I am not sure if it is good in Ferrite, because we use this name almost nowhere. To be specific, now we have `cell`, `entity` and `element` refering to the same concept (, sometimes).

termi-official (commented):
Looks already quite good! Found some points which we should discuss.

src/Grid/grid.jl
```diff
@@ -109,33 +109,33 @@ getdim(::Cell{dim}) where dim = dim
...
 
 struct ExclusiveTopology <: AbstractTopology
-    vertex_to_cell::Vector{Vector{Int}}
+    vertex_to_cell::Dict{Int,Vector{Int}}
     cell_neighbor::Vector{EntityNeighborhood{CellIndex}}
     face_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
     vertex_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
     edge_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
-    vertex_vertex_neighbor::Vector{EntityNeighborhood{VertexIndex}}
+    vertex_vertex_neighbor::Dict{Int,EntityNeighborhood{VertexIndex}}
```
termi-official: What was the rationale behind this change?

termi-official (commented):
Forgot what the rationale of this change was - also, could you mark this feature as experimental?

src/Grid/grid.jl
koehlerson: Vertices do not have a linear index for higher order cases. Up till now I used the vector index and mapped it to the neighbors, but this isn’t possible for e.g. a quadratic quadrilateral where your First vertex is global id 1 and your second vertex is global id 3. that’s why we need an actual map

test/test_grid_dofhandler_vtk.jl
```diff
@@ -164,3 +164,103 @@ end
...
+#                   +-----+-----+
+#                   |\  2 |\  4 |
+#                   |  \  |  \  |
+#                   |  1 \| 3  \|
+#                   +-----+-----+
+# test for multiple corner_neighbors as in e.g. ele 3, local corner 3 (middle node)
+    trigrid = generate_grid(Triangle,(2,2))
+    topology = trigrid.topology
+    @test topology.corner_neighbor[3,3] == Ferrite.Neighbor([VertexIndex(5,2),VertexIndex(6,1),VertexIndex(7,1)])
+
```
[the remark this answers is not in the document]
termi-official: Resolved below in the test suite.

termi-official (approved):
I think this one looks really great as a first attempt to integrate topology information!

Ping @fredrikekre - I think this one is ready so far.

kimauth:
I've been using this to implement a simple mesh adaptivity algorithm for my students. I must say big thanks, this PR made it so easy! 🎉 

Small remark: Would a singleton type `NoTopology` be a better default than an empty topology in case no topology is build? I wanted to dispatch such that my refinement algorithm only takes in grids which have a topology built, but right now an empty topology is constructed instead. (I think that should also have an even smaller overhead when no topology is needed.)

koehlerson:
I'll add this, if you post those nice adaptivity pictures for the flex inside here :sunglasses:

kimauth:
Here are the pictures 🙂 

anim_grid_refinement 0000
anim_grid_refinement 0001
anim_grid_refinement 0002
anim_grid_refinement 0003
anim_grid_refinement 0004
anim_grid_refinement 0005
anim_grid_refinement 0006
anim_grid_refinement 0007
anim_grid_refinement 0008
anim_grid_refinement 0009

src/Grid/grid_generators.jl
```diff
@@ -22,7 +22,7 @@ e.g. `Triangle` or `Hexahedron`. `nel` is a tuple of the number of elements in e
 generate_grid
 
 # Line
-function generate_grid(::Type{Line}, nel::NTuple{1,Int}, left::Vec{1,T}=Vec{1}((-1.0,)), right::Vec{1,T}=Vec{1}((1.0,))) where {T}
+function generate_grid(::Type{Line}, nel::NTuple{1,Int}, left::Vec{1,T}=Vec{1}((-1.0,)), right::Vec{1,T}=Vec{1}((1.0,)); build_topology=false) where {T}
```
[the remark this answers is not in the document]
koehlerson: Currently the constructor of `ExclusiveTopology` builds the struct based on an `Vector{C} where C<: AbstractCell`
https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files/b621afe33ddea62301ee9f5838dc0d69a02739d5#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90R134

So any kind of grid can get a topology if the cells are of type `<:AbstractCell` and of course they need to fulfill the `AbstractCell` interface that is `faces,edges,vertices`

src/Grid/grid.jl
```diff
@@ -68,12 +68,200 @@ struct VertexIndex <: BoundaryIndex
...
+edge_npoints(::Cell{3,4,1}) = 2 #not sure how to handle embedded cells e.g. Quadrilateral3D
+face_npoints(::Cell{3,N,6}) where N = 4
+face_npoints(::Cell{3,N,4}) where N = 3
+edge_npoints(::Cell{3,N,M}) where {N,M} = 2
+
+getdim(::Cell{dim}) where dim = dim
+
+abstract type AbstractTopology end
+
+struct ExclusiveTopology <: AbstractTopology
```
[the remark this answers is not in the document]
koehlerson: In the beginning I only had https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files/b621afe33ddea62301ee9f5838dc0d69a02739d5#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90R113-R116

as the struct's field which maps an entity to something the entity is connected exclusively and thus, the name. However things changed a bit with the `vertex_to_cell` , `vertex_vertex_neighbor`  and `face_skeleton` list, but still things could be saved differently by e.g. saving all kinds of neighborhoods in the `SparseMatrixCSC` entries. 

To give you an example by what I mean with exclusive neighborhood: https://github.com/Ferrite-FEM/Ferrite.jl/pull/363/files/b621afe33ddea62301ee9f5838dc0d69a02739d5#diff-796a5ea6f5cdd76d08a6b0a07f1bd1813fa58607624b572ef71f1cd8ffdddf8aR219-R231

In this example cell 1 local vertex 3 has the exclusive neighborhood with cell 4 local vertex 1. Of course, local vertex 3 of cell 1 contributes to other neighborhoods as well but this due to higher dimensional things such as faces in this case. Now, lets take a look at cell 1 local vertex 2. There we don't have an exclusive vertex neighborhood, but rather an exclusive face neighborhood. So, checking vertex_neighbor[1,2] will give you nothing, but face_neighbor[2,2] will return `FaceIndex(2,4)`

docs/src/reference/grid.md
```diff
@@ -2,4 +2,8 @@
 DocTestSetup = :(using Ferrite)
 ```
 
-# Grid
\ No newline at end of file
+# Grid
+```@docs
+Ferrite.full_neighborhood
+Ferrite.faceskeleton
```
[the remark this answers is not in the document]
koehlerson: If you want to do adaptive stuff, you need to compute entities along the faces of the elements, but you need to iterate over all faces without duplication, because otherwise things will cancel out each other (e.g. due to the normal within the error formulation) 

So, what you can do with it is something like this

```julia
for face in faceskeleton(grid)
    reinit!(fv,face,grid)
    neighbors = full_neighborhood(grid,face)
    for neighbor in neighbors
        reinit!(fv_neighbors,neighbor,grid)
        for qp in 1:getnquadpoints(fv_neighbors)
            gradu_jump = (function_gradient(fv,u,qp) - function_gradient(fv_neighbor,u,qp)) # more stuff like normal and detJdA
        end
    end
end
```

what needs to be added to get a nice interface is some sort of new facevalue type that has two facevalues and reverts the list of the quadrature points for one facevalue field, since they will be flipped against each other and thus spatial dependencies could be wrong if e.g. a spatial dependent elasticity,conductivity (or whatever) matrix is needed. To visualize the problem, what the current implementation (based on local face indices) of facevalues would compare
```
                       +-----+-----+
                       | (1) x (2) |
                       |     |     |
                       |     |     |
                       | (2) x (1) |
                       +-----+-----+
                            
```

and what we actually need:
```
                       +-----+-----+
                       | (1) x (1) |
                       |     |     |
                       |     |     |
                       | (2) x (2) |
                       +-----+-----+
```
`x`'s correspond to integration points on the face

src/Grid/grid.jl
```diff
@@ -68,12 +68,200 @@ struct VertexIndex <: BoundaryIndex
     idx::Tuple{Int,Int} # cell and side
 end
 
+struct EntityNeighborhood{T<:Union{BoundaryIndex,CellIndex}}
+    neighbor_info::Vector{T}
+end
+
+EntityNeighborhood(info::T) where T <: BoundaryIndex = EntityNeighborhood([info])
+Base.zero(::Type{EntityNeighborhood{T}}) where T = EntityNeighborhood(T[])
```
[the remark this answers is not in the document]
koehlerson: yes

src/Grid/grid.jl
```diff
@@ -68,12 +68,200 @@ struct VertexIndex <: BoundaryIndex
...
+abstract type AbstractTopology end
+
+struct ExclusiveTopology <: AbstractTopology
+    vertex_to_cell::Dict{Int,Vector{Int}}
+    cell_neighbor::Vector{EntityNeighborhood{CellIndex}}
+    face_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
+    vertex_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
+    edge_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}
+    vertex_vertex_neighbor::Dict{Int,EntityNeighborhood{VertexIndex}}
+    face_skeleton::Vector{FaceIndex}
```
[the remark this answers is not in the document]
koehlerson: will do that soon

src/Grid/grid_generators.jl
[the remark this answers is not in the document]
koehlerson: @kimauth's adapted grids are not from a `grid_generator` so just as an example: https://github.com/kimauth/FEMSolids.jl/pull/3/files#diff-33cf25c25ddedccb68eb82b2d8187e39133d24f70794eabdffa8e6053630073bR119

koehlerson:
## Tuple vs Vector Container for `EntityNeighborhood`

### Tuple based version (not pushed)
```
julia> @benchmark generate_grid(Hexahedron,(10,10,10);build_topology=true)
BenchmarkTools.Trial: 239 samples with 1 evaluation.
 Range (min … max):  27.914 ms … 82.874 ms  ┊ GC (min … max):  0.00% … 18.73%
 Time  (median):     40.817 ms              ┊ GC (median):    19.51%
 Time  (mean ± σ):   41.927 ms ±  8.686 ms  ┊ GC (mean ± σ):  18.67% ±  6.19%

         ▇▂    ▁▃█▄
  ▄▃▁▁▃▄▇███▆▆▅████▇▇▇▄▃▃▅▃▃▄▃▄▃▃▁▄▃▃▃▁▁▃▁▁▃▁▃▁▃▁▁▁▃▁▃▁▁▁▁▁▁▃ ▃
  27.9 ms         Histogram: frequency by time        79.3 ms <

 Memory estimate: 39.91 MiB, allocs estimate: 472875.

julia> @benchmark generate_grid(Hexahedron,(100,100,10);build_topology=true)
BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  7.166 s …    7.329 s  ┊ GC (min … max): 40.14% … 40.53%
 Time  (median):     7.248 s               ┊ GC (median):    40.33%
 Time  (mean ± σ):   7.248 s ± 115.748 ms  ┊ GC (mean ± σ):  40.33% ±  0.28%

  █                                                        █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  7.17 s         Histogram: frequency by time         7.33 s <

 Memory estimate: 4.06 GiB, allocs estimate: 53385551.
```

### Vector based version (current version of this branch)

```
julia> @benchmark generate_grid(Hexahedron,(10,10,10);build_topology=true)
BenchmarkTools.Trial: 330 samples with 1 evaluation.
 Range (min … max):  25.484 ms … 38.492 ms  ┊ GC (min … max):  0.00% … 14.66%
 Time  (median):     30.147 ms              ┊ GC (median):    15.61%
 Time  (mean ± σ):   30.280 ms ±  2.162 ms  ┊ GC (mean ± σ):  14.94% ±  5.75%

    ▂              ▃█▆▆▆▆▆▅▁▃▇▄▁  ▁  ▄▄▁ ▅
  ▅▄█▆▄▅▄▃▁▄▁▃▁▁▃▄▆██████████████▅██▄█████▇▅▇▃▁▅▄▃▁▄▁▄▃▃▁▃▃▁▃ ▄
  25.5 ms         Histogram: frequency by time          36 ms <

 Memory estimate: 39.01 MiB, allocs estimate: 444226.

julia> @benchmark generate_grid(Hexahedron,(100,100,10);build_topology=true)
BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  6.175 s …    6.736 s  ┊ GC (min … max): 40.28% … 44.57%
 Time  (median):     6.455 s               ┊ GC (median):    42.52%
 Time  (mean ± σ):   6.455 s ± 396.190 ms  ┊ GC (mean ± σ):  42.52% ±  3.03%

  █                                                        █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  6.18 s         Histogram: frequency by time         6.74 s <

 Memory estimate: 4.00 GiB, allocs estimate: 50561142.

```

Probably only due to the less allocs, which are too high anyway. Need to go through in detail where things can be optimized

termi-official:
>  Need to go through in detail where things can be optimized

I do not think we need to go to optimization yet. Would you agree that we should gather user feedback on the design before optimizing stuff?

koehlerson:
sure, it means less work right now :D

src/Grid/grid.jl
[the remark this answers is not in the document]
fredrikekre: Perhaps it would be easier to just use a `Dict` then? I assume no linear algebra operations etc are performed on this matrix anyway?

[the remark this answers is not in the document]
koehlerson: I can change `SparseMatrixCSC{EntityNeighborhood,Int}` to `Dict{Index,EntityNeighborhood}` where `Index = Union{CellIndex,BoundaryIndex}`. Should the dict version be faster than accessing a sparse matrix?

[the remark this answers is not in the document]
termi-official: Technically speaking we can abuse the sparse matrix for GraphBLAS-type shenanigans (e.g. patch extraction), but I do not think it is super important to implement these operations via LinA.

bhaveshshrimali:
Hey @koehlerson , any ideas as to when this could be merged into master?

koehlerson:
From my perspective this is good to go as an experimental feature, but final approvement of @fredrikekre would be good. Otherwise a roadmap what should be changed such that I can finalize it to the required changes is needed

docs/src/reference/grid.md
```diff
@@ -113,3 +115,4 @@ end
 Now, you would be able to assemble the heat equation example over the new custom `SmallGrid` type.
 Note that this particular subtype isn't able to handle boundary entity sets and so, you can't describe boundaries with it.
 In order to use boundaries, e.g. for Dirichlet constraints in the ConstraintHandler, you would need to dispatch the `AbstractGrid` sets utility functions on `SmallGrid`.
+>>>>>>> master
```
fredrikekre, line 118:
```suggestion
```

fredrikekre:
Thoughts on https://github.com/Ferrite-FEM/Ferrite.jl/pull/363#discussion_r767718922 ?

Other than that, I guess https://github.com/Ferrite-FEM/Ferrite.jl/pull/363#discussion_r767713632 could be reworked later. I just think it is perhaps a bit strange to put all of that in a sparse matrix, but doesn't matter much.

koehlerson:
> Thoughts on [#363 (comment)](https://github.com/Ferrite-FEM/Ferrite.jl/pull/363#discussion_r767718922) ?
> 
> Other than that, I guess [#363 (comment)](https://github.com/Ferrite-FEM/Ferrite.jl/pull/363#discussion_r767713632) could be reworked later. I just think it is perhaps a bit strange to put all of that in a sparse matrix, but doesn't matter much.

I can also change it to a Dict if you prefer it. I started with a matrix because it felt naturally due to two indices that are mapped (cell index and local entity index). But I guess we can just set up the dict with an `Union{CellIndex,<:BoundaryIndex}`. Interface for users with `getneighbors` can withstand the change

fredrikekre:
Yea, let's leave it like this for now. I just mean that sparse matrices are not really meant for general data storage (since you need to define `zero` etc), so maybe a dictionary would be nicer in that sense.

Final comment, maybe `getneighborhood` would be nicer? :D

koehlerson:
Yeah I see the point and changing to a specific key-type for the dict does not make it any harder to implement this change :)

> Final comment, maybe `getneighborhood` would be nicer? :D

Ok sure, give me a sec :D

src/Grid/grid.jl
```diff
@@ -326,18 +326,18 @@ end
...
-    getneighbors(grid::Grid{dim,C,T,ExclusiveTopology}, faceidx::FaceIndex, include_self=false)
-    getneighbors(grid::Grid{dim,C,T,ExclusiveTopology}, vertexidx::VertexIndex, include_self=false)
-    getneighbors(grid::Grid{dim,C,T,ExclusiveTopology}, EdgeIndex::EdgeIndex, include_self=false)
+    getneighborhood(grid::Grid{dim,C,T,ExclusiveTopology}, cellidx::CellIndex, include_self=false)
+    getneighborhood(grid::Grid{dim,C,T,ExclusiveTopology}, faceidx::FaceIndex, include_self=false)
+    getneighborhood(grid::Grid{dim,C,T,ExclusiveTopology}, vertexidx::VertexIndex, include_self=false)
+    getneighborhood(grid::Grid{dim,C,T,ExclusiveTopology}, EdgeIndex::EdgeIndex, include_self=false)
 
 Returns all directly connected entities of the same type, i.e. calling the function with a `VertexIndex` will return
 a list of directly connected vertices (connected via face/edge). If `include_self` is true, the given `*Index` is included
```
termi-official, line 335: Just as a reminer, with the current interface we cannot directly control which kind of neighborhood we want? E.g.for vertices just the edge or just the face neighborhood.

termi-official:
Also, do you know how this behaves on NC meshes?

src/Grid/grid.jl
koehlerson: No, but implementation on top of it is possible, e.g. given a vertex get all directly connected edges would require to get all edges of the current cell and the cell neighbors were the given vertex is part of a cell edge, etc.

koehlerson:
> Also, do you know how this behaves on NC meshes?

Kind of yes, given a hanging node like this
```
         (7)    (8)
(6) +-----+-----+
    |  1  |     |
(4) +-----+  3  |
    |  2  |     |
    +-----+-----+
  (1)    (2)    (3)
```

```
julia> using Ferrite

julia> nodes = [Ferrite.Node((0,0)) for i in 1:8]
8-element Vector{Node{2, Int64}}:
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])
 Node{2, Int64}([0, 0])

julia> cells = [Quadrilateral((4,5,7,6)),Quadrilateral((1,2,5,4)),Quadrilateral((2,3,8,7))]
3-element Vector{Quadrilateral}:
 Quadrilateral((4, 5, 7, 6))
 Quadrilateral((1, 2, 5, 4))
 Quadrilateral((2, 3, 8, 7))

julia> top = Ferrite.ExclusiveTopology(cells)
```
you would detect a the conforming face neighbor
```
julia> top.face_neighbor
2×3 SparseArrays.SparseMatrixCSC{Ferrite.EntityNeighborhood, Int64} with 2 stored entries:
 EntityNeighborhood{FaceIndex}(FaceIndex[FaceIndex((2, 3))])  …                              ⋅
                             ⋅                                   EntityNeighborhood{FaceIndex}(FaceIndex[FaceIndex((1, 1))])

julia> top.face_neighbor[2,3]
FaceIndex((1, 1))


julia> top.face_neighbor[1,1]
FaceIndex((2, 3))
```
and the non-conforming part is now wrongly encoded in the vertex_neighborhood
```
julia> top.vertex_neighbor
3×4 SparseArrays.SparseMatrixCSC{Ferrite.EntityNeighborhood, Int64} with 4 stored entries:
                             ⋅                                      …                              ⋅
                             ⋅                                                                     ⋅
 EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((2, 2))])     EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((1, 3))])
```
This happens, since I'm checking for the length of the matching entities. Consider the given cell with cell index 1, the construction sees the neighborhood with length 2 at the local face id 1. However, at the local face id 2  the length of the matching entities is 1, since only 1 vertex matches of the current cells and thus, is safed as a vertex neighbor for cell 3 in local vertex index 1 and 4 (see `top.vertex_neighbor`).

src/Grid/grid.jl
termi-official: I know that it is possible to do. :) My question aimed at our abstraction layer, because we want to leave the option open to utilize different topologies, which may be optimized for specific use cases. However, we can develop some algorithms on top of a common interface which are independent of the actual topology.

koehlerson: but this is still open, if you dispatch simply on a different topology type, instead of `ExclusiveTopology`

termi-official: Right, but currently we have the contract that "a list of directly connected vertices (connected via face/edge)" is returned, which is ambiguous. E.g. a new topology may return the vertices connected via edges on hexahedral elements which another one may return the vertices connected via faces. Both have legitimate use cases, but the contract is too wide imho. Similar issues arise due to ambiguities in NC meshes. Not sure how to improve the docs here.

koehlerson: is directly connected really ambigious? With directly connected I mean all vertices connected by a face in 2D or an edge in 3D, i.e. where you have a direct line in between

koehlerson: from the example from my comment below about NC meshes, e.g. node 7
```
julia> grid =  Grid(cells,nodes;topology=top)
Grid{2, Quadrilateral, Int64, Ferrite.ExclusiveTopology} with 3 Quadrilateral cells and 8 nodes

julia> Ferrite.getneighborhood(grid,VertexIndex(1,3))
4-element Vector{VertexIndex}:
 VertexIndex((1, 2))
 VertexIndex((1, 4))
 VertexIndex((3, 1))
 VertexIndex((3, 3))
```

termi-official: Yes, I think it is ambiguous. 
1. we have no definition of what a face/edge is (if you remember the discussion on slack). 
2. currently we use face and edge in 2D to **sometimes** refer to the same entity and sometimes not (due to the missing definitions)
3. while I intuitively understand what you want to say with "connected by a direct line", you can also draw a line over faces to connect the "diagonal" in hex faces. Not sure how to precise this one.

The issue is clearer for the nodes 2 and 7. Is 5 really a neighbor for them? And how relates 3 to them? However, I would leave the NC issues out of the PR for now to not overcomplicate it.

koehlerson: I don't see how this results in something ambiguous in this particular example if the definitions of `faces, edges` and `vertices`is used from: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L459-L476
in 2D the line is the face, while in 3D the line is the edge of a three dimensional cell. The face of a two dimensional connected to a 3D cell results in a face/edge neighborhood in my code, while a truly embedded 2D cell in 3D has again a edge neighborhood https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L475. Such that we end up with something that can distinguish between the "line entities" of a given cell, due such if clauses https://github.com/Ferrite-FEM/Ferrite.jl/blob/mk/topologyinfo/src/Grid/grid.jl#L209 in combination with the dispatching of the embedded cells in e.g. https://github.com/Ferrite-FEM/Ferrite.jl/blob/mk/topologyinfo/src/Grid/grid.jl#L718 and https://github.com/Ferrite-FEM/Ferrite.jl/blob/mk/topologyinfo/src/Grid/grid.jl#L724

So truly I'm relying on the given definitions. If it's a problem of the definitions, then it's not the related to this PR. However the PR or the implementation *should* rely always on these definitions, since they allow for a user defined cell extension. 

To finalize it, the situation is very clear (at least for me) for node 5, 5 is a vertex neighbor of those nodes, why not? It is directly connected by the corresponding line entity of the cell, i.e. face
```
julia> nodes = [Ferrite.Node((0,0)) for i in 1:8];

julia> cells = [Quadrilateral((4,5,7,6)),Quadrilateral((1,2,5,4)),Quadrilateral((2,3,8,7))];

julia> top = Ferrite.ExclusiveTopology(cells);

julia> grid =  Grid(cells,nodes;topology=top);

julia> Ferrite.getneighborhood(grid,VertexIndex(3,4))
4-element Vector{VertexIndex}:
 VertexIndex((1, 2))
 VertexIndex((1, 4))
 VertexIndex((3, 1))
 VertexIndex((3, 3))

julia> Ferrite.getneighborhood(grid,VertexIndex(1,3))
4-element Vector{VertexIndex}:
 VertexIndex((1, 2))
 VertexIndex((1, 4))
 VertexIndex((3, 1))
 VertexIndex((3, 3))

julia> Ferrite.getneighborhood(grid,VertexIndex(2,2))
4-element Vector{VertexIndex}:
 VertexIndex((2, 1))
 VertexIndex((2, 3))
 VertexIndex((3, 2))
 VertexIndex((3, 4))

julia> Ferrite.getneighborhood(grid,VertexIndex(3,1))
4-element Vector{VertexIndex}:
 VertexIndex((2, 1))
 VertexIndex((2, 3))
 VertexIndex((3, 2))
 VertexIndex((3, 4))

```

termi-official: > [...] If it's a problem of the definitions, then it's not the related to this PR. [...]

Agreed. Since this is quite complicated let us leave the disambiguation issue for now. Maybe we should leave a comment on the strict dependency on the mentioned interface to vertices,edges and faces in the topology docs?

> To finalize it, the situation is very clear (at least for me) for node 5, 5 is a vertex neighbor of those nodes, why not? [...]

Okay, I was not precise enough in the problem description here, so let me leave a clarification for future reference. Depending on how you define "face" and "line" in NC meshes along the "nonconforming face", you can have the following neighborhoods for vertex 3:
* {1,3,5,7}
* {1,3,7}
* {1,3,5}

Maybe, after the last design decisions for this PR are settled, we should move some parts of your examples to the docs (since I think they are quite nice for new users to understand the interface).

termi-official:
Can you merge the current master? I will make the next review afterwards and thinks this should be good to go. What do you think @fredrikekre ?

src/Grid/grid.jl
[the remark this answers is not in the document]
termi-official: Can we add some basic documentation for the struct? I think this could be not super easy for new users to grasp.

src/Grid/grid.jl
```diff
@@ -127,7 +311,65 @@ end
...
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, cellidx::CellIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, faceidx::FaceIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, vertexidx::VertexIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, EdgeIndex::EdgeIndex, include_self=false)
+
+Returns all directly connected entities of the same type, i.e. calling the function with a `VertexIndex` will return
+a list of directly connected vertices (connected via face/edge). If `include_self` is true, the given `*Index` is included 
+in the returned list.
+
+**Warning:** this feature is highly experimental and very likely subjected to interface changes in the future.
```
termi-official, line 324: I think we could highlight this via alert box. (see https://github.com/JuliaDocs/Documenter.jl/blob/cd57c41ab1ff8a8dacf5c21a7d8363b777a6ad9a/src/Documenter.jl#L874-L879)

src/Grid/grid.jl
```diff
@@ -127,7 +311,65 @@ end
 ##########################
 # Grid utility functions #
 ##########################
-@inline getdim(grid::AbstractGrid{dim}) where {dim} = dim
+"""
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, cellidx::CellIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, faceidx::FaceIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, vertexidx::VertexIndex, include_self=false)
+    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, EdgeIndex::EdgeIndex, include_self=false)
```
termi-official, line 318:
```suggestion
    getneighborhood(top::ExclusiveTopology, grid::Grid{dim,C,T}, edgeidx::EdgeIndex, include_self=false)
```

termi-official (changes requested):
I have some minor points regarding the documentation. After these changes I think we should be fine merging this and I will follow up with a simple linear FSI example.

src/Grid/grid.jl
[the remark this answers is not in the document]
termi-official: I think your change is already enough. I just imagined the situation where a new user tries to obtain the connectivity data, but the term topology (which I regularly use for this) might be not the first mental connection they would make. Documenting the requirements on a cell is also great, as well as some internals, almost forgot this!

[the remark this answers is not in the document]
termi-official: Okay, maybe we should add another sentence that just tells that this is the struct to access the connectivity. This way a search in the documentation (or an external search machine) directly points to this. Or what do you think?

termi-official (approved):
LGTM.

PR #366, opened 2021-08-19
Remove unused type arguments that cause allocations in Ferrite.cellnodes!()

lijas:
This seems like a bug in julia? Feels weird, hopefully i am not benchmarking it wrong in some way :sweat_smile: 

`@btime Ferrite.cellnodes!($nodes, $dh, $1)`

Before:
`25.734 ns (1 allocation: 112 bytes)`

Now:
`5.311 ns (0 allocations: 0 bytes)`


KristofferC:
https://github.com/JuliaLang/julia/issues/29393

KristofferC (approved):

PR #368, opened 2021-08-31
Simple Example for DifferentialEquations integration

termi-official:
Turns out implementing a consistent multi-level Newton-Raphson for DifferentialEquations is quite tricky ( #325  is blocked for now ). Since several users asked about examples for the integration, here a simpler example.

## Open TODOs

* [x] Implement incompressibles Stokes flow
* [x] ~~Fix DAE initialization~~ Correct initialization of the algebraic portion is nontrivia, since the semi-discretization has Hessenberg index 2. DiffEq just has methods for index 1 dae initialization, hence we just start with a trivial (zero) initial condition.
* [x] Implement incompressible Navier-Stokes flow
* [x] ~~Simple mesh loader to change regular grid with a domain with hole~~ This can be done as a follow up PR. Currently the mesh is generated manually.
* [x] Fix CI pipeline (update somehow broke plotting)
* [ ] ~~new apply_zero! dispatch to zero out just a sub-field~~ -> future work... 

## Non-addressed issues
* Mesh IO: the example currently generates the mesh manually, which is suboptimal. Improving this situation is left for follow up  PR and out of scope for this PR.
* Solver: A direct sparse solver has been utilized without bandwith optimizing reordering of the matrix, leading to an immense amount of memory necessary. Also the Stokes operator is just semi-definite, which may lead to additional issues with the direct solver. A preconditioned GMRes should be deployed through a future PR.


termi-official:
Vortex shedding works. :) Cleanup and documentation will follow.
vortex-shedding

KristofferC:
Epic!

termi-official:
Almost finished. This example basically shows how timestep with DifferentialEquations.jl with a semi-discretization of problems which can be stated in mass matrix form and time-dependent Dirichlet boundary conditions.

I think I may need one or two more iterations on the explanation. A major weakness for now is that I have to overwrite one of the internal methods from the OrdinaryDiffEq package (OrdinaryDiffEq.initialize!) to apply the Dirichlet BC correctly in the time-dependent case. If the Dirichlet BC does not depend on time, everything should also work fine without hooking in this method. Also, I can not guarantee that my patch of OrdinaryDiffEq.initialize! works for all timesteppers as it may mess with assumptions taken in the corresponding OrdinaryDiffEq.perform_step! implementations.

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+#
+# ![](https://user-images.githubusercontent.com/9196588/134514213-76d91d34-19ab-47c2-957e-16bb0c8669e1.gif)
+#
+#
+# In this example we focus on a simple but visually appealing problem from
+# fluid dynamics, namely vortex shedding, which is also known as
+# von-Karman vortex streets, to show how to utilize [DifferentialEquations.jl]()
+# in tandem with [Ferrite.jl](). To keep things simple we use a general approach
+# to discretize the system, hence we refrain from Chorin's more performant
+# projection method.
```
koehlerson, line 11:
```suggestion
# In this example we focus on a simple but visually appealing problem from
# fluid dynamics, namely vortex shedding. This problem is also known as
# [von-Karman vortex streets](https://en.wikipedia.org/wiki/K%C3%A1rm%C3%A1n_vortex_street). Within this example, we show how to utilize [DifferentialEquations.jl]()
# in tandem with [Ferrite.jl](). To keep things simple we use a general approach
# to discretize the system: [Chorin's method](https://en.wikipedia.org/wiki/Projection_method_(fluid_dynamics)#Chorin's_projection_method). A performant projection method for the Navier-Stokes equations.
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+#
+# ## Remarks on DifferentialEquations.jl
```
koehlerson, line 13: maybe add in this section a dummy example how mass matrix form looks like

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+#    \partial_t v &= \nu \Delta v - (v \cdot \nabla) v - \nabla p \\
+#               0 &= \nabla \cdot v
+#  \end{aligned}
+# ```
+#
+# where $v$ is the unknown velocity field, $p$ the unknown pressure field
+# and $\nu$ the dynamic viscosity. We assume a constant density of 1 for the fluid
+# and negligible coupling between the velocity components.
+# Finally we see that the pressure term appears only in combination with the gradient
+# operator, so for any solution p the function p + c is also an admissible solution, if
```
koehlerson, line 37:
```suggestion
# operator, so for any solution $p$ the function $p + c$ is also an admissible solution, if
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+# we do not impose Dirichlet conditions on the pressure. To resolve this we introduce the
+# implicit constraint that $ \int_\Omega p = 0 $.
+#
+# Our setup is derived from [Turek's DFG benchmark](http://www.mathematik.tu-dortmund.de/~featflow/en/benchmarks/cfdbenchmarking/flow/dfg_benchmark1_re20.html).
+# We model a channel with size $0.41 \times 2.2$ and a hole of radius $0.05$ centered at $(0.2, 0.2)$.
+# The left side has a parabolic inflow profile, which is ramped up over time, modeled as the Dirichlet condition
+# $v_x(t) \mapsto 4*v_{in}(t)*y*(0.41-y)/0.41^2$ where $v_{in}(t) = clamp(t, 0.0, 1.0)$. With a viscosity of $\nu = 0.001$
+# this is enough to induce turbulence behind the cylinder which leads to vortex shedding. The top and bottom of our
+# channel have no-slip conditions, i.e. v = (0,0), while the right boundary has the do-nothing boundary condtion
```
koehlerson, line 46:
```suggestion
# channel have no-slip conditions, i.e. $v = (0,0)$, while the right boundary has the do-nothing boundary condtion
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+#
+#
+# ## Commented Program
+#
+# Now we solve the problem in Ferrite with DifferentialEquations.jl. What follows is a program spliced with comments.
+# The full program, without comments, can be found in the next [section](@ref ns_vs_diffeq-plain-program).
+#
+# First we load Ferrite, and some other packages we need
+using Ferrite, SparseArrays, BlockArrays, LinearAlgebra, UnPack
+# Since we do note need the complete DifferentialEquations suite just load the required ODE infrastructure, which can also handle
```
koehlerson, line 75:
```suggestion
# Since we do not need the complete DifferentialEquations suite, we just load the required ODE infrastructure, which can also handle
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+# The outflow boundary condition has been applied on the right side of the
+# cylinder when the weak form has been derived by setting the boundary integral
+# to zero. It is also called the do-nothing condition. Other outflow conditions
+# are also possible.
+∂Ω_free = getfaceset(grid, "right");
+
+close!(ch)
+update!(ch, 0.0);
+
+# ### Linear System Assembly
```
koehlerson, line 167: maybe add here once again the weak form and mark with `underbrace` which parts correspond to stokes, advective and mass terms

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+        assemble!(mass_assembler, celldofs(cell), Mₑ)
+    end
+
+    return M
+end;
+
+# Next we discuss the assembly of the Stokes matrix.
+# Remember that we use the same function spaces for trial and test, hence the
+# matrix has the following block form
+# ```math
```
koehlerson, line 210: this part isn't displayed properly in the docs deployment, maybe try `bmatrix`?

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+            # Viscosity term "A"
+            #+
+            for i in 1:n_basefuncs_v
+                ∇φᵢ = shape_gradient(cellvalues_v, q_point, i)
+                for j in 1:n_basefuncs_v
+                    ∇φⱼ = shape_gradient(cellvalues_v, q_point, j)
+                    Kₑ[BlockIndex((v▄, v▄), (i, j))] -= ν * ∇φᵢ ⊡ ∇φⱼ * dΩ
+                end
+            end
+            # Pressure + Incompressibility term B - note the symmetry.
```
koehlerson, line 247:
```suggestion
            # Pressure + Incompressibility term $B$ - note the symmetry.
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+
+    @inbounds for cell in CellIterator(dh)
+        fill!(Kₑ, 0)
+
+        Ferrite.reinit!(cellvalues_v, cell)
+        Ferrite.reinit!(cellvalues_p, cell)
+
+        for q_point in 1:getnquadpoints(cellvalues_v)
+            dΩ = getdetJdV(cellvalues_v, q_point)
+            # Viscosity term "A"
```
koehlerson, line 238:
```suggestion
            # Viscosity term $A$
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+end;
+
+# For the linear equations we can cleanly integrate with the linear solver
+# interface provided by the DifferentialEquations ecosystem. We use a direct
+# solver for simplicity, altough it comes with some issues. Implementing
+# GMRES with efficient preconditioner is left open for future work.
+mutable struct FerriteLinSolve{CH,F}
+    ch::CH
+    factorization::F
+    A
```
koehlerson, line 326:
```suggestion
    A::Union{Nothing, SparseMatrixCSC}
```
I guess, can also be parametrized, at least some information should be beneficial for the compiler, see https://docs.julialang.org/en/v1/manual/performance-tips/#Type-declarations

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+end
+FerriteLinSolve(ch) = FerriteLinSolve(ch,lu,nothing)
```
koehlerson, line 328: one white line between each function declaration would look better imho. Does the function need to have this `p::` closure? It looks not really beginner friendly, but I guess we are constrained to this since DiffEq calls this?

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+    cache.tstep = integrator.t + nlsolver.c * dt
+    OrdinaryDiffEq.calculate_residuals!(weight, fill!(weight, one(eltype(u))), uprev, u,
+                         opts.abstol, opts.reltol, opts.internalnorm, t);
+
+    # Before starting the nonlinear solve we have to set the time correctly.
+    # Note that ch is a global variable for now.
+    #+
+    update!(ch, cache.tstep);
+
+    # The update of u takes uprev + z or tmp + z most of the time, so we have
```
koehlerson, line 308:
```suggestion
    # The update of `u` takes `uprev + z` or `tmp + z` most of the time, so we have
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+        ## Update factorization
+        p.A = p.factorization(A)
+    end
+    ldiv!(x, p.A, b)
+    apply_zero!(x, p.ch)
+    return nothing
+end;
+
+# DifferentialEquations assumes dense matrices by default, which is not
+# feasible for semi-discretization of finize element models. We communicate
```
koehlerson, line 345:
```suggestion
# feasible for semi-discretization of finite element models. We communicate
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+        all_celldofs = celldofs(cell)
+        v_celldofs = all_celldofs[dof_range(dh, :v)]
+        v_cell = u[v_celldofs]
+        for q_point in 1:getnquadpoints(cellvalues_v)
+            dΩ = getdetJdV(cellvalues_v, q_point)
+            ∇v = function_gradient(cellvalues_v, q_point, v_cell)
+            v = function_value(cellvalues_v, q_point, v_cell)
+            for j in 1:n_basefuncs
+                φⱼ = shape_value(cellvalues_v, q_point, j)
+                # Note that the order the gradient term is now on the left, which is the correct thing to do here.
```
koehlerson, line 372:
```suggestion
                # Note that the other the gradient term is now on the left, which is the correct thing to do here.
```
not sure, but I guess `other`

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,463 @@
...
+                # It can be quickly shown through index notation.
+                # ```math
+                # [(v \cdot \nabla) v]_i = v_j \partial_j v_i = \partial_j v_i v_j = (\nabla v) v
```
koehlerson, line 375:
```suggestion
                # [(v \cdot \nabla) v] = v_i \partial_i v_j = \partial_i v_i v_j = (\nabla v) v
```

Otherwise you'd have to write 
```md
               # [(v \cdot \nabla) v]_{ij} = v_i \partial_i v_j = \partial_i v_i v_j = (\nabla v) v
```

Because in tensor notation you only apply indices to a bracket if you deconstruct every inner running index, in order to show the order of the tensor. At least this is what is done in the mechanics community and "canonical" in the sense of tensor notation. Sometimes, also utilized to denote a derivative with running index to a bracket by e.g. `(v_i v_j)_{,k}` would be the derivative with running index `k` to the inner arguments. Nonetheless, the indices should start first with i and secondly with j, by that you get naturally the correct nabla formulation, you can also be very explicit by 

```md
               # [(v \cdot \nabla) v] = v_i \boldsymbol{e}_i \cdot \partial_j \boldsymbol{e}_j v_k \boldsymbol{e}_k = v_i \partial_j \delta_{ij} v_k \boldsymbol{e}_k = v_i \partial_i \boldsymbol{e}_k v_k \boldsymbol{e}_k = \partial_i v_i v_j = (\nabla v) v
```

koehlerson (commented):
added a few corrections and suggestions :)

fredrikekre:
> A major weakness for now is that I have to overwrite one of the internal methods from the OrdinaryDiffEq package (OrdinaryDiffEq.initialize!) to apply the Dirichlet BC correctly in the time-dependent case.

Would it make sense to make a feature request about this then?

termi-official:
Thanks for the feedback @koehlerson , I will address your points later.

@fredrikekre regarding the feature request I am not sure what exactly I should request here. Technically it is at the current state already possible to not overwrite the initialize! method. I found here 3 ways
A) useing Cassette.jl and patch the initialize! method (for Newton) instead
B) developing a wrapper for DiffEq's Newton
C) utilize operators for the BCs

I think A is a bit too hacky, so I didn't wanted to go this route.

I tried B but failed for now, because the NL solver appears at many points in the code. Should not be impossible tho, but will take more time. A problem with this approach is, that not every solver utilizes the this Newton loop. The NL solve framework is quite fragmented atm, as I discussed with the maintainer (Chris) on Slack, and there is a chance that this will be refactored in the future. For the time steppers which do not utilize the internal NL solve framework (e.g. explicit and Rosenbrock-type methods) I could not find a way to enforce time-dependent DBCs at all. I also cannot guarantee in the current implementation, that all time steppers which utilize the Newton will work, because I rely on the assumption that the time stepper updates the solution $u$ either via $uprev + dt_{stage}*z$ or $tmp + dt_{stage}*z$.

Regarding C: For linear problems we could adopt the trick from [DiffEqOperators.jl](https://github.com/SciML/DiffEqOperators.jl), but I also could not find a way to generalize this to nonlinear problems.


To elaborate a bit more on why I pursuit B over the alternatives. The goal of #325 is to develop a staggered/multilevel-type Newton solver integration. These algorithms are efficient when we have nonlinear equations at the integration points, as e.g. in several plasticity models. With this PR I wanted to explore the simpler case, when we do not have any additional equations at the integration points.

docs/src/literate/ns_vs_diffeq.jl
termi-official: For the second call we need this closure, because the function is stateful, carrying the sparse LU decomposition and the constraint handler.

docs/src/literate/ns_vs_diffeq.jl
termi-official: Also, for some reason DiffEq chokes if we do this.

koehlerson: hm but then it should still be possible to constraint the type by some parametrization e.g. `Union{Nothing,T}` and T is then `<: Union{MatrixFree,AbstractMatrix}`. Having no constraint doesn't feel good to me, since beginner will take the code and most likely copy paste it

docs/src/literate/ns_vs_diffeq.jl
koehlerson: Sorry, question was misleading, so we define here something like a functor right? I'm just wondering if we can instead use a normal function, otherwise we should at least explain why this is needed and how it works with a link to https://docs.julialang.org/en/v1/manual/methods/index.html#Function-like-objects-1

Beginners will have a hard time figuring out what is happening in the 
```julia
function (p::FerriteLinSolve)()
```
part

termi-official: I tried to put in a normal function for the first one and got a very strange type error. Not sure what exactly is breaking. If you want you can take a closer look on this one, I will push my remaining fixes in a few minutes.

For this I just derived some quick and dirty example from https://diffeq.sciml.ai/stable/features/linear_nonlinear/#Implementing-Your-Own-LinSolve:-How-LinSolveFactorize-Was-Created

docs/src/literate/ns_vs_diffeq.jl
termi-official: No constraints should be fine here. I do not think that possible type instabilities here could lead to bottlenecks - usually either the evaluation of the right hand side, assembly or the solve is the bottleneck.

docs/src/literate/ns_vs_diffeq.jl
termi-official: Is this also resolved?

docs/src/literate/ns_vs_diffeq.jl
koehlerson: It's primarly about providing good pratical code examples and having unconstrained types in structs is not advised according to the docs. I benchmarked it, its just 4s difference 56s vs 52s with parametrization. But still, 7% by a trivial change. 

In case you want to change it:
```julia
FerriteLinSolve(ch) = FerriteLinSolve(ch,lu,lu(sparse(ones(1,1))))

mutable struct FerriteLinSolve{CH,F,T<:Factorization}
    ch::CH
    factorization::F
    A::T
end
```

docs/src/literate/ns_vs_diffeq.jl
koehlerson: yes :)

docs/src/literate/ns_vs_diffeq.jl
termi-official: Okay, I agree on this point, thanks!

docs/src/literate/ns_vs_diffeq.jl
koehlerson: We can just leave it as it is. But I would recommend to mention the link to the docs or at least mention functor somewhere in between

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,507 @@
...
+# [von-Karman vortex streets](https://en.wikipedia.org/wiki/K%C3%A1rm%C3%A1n_vortex_street). Within this example, we show how to utilize [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl)
+# in tandem with Ferrite.jl. To keep things simple we use a naive approach
+# to discretize the system.
+#
+# ## Remarks on DifferentialEquations.jl
+#
+# Many "time step solvers" of [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl) assume that that the
+# problem is provided in mass matrix form. The incompressible Navier-Stokes
+# equations as stated above yield a DAE in this form after applying a spatial
+# discretization technique - in our case FEM. The mass matrix form a ODEs and DAEs
```
koehlerson, line 17:
```suggestion
# discretization technique - in our case FEM. The mass matrix form of ODEs and DAEs
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,507 @@
...
+#
+# ### Derivation of Semi-Discrete Weak Form
+#
+# By multiplying test functions $\varphi$ and $\psi$ from a suitable test function space on the strong form,
+# followed by integrating over the domain and applying partial integration to the pressure and viscosity terms
+# we can obtain the following weak form
+# ```math
+#  \begin{aligned}
+#    \int \partial_t v \cdot \varphi &= - \int \nu \nabla v : \nabla \varphi - \int (v \cdot \nabla) v \cdot \varphi + \int p (\nabla \cdot \varphi) + \int_{\partial \Omega_{N}} \underbrace{(\nu \partial_n v - p n )}_{=0} \cdot \varphi \\
+#                                  0 &= (\int \nabla \cdot v) \psi
```
koehlerson, line 75:
```suggestion
#                                  0 &= \left(\int \nabla \cdot v \right) \psi
```
not necessary but it's r/mildlyinfuriating without `\left\right`:D

koehlerson: oh wait, is the opening bracket actually intended before the integral? Or should it be in front of the nabla?

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,507 @@
...
+# of "$d_t p_i = 0$", because otherwise the equation system gets singular. This
+# is obvious when we also take into account that our mass matrix is zero for these
+# dofs, such that we obtain the equation $0 \cdot d_t p_i = p_i$.
+struct RHSparams
+    K::SparseMatrixCSC
+    ch::ConstraintHandler
+    dh::DofHandler
+    cellvalues_v::CellVectorValues
+end
+p = RHSparams(K, ch, dh, cellvalues_v)
```
koehlerson, line 364:
```suggestion
p = RHSparams(K, ch, dh, cellvalues_v)

```
white line, otherwise function declaration and RHSparams stuff is in one Documenter code block

koehlerson: grafik

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,507 @@
...
+# enable the progress bar with the `progess` and `progress_steps` arguments.
+# Finally we have to communicate the time step length and initialization
+# algorithm. Since we start with a valid initial state we do not use one of
+# DifferentialEquations.jl initialization algorithms.
+# NOTE: At the time of writing this [no Hessenberg index 2 initialization is implemented](https://github.com/SciML/OrdinaryDiffEq.jl/issues/1019).
+#
+# To visualize the result we export the grid and our fields
+# to VTK-files, which can be viewed in [ParaView](https://www.paraview.org/)
+# by utilizing the corresponding pvd file.
+#timestepper = ImplicitEuler(linsolve=FerriteLinSolve(ch))
```
koehlerson, line 435:
```suggestion

```
legacy

docs/src/literate/ns_vs_diffeq.jl
termi-official: The integral should be on the outside

koehlerson:
LGTM :)

termi-official:
Thanks for the detailed review Maxi! This one should be ready - however some future work is left open.

1. Mesh IO / refactoring of the mesh generator: Currently I just hack together a holed domain. Probably not the best way. Should definitely be addressed once we have some mesh IO. Or the mesh generator should be refactored.
2. Preconditioner+iterative solver: Once this is merged I will look into the dof handler for a follow up PR to address the assembly mode. Currently we assemble blocked systems as `ABABAB....`, where `A` and `B` are representing the fields. In the case of incompressible Navier-Stokes A=velocity and B=pressure. However, some preconditioners require the system to be in the form `AAAA.....BBBB....`, see e.g. https://www.dealii.org/current/doxygen/deal.II/step_57.html#TheSolverandPreconditioner

KristofferC:
> he example has tests included, but for now I didn't wanted to introduce the fat DifferentialEquations dependency into Ferrite just for the test (currently the dep is just in the doc).

You can add test specific dependencies: https://github.com/Ferrite-FEM/Ferrite.jl/blob/b44132c6d9a2e4b0dbadc582635a6bb3dff086b6/Project.toml#L18-L33

fredrikekre:
Good to go? Nice that you managed without overwriting DiffEq internals!

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,500 @@
...
+    #compress=false flag because otherwise each vtk file will be stored in memory
+    vtk_grid("vortex-street-$t.vtu", dh; compress=false) do vtk
+        vtk_point_data(vtk,dh,u)
+        vtk_save(vtk)
+        pvd[t] = vtk
+    end
+end
+vtk_save(pvd);
+
+# Test the result for full proper development of the flow                   #hide
```
koehlerson, line 453: renders in docs, should be hidden

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,500 @@
...
+# of $v$ and $p$ respectively, while $\hat{\varphi}$ is the choice for the test function in the discretization. The hats are dropped
+# in the implementation and only stated for clarity in this section.
+#
+#
+# ## Commented Implementation
+#
+# Now we solve the problem with Ferrite and [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl). What follows is a program spliced with comments.
+# The full program, without comments, can be found in the next [section](@ref ns_vs_diffeq-plain-program).
+#
+# First we load Ferrite, and some other packages we need
```
koehlerson, line 115:
```suggestion
# First we load Ferrite and some other packages we need
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,500 @@
...
+for facesetname in keys(grid.facesets)
+    grid.facesets[facesetname] = Set(map(fi -> FaceIndex( cell_indices_map[fi.idx[1]] ,fi.idx[2]), collect(grid.facesets[facesetname])))
+end;
+
+# We test against full development of the flow - so regenerate the grid                              #src
+grid = generate_grid(Quadrilateral, (x_cells, y_cells), Vec{2}((0.0, 0.0)), Vec{2}((0.55, 0.41)));   #hide
+
+# ### Function Space
+# To ensure stability we utilize the Taylor-Hood element pair Q2-Q1.
+# We have to utilize the same quadrature rule because in the weak form the
```
koehlerson, line 159:
```suggestion
# We have to utilize the same quadrature rule for the velocity as for the pressure, because in the weak form the
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,500 @@
...
+
+# To apply the nonlinear portion of the Navier-Stokes problem we simply hand
+# over the dof handler and cell values to the right hand side as a parameter.
+# Further the pre-assembled linear part (which is time independent) is
+# passed to save some runtime. To apply the time-dependent Dirichlet BCs, we
+# also hand over the constraint handler.
+# The basic idea to apply the Dirichlet BCs consistently is that we copy the
+# current solution `u`, apply the Dirichlet BCs on the copy, evaluate the
+# discretized right hand side of the Navier-Stokes equations with this vector
+# and finally set the rhs to zero on every constraint. This way we obtain a
```
koehlerson, line 340:
```suggestion
# and finally set the r.h.s. to zero on every constraint. This way we obtain a
```

docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -0,0 +1,500 @@
...
+jac_sparsity = sparse(K);
+
+# To apply the nonlinear portion of the Navier-Stokes problem we simply hand
+# over the dof handler and cell values to the right hand side as a parameter.
+# Further the pre-assembled linear part (which is time independent) is
+# passed to save some runtime. To apply the time-dependent Dirichlet BCs, we
+# also hand over the constraint handler.
+# The basic idea to apply the Dirichlet BCs consistently is that we copy the
+# current solution `u`, apply the Dirichlet BCs on the copy, evaluate the
+# discretized right hand side of the Navier-Stokes equations with this vector
```
koehlerson, line 339:
```suggestion
# discretized right hand side (r.h.s.) of the Navier-Stokes equations with this vector
```

koehlerson (commented):
last few bits before merging

termi-official:
Should be fine @fredrikekre . Docs builds properly on my machine. Dunno why the the last push did not trigger the CI. Once this is merged I will open up the follow up issues.

fredrikekre:
Seems to be a github issue: https://www.githubstatus.com/

koehlerson (approved):

fredrikekre:
"Simple Example" 😛

KristofferC:
7drHiqr

PR #373, opened 2021-09-10
Reduce the ouput in quasi-incompressible hyperelasticity example

bhaveshshrimali:
Sorry for the hurried implementation in #370. While there are no major changes to the code anywhere,
as [discussed](https://github.com/Ferrite-FEM/Ferrite.jl/pull/370#issuecomment-916266027) in #370, the current output to screen is humongous. Since all of it is not necessary, it suffices to output just the overall progress at specific time steps. The current code outputs [this](https://github.com/Ferrite-FEM/Ferrite.jl/pull/370#issuecomment-916519967) which is both terse and informative

### DIFF
--- a/docs/src/literate/quasi_incompressible_hyperelasticity.jl
+++ b/docs/src/literate/quasi_incompressible_hyperelasticity.jl
@@ -23,16 +23,16 @@
 # ```math
 #     \Psi(u) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
 # ```
-# where I₁=F:F≡FᵢⱼFᵢⱼ and J = det(F) denote the standard invariants of the deformation gradient tensor F = 𝛪 + ∇u.
+# where $I_1 = F:F = F_{ij}F_{ij}$ and J = det(F) denote the standard invariants of the deformation gradient tensor $F = \mathbb{I}+\nabla u$.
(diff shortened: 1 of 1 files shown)


docs/src/literate/quasi_incompressible_hyperelasticity.jl
```diff
@@ -332,15 +332,13 @@ function solve(interpolation_u, interpolation_p)
...
 
         ## Save the solution fields
-        @timeit "export" begin
+        begin
             vtk_grid("hyperelasticity_incomp_mixed_$t.vtu", dh, compress=false) do vtkfile
                 vtk_point_data(vtkfile, dh, w)
                 vtk_save(vtkfile)
                 pvd[t] = vtkfile
             end
         end
```
fredrikekre, line 341: This `begin`/`end` can be removed now:
```suggestion
        vtk_grid("hyperelasticity_incomp_mixed_$t.vtu", dh, compress=false) do vtkfile
            vtk_point_data(vtkfile, dh, w)
            vtk_save(vtkfile)
            pvd[t] = vtkfile
        end
```

fredrikekre (commented):
If you still want the timer output you can probably just move the printing outside the loop and print it ones instead.

bhaveshshrimali:
> If you still want the timer output you can probably just move the printing outside the loop and print it ones instead.

Would be fine either way but adding the timer outputs wouldn't add anything additional to the demonstration per-se.

PR #376, opened 2021-10-05
Update plasticity example to use immutable structs for material states.

fredrikekre:


docs/src/literate/plasticity.jl
```diff
@@ -398,7 +376,7 @@ xlabel!("Maximum deflection [m]")
 
 ## test the result                       #src
 using Test                               #src
-@test norm(u_max[end]) ≈ 0.2544526451    #src
+@test norm(u_max[end]) ≈ 0.255056309     #src
```
fredrikekre, line 379: :)))))))

I guess `symmetric(function_gradient(..))` -> `function_symmetric_gradient(...)` caused this.

kimauth: Shouldn't that literally be the same? https://github.com/Ferrite-FEM/Ferrite.jl/blob/c585e1c29efe2173fd91bfc17d824ae873db1ee9/src/FEValues/common_values.jl#L210:L213

fredrikekre: Huh, then I don't know why there would be a difference.

docs/src/literate/plasticity.jl
```diff
@@ -346,8 +324,8 @@ function solve()
         end
 
         ## Update all the material states after we have reached equilibrium
-        for cell_states in states
-            foreach(update_state!, cell_states)
+        for i in eachindex(states, states_old)
+            states[i] .= states_old[i]
```
KristofferC, line 328: `copy!(states, states_old)`?

fredrikekre: Doesn't work recursively. `states is `Vector{Vector{MaterialState}}`.

KristofferC: Make it a `Matrix` :D

kimauth: Haha, I feel a bit guilty for the `Vector{Vector{MaterialState}}` now. 😅  What is the advantages of using a Matrix?

KristofferC: If you have a `Matrix` you know that all the "rows" are of equal length. All the data is also stored contiguously. You also only make one allocation instead of one per element. But the drawback is of course that you need to have the same number of GPs per element.

kimauth: Does it really matter in practice if data for cells is stored contiguously? E.g. if I'm using a multithreaded assembly, I access the cells in a potentially random order and it doesn't matter if data for cell 1 is next to data for cell 2, does it? Is should only matter that all integration points of a cell store their data contiguously.

KristofferC: Not sure. That's when I would benchmark :p

termi-official: Having a vector of vectors can be beneficial for multithreading to suppress potential false sharing if you have complete random accesses. From my experience on multithreaded applications it can be helpful to have thread local caches on which each thread operates. So, if you are working on a single thread I would bet that having the data contiguously stored is faster. But I agree, in the end I would also opt for benchmarking everything.

Also, I since we discuss this here, maybe it could be helpful for users to have a thin abstraction on these L2 variables/quadrature spaces - this way we can also potentially switch between different data structures in the backend. I ran into a similar issue in PR 325 - there I can not find a good way to pass the local problems to DiffEq. I think it might be further useful when working with the `MixedDofHandler` to avoid potential issues (and code duplication) with subdomains.

PR #379, opened 2021-10-06
DofHandler with any <: AbstractGrid

koehlerson:
This PR is in the same spirit of https://github.com/Ferrite-FEM/Ferrite.jl/pull/309 and not such a crowded catastrophe as https://github.com/Ferrite-FEM/Ferrite.jl/pull/307

I added `cellnodes!` and `cellcoords!` dispatches for `AbstractGrid` that only call the interface function of `<:AbstractGrid`. I noticed that this is needed when we did the first `<:AbstractGrid` prototype. I'm not sure if these definitions should better live in grid.jl. They assume a concrete homogeneous `cells` vector right now, since they are only called by `DofHandler`.

I'd like to do a follow up PR such that a `CellIterator` also can handle any <:AbstractGrid. Then, we would only need consistent usage of the interface functions and everything should work ootb for properly subtyped grids


koehlerson:
grafik

koehlerson:
grafik

kimauth (approved):

koehlerson:
Comparing heat equation on master vs this branch on my laptop (subject to a plethora of fluctuations) in order to see if there is a significant performance drawback. Note this is not intended as performance benchmark but rather as a rough estimate if something was utterly needed w.r.t. type instability

## PR
200 x 200 elements
julia> @time include("docs/src/examples/heat_equation.jl")
  1.148711 seconds (187.30 k allocations: 113.092 MiB, 2.27% gc time, 19.11% compilation time)
 
500 x 500 elements
julia> @time include("docs/src/examples/heat_equation.jl")
  6.478253 seconds (400.90 k allocations: 702.862 MiB, 4.93% gc time, 3.64% compilation time)
  
 ## Master
200 x 200 elements
julia> @time include("docs/src/examples/heat_equation.jl")
  1.152864 seconds (201.76 k allocations: 113.971 MiB, 11.94% gc time, 21.89% compilation time)
  
500 x 500 elements
julia> @time include("docs/src/examples/heat_equation.jl")
  6.244453 seconds (401.77 k allocations: 702.985 MiB, 3.77% gc time, 3.58% compilation time)

koehlerson:
ping @fredrikekre

src/Dofs/DofHandler.jl
```diff
@@ -288,6 +288,23 @@ function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::Grid{dim,C}, i::In
...
+    return global_nodes
+end
+
+function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid{dim}, i::Int) where {dim,T}
+    C = getcelltype(grid)
+    @assert length(global_coords) == nnodes(C)
+    global_coords .= getcoordinates(grid, i)
+    return global_coords
+end
+
```
fredrikekre, line 307: Not needed after the change in https://github.com/Ferrite-FEM/Ferrite.jl/pull/425/files#diff-8db5ed06971ce1004169862f3e499f55666dd630868843be8a53d30ea385e505?

fredrikekre (commented):
Rebase?

src/Dofs/DofHandler.jl
koehlerson: what we really need is: https://github.com/Ferrite-FEM/Ferrite.jl/blob/81440fabd5434820e6b3fb6b7a4529830d90a111/src/Dofs/DofHandler.jl#L315-L329 from Kim's https://github.com/Ferrite-FEM/Ferrite.jl/pull/445 

It's supposed to handle the information for the `reinit!` which needs the actual coords for any grid. The functionality can be recovered by using the already defined functions on `AbstractGrid`, so `getcoordinates` etc.

koehlerson: I can either delete the functions completely and wait for Kim's PR or we take the ones from Kim's PR with the slight improvement suggestion I made over there

kimauth: We almost have this since #425 though (see Fredriks comment above). Why not just make the changes to that one? Allowing `AbstractGrid` in the function signature and using `getcells` + `getnodes` should literally have no impact on using the function with the regular `Grid`.

koehlerson: Sorry, I think I misunderstood the comment in the beginning. I changed in the latest commit the functions to work for any `<: AbstractGrid` (I hope you meant this @kimauth, otherwise I'll revert it) I had to use the `getcoordinates` dispatch for the `Node` type as an abstraction in between, otherwise grids always need to use `Node` as their node type. In the test example I only used a `NTuple{dim}`, but we can also just assume that `Node` is always used, if you prefer this.

kimauth: Looks like what I meant. 
I think assuming that a node always has a field `x` for the coordinate is not too much of a stretch - but considering that we have `getcoordinates(node)` it's good to use I guess.

PR #388, opened 2021-10-12
Performance regression tests

### LINKED ISSUE #94: Benchmarking
https://github.com/JuliaCI/PkgBenchmark.jl

termi-official:
Introduce performance regression benchmarks through PkgBenchmark.jl on the most important pieces of Ferrite. Resolves #94 .

## Benchmark Coverage

* Common Lagrange assembly loops
* Available hyperrectangular mesh generators
* `close!(dh)` for the most common Lagrange problems (DofHandler + MixedDofHandler)
* Application of Dirichlet BCs with both available strategies
 
## What is missing

* CI integration: Doing this should be addressed by a follow up PR. [Currently a problem is that we cannot select subsets of the benchmark suite](https://github.com/JuliaCI/PkgBenchmark.jl/issues/132) and running the complete suite really takes some time. Integration can be achieved via BenchmarkCI.jl
* Micobenchmarks for utils: We have a plethora of utility functions which are not covered yet. We should at least cover non-lookup based utilities in future benchmarks.
* Embedded elements in 2D and 3D due to missing capabilities


koehlerson (commented):
nice, maybe add also an assembly where not the `CellIterator` is used, but rather

```julia
for cellid in 1:getncells(grid)
    # assemble code
    assemble!(assembler, celldofs(dh,cellid), Me, fe)
end
```

KristofferC:
> 1h tuning and I have not run the benchmark further than tuning yet.

Yikes, that seems a bit much. I think this is perhaps a little bit excessive to try all different combinations of dimensions etc. They all run pretty much the same code so it seems quite unlikely that one of them will randomly be super slow.

Very slow thorough benchmarks are worse than fast not so thorough benchmarks because slow benchmarks will never be run.

termi-official:
I agree on most points and I think having the option to run just some benchmarks would certainly help, but this feature is not yet implemented in PkgBenchmark (see https://github.com/JuliaCI/PkgBenchmark.jl/issues/132).

The rationale behind this was to provide infrastructure for possible new dispatches on close! and benchmarks with different dof handlers/grid types/.... I could reduce the loops to just include something like "2^dim element" grid in 3d with field-dim = spatial-dim (e.g. commonly incompressible elasticity). I would leave the current loops commented out such that we can reuse them once the feature is implemented in PkgBenchmark.jl

termi-official:
I took a look into implementing running subsets of the benchmark but cannot find a good angle on this. Problem is consistent integration with the tuning procedure. Nonetheless, I think this implementation is helpful to compare different approaches and to catch performance regressions, especially since we currently refactor a bit and add quite a bit new features. I think I can just add two additional CI jobs to cover such benchmarks (for nightly and LTS) via BenchmarkCI.jl.

Project.toml
```diff
@@ -23,6 +23,7 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 IterativeSolvers = "42fd0dbc-a981-5370-80f2-aaf504508153"
 KrylovMethods = "9a2cd570-f05c-5dc1-9209-93ad6f5727f7"
 NBInclude = "0db19996-df87-5ea3-a455-e3a50d440464"
+PkgBenchmark = "32113eaa-f34f-5b0d-bd6c-c81e245fc73d"
```
KristofferC, line 26: Should probably be in `benchmark/Project.toml`?

termi-official: Sorry to ask, but how can I ensure that this subproject uses the correct Ferrite branch as dependency? `add ..` seems to fix the branch
```
...
[c061ca5d] + Ferrite v0.3.8 `..#do/performance-regression-tests`).
...
```

termi-official:
I added a makefile for the most common tasks and some docs. Also, I manually split up the benchmarks into small groups which can be executed individually for more fine-grained control (since PkgBenchmark.jl does not have such a feature).

docs/src/index.md
```diff
@@ -64,3 +64,50 @@ Ferrite.debug_mode(enable=false)
 ```
 
 also followed by restarting the Julia process.
+
+## Performance Analysis
```
fredrikekre, line 68: Could (should?) move to a new page in devdocs.

termi-official: Give me 5 minutes.

PR #390, opened 2021-10-12
Implement export of projected fields using the projector

fredrikekre:
Implement export of projected fields using the projector and simplify `vtk_point_data` with tensor output.


kimauth:
Changing the export order of regular tensors is sort of breaking, is it? In Paraview the entries will be "shuffled" up compared to the previous export and it isn't obvious which one is which from what you can see in Paraview.

Other than that, this looks good.

fredrikekre:
I made a PR to label non-symmetric tensors too in ParaView: https://gitlab.kitware.com/paraview/paraview/-/merge_requests/5442

Issue #398, opened 2021-10-21
Infrastructure for matrix-valued problem


termi-official:
PR #392 should have the batteries for DG - but just for scalar problems. If we have vector valued problems, then our flux space becomes matrix-valued (think about displacement and stress). For convenience we could provide some handlers for handling matrix valued problems. This way DG can be implemented in a cleaner way than reformulating everything in Voigt notation beforehand.

PR #404, opened 2022-01-04
WIP: Traits for `ScalarValues` / `VectorValues`

kimauth:
A suggestion on using traits instead of Unions for defining `ScalarValues` and `VectorValues`.

I want to use it for subtyping `Values` without copying a bunch of code from Ferrite. Could also be useful for custom definition of (less common?) problems like the one mentioned in #398. 

I'm happy to hear some feedback, ~~if this is appreciated I'll add some docs as well.~~

### DIFF
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -116,9 +116,8 @@ nodal values of ``\\mathbf{u}``.
 function_value(fe_v::T, q_point, u, dof_range) where T = function_value(FieldTrait(T), fe_v, q_point, u, dof_range)
 function_value(fe_v::T, q_point, u) where T = function_value(FieldTrait(T), fe_v, q_point, u)
 
-function function_value(ft::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
-    n_base_funcs = getn_scalarbasefunctions(fe_v)
-    isa(ft, VectorValued) && (n_base_funcs *= dim)
+function function_value(::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
+    n_base_funcs = getnbasefunctions(fe_v)
     @assert length(dof_range) == n_base_funcs
     @boundscheck checkbounds(u, dof_range)
     val = zero(_valuetype(fe_v, u))
@@ -167,9 +166,8 @@ where ``\\mathbf{u}_i`` are the nodal values of ``\\mathbf{u}``.
 function_gradient(fe_v::T, q_point, u) where T = function_gradient(FieldTrait(T), fe_v, q_point, u)
 function_gradient(fe_v::T, q_point, u, dof_range) where T = function_gradient(FieldTrait(T), fe_v, q_point, u, dof_range)
 
-function function_gradient(ft::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
-    n_base_funcs = getn_scalarbasefunctions(fe_v)
-    isa(ft, VectorValued) && (n_base_funcs *= dim)
+function function_gradient(::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
+    n_base_funcs = getnbasefunctions(fe_v)
     @assert length(dof_range) == n_base_funcs
     @boundscheck checkbounds(u, dof_range)
     grad = zero(_gradienttype(fe_v, u))
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -251,7 +251,6 @@ function function_divergence(::IsScalarValued, fe_v::Values{dim}, q_point::Int, 
     return diverg
 end
 
-# See https://github.com/Ferrite-FEM/Ferrite.jl/issues/336
 function_divergence(::IsVectorValued, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T} =
     tr(function_gradient(fe_v, q_point, u, dof_range))
 
@@ -288,7 +287,6 @@ function Base.show(io::IO, ::MIME"text/plain", fe_v::Values)
 end
 
 # copy
-# TODO: needs to be adapted to new traits
 for ValueType in (CellScalarValues, CellVectorValues, FaceScalarValues, FaceVectorValues)
     args = [:(copy(cv.$fname)) for fname in fieldnames(ValueType)]
     @eval begin
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -2,14 +2,23 @@
 
 using Base: @propagate_inbounds
 
-const ScalarValues{dim,T,shape} = Union{CellScalarValues{dim,T,shape},FaceScalarValues{dim,T,shape},PointScalarValues{dim,T,shape}}
-const VectorValues{dim,T,shape} = Union{CellVectorValues{dim,T,shape},FaceVectorValues{dim,T,shape}}
+abstract type FieldTrait end
+struct IsVectorValued <: FieldTrait end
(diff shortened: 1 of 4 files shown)


src/FEValues/common_values.jl
```diff
@@ -103,9 +113,12 @@ where ``u_i`` are the value of ``u`` in the nodes. For a vector valued function
...
 nodal values of ``\\mathbf{u}``.
 """
-function function_value(fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
+function_value(fe_v::T, q_point, u, dof_range) where T = function_value(FieldTrait(T), fe_v, q_point, u, dof_range)
+function_value(fe_v::T, q_point, u) where T = function_value(FieldTrait(T), fe_v, q_point, u)
+
+function function_value(ft::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
     n_base_funcs = getn_scalarbasefunctions(fe_v)
-    isa(fe_v, VectorValues) && (n_base_funcs *= dim)
+    isa(ft, IsVectorValued) && (n_base_funcs *= dim)
```
termi-official, line 121: Sorry for the question, but isn't this just `getnbasefunctions` ?

kimauth: Good point, I think `getnbasefunctions` should do the job here. Any reason that wasn't used @fredrikekre ?

src/FEValues/common_values.jl
```diff
@@ -2,14 +2,23 @@
...
-const VectorValues{dim,T,shape} = Union{CellVectorValues{dim,T,shape},FaceVectorValues{dim,T,shape}}
+abstract type FieldTrait end
+struct IsVectorValued <: FieldTrait end
+struct IsScalarValued <: FieldTrait end
+
+FieldTrait(::Type{<:CellScalarValues}) = IsScalarValued()
+FieldTrait(::Type{<:FaceScalarValues}) = IsScalarValued()
+FieldTrait(::Type{<:PointScalarValues}) = IsScalarValued()
+FieldTrait(::Type{<:CellVectorValues}) = IsVectorValued()
+FieldTrait(::Type{<:FaceVectorValues}) = IsVectorValued()
```
termi-official, line 13: I really like the idea. 

Maybe we should be a bit more precise here and adjust the name? I mean, we could utilize the FieldTraits for more things like e.g. the conformity-class (e.g. H1, L2, H(div), H(curl), ...).

My first suggestion would be `FieldValueTrait` - however I also do not like this name too much.

termi-official: To give one example, H(curl) and H(div) need different transformations from and to the reference element than L2 and H1 conforming formulations. See e.g. https://www.cambridge.org/core/services/aop-cambridge-core/content/view/4B91C5FDC5131BE591D41F8577DEB4F3/S0764583X10000348a.pdf/hexahedral-hdiv-and-hcurl-finite-elements-.pdf .

However, this was primarily ment to give an example why the name "field trait" may be ambiguous. :)

kimauth: Wouldn't that be handled by having special `reinit!` functions for e.g. some `HcurlValues`? 

I'm open to a better name suggestion though, no hard opinions there. `FieldTrait` would be internal to Ferrite anyways, so technically non-breaking to change (even later) it if there ever would be a hard need for a different name.

termi-official: Depends on how we want to design it. Technically speaking a CellVectorValues could do the job here already, right @koehlerson ?

koehlerson: If the `reinit!` would dispatch on the the trait: yes. Otherwise we would need a new cellvalue type

koehlerson: on the other hand one could just dispatch on the cellvalue type, mhm don't know :man_shrugging:

kimauth: My cohesive zone elements are why I want to do this at all.

I have a `SurfaceVectorValues` type that has slightly different fields and a custom `reinit!`, but I want to use all the other stuff which is defined for `VectorValues`. But before this PR, there is no way to tell Ferrite that `SurfaceVectorValues` are `VectorValues`.

Not sure if using a trait for the `reinit!` function would be the right design. As it is now this PR is just meant to allow extending the `VectorValues` and `ScalarValues`, but doesn't change anything about the programming logic of them. Perhaps we should open as separate issue with a concrete example for you other elements in order to discuss general design stuff?

koehlerson: Okay sorry I was confused, I totally see the point now. I think it's worth to discuss this further in an issue :)

src/FEValues/common_values.jl
```diff
@@ -103,9 +113,12 @@ where ``u_i`` are the value of ``u`` in the nodes. For a vector valued function
 ``\\mathbf{u}(\\mathbf{x}) = \\sum\\limits_{i = 1}^n N_i (\\mathbf{x}) \\mathbf{u}_i`` where ``\\mathbf{u}_i`` are the
 nodal values of ``\\mathbf{u}``.
 """
-function function_value(fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
+function_value(fe_v::T, q_point, u, dof_range) where T = function_value(FieldTrait(T), fe_v, q_point, u, dof_range)
+function_value(fe_v::T, q_point, u) where T = function_value(FieldTrait(T), fe_v, q_point, u)
+
+function function_value(ft::FieldTrait, fe_v::Values{dim}, q_point::Int, u::AbstractVector{T}, dof_range = eachindex(u)) where {dim,T}
```
lijas, line 119: Is it possible to keep to old function argument here since both scalarvalues and vectorvalues use the same function?

src/FEValues/common_values.jl
lijas, line 121: Yeah, I think it is possible to call getnbasefunctions directly instead (but double check this :P )

lijas (commented):
I think it looks good, and can be merged...

But we can use the traits in the reinit! methods aswell now: https://github.com/Ferrite-FEM/Ferrite.jl/blob/6684cb60cd3abc9f7d9e5415a970e62752876d2f/src/FEValues/cell_values.jl#L147

src/FEValues/common_values.jl
kimauth: I think this is needed due to the specialized version of `function_value` for vector values further down. Correct me, if I'm wrong.

src/FEValues/common_values.jl
```diff
@@ -2,14 +2,23 @@
 
 using Base: @propagate_inbounds
 
-const ScalarValues{dim,T,shape} = Union{CellScalarValues{dim,T,shape},FaceScalarValues{dim,T,shape},PointScalarValues{dim,T,shape}}
-const VectorValues{dim,T,shape} = Union{CellVectorValues{dim,T,shape},FaceVectorValues{dim,T,shape}}
+abstract type FieldTrait end
+struct IsVectorValued <: FieldTrait end
+struct IsScalarValued <: FieldTrait end
```
fredrikekre, line 7:
```suggestion
struct VectorValued <: FieldTrait end
struct ScalarValued <: FieldTrait end
```
looks better IMO.

src/FEValues/common_values.jl
fredrikekre: Maybe, if the result of that function has changed since https://github.com/Ferrite-FEM/Ferrite.jl/commit/894f4d523a4a3371868ae1f66df57d9746bd1b7a

kimauth: It has in #125

lijas:
Just a general question... why not use VectorValues as supertype instead of Face/CellValues. What dictates what trait/property should be used as supertype, and what should be implemented as traits?

src/FEValues/common_values.jl
lijas: Ok, keep it like this then

kimauth:
The reason here was to literally replace the old scalar/vector value unions. Sure could be the other way around, but that would be a larger redesign. Would you argue that it really should be the other way around?

lijas:
Not worth redesigning I think, I was just wondering if it would be better and if there is some helpful rule to decide what should be used as traits/supertypes

PR #405, opened 2022-01-06
Fixed Helmholtz example - Neumann Boundary Condition

blaszm:
The weak form formulation contained an error with the Neumann Boundary condition, which should vanish because of partial integration. Addtionally, the sign of g when added to the local force vector was incorrect. We fixed both these errors and also the documentation and the test at the end. 

Please note that the influence of this change is not directly visible in ParaView, as the used analytical function has three peaks in the middle of the region, thus the boundary condition has barely any influence. By changing the center to xs = (Vec{2}((-0.95,  0.95)),          Vec{2}((-0.95, -0.95)), Vec{2}(( 0.5,  -0.5))), the effect can be seen:
PV_helmholtz

Here, the top example is the old implementation, which shows oscillations in the bottom and a deformation in the top left corner, while the improved version looks much smoother and as one would expect the solution to look like.

### DIFF
--- a/docs/src/literate/helmholtz.jl
+++ b/docs/src/literate/helmholtz.jl
@@ -26,7 +26,7 @@
 # \int \nabla δu \cdot \nabla u d\Omega
 # + \int δu \cdot u d\Omega
 # - \int δu \cdot f d\Omega
-# - \int δu \cdot g_2 d\Gamma_2 = 0 \forall δu
+# - \int δu g_2 d\Gamma_2 = 0 \forall δu
 # ```
 #
 # where $δu$ is a suitable test function that satisfies:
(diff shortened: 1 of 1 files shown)


termi-official (approved):
LGTM.

docs/src/literate/helmholtz.jl
```diff
@@ -26,7 +26,7 @@
 # \int \nabla δu \cdot \nabla u d\Omega
 # + \int δu \cdot u d\Omega
 # - \int δu \cdot f d\Omega
-# + \int δu \cdot (n \cdot \nabla u - g_2) d\Gamma_2
+# - \int δu \cdot g_2 d\Gamma_2 = 0 \forall δu
```
kimauth, line 29: I think there is no need for the \cdot here, is there? Seems to me like these are all scalars (same for the two terms above).

kimauth: Maybe it would be easier to read if it was `int_{\Gamma_2}` as well (and similar for the other terms), once we're on changing it anyways. What do you think?

kimauth (commented):
Good catch, thank you!

PR #406, opened 2022-01-12
Export tensors with component names.

fredrikekre:
See https://github.com/jipolanco/WriteVTK.jl/pull/97


kimauth (approved):
Looks good to me :)

PR #411, opened 2022-01-27
FaceValues boundscheck on reinit!

kimauth:
check that `face` is a valid faceid, similar to #407


src/FEValues/face_values.jl
```diff
@@ -262,3 +263,10 @@ function spatial_coordinate(bcv::BCValues, q_point::Int, xh::AbstractVector{Vec{
     end
     return x
 end
+
+nfaces(fv) = size(fv.N)[3]
```
fredrikekre, line 267:
```suggestion
nfaces(fv) = size(fv.N, 3)
```

PR #412, opened 2022-01-28
Add copy(u) to not modify internal DiffEq state.

fredrikekre:


docs/src/literate/ns_vs_diffeq.jl
```diff
@@ -469,7 +469,8 @@ function compute_divergence(dh, u, cellvalues_v)                            #hid
     return divv                                                             #hide
 end                                                                         #hide
 @testset "INS OrdinaryDiffEq" begin                                         #hide
-    u = integrator.integrator.u                                             #hide
+    u = copy(integrator.integrator.u)                                       #hide
+    apply!(u, ch)                                                           #hide
```
fredrikekre, line 473: Also needs to apply here since before we mutated the internals and had the correct solution.

termi-official:
Good catch here! Accidentally introduced the reference when trying to optimize this copy. 

Changes look fine. :)

PR #415, opened 2022-02-03
Move applyzero check out of loop in apply.

fredrikekre:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/415/files?&w=1 for review


KristofferC:
Lgmt

PR #416, opened 2022-02-04
Remove export of ScalarValues and VectorValues

KnutAM:
`ScalarValues` and `VectorValues` are exported but don't seem to be defined.
I also added a test to check that no symbols are exported that haven't been defined. Not sure where to put this test though...

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,4 +22,4 @@ include("test_pointevaluation.jl")
 # include("test_notebooks.jl")
 include("test_apply_rhs.jl")
 include("test_examples.jl")
-@test isa([eval(type) for type in names(Ferrite)], Vector)  # Test that all exported symbols are defined
+@test isa([Core.eval(Ferrite,type) for type in names(Ferrite)], Vector)  # Test that all exported symbols are defined
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,3 +22,4 @@ include("test_pointevaluation.jl")
 # include("test_notebooks.jl")
 include("test_apply_rhs.jl")
 include("test_examples.jl")
+@test isa([eval(type) for type in names(Ferrite)], Vector)  # Test that all exported symbols are defined
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -22,4 +22,4 @@ include("test_pointevaluation.jl")
 # include("test_notebooks.jl")
 include("test_apply_rhs.jl")
 include("test_examples.jl")
-@test isa([Core.eval(Ferrite,type) for type in names(Ferrite)], Vector)  # Test that all exported symbols are defined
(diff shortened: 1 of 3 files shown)


fredrikekre:
Removed in https://github.com/Ferrite-FEM/Ferrite.jl/pull/404. Docs look more ugly now, it was nice to be able to reference both `CellVectorValues` and `FaceVectorValues` as `VectorValues`:P

KnutAM:
Not sure why "curl" fails, the new test passes when I run `]test` on my computer

kimauth:
We could document the `VectorValued` and `ScalarValued` Traits and refer to them in the docs instead. Though the question is where in the docs that should go, it's not really supposed to be exposed to the user.

Not sure if I like writing out `CellScalarValues` and `FaceScalarValues` everywhere, e.g. `PointScalarValues` uses some of these functions as well. That list easily becomes lengthy.

test/runtests.jl
```diff
@@ -22,3 +22,4 @@ include("test_pointevaluation.jl")
 # include("test_notebooks.jl")
 include("test_apply_rhs.jl")
 include("test_examples.jl")
+@test isa([Core.eval(Ferrite,type) for type in names(Ferrite)], Vector)  # Test that all exported symbols are defined
```
fredrikekre, line 25: I think there is a standard way of testing this, but can't find it right away.

KnutAM:
Could one write e.g "where `u` is [`VectorValued`](https://github.com/Ferrite-FEM/Ferrite.jl/pull/404)" where the link is to a page documenting what VectorValued means (but not giving this page in the menu). And also documenting `ScalarValued` on that page?

test/runtests.jl
KnutAM: https://github.com/JuliaLang/julia/issues/18780?

fredrikekre: No, I was thinking of https://github.com/JuliaTesting/Aqua.jl/blob/master/src/exports.jl perhaps, but that seems a bit overkill, perhaps
```suggestion
@test all(x -> isdefined(Ferrite, x), names(Ferrite))  # Test that all exported symbols are defined

```
is sufficient.

kimauth:
Seems not so nice to me to hide it away like that. Perhaps just an introductory section at the top of https://ferrite-fem.github.io/Ferrite.jl/stable/reference/fevalues/ which explains that there are `ScalarValues` and `VectorValues`, where the `VectorValues` are zero-padded versions of the `ScalarValues`? I remember that I found this hard to understand in the beginning.

KnutAM:
Yes, that would be even better!

kimauth:
No need to do it in this PR, the PR is after all about removing the export that I forgot to remove in #404 . I can draft something for the `VectorValues` and `ScalarValues`.

kimauth (approved):

PR #417, opened 2022-02-07
Small optimimization of constraint condensation

fredrikekre:
Use a dict of size n constraints instead of a vector of size ndofs.


lijas:
I guess it is more memory efficient, but is it faster?

lijas:
Ok makes sense. I think I use the same trick in create_constraint_matrix, so you can update that as well maybe.
Otherwise this looks good to me.

PR #418, opened 2022-02-09
Periodic Dirichlet boundary conditions

fredrikekre:


lijas (commented):
I added some short comment. I did not really check the `_add!` - function :P 

It would also be nice if you can inhomogeneities to the `PeriodicConstraint` via a function (similarly to how `Dirichlet` does it). I think you are enforcing [|um|] = 0, but I am enforcing [|u|] = [|uM|]

src/Dofs/ConstraintHandler.jl
```diff
@@ -812,3 +817,201 @@ create_symmetric_sparsity_pattern(dh::DofHandler,      ch::ConstraintHandler) =
...
+``[[\\bullet]] = u(x_{+}) - u(x_{-})`` is the "jump operator". This means that the field
+``u`` on the mirror boundary is constrained to the value of ``u`` on the image boundary.
+
+Non-homogeneouos periodic boundary conditions, ``[[u]] = C``, for some offset ``C`` are
+also supported.
+"""
+struct PeriodicDirichlet
+    field_name::Symbol
+    components::Vector{Int} # components of the field
+    face_pairs::Vector{Pair{String,String}}
```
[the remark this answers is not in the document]
fredrikekre: Not really necessary here, this struct is replaced by a number of `AffineConstraints` and never added to the dof handler.

src/Dofs/ConstraintHandler.jl
```diff
@@ -281,6 +270,22 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcfaces::Set{Index}, inter
     return ch
 end
 
+# Calculate which local dof index live on each face:
+# face `i` have dofs `local_face_dofs[local_face_dofs_offset[i]:local_face_dofs_offset[i+1]-1]
+function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, boundaryfunc::F=faces) where F
```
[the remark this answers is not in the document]
fredrikekre: Yeee

fredrikekre:
> It would also be nice if you can inhomogeneities [...]

Done.

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
+# # Computational homogenization
+#
+# ![](rve_homogenization.png)
+#
+# Figure 1: *Von Mise stress in an RVE with 5 stiff inclusions embedded in a softer matrix
```
KnutAM, line 5:
```suggestion
# Figure 1: *von Mises stress in an RVE with 5 stiff inclusions embedded in a softer matrix
```

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+# material loaded in shear. The problem is solved with homogeneous Dirichlet boundary
```
KnutAM, line 6:
```suggestion
# for a material that is loaded in shear. The problem is solved by using homogeneous Dirichlet boundary
```

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+# conditions (left) and (strong) periodic boundary conditions (right).*
+#
+# ## Introduction
+#
+# In this example we will solve the RVE problem for computational homogenization of
```
KnutAM, line 11:
```suggestion
# In this example we will solve the RVE (Representative Volume Element) problem for computational homogenization of
```

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+# Note that, since ``\boldsymbol{u} = \bar{\boldsymbol{\varepsilon}} \cdot [\boldsymbol{x} -
+# \bar{\boldsymbol{x}}] + \boldsymbol{u}^\mu``, this problem is equivalent to solving for
+# ``\boldsymbol{u}^\mu \in \mathbb{U}_\Box^{\mathrm{D},0}``, which is what we will do in
+# the implementation.
+#
+# **Periodic boundary conditions**
+#
+# The RVE problem pertaining to periodic boundary conditions is obtained by restricting
+# ``\boldsymbol{u}^\mu`` to be periodic, and ``\boldsymbol{t}`` anti-periodic across the
+# RVE. Similarly as for Dirichlet boundary conditions, Eq. ``(1\mathrm{b})`` is directly
```
KnutAM, line 100: "and ``\boldsymbol{t}`` anti-periodic across the RVE"
Is this really true that we enforce, doesn't this just happen to be the case if the structure is periodic?

KnutAM (commented):
As much as I had time for now, will continue tomorrow!

docs/src/literate/computational_homogenization.jl
fredrikekre: Good question. I believe it is actually just an assumption, but never enforced, which is used to get rid of the second equation.

KnutAM: I suppose it must hold true for the integral equation due to equilibrium, but I'm not sure about the point-wise anti-periodicity...

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+# compute the Voigt and Reuss bounds:
+
+vm = 0.648 # Volume fraction for the matrix material
+E_voigt = vm * Em + (1-vm) * Ei
+E_reuss = inv(vm * inv(Em) + (1-vm) * inv(Ei));
+
+# We can now compare the eigenvalues of the tensors. We expect
+# ``E_\mathrm{Reuss} \leq E_\mathrm{PeriodicBC} \leq E_\mathrm{DirichletBC} \leq E_\mathrm{Voigt}``, which is what we get:
+t = #hide
+(first ∘ eigvals).((E_reuss, E_periodic, E_dirichlet, E_voigt))
```
KnutAM, line 477: For my understanding: Shouldn't this also hold for the 2nd eigenvalue?

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+        σ̄.periodic[3][i, j]
+    end
+end);
+
+# We can check that the result are what we expect, namely that the stiffness with Dirichlet
+# boundary conditions is higher than when using periodic boundary conditions, and that
+# the Reuss assumption is an lower bound, and the Voigt assumption a upper bound. We first
+# compute the Voigt and Reuss bounds:
+
+vm = 0.648 # Volume fraction for the matrix material
```
KnutAM, line 470: Would it be nice to calculate the volume fraction automatically instead, should only be a few extra lines?

docs/src/literate/computational_homogenization.jl
```diff
@@ -0,0 +1,548 @@
...
+            ĒΩ += Ē′                                                           #src
+        end                                                                    #src
+    end                                                                        #src
+    return ĒΩ / Ω                                                              #src
+end                                                                            #src
+
+@test homogenize_test(reduce(hcat, u.dirichlet), dh, cellvalues, 10*C, C) ≈ E_dirichlet #src
+@test homogenize_test(reduce(hcat, u.periodic), dh, cellvalues, 10*C, C) ≈ E_periodic #src
+
+#md # ## [Raw source](@id homogenization-raw-code)
```
KnutAM, line 539:
```suggestion
#md # ## [Plain program](@id homogenization-raw-code)
```
to be consistent with the other examples?

KnutAM (commented):
Nice example!

docs/src/literate/computational_homogenization.jl
fredrikekre: Yea I could do that.

docs/src/literate/computational_homogenization.jl
fredrikekre: I copied this verbatim from some other example, so I think we use different, but I like Plain program better.

docs/src/literate/computational_homogenization.jl
KnutAM: It looks like we assume anti-periodicity, and by inserting that assumption in equation 9 it is effectively enforced. Without this assumption, the problem doesn't have sufficient bc because the traction would be unknown for both boundaries. At least that is how it all makes more sense to me now...

PR #420, opened 2022-02-17
Make show(grid) work for user defined celltypes

KnutAM:
REPL printing of a grid relied on the const dictionary celltypes. 
This suggestion uses the `repr` of each `typeof(cell)` instead, which makes it easy to have REPL printing of grid not throw an error for grids containing cells that are not builtin to Ferrite. 

Slight change of output for some currently implemented cells (but in my opinion better as it matches the type aliases). 

```julia
for (key, val) in Ferrite.celltypes
  println(val, " => ", repr(key))
end
3D-Quadrilateral => Quadrilateral3D
Cell{3,20,6} => Cell{3, 20, 6}
3D-Line => Line3D
Tetrahedron => Tetrahedron
Hexahedron => Hexahedron
Line => Line
QuadraticTetrahedron => QuadraticTetrahedron
2D-Line => Cell{2, 2, 2}
QuadraticTriangle => QuadraticTriangle
Triangle => Triangle
QuadraticQuadrilateral => QuadraticQuadrilateral
QuadraticLine => QuadraticLine
Quadrilateral => Quadrilateral
```
"2D-Line" is different because (wrongly) defined as Cell{2,2,2} in celltypes, but Line2D = Cell{2,21}


KnutAM:
Is Ferrite.celltypes used for anything but printing (I only see it elsewhere in one VTK test case)? 
If not, could it be removed or simplified to just a list of built-in cell types?

fredrikekre:
> Is Ferrite.celltypes used for anything but printing

I don't think so, it was introduced because old Julia versions didn't use type-aliases for printing.

KnutAM:
OK, then I've removed `celltypes` and replaced it with a tuple "`implemented_celltypes`" (to retain this information in a more compact form). 
I was thinking it's good to have such a list that could be used for (1) testing that all required functions are implemented and (2) for automatic generation of documentation for implemented celltypes (not sure how to do the latter, but I guess it's possible).

PR #425, opened 2022-02-22
Arbitrary evaluation in points

fredrikekre:


src/PointEval/PointEvalHandler.jl
```diff
@@ -273,3 +273,41 @@ function get_func_interpolations(dh::DH, fieldname) where DH<:MixedDofHandler
...
+
+# Iteration of PointEvalHandler
+struct PointIterator{PH<:PointEvalHandler}
+    ph::PH
+    current_idx::ScalarWrapper{Int}
+    nodes::Vector{Int}
+    coords::Vector{Vec{2,Float64}}
+end
+function PointIterator(ph::PointEvalHandler{G}) where {D,C,T,G<:Grid{D,C,T}}
+    @assert isconcretetype(C) # Check mate MixedDofHandler
```
kimauth, line 285: `PointIterators` only for non-mixed grids? :(

fredrikekre: image

koehlerson, line 285:
```suggestion
    C = getcelltype(grid)
    @assert isconcretetype(C) # Check mate MixedDofHandler
```

is there a benefit of restricting it to `Grid` subtypes in general or can we directly use `<:AbstractGrid` and use the interface?

KristofferC: Mixed grids are ungodly. Triangles are all we need!

src/PointEval/PointEvalHandler.jl
```diff
@@ -273,3 +273,42 @@ function get_func_interpolations(dh::DH, fieldname) where DH<:MixedDofHandler
...
+# Iteration of PointEvalHandler
+struct PointIterator{PH<:PointEvalHandler}
+    ph::PH
+    current_idx::ScalarWrapper{Int}
+    nodes::Vector{Int}
+    coords::Vector{Vec{2,Float64}}
+end
+function PointIterator(ph::PointEvalHandler{G}) where {D,C,T,G<:Grid{D,C,T}}
+    # @assert isconcretetype(C) # Check mate MixedDofHandler
+    N = nnodes_per_cell(ph.grid, 1)
```
kimauth, line 286: This will still not work for mixed grids. The principle of the `MixedDofHandler` is that you should go one `FieldHandler` at a time, so perhaps the `PointIterator` should follow that similarily to the `CellIterator` and take in a cellset. Then it would iterate only over the points which belong to that cellset.

fredrikekre: I'm `resize!`-ing: https://github.com/Ferrite-FEM/Ferrite.jl/pull/425/files#diff-431dc655f343b9166f1e184fc15699963fa4dcbba696fee98fdcb4354e840d3eR303

kimauth: Okay, but how do you handle choosing the correct `PointValues` then? Up to the user?

fredrikekre: Yea I think so. You would have something like
```julia
for point in PointIterator(ph)
    point === nothing && continue # point was not found
    cid = cellid(point)
    if cid in triangleset
        reinit!(triangle_pv, point)
        function_gradient(pv, ...)
    elseif cid in quadset
        reinit!(quad_pv, point)
        function_gradient(pv, ...)
    end
end
```

fredrikekre: Since the dofhandlers currently are not involved at all in PointEvalHandler I think this is fine.

PR #426, opened 2022-02-22
start docs for MixedDofHandler

kimauth:
A start to add some documentation for the `MixedDofHandler`. The `Common methods` section could still use some love, just added the docstrings that were already there in the code.

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -1,15 +1,30 @@
+"""
+Field(name::Symbol, interpolation::Interpolation, dim::Int)
 
+Construct `dim`-dimensional `Field` called `name` which is approximated by `interpolation`.
+
+The interpolation is used for distributing the degrees of freedom.
+"""
+struct Field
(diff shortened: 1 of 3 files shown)


src/Dofs/MixedDofHandler.jl
```diff
@@ -1,15 +1,30 @@
+"""
+Field(name::Symbol, interpolation::Interpolation, dim::Int)
```
KnutAM, line 2:
```suggestion
    Field(name::Symbol, interpolation::Interpolation, dim::Int)
```
Noticed formatting missing while reading the preview doc...

termi-official:
I guess the long term goal is now to keep both handlers as separate entities?

kimauth:
This PR isn't supposed to make a decision on that, it would still be nice to join them to one `DofHandler`.
But for now I don't think anyone is actively working on that and the `MixedDofHandler` is completely undocumented while allowing for some really nice functionality.

fredrikekre (approved):

PR #427, opened 2022-02-23
add!(::ConstraintHandler, ::Dirichlet) for nonconcrete celltypes

KnutAM:
Allow the following syntax also for a Dirichlet condition covering nonconcrete celltypes
```julia
add!(ch, dbc)
```
Before, this had to be done for each fieldhandler in `ch::ConstraintHandler` with a unique `dbc::Dirichlet` for each fieldhandler.


KnutAM:
Could you review @kimauth?

KnutAM:
The current version doesn't support the strategy from https://github.com/Ferrite-FEM/Ferrite.jl/pull/459#issuecomment-1361911806. 
Todo: Change deepcopy to just copying the set and not the entire Dirichlet struct.

Project.toml
```diff
@@ -43,6 +43,7 @@ SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
```
fredrikekre, line 46: `@test_logs` comes from `Test` so I don't think this is needed?

src/Dofs/ConstraintHandler.jl
```diff
@@ -855,8 +855,30 @@ end
...
 #Function for adding constraint when using multiple celltypes
-function add!(ch::ConstraintHandler, fh::FieldHandler, dbc::Dirichlet)
-    _check_cellset_dirichlet(ch.dh.grid, fh.cellset, dbc.faces)
+function add!(ch::ConstraintHandler{<:MixedDofHandler}, dbc::Dirichlet)
+    dbc_added = false
+    for fh in ch.dh.fieldhandlers
+        if overlaps(fh, dbc)
+            # If dbc have dofs not in fh, then these will be removed from dbc, hence deepcopy
+            # In this case, add! will warn, unless warn_check_cellset=false
+            add!(ch, fh, deepcopy(dbc), warn_check_cellset=false)
```
fredrikekre, line 864: I believe I asked @kimauth this some time -- but why is there a need to copy at all? See also https://github.com/Ferrite-FEM/Ferrite.jl/blob/70f69dd304ee58964d2d31d179d7f15babf36019/src/Dofs/ConstraintHandler.jl#L43 which is what I might have asked about.

KnutAM: The deepcopy is required due to https://github.com/Ferrite-FEM/Ferrite.jl/blob/70f69dd304ee58964d2d31d179d7f15babf36019/src/Dofs/ConstraintHandler.jl#L318

KnutAM: But perhaps it would be better to rather recreate the entire Dirichlet like in your link, but perhaps here: https://github.com/Ferrite-FEM/Ferrite.jl/blob/70f69dd304ee58964d2d31d179d7f15babf36019/src/Dofs/ConstraintHandler.jl#L313 instead, and then getting rid of that copy upon construction too?

Project.toml
KnutAM: Hm, must have missed that when testing (probably tried without `using Test`) :)
Will remove

src/Dofs/ConstraintHandler.jl
fredrikekre: Yea I don't really understand why this doesn't just skip.

Project.toml
KnutAM: Edit: It seems actually to be required for `min_level=Logging.Warn`

lijas:
I recently ran to the issue which this PR resolves (I wanted to add a constraint on a surface/faceset which had cells in different fieldhandlers). Looks good to me :)

PR #431, opened 2022-02-26
AffineConstraints: Fix usage with get_rhs_data.

### LINKED ISSUE #419: apply_rhs! doesn't work with affine constraints
I believe that https://github.com/Ferrite-FEM/Ferrite.jl/blob/9bfc3ad57a9a31728a9853f5f63625da6e43cff2/src/Dofs/ConstraintHandler.jl#L672-L678 needs to be done inside `apply_rhs!`.

fredrikekre:
This applies affine constraints also in apply_rhs! for solving a system
with multiple right hand sides. Also adds tests, and removes the
workarounds for this bug in computational homogenization example.

Fixes #419.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -111,22 +111,31 @@ See also: [`get_rhs_data`](@ref).
 """
 function apply_rhs!(data::RHSData, f::AbstractVector, ch::ConstraintHandler, applyzero::Bool=false)
     K = data.constrained_columns
-    @assert length(f) == 0 || length(f) == size(K, 1)
-    @boundscheck length(f) == 0 || checkbounds(f, ch.prescribed_dofs)
-
+    @assert length(f) == size(K, 1)
+    @boundscheck checkbounds(f, ch.prescribed_dofs)
     m = data.m
+
+    # TODO: Can the loops be combined or does the order matter?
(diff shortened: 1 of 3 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -111,22 +111,31 @@ See also: [`get_rhs_data`](@ref).
...
-        if length(f) != 0
-            vz = applyzero ? zero(eltype(f)) : v
-            f[d] = vz * m
+    end
+    @inbounds for ac in ch.acs
+        global_dof = ac.constrained_dof
+        for (d, v) in ac.entries
+            f[d] += f[global_dof] * v
         end
+        f[global_dof] = 0
```
lijas, line 132: Here you set it f[global_dof] = 0, but on line 138 you set it to vz*m... can you merge these loops?

Perhaps from line132:
```
vz = applyzero ? zero(eltype(f)) : ac.inhomogenity
f[global_dof] = vz*m
```

and remove the second loop?

fredrikekre: Not sure, this loops over the constraints, and the other loop is over all constrained dofs. Would have to do a lookup whether the dof is a affine-constrained dof or not in that case. I think it is pretty fast to iterate through these anyway. (And as it is now it matches the code of `apply!(K, f, ch)`, which does `_condense!` in between.

PR #432, opened 2022-02-27
Error when pushing fields to cells which already are in `MixedDofHandler`

### LINKED ISSUE #429: Adding mixed fields on a per field bases doesn't work
When adding multiple fields to a grid, but not everywhere, one method fails silently, as shown below

```julia
using Ferrite
# Mesh
# Node nr: n
# Element nr: (n)
    # 4---5---6
    # |   |   |
    # |(1)|(2)|
    # |   |   |
    # 1---2---3
# Field :u on both cells, field :v only on cell 1
# Should have total 10 dofs
# Cell 1 connected to 8, and cell 2 to 4

nodes = [Node(Vec((i,j))) for (i,j) in ((0.0, 0.0), (1.0, 0.0), (2.0, 0.0), (0.0, 1.0), (1.0, 1.0), (2.0, 1.0))]
cells = [Quadrilateral((1,2,5,4)), Quadrilateral((2,3,6,5))]
cellsets = Dict("u" => Set(1:2), "v" => Set(1:1), "uonly" => Set(2:2), "uandv" => Set(1:1))
grid = Grid(cells, n; cellsets=cellsets)

dim = Ferrite.getdim(grid)
ip = Lagrange{dim,RefCube,1}()

function mdh_push_per_field(grid, ip)
    dh = MixedDofHandler(grid)
    push!(dh, FieldHandler([Field(:u, ip, 1)], getcellset(grid, "u")))
    push!(dh, FieldHandler([Field(:v, ip, 1)], getcellset(grid, "v")))
    close!(dh)
    return dh
end

function mdh_push_per_cellset(grid, ip)
    dh = MixedDofHandler(grid)
    push!(dh, FieldHandler([Field(:u, ip, 1), Field(:v, ip, 1)], getcellset(grid, "uandv")))
    push!(dh, FieldHandler([Field(:u, ip, 1)], getcellset(grid, "uonly")))
    close!(dh)
end

println("mdh_push_per_field gives incorrect result")
dh_per_field = mdh_push_per_field(grid, ip)
for cell in CellIterator(dh_per_field)
    println(celldofs(cell))
end

println("mdh_push_per_field gives correct result")
dh_per_cellset = mdh_push_per_cellset(grid, ip)
#cell in CellIterator(dh_per_field) # Throws error because of different dofs per cell 
#    println(celldofs(cell))
#end
cell = CellIterator(dh_per_cellset, [1]); reinit!(cell, 1); println(celldofs(cell))
cell = CellIterator(dh_per_cellset, [2]); reinit!(cell, 1); println(celldofs(cell))
```

kimauth:
closes #429 

The current way of distributing degrees of freedom does not allow to add fields to the same cell in different `FieldHandler`s, so we should throw an error when someone tries to do that.

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -113,6 +113,7 @@ nfields(dh::MixedDofHandler) = length(getfieldnames(dh))
 function Base.push!(dh::MixedDofHandler, fh::FieldHandler)
     @assert !isclosed(dh)
     _check_same_celltype(dh.grid, collect(fh.cellset))
+    _check_cellset_intersections(dh, fh)
     # the field interpolations should have the same refshape as the cells they are applied to
     refshapes_fh = getrefshape.(getfieldinterpolations(fh))
     # extract the celltype from the first cell as the celltypes are all equal
@@ -124,6 +125,12 @@ function Base.push!(dh::MixedDofHandler, fh::FieldHandler)
 
     push!(dh.fieldhandlers, fh)
     return dh
+end
+
+function _check_cellset_intersections(dh::MixedDofHandler, fh::FieldHandler)
+    for _fh in dh.fieldhandlers
+        isempty(intersect(_fh.cellset, fh.cellset)) || error("Each cell can only belong to a single FieldHandler.")
+    end
 end
 
 function Base.push!(dh::MixedDofHandler, name::Symbol, dim::Int)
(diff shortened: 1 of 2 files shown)


src/Dofs/MixedDofHandler.jl
```diff
@@ -153,6 +155,12 @@ function Base.push!(dh::MixedDofHandler, fh::FieldHandler)
     return dh
 end
 
+function _check_cellset_intersections(dh::MixedDofHandler, fh::FieldHandler)
+    for _fh in dh.fieldhandlers
+        isempty(intersect(_fh.cellset, fh.cellset)) || error("Each cell can only belong to a single FieldHandler.")
```
KnutAM, line 160:
```suggestion
        isdisjoint(_fh.cellset, fh.cellset) || error("Each cell can only belong to a single FieldHandler.")
```
Not important, but this is supported from Julia 1.5

KnutAM (approved):

PR #435, opened 2022-03-07
ConstraintHandler: More flexible update! for runtime parameters

### LINKED ISSUE #207: Having to send a time into the function for constraints is a design failure.

fredrikekre:
This allows functions for Dirichlet boundary conditions to accept a
third argument with parameters that are not available when defining
the constraint (e.g. data that are computed/updated/changed during the
simulation). `update!` can now be called with a third argument which
is passed along to the constraint function `f`, for example
`update!(ch, t, p)` will call `f(x, t, p)`.

Fixes #207, closes #213.

### DIFF
--- a/docs/src/literate/computational_homogenization.jl
+++ b/docs/src/literate/computational_homogenization.jl
@@ -517,14 +517,14 @@ round.(ev; digits=-8)
 # also compute the macroscopic part of the displacement.
 
 chM = ConstraintHandler(dh)
-add!(chM, Dirichlet(:u, Set(1:getnnodes(grid)), (x, t) -> εᴹ[Int(t)] ⋅ x, [1, 2]))
+add!(chM, Dirichlet(:u, Set(1:getnnodes(grid)), (x, _, εᴹ) -> εᴹ ⋅ x, [1, 2]))
 close!(chM)
 uM = zeros(ndofs(dh))
 
 vtk_grid("homogenization", dh) do vtk
     for i in 1:3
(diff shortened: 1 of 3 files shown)


lijas:
What if you have bcs that need different input parameters?, then this wont work i guess?

I would prefer some sort of grouping instead

```
add!(ch, Dirichlet(...), :static_dbc)
add!(ch, Dirichlet(...), :complicated_bc_group)

update!(ch, t, :static_dbc)
update!(ch, t, para1, para2, :complicated_bc_group)
```

fredrikekre:
I was thinking you would encode that in your function, something like
```julia
dbc1 = Dirichlet(..., (x, t, p) -> x + t + p.dbc1)
dbc2 = Dirichlet(..., (x, t, p) -> x + t + p.dbc2)

dbc_params = (dbc1 = 1, dbc2 = 2)
update!(ch, t, dbc_params)
```

With your approach you have to specify things at two places: with the symbol and with the params you pass.

lijas:
This approach requires less modifications so I am ok with it. But it feels "wrong"/like a design flaw that all bcs get a bunch parameters they dont need.

kimauth (approved):
I need something similar to this. The PR looks good to me, how about merging it? 

Regarding passing unnecessary parameters: The user never sees this and can even mix conditions `f(x,t)` and `f(x,t,params)` in one `ConstraintHandler`. So it doesn't seem to be that much of a draw-back to me.

fredrikekre:
#459

PR #439, opened 2022-03-19
Polish hyperelasticity examples

termi-official:
The basic hyperelasticity example lacks some information. This PR adds **some** of the lacking information, namely a rough outline of the utilized Newton-Raphson. For now I decided against including the derivation of the discrete Newton-Raphson, as I think this is more a distraction than of help. Additionally I polished the wording and notation at some points.

### DIFF
--- a/docs/src/literate/hyperelasticity.jl
+++ b/docs/src/literate/hyperelasticity.jl
@@ -35,7 +35,7 @@
 # domain's boundary, and where ``\mathbb{U}`` and ``\mathbb{U}^0`` are suitable trial and test sets.
 # ``\Omega`` denotes the reference domain, which is also called reference or material domain.
 # Gradients are defined with respect to the reference domain, here denoted with  an ``X``. 
-# Formally this is expressed as ``\nabla_{\mathbf{X}} \bullet := \sum_{i=1}^{dim} \frac{\partial(\bullet)}{\partial X_i}``.
+# Formally this is expressed as ``(\nabla_{\mathbf{X}} \bullet)_{ij} := \frac{\partial(\bullet)_i}{\partial X_j}``.
 # Note that for large deformation problems it is also possibile that gradients and integrals
 # are defined on the deformed domain, which is also called the current or spatial domain, depending
 # on the specific formulation.
--- a/docs/src/literate/hyperelasticity.jl
+++ b/docs/src/literate/hyperelasticity.jl
@@ -34,8 +34,9 @@
 # on the reference domain, ``\mathbf{t}`` is the traction acting on the Neumann part of the reference
 # domain's boundary, and where ``\mathbb{U}`` and ``\mathbb{U}^0`` are suitable trial and test sets.
 # ``\Omega`` denotes the reference domain, which is also called reference or material domain.
-# Gradients (i.e. ``\nabla_{\mathbf{X}}``) are defined with respect to the reference domain, here denoted with
-# an ``X``. Note that for large deformation problems it is also possibile that gradients and integrals
+# Gradients are defined with respect to the reference domain, here denoted with  an ``X``. 
+# Formally this is expressed as ``\nabla_{\mathbf{X}} \bullet := \sum_{i=1}^{dim} \frac{\partial(\bullet)}{\partial X_i}``.
+# Note that for large deformation problems it is also possibile that gradients and integrals
 # are defined on the deformed domain, which is also called the current or spatial domain, depending
 # on the specific formulation.
 #
@@ -117,7 +118,7 @@ end;
 # \mathbf{u}^{k+1} = \mathbf{u}^{k} - \Delta \mathbf{u}^{k}
 # ```
 #
-# util some termination condition has been met. Therin we determine ``\Delta \mathbf{u}^{k}``
+# until some termination condition has been met. Therein we determine ``\Delta \mathbf{u}^{k}``
 #
 # ```math
 # \mathbf{K}(\mathbf{u}^{k}) \Delta \mathbf{u}^{k} = \mathbf{g}(\mathbf{u}^{k})
--- a/docs/src/literate/hyperelasticity.jl
+++ b/docs/src/literate/hyperelasticity.jl
@@ -24,8 +24,8 @@
 # as follows: Find ``u \in \mathbb{U}`` such that
 #
 # ```math
-# \int_\Omega \nabla \delta \mathbf{u} : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
-# \int_\Omega \delta \mathbf{u} \cdot \mathbf{b}\ \mathrm{d}\Omega + \int_{\Gamma^\mathrm{N}}
+# \int_{\Omega} \nabla_{\mathbf{X}} \delta \mathbf{u} : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
+# \int_{\Omega} \delta \mathbf{u} \cdot \mathbf{b}\ \mathrm{d}\Omega + \int_{\Gamma^\mathrm{N}}
 # \delta \mathbf{u} \cdot \mathbf{t}\ \mathrm{d}\Gamma
 # \quad \forall \delta \mathbf{u} \in \mathbb{U}^0,
 # ```
@@ -34,9 +34,10 @@
 # on the reference domain, ``\mathbf{t}`` is the traction acting on the Neumann part of the reference
 # domain's boundary, and where ``\mathbb{U}`` and ``\mathbb{U}^0`` are suitable trial and test sets.
 # ``\Omega`` denotes the reference domain, which is also called reference or material domain.
-# Gradients are defined with respect to the reference domain. Note that for large deformation problems
-# it is also possibile that gradients and integrals are defined on the deformed domain, which is also
-# called the current or spatial domain, depending on the specific formulation.
+# Gradients (i.e. ``\nabla_{\mathbf{X}}``) are defined with respect to the reference domain, here denoted with
+# an ``X``. Note that for large deformation problems it is also possibile that gradients and integrals
+# are defined on the deformed domain, which is also called the current or spatial domain, depending
+# on the specific formulation.
 #
 
 using Ferrite, Tensors, TimerOutputs, ProgressMeter
@@ -46,7 +47,7 @@ import KrylovMethods, IterativeSolvers
 #
 # The stress can be derived from an energy potential, defined in
 # terms of the right Cauchy-Green tensor ``\mathbf{C} = \mathbf{F}^{\mathrm{T}} \mathbf{F}``,
-# where ``\mathbf{F} = \mathbf{I} + \nabla \mathbf{u}`` is the deformation gradient.
+# where ``\mathbf{F} = \mathbf{I} + \nabla_{\mathbf{X}} \mathbf{u}`` is the deformation gradient.
 # We shall use a neo-Hookean model, where the potential can be written as
 #
 # ```math
@@ -126,12 +127,12 @@ end;
 # the Jacobi matrix, such that
 #
 # ```math
-# K_{ij} = \int_{\Omega} \nabla \delta u_{i} : \frac{\partial \mathbf{P}}{\partial \mathbf{F}} : \nabla \delta u_{j} \, \mathrm{d} \Omega
+# K_{ij} = \int_{\Omega} \nabla_{\mathbf{X}} \delta \mathbf{u}_{i} : \frac{\partial \mathbf{P}}{\partial \mathbf{F}} : \nabla_{\mathbf{X}} \delta \mathbf{u}_{j} \, \mathrm{d} \Omega
 # ```
 # and
 #
 # ```math
-# g_{i} = \int_{\Omega} \nabla \delta u_{i} : \mathbf{P} - \delta u_{i} \cdot \mathbf{b} \, \mathrm{d} \Omega
+# g_{i} = \int_{\Omega} \nabla_{\mathbf{X}} \delta \mathbf{u}_{i} : \mathbf{P} - \delta \mathbf{u}_{i} \cdot \mathbf{b} \, \mathrm{d} \Omega
 # ```
 #
 # A detailed derivation can be found in every continuum mechanics book, which has a
--- a/docs/src/literate/hyperelasticity.jl
+++ b/docs/src/literate/hyperelasticity.jl
@@ -30,13 +30,13 @@
 # \quad \forall \delta \mathbf{u} \in \mathbb{U}^0,
 # ```
 #
-# where ``\mathbf{u}`` is the unknown displacement field, ``\mathbf{b}`` is the body force, ``\mathbf{t}``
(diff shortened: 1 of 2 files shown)


docs/src/literate/quasi_incompressible_hyperelasticity.jl
```diff
@@ -11,60 +11,65 @@
...
+#
 # For clarity of presentation we ignore any non-zero surface tractions and body forces and instead consider only
 # applied displacements (i.e. non-homogeneous dirichlet boundary conditions). Moreover we stick our attention to the standard Neo-Hookean stored energy density
 #
 # ```math
-#     \Psi(u) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
+#     \Psi(\mathbf{u}) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
 # ```
-# where $I_1 = F:F = F_{ij}F_{ij}$ and J = det(F) denote the standard invariants of the deformation gradient tensor $F = \mathbb{I}+\nabla u$.
+# where $I_1 = \mathbf{F}:\mathbf{F} = F_{ij}F_{ij}$ and $J = \det(\mathbf{F})$ denote the standard invariants of the deformation gradient tensor $\mathbf{F} = \mathbf{I}+\nabla \mathbf{u}$.
```
KnutAM, line 29: Not related to this PR, but normally the double contraction of a tensor is the second invariant. In this case, I believe the reason for the naming is that I1 is the first invariant (trace) of C=F^T F? (To me, this took a few minutes to figure out why it was written like this). Perhaps good to mention somewhere?

KnutAM, line 29: Also, grad_0?

docs/src/literate/quasi_incompressible_hyperelasticity.jl
```diff
@@ -11,60 +11,65 @@
...
+# The Euler-Lagrange equations corresponding to the above energy give us our governing PDEs in the weak form, namely
 # ```math
-#   \int_\Omega \frac{\partial\Psi^\star}{\partial F}:\delta F \ \mathrm{d}\Omega = 0
+#   \int_\Omega \frac{\partial\Psi^\star}{\partial \mathbf{F}}:\delta \mathbf{F} \ \mathrm{d}\Omega = 0
 # ```
 # and
 # ```math
 #   \int_\Omega \frac{\partial \Psi^\star}{\partial p}\delta p \ \mathrm{d}\Omega = 0,
 # ```
 # where δF = δ∇u = ∇(δu) and δu and δp denote arbitrary variations with respect to displacement and pressure (or the test functions). See the references
```
KnutAM, line 62:
```suggestion
# where $\delta \mathrm{F} = \delta \mathrm{grad}_0(\mathbf{u}) = \mathrm{grad}_0(\delta \mathbf{u})$ and $\delta \mathbf{u}$ and $\delta p$ denote arbitrary variations with respect to displacement and pressure (or the test functions). See the references
```

KnutAM (commented):
Nice work!

docs/src/literate/hyperelasticity.jl
```diff
@@ -18,36 +18,42 @@
...
-# method, and for solving the Newton increment we use conjugate gradient.
+# method, and for solving the Newton increment we use conjugate gradients.
 #
-# The weak format is expressed in terms of the first Piola-Kirchoff stress ``\mathbf{P}``
+# The weak form is expressed in terms of the first Piola-Kirchoff stress ``\mathbf{P}``
 # as follows: Find ``u \in \mathbb{U}`` such that
 #
 # ```math
-#  \int_\Omega [\delta \mathbf{u} \otimes \nabla] : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
+# \int_\Omega \nabla \delta \mathbf{u} : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
```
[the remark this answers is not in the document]
koehlerson: but then it should rather be `Grad` instead of `grad`, shouldn't it?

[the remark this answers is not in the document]
termi-official: Thanks for the feedback! Maxi is right, it should be `Grad` then - which is why I decided against this notation, trying to avoid confusion.

Regarding `\delta \mathbf{u} \otimes \nabla`, this notation is super problematic, because we change the "direction" of the operator acting. Making a pathological example, what would be `\delta \mathbf{u} \otimes \nabla \otimes \delta \mathbf{v}`? From a pure technical side "\otimes" (and also other products) is not commutative anymore when we involve these kinds of operators. It just happens that some groups define `\delta \mathbf{u} \otimes \nabla := \nabla \otimes \delta \mathbf{u}` - which is fine, but super confusing. I can rewrite it as `\nabla \otimes \delta \mathbf{u}`, if you like this one better. :) 

To give a more realistic example which almost always leads to confusion: Take the Navier-Stokes equations. They have `(v \cdot \nabla) v` and they have, in the incompressible case, `\nabla \cdot v = 0`. However, this does not make the term `(v \cdot \nabla) v` zero, because the inner product is not communative on this space. 

Does this make sense for you?

docs/src/literate/hyperelasticity.jl
```diff
@@ -97,6 +103,35 @@ end;
...
+# \mathbf{u}^{k+1} = \mathbf{u}^{k} - \Delta \mathbf{u}^{k}
+# ```
+#
+# util some termination condition has been met. Therin we determine ``\Delta \mathbf{u}^{k}``
+#
+# ```math
+# \mathbf{K}(\mathbf{u}^{k}) \Delta \mathbf{u}^{k} = \mathbf{g}(\mathbf{u}^{k})
+# ```
+#
+# where $K$ is the Jacobi matrix and $g$ the global residual, such that
```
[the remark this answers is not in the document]
termi-official: Good point, thanks!

docs/src/literate/quasi_incompressible_hyperelasticity.jl
termi-official: Nice catch again! Dunno how I missed that.

docs/src/literate/hyperelasticity.jl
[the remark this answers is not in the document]
KnutAM: Sure, I think `Grad` is just another way of specifying wrt. to initial coordinates (as opposed to subscript 0 or X)? As long as we write that, I think that is fine and the most safe option.

Regarding the issues with using the open product, I agree that this is problematic. But when groups use `\nabla \otimes \delta \mathbf{u}`, this only works for symmetric tensors. To ensure both correct order wrt differentiation and which tensor leg it acts on, I guess it should be `grad(u)=[\nabla \otimes u]^T`. 
Regarding you example from NS, I would write that as `v \cdot grad(v) = v \cdot [v \otimes \nabla]` if I understand the notation correctly?
(Just so we are not talking past each other, I define `\nabla = [\partial / \partial x_i] e_i` where `e_i` are the basis vectors, or conversely \nabla_0 with X_i instead. ).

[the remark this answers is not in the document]
termi-official: >Sure, I think Grad is just another way of specifying wrt. to initial coordinates (as opposed to subscript 0 or X)?

Yes.

>[...] But when groups use \nabla \otimes \delta \mathbf{u}, this only works for symmetric tensors. To ensure both correct order wrt differentiation and which tensor leg it acts on, I guess it should be grad(u)=[\nabla \otimes u]^T.

Definitely correct. And I also think that this is where this notation comes from ("transpose applied").

> Regarding you example from NS, I would write that as v \cdot grad(v) = v \cdot [v \otimes \nabla] if I understand the notation correctly?

`(v \cdot \nabla) v` becomes `v_i \partial_i v_j e_j` in index notation while `v \cdot grad(v)` becomes `v_i e_i \cdot (\partial_j e_j \otimes v_k e_k)^T = v_i v_i \partial_j e_j`. So it should be  rather`v \cdot grad(v)^T` I guess.

docs/src/literate/hyperelasticity.jl
```diff
@@ -18,36 +18,44 @@
...
 #
-# The weak format is expressed in terms of the first Piola-Kirchoff stress ``\mathbf{P}``
+# The weak form is expressed in terms of the first Piola-Kirchoff stress ``\mathbf{P}``
 # as follows: Find ``u \in \mathbb{U}`` such that
 #
 # ```math
-#  \int_\Omega [\delta \mathbf{u} \otimes \nabla] : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
-# \int_\Omega \delta \mathbf{u} \cdot \mathbf{b}\ \mathrm{d}\Omega + \int_{\Gamma^\mathrm{N}}
-# \mathbf{u} \cdot \mathbf{t}\ \mathrm{d}\Gamma
+# \int_{\Omega} \nabla_{\mathbf{X}} \delta \mathbf{u} : \mathbf{P}(\mathbf{u})\ \mathrm{d}\Omega =
```
fredrikekre, line 27: Perhaps define `\nabla_{\mathbf{X}} \bullet` somewhere since it was discussed?

docs/src/literate/hyperelasticity.jl
```diff
@@ -18,36 +18,44 @@
...
 #
-# where ``\mathbf{u}`` is the unknown displacement field, ``\mathbf{b}`` is the body force, ``\mathbf{t}``
-# is the traction on the Neumann part of the boundary, and where ``\mathbb{U}`` and ``\mathbb{U}^0`` are
-# suitable trial and test sets.
+# where ``\mathbf{u}`` is the unknown displacement field, ``\mathbf{b}`` is the body force acting
+# on the reference domain, ``\mathbf{t}`` is the traction acting on the Neumann part of the reference
+# domain's boundary, and where ``\mathbb{U}`` and ``\mathbb{U}^0`` are suitable trial and test sets.
+# ``\Omega`` denotes the reference domain, which is also called reference or material domain.
+# Gradients (i.e. ``\nabla_{\mathbf{X}}``) are defined with respect to the reference domain, here denoted with
+# an ``X``. Note that for large deformation problems it is also possibile that gradients and integrals
```
fredrikekre, line 38: ... perhaps the definition could fit here.

docs/src/literate/hyperelasticity.jl
```diff
@@ -97,6 +105,40 @@ end;
...
+# the problem such that we can apply Newton's method, and then apply the FEM to
+# discretize the problem. Skipping a detailed derivation, Newton's method can
+# be expressed as:
+# Given some initial guess ``\mathbf{u}^0``, find a sequence ``\mathbf{u}^{k}`` by iterating
+#
+# ```math
+# \mathbf{u}^{k+1} = \mathbf{u}^{k} - \Delta \mathbf{u}^{k}
+# ```
+#
+# util some termination condition has been met. Therin we determine ``\Delta \mathbf{u}^{k}``
```
fredrikekre, line 120:
```suggestion
# until some termination condition has been met. Therein we determine ``\Delta \mathbf{u}^{k}``
```

docs/src/literate/quasi_incompressible_hyperelasticity.jl
```diff
@@ -11,60 +11,65 @@
...
+#
 # For clarity of presentation we ignore any non-zero surface tractions and body forces and instead consider only
 # applied displacements (i.e. non-homogeneous dirichlet boundary conditions). Moreover we stick our attention to the standard Neo-Hookean stored energy density
 #
 # ```math
-#     \Psi(u) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
+#     \Psi(\mathbf{u}) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
 # ```
-# where $I_1 = F:F = F_{ij}F_{ij}$ and J = det(F) denote the standard invariants of the deformation gradient tensor $F = \mathbb{I}+\nabla u$.
+# where $I_1 = \tf{\mathbf{C}} = F_{ij}F_{ij} = \tr(F)$ and $J = \det(\mathbf{F})$ denote the standard invariants of the deformation gradient tensor $\mathbf{F} = \mathbf{I}+\nabla \mathbf{u}$.
```
fredrikekre, line 29: This doesn't render (https://ferrite-fem.github.io/Ferrite.jl/previews/PR439/examples/quasi_incompressible_hyperelasticity/). I guess `\tf` and/or `\tf` isn't available.

KnutAM: Shouldn't it be `\tr` on the first? The last step I don't get, perhaps \tr(F^T F)?

termi-official: Right, it is the trace and especially trace of F^T F, as Knut notes. "tf" is a typo.

docs/src/literate/hyperelasticity.jl
termi-official: Sure.

docs/src/literate/quasi_incompressible_hyperelasticity.jl
```diff
@@ -26,7 +26,7 @@
 # ```math
 #     \Psi(\mathbf{u}) = \frac{\mu}{2}\left(I_1 - 3 \right) - \mu \log(J) + \frac{\lambda}{2}\left( J - 1\right){}^2,
 # ```
-# where $I_1 = \tf{\mathbf{C}} = F_{ij}F_{ij} = \tr(F)$ and $J = \det(\mathbf{F})$ denote the standard invariants of the deformation gradient tensor $\mathbf{F} = \mathbf{I}+\nabla \mathbf{u}$.
+# where $I_1 = \mathrm{tr}(\mathbf{C}) = \mathrm{tr}(F^T F) = F_{ij}F_{ij}$ and $J = \det(\mathbf{F})$ denote the standard invariants of the deformation gradient tensor $\mathbf{F} = \mathbf{I}+\nabla \mathbf{u}$.
```
KnutAM, line 29:
```suggestion
# where $I_1 = \mathrm{tr}(\mathbf{C}) = \mathrm{tr}(\mathbf{F}^\mathrm{T} \mathbf{F}) = F_{ij}F_{ij}$ and $J = \det(\mathbf{F})$ denote the standard invariants of the deformation gradient tensor $\mathbf{F} = \mathbf{I}+\nabla_{\mathbf{X}} \mathbf{u}$.
```
Just to be picky 😉, but otherwise looks good to me now! Nice doc improvement!

KnutAM (approved):

docs/src/literate/quasi_incompressible_hyperelasticity.jl
termi-official: Not picky, you are correct. :)

termi-official:
Thanks for taking so much time to review!

PR #441, opened 2022-04-07
allow node renumbering during vtk export

kimauth:
`nodes_to_vtknodes` can be defined for custom cell types that don't follow the vtk node numbering in order to allow correct export.


lijas:
Nice, but maybe the function name should include "ordering" or similar?

kimauth:
`nodes_to_vtkorder` perhaps?

termi-official:
Is this function really everything that must be customized for the exporter to function properly?

kimauth:
Yes, I think so. Do you see something else that would need changing?

termi-official:
Just by quickly going over the code I do not see something else. I was just suprised that the solution might be so easy. :)

lijas (approved):

PR #442, opened 2022-04-08
Rename internal curl to curl_from_gradient

KnutAM:
`Ferrite.curl` and `Tensors.curl` cause a naming conflict, even if `Ferrite.curl` is not exported directly, because `Tensors` exports `curl`. 
From discussions on slack, this PR renames the `Ferrite.curl` function to `curl_from_gradient`, since it doesn't do any differentiation, but takes in the gradient and outputs the curl of the vector for which the gradient is given.


KnutAM:
Can you review @lijas ?

lijas (approved):

PR #443, opened 2022-04-08
Remove the copy!! function and use copy! from Base

KnutAM:
`Base.copy!` is supported from julia 1.1 and is equivalent to `Ferrite.copy!!`. The use of `copy!!` in the constraint handler confused me at first when trying to understand Ferrite, so I propose to use the base implementation instead now that support for 1.0 is dropped. 
Edit: typo in the commit message, but I guess this disappears when squashed upon merge...


KnutAM:
Can you review @fredrikekre?

Issue #444, opened 2022-04-09
`Ferrite.nfields` should be called differently


kimauth:
```julia
grid = generate_grid(Quadrilateral, (1,1))
dh = DofHandler(grid)
push!(dh, :u, 2)
close!(dh)
```
```julia
julia> nfields(dh)
9

julia> Ferrite.nfields(dh)
1
```

The first one calls Julias [`nfields`](https://docs.julialang.org/en/v1/base/base/#Core.nfields).

termi-official:
Addressed in #486 because it has been an issue several times.

PR #445, opened 2022-04-10
RFC: Introducing a `MixedGrid`

kimauth:
Following up on our discussion from last week, here comes a suggestion on how to handle mixed grids in a more data-oriented manner.

The main gripes about the current implementation are:
- Grids with several celltypes store their cells as `Vector{Union{CellType1, CelType2,...}}` or even worse as `Vector{AbstractCell}`. This makes extracting a cell based on its global id inherently type unstable.
- Due to the above reason, the `MixedDofHandler` stores its own grid representation so that it can efficiently access cell nodes and cell coordinates.
- `DofHandler` & `MixedDofHandler`. Mixed Celltypes in a grid are now reflected by need to use `MixedDofHandler`. It would be nice to reflect this in the `Grid` datatype instead and use a single `DofHandler` no matter what the grid type. _(Currently the `DofHandler` additionally does not support fields on subsets of the domain, while the `MixedDofHandler` does.)_   
- When assembling with a `MixedDofHandler`, it is crucial for performance to not iterate over all cells at once. Instead a function barrier should be set for each `FieldHandler`. This is not mentioned in our docs.

This PR attempts to solve the first two points and lay ground for solving the 3rd in the future.

# MixedGrid type
It introduces a `MixedGrid` that stores its cells as `Tuple{Vector{CellType1}, Vector{CellType2}, ...}`.
The cells in the grid are indexed by 
```julia
struct CellId{I}
    i::Int
end
```
where `I` is the index within the tuple and `i` is the index within the `Vector`. This can be seen as local cell id. Cells also have global cell ids (`Int`s), which count through the cells tuple from beginning to end. Global ids can thus easily computed from local ids. 

This means that cells will likely be renumbered when importing a mixed grid from an external grid generator. I don't see this as a huge problem, but an alternative approach could be to store a mapping local id -> global id.

# MixedDofHandler + MixedGrid
A grid with mixed celltypes should always be represented as `MixedGrid` and always be indexed by `CellId`. That is, the `FieldHandler`s then store a (concretely typed!) `Set{CellId{I}}` and all cellsets + facesets in the mixed grid should reference cells by `CellId` instead of their global id (the latter isn't implemented yet for this PR).

The `MixedDofHandler` does not store cell nodes and cell coordinates anymore. Methods for extracting them fall back to the implementations for the underlying grid.

The `MixedDofHandler` allows to use either `Grid` or `MixedGrid` with it. Therefore, the `celldofs` are still stored by global cell ids.

# DofHandler methods + Benchmarking
Here is a list of all functions that use the DofHandlers (to get an overview). Some methods are defined for `AbstractDofHandler`, but will only work for the `DofHandler`, thus they're listed on the left.

| DofHandler | MixedDofHandler |
| --- | --- |
| `getfielddim(dh::DofHandler, field_idx::Int)` | no exact equivalent |
| ` getfielddim(dh::DofHandler, name::Symbol)` |`getfielddim(dh::MixedDofHandler, name::Symbol)`|
| `ndim(dh::AbstractDofHandler, field_name::Symbol)`, same functionality as getfielddim | doesn't exist |
| `nfields(dh::AbstractDofHandler)`, see #444| `nfields(dh::MixedDofHandler)` |
| `ndofs_per_cell(dh::AbstractDofHandler, cell::Int=1)`|`ndofs_per_cell(dh::MixedDofHandler, cell::Int=1)` | 
| `getfieldnames(dh::AbstractDofHandler)` | `getfieldnames(dh::MixedDofHandler)`, `getfieldnames(fh::FieldHandler)` |
| | |
| `find_field(dh::DofHandler, field_name::Symbol)` | `find_field(fh::FieldHandler, field_name::Symbol)` |
| `field_offset(dh::DofHandler, field_name::Symbol)`|`field_offset(fh::FieldHandler, field_name::Symbol)` | 
| `dof_range(dh::DofHandler, field_name::Symbol)` | `dof_range(fh::FieldHandler, field_name::Symbol)` |
| `getfieldinterpolation(dh::DofHandler, field_idx::Int)` | no exact equivalent | 
| no excact equivalent | `getfieldinterpolations(fh::FieldHandler)`|
| | |
| ` cellcoords!(global_coords::Vector{<:Vec}, dh::DofHandler, i::Int)` | `cellcoords!(global_coords::Vector{Vec{dim,T}}, dh::MixedDofHandler, i::Int) where {dim,T}` |
| `celldofs!(global_dofs::Vector{Int}, dh::DofHandler, i::Int)` | `celldofs!(global_dofs::Vector{Int}, dh::MixedDofHandler, i::Int)` |
| `celldofs(dh::DofHandler, i::Int)` | `celldofs(dh::MixedDofHandler, i::Int)` |
| use grid version | `cellnodes!(global_nodes::Vector{Int}, dh::MixedDofHandler, i::Int)`|

Most of these are used for dof distribution, I'd expect that only `celldofs!` and `cellcoords!` are used in performance relevant code, so let's focus on these for some micro benchmarks. The benchmarks extract cellcoords and celldofs for a `Quadrilateral` in a mixed grid with `Triangle`s and `Quadrilateral`s.  

| | `DofHandler` | `MixedDofHandler#master` <br> + mixed grid |  `MixedDofHandler#thisPR` <br> + mixed grid | `MixedDofHandler#thisPR` <br> + concrete grid |
|---|---|---|---|---|
| `cellcoords!` | 4.100 ns (0 allocations: 0 bytes)| 9.300 ns (0 allocations: 0 bytes) | 4.300 ns (0 allocations: 0 bytes) | 4.100 ns (0 allocations: 0 bytes) |
| `celldofs!` | 9.500 ns (0 allocations: 0 bytes) | 9.810 ns (0 allocations: 0 bytes) | 9.300 ns (0 allocations: 0 bytes) | 9.300 ns (0 allocations: 0 bytes) |

For this case, the `MixedDofHandler` now performs equally well as the `DofHandler`, no matter if it is used with `Grid` or with `MixedGrid`. 

Technically, it was possible to use the grid version of `cellcoords!` before this PR (and thus get rid of the grid representation in the `MixedDofHandler`), however that can be (very) slow:
|cell type| `Quadrilateral` <br> (concretly typed) | `Union{Triangle, Quadrilateral}` | `AbstractCell` |
|---|---|---|---|
| `cellcoords!`<br>`(xe, grid, i)` | 3.700 ns (0 allocations: 0 bytes) | 6.700 ns (0 allocations: 0 bytes) | 1.710 μs (22 allocations: 720 bytes)|

# Usage
The usage pattern remains exactly the same as before. 
```julia
# e.g. for the quads and tris example above:
buffers_quads = (xe_quads, dofs_quads, cv_quads, ke_quads, re_quads)
buffers_tris = (xe_tris, dofs_tris, cv_tris, ke_tris, re_tris)
buffers = [buffers_quads, buffers_tris]

for (fh_idx, fh) in enumerate(dh.fieldhandlers)
    xe, dofs, cv, ke, re = buffers[fh_idx]
    loop_fh(dh, fh, ke, re, xe, dofs, cv)
end

function loop_fh(dh, fh, ke, re, xe, dofs, cv)
    for cellid in fh.cellset
        cellcoords!(xe, dh, cellid)
        celldofs!(dofs, dh, cellid)
        # ... do stuff like element routine
        # @views ue = u[dofs] 
        # assemble_cell!(ke, re, xe, cv, ue)
    end
end
```

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -312,8 +312,9 @@ function cellnodes!(global_nodes::Vector{Int}, grid::Grid{dim,C}, i::Int) where 
     return global_nodes
 end
 
-function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::Grid{dim,C}, i::Int) where {dim,C,T}
-    nodes = grid.cells[i].nodes
+# shared implementation for all grids
+function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid, cell::C) where {dim,C<:Ferrite.AbstractCell,T}
+    nodes = cell.nodes
     N = length(nodes)
     @assert length(global_coords) == N
     for j in 1:N
@@ -321,6 +322,12 @@ function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::Grid{dim,C}, i::In
     end
     return global_coords
 end
+
+function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid, i) where {dim,T}
+    cell = getcells(grid, i)
+    cellcoords!(global_coords, grid, cell)
+end
+
 cellcoords!(global_coords::Vector{<:Vec}, dh::DofHandler, i::Int) = cellcoords!(global_coords, dh.grid, i)
 
 function celldofs(dh::DofHandler, i::Int)
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -23,9 +23,9 @@ A `FieldHandler` must fullfill the following requirements:
 
 Notice that a `FieldHandler` can hold several fields.
 """
-mutable struct FieldHandler
+mutable struct FieldHandler{CI}
     fields::Vector{Field}
-    cellset::Set{Int}
+    cellset::Set{CI}
 end
 
 struct CellVector{T}
@@ -46,18 +46,16 @@ Construct a `MixedDofHandler` based on `grid`. Supports:
 - `Grid`s with or without concrete element type (E.g. "mixed" grids with several different element types.)
 - One or several fields, which can live on the whole domain or on subsets of the `Grid`.
 """
-struct MixedDofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
+struct MixedDofHandler{dim,G<:AbstractGrid{dim}} <: AbstractDofHandler
     fieldhandlers::Vector{FieldHandler}
     cell_dofs::CellVector{Int}
-    cell_nodes::CellVector{Int}
-    cell_coords::CellVector{Vec{dim,T}}
     closed::ScalarWrapper{Bool}
     grid::G
     ndofs::ScalarWrapper{Int}
 end
 
-function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
-    MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], CellVector(Int[],Int[],Int[]), CellVector(Int[],Int[],Int[]), CellVector(Vec{dim,T}[],Int[],Int[]), Ferrite.ScalarWrapper(false), grid, Ferrite.ScalarWrapper(-1))
+function MixedDofHandler(grid::G) where {dim, G<:AbstractGrid{dim}}
+    return MixedDofHandler{dim,G}(FieldHandler[], CellVector(Int[],Int[],Int[]), ScalarWrapper(false), grid, ScalarWrapper(-1))
 end
 
 getfieldnames(fh::FieldHandler) = [field.name for field in fh.fields]
@@ -73,26 +71,14 @@ function celldofs!(global_dofs::Vector{Int}, dh::MixedDofHandler, i::Int)
     unsafe_copyto!(global_dofs, 1, dh.cell_dofs.values, dh.cell_dofs.offset[i], length(global_dofs))
     return global_dofs
 end
+celldofs!(global_dofs::Vector{Int}, dh::MixedDofHandler, i::CellId) = celldofs!(global_dofs, dh, globalid(dh.grid, i))
 
 function celldofs(dh::MixedDofHandler, i::Int)
     @assert isclosed(dh)
     return dh.cell_dofs[i]
 end
 
-function cellcoords!(global_coords::Vector{Vec{dim,T}}, dh::MixedDofHandler, i::Int) where {dim,T}
-    @assert isclosed(dh)
-    @assert length(global_coords) == nnodes_per_cell(dh, i)
-    unsafe_copyto!(global_coords, 1, dh.cell_coords.values, dh.cell_coords.offset[i], length(global_coords))
-    return global_coords
-end
-
-function cellnodes!(global_nodes::Vector{Int}, dh::MixedDofHandler, i::Int)
-    @assert isclosed(dh)
-    @assert length(global_nodes) == nnodes_per_cell(dh, i)
-    unsafe_copyto!(global_nodes, 1, dh.cell_nodes.values, dh.cell_nodes.offset[i], length(global_nodes))
-    return global_nodes
-end
-
+cellcoords!(global_coords::Vector{<:Vec}, dh::MixedDofHandler, i) = cellcoords!(global_coords, dh.grid, i)
 
 """
     getfieldnames(dh::MixedDofHandler)
@@ -140,12 +126,12 @@ Add all fields of the [`FieldHandler`](@ref) `fh` to `dh`.
 """
 function Base.push!(dh::MixedDofHandler, fh::FieldHandler)
     @assert !isclosed(dh)
-    _check_same_celltype(dh.grid, collect(fh.cellset))
+    # _check_same_celltype(dh.grid, collect(fh.cellset))
     _check_cellset_intersections(dh, fh)
     # the field interpolations should have the same refshape as the cells they are applied to
     refshapes_fh = getrefshape.(getfieldinterpolations(fh))
     # extract the celltype from the first cell as the celltypes are all equal
-    cell_type = typeof(dh.grid.cells[first(fh.cellset)])
+    cell_type = typeof(getcells(dh.grid, first(fh.cellset)))
     refshape_cellset = getrefshape(default_interpolation(cell_type))
     for refshape in refshapes_fh
         refshape_cellset == refshape || error("The RefShapes of the fieldhandlers interpolations must correspond to the RefShape of the cells it is applied to.")
@@ -227,7 +213,8 @@ function __close!(dh::MixedDofHandler{dim}) where {dim}
     @debug "\n\nCreating dofs\n"
     for fh in dh.fieldhandlers
         # sort the cellset since we want to loop through the cells in a fixed order
-        cellnumbers = sort(collect(fh.cellset))
+        # now only works for MixedGrid
+        cellnumbers = get_sorted_cellset(fh)
         nextdof = _close!(
             dh,
             cellnumbers,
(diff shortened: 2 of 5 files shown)


KnutAM:
Is there a performance reason for keeping the `FieldHandler` lightweight? If not, would it be a convenient way to define
```julia
mutable struct FieldHandler{CI,T}
    fields::Vector{Field}
    cellset::Set{CI}
    data::T
end
```
allowing a `FieldHandler`, if desired (default to `data=nothing`), to store e.g. cellvalues, buffers, material struct, etc. This implies that the user doesn't need to take care of storing buffers in the same order as `FieldHandler`s are pushed to the `MixedDofHandler`. 
Perhaps bad for parallelization? 
Maybe also not part of this PR, but it does relate to handling mixed grids, so just throwing this in here...

src/Dofs/DofHandler.jl
```diff
@@ -312,15 +312,22 @@ function cellnodes!(global_nodes::Vector{Int}, grid::Grid{dim,C}, i::Int) where
...
 
-function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::Grid{dim,C}, i::Int) where {dim,C,T}
-    nodes = grid.cells[i].nodes
+# shared implementation for all grids
+function cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid, cell::C) where {dim,C<:Ferrite.AbstractCell,T}
+    nodes = cell.nodes
     N = length(nodes)
     @assert length(global_coords) == N
     for j in 1:N
         global_coords[j] = grid.nodes[nodes[j]].x
```
koehlerson, line 321: There is an interface function which should be used, because the functionality of `cellcoords!` can be recovered by the already defined interface functions. I think it would be worth considering to put this function in another file, e.g. in the grid.jl file, doesn't seem to be exclusively related the dofhandler
```suggestion
        global_coords[j] = getnodes(grid,nodes[j]).x
```

src/Grid/mixed_grid.jl
```diff
@@ -0,0 +1,51 @@
...
+    ncells_per_vector::Vector{Int}
+    nodes::Vector{Node{dim,T}}
+    # Sets
+    cellsets::Dict{String,Set{Int}}
+    nodesets::Dict{String,Set{Int}}
+    facesets::Dict{String,Set{FaceIndex}} 
+    edgesets::Dict{String,Set{EdgeIndex}} 
+    vertexsets::Dict{String,Set{VertexIndex}} 
+    # Boundary matrix (faces per cell × cell)
+    boundary_matrix::SparseMatrixCSC{Bool,Int}
```
koehlerson, line 12: @termi-official is there a better way to encode the boundary information into the grid? Would be a good opportunity to include it in this struct

koehlerson (commented):
I like this PR; included a small suggestion. Would be a nice working example of using the grid interface. We should take care that functionality doesn't diverge too much. `DofHandler` cannot use any other grid than `Grid`. https://github.com/Ferrite-FEM/Ferrite.jl/pull/379 tried to address it, but seems to be stalled. So with including something like this, it would be quite unnatural for new or not so experienced users to have a `DofHandler` who can't take the `MixedGrid` (which isn't probably that useful). Therefore, a long term strategy would be nice how this problem will be resolved.

src/Grid/mixed_grid.jl
termi-official: Yes and no. It really depends on your use case. You can get away without using the `boundary_matrix` and I honestly think that it should not be in any grid struct, but in a separate (boundary information) data structure.

termi-official: I mean we have the facesets data structure anyway to distinguish different portions of the boundary - and a set lookup should be O(1). If we want to check if we are on the interior, then the upcoming topology PR should handle this better.

termi-official: Maybe @fredrikekre does know more.

src/Dofs/DofHandler.jl
kimauth: Using the interface function is definitely the right way to go. 

I also agree that this should live in the grid file. I simply didn't move it yet so that it's easier to see the actual diff caused by this PR (and not the one by shuffling around functions).

src/Grid/mixed_grid.jl
termi-official: Note that removing the field is definitely a breaking change, so I do not think we can remove it immediately. (See https://github.com/Ferrite-FEM/Ferrite.jl/blob/ea716bc7a73da0b7970f81187b9b8b26ea502598/src/iterators.jl#L90 which is exported.) This is also why I asked Frederik for a comment.

kimauth: I guess the reason this exists is that it is faster than the set-lookup:
```julia
julia> grid = generate_grid(Triangle, (20, 20));

julia> ci = CellIterator(grid); reinit!(ci, 1);

julia> @btime onboundary($ci, $1);
  4.300 ns (0 allocations: 0 bytes)

julia> faceindex = FaceIndex(1,1)
FaceIndex((1, 1))

julia> @btime in($faceindex, $(grid.facesets["bottom"]))
  16.032 ns (0 allocations: 0 bytes)
```
So from that perspective, I'd suggest leaving it as it is for now and eventually tackle if it should be in the grid or not in another PR?

src/Grid/mixed_grid.jl
```diff
@@ -0,0 +1,51 @@
...
+end
+
+function MixedGrid(cells::C,
+        nodes::Vector{Node{dim,T}};
+        cellsets::Dict{String,Set{Int}}=Dict{String,Set{Int}}(),
```
lijas, line 17: I guess `cellsets` should not be a vector of Ints in the MixedGrid?

src/Grid/mixed_grid.jl
```diff
@@ -0,0 +1,51 @@
...
+        facesets::Dict{String,Set{FaceIndex}}=Dict{String,Set{FaceIndex}}(),
+        edgesets::Dict{String,Set{EdgeIndex}}=Dict{String,Set{EdgeIndex}}(),
+        vertexsets::Dict{String,Set{VertexIndex}}=Dict{String,Set{VertexIndex}}(),
+        boundary_matrix::SparseMatrixCSC{Bool,Int}=spzeros(Bool, 0, 0)) where {dim,C,T}
+        ncells_per_type = collect(length.(cells))
+    return MixedGrid(cells, ncells_per_type, nodes, cellsets, nodesets, facesets, edgesets, vertexsets, boundary_matrix)
+end
+
+struct CellId{I}
+    i::Int
```
lijas, line 28: Having the cell id as a type parameter is clever, but can give some trouble if one has a cellset with different celltypes: https://docs.julialang.org/en/v1/manual/performance-tips/#The-dangers-of-abusing-multiple-dispatch-(aka,-more-on-types-with-values-as-parameters)

I dont think that is necessarily that common though...

lijas (commented):
Seems like a good PR to get a type-stable grid with mixed celltypes, and getting rid of the grid-representation in the MixedDofhandler. The MixedGrid is quite minimalistic, and fits into the grid api without changing other parts of the code too much. 

Just one thought: 

It seems like just type-annotating the normal `Grid` (when it includes mixed celltypes) gives similar benchmarks when calling `cellcords!(...)`, as compared to with your PR, e.g.:
```

function f1(coords, grid::Grid{2,Ferrite.AbstractCell,Float64}, cellset::Set{Int})
    for i in cellset
        cell::Quadrilateral = grid.cells[i]; #type annotate
        Ferrite.cellcoords!(coords, grid, cell)
        #...
    end
end
```


I think it is typically the case that one can type-annotate their cell-loops (that is at least the case in my code), so It is worth considering if we want want to add a completely new Grid-type, if we dont gain that much?
Personally I would probably use the type-annotation approach above, but this PR supports that as well, so that is nice.

src/Dofs/MixedDofHandler.jl
```diff
@@ -244,24 +231,14 @@ function __close!(dh::MixedDofHandler{dim}) where {dim}
...
-        push!(dh.cell_nodes.length, length(cell.nodes))
-        push!(dh.cell_coords.length, length(cell.nodes))
-    end
-
     return dh, vertexdicts, edgedicts, facedicts
 
 end
 
+get_sorted_cellset(fh::FieldHandler{Int}) = sort!(collect(fh.cellset))
+get_sorted_cellset(fh::FieldHandler{CellId{I}}) where I = sort!(collect(fh.cellset), by=c->c.i)
```
lijas, line 239: Define `>` and `<` for a `CellId` instead?

src/iterators.jl
```diff
@@ -41,7 +41,7 @@ struct CellIterator{dim,C,T,DH<:Union{AbstractDofHandler,Nothing}}
     dh::Union{DH,Nothing}
     celldofs::Vector{Int}
```
KnutAM, line 43: Seems like the CellIterator will still have an Abstract celltype? 
This PR doesn't seem necessary for fixing that but would enable lifting the type instability higher up (i.e. making the creation of CellIterator type-stable also) !

src/Grid/mixed_grid.jl
fredrikekre: It might make more sense to have it be a lookup for whether the cell has any face on a boundary. Then you can check once per cell, instead of once epr face. (This also works better with multi-celltype grids).

src/Grid/mixed_grid.jl
kimauth: No, working on that :)

src/iterators.jl
kimauth: Haven't done anything on the `CellIterator` yet, I'd hope fixing the `CellIterator` for mixed grids would come naturally once we agree on a grid format.

src/Grid/mixed_grid.jl
```diff
@@ -0,0 +1,51 @@
+mutable struct MixedGrid{dim, C, T<:Real} <: Ferrite.AbstractGrid{dim}
+    cells::C # Tuple of concretely typed cell vectors
+    ncells_per_vector::Vector{Int}
```
KnutAM, line 3: Is `ncells_per_vector` needed? 
Did some quick benchmarks for `getncells` and couldn't see that it would be faster (actually slower but so fast that it wouldn't matter anyway)

termi-official:
I see that this PR has quite some impact on the microbenchmarks, but does it really have a significant impact on overall assembly performance for non-trivial assembly loops? I mean first, we cut down the cost of calls which are usually not the most costly ones by several nanoseconds. Second, this PR feels a bit contrary to the goal of merging the DofHandler and MixedDofhandler, since we diversify the code base further. From a conceptual point, f we want to merge this code, wouldn't it be more beneficial if the MixedGrid replaces the current Grid? This also cuts maintainance cost.

fredrikekre:
Superseded by the DofHandler merge.

PR #448, opened 2022-04-14
SubDofHandler/MixedGrid/PorousMedia example

### LINKED ISSUE #355: No examples with `MixedDofHandler`
We don't seem to have any examples that use the `MixedDofHandler`. Any ideas of simple applications?
Main points that the `MixedDofHandler` can do that the `DofHandler` can't would be

- a grid with different celltypes in it 
- a problem with several fields where one of the fields lives only on a part of the domain.

KnutAM:
Closes #355 

Example for using both mixed grid and fields on subsets of the domain: Porous media with embedded impermeable particles. [Link to example in docs preview](https://ferrite-fem.github.io/Ferrite.jl/previews/PR448/tutorials/porous_media/)


KnutAM:
This is my naive attempt for an example that I would have liked to have when starting using the `MixedDofHandler`
Probably a few blunders, so I would be happy to get your opinions @kimauth, @lijas, and @fredrikekre, on if this is reasonable and what should be changed in the implementation. Then I can spend more time on beautifying and comment/theory clarifications! 
(The loading case should change to highlight more that it is a coupled problem, but that shouldn't affect the code)

KnutAM:
Note for later, if to be included in Ferrite.jl: 
A simpler and shorter theory section, different boundary conditions, and using AD for the stiffness, can be found [here](https://knutam.github.io/FerriteProblems.jl/dev/examples/porous_media/). These changes can be made here as well.

docs/download_resources.jl
```diff
@@ -7,6 +7,7 @@ mkpath(directory)
 for (file, url) in [
         "periodic-rve.msh" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/periodic-rve.msh",
         "periodic-rve-coarse.msh" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/periodic-rve-coarse.msh",
+        "porous_media_0p25.inp" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/kam/MixedDofHandlerExample/docs/src/literate/porous_media_0p25.inp",
```
termi-official, line 10: Wrong branch? Also, you could add the other abaqus mesh to the branch, too.

docs/src/literate/porous_media.jl
```diff
@@ -0,0 +1,340 @@
...
+    ip4_quad = Lagrange{dim, RefCube, 2}()
+    qr3 = QuadratureRule{dim, RefTetrahedron}(1)
+    qr4 = QuadratureRule{dim, RefCube}(2)
+
+    ## Setup the MixedDofHandler
+    dh = MixedDofHandler(grid)
+    push!(dh, FieldHandler([Field(:u, ip3_lin, dim)], getcellset(grid,"solid3")))
+    push!(dh, FieldHandler([Field(:u, ip4_lin, dim)], getcellset(grid,"solid4")))
+    push!(dh, FieldHandler([Field(:u, ip3_quad, dim), Field(:p, ip3_lin, 1)], getcellset(grid,"porous3")))
+    push!(dh, FieldHandler([Field(:u, ip4_quad, dim), Field(:p, ip4_lin, 1)], getcellset(grid,"porous4")))
```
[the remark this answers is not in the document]
KnutAM: Yes, until https://github.com/Ferrite-FEM/Ferrite.jl/pull/644#issuecomment-1493932715 is done I think so (except it should be `add!`)
I'll wait to update this example until those changes are merged.

docs/Manifest.toml
```diff
@@ -349,6 +349,12 @@ version = "0.1.3"
 [[deps.FileWatching]]
 uuid = "7b1f6079-737a-58dc-b8bc-7a2ca5c1b5ee"
 
+[[deps.FerriteMeshParser]]
```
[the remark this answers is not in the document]
KnutAM: Sorry, but I don't follow what the advantage of having different packages would be? (And what that would have to do with Julia 1.9 becoming LTS)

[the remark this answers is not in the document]
termi-official: Usually when you have one mesh format which you want to load, then you do not need all the parsers for all the mesh formats under the sun. :P 1.9 Brings extensions, so you can be more selective about this part by e.g. explicitly installing the dependency which contains the actual parser stuff (which is sometimes an external C library).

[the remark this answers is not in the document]
KnutAM: Good point! 
For that, I would say that in `FerriteMeshParser.jl`, all parser functionality in pure julia that doesn't have heavy-weight dependencies can be included directly. And other parsers should be added as extensions.
To support julia < 1.9, we could use Requires.jl, or simply just not support julia <1.9 for newer versions after some time.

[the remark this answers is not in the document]
KnutAM: And if you have some external parsers in mind, feel free to open an issue/pr of what could/should be added:)

src/Dofs/DofHandler.jl
```diff
@@ -823,18 +823,49 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::Vector, fieldname::Symbol)
...
 # Internal method that have the vtk option to allocate the output differently
 function _evaluate_at_grid_nodes(dh::DofHandler, u::Vector{T}, fieldname::Symbol, ::Val{vtk}=Val(false)) where {T, vtk}
     # Make sure the field exists
     fieldname ∈ getfieldnames(dh) || error("Field $fieldname not found.")
     # Figure out the return type (scalar or vector)
     field_idx = find_field(dh, fieldname)
-    ip = getfieldinterpolation(dh, field_idx)
-    RT = ip isa ScalarInterpolation ? T : Vec{n_components(ip),T}
+    n_comp = n_components(dh, fieldname)
+    RT = physical_field_type(dh, fieldname, T)
```
KnutAM, line 865: @fredrikekre: This solves the bug I hit earlier. Will branch out to separate PR once the example is finished, the functions above should probably be refined and checks moved to close if not already checked...

src/Dofs/DofHandler.jl
```diff
@@ -849,8 +849,8 @@ function _evaluate_at_grid_nodes(dh::DofHandler, u::Vector{T}, fieldname::Symbol
         CT = getcelltype(get_grid(dh), first(sdh.cellset))
         ip_geo = default_interpolation(CT)
         local_node_coords = reference_coordinates(ip_geo)
-        qr = QuadratureRule{getrefshape(ip)}(zeros(length(local_node_coords)), local_node_coords)
         ip = getfieldinterpolation(sdh, field_idx)
+        qr = QuadratureRule{getrefshape(ip)}(zeros(length(local_node_coords)), local_node_coords)
```
KnutAM, line 853: Note: This bug was detected while running the example, and I decided to fix it here instead of doing a separate PR since it is such a small change.

termi-official: Could you also add a regression test on this?

KnutAM: Done

termi-official:
Is there something blocking this PR from merge?

KnutAM:
> Is there something blocking this PR from merge?

Missing review from my perspective...

src/Dofs/DofHandler.jl
KnutAM: Note to when this is merged; this bug was now fixed and tested by #800. 

Make sure to not duplicate fix and tests. (I can merge with master when someone tells me they are ready to review this one)

KnutAM:
Looks like the increased CI doc time is due to updated packages after updating the docs project (need to upgrade FerriteMeshParser to 0.1.7 to support the master branch).

Issue #195, opened 2018-04-16
Support arbitrary precision


mohamed82008:
Only Float64 precision is fully supported. I will make a PR for this.

PR #451, opened 2022-05-14
RFC: Attempt at complex dofs with Helmholtz eqn

jondea:
Hi, first I want to say thank you for creating this package!

I was looking to use it to solve the complex valued Helmholtz equation, however I can't find any references to using complex numbers with the library. One solution is to register the real and imaginary parts separately as 2 real unknowns, however, this is not as elegant and produces a matrix twice as large as it needs to be.

So I hacked around with the library and got it to accept complex numbers. I modified the Helmholtz example to demonstrate the solution, although I suspect it would be better to create another to demonstrate complex values. My solution doesn't feel particularly correct so I wanted to submit it as an RFC to hopefully get some pointers. In particular, I feel like the `DofHandler` should know the nodal type, and should pass this through to the assembly process and `ConstraintHandler`.

NB: Support for mixed precision has been discussed (#195), but I couldn't find any code related to it.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -67,14 +67,14 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     free_dofs::Vector{Int}
     inhomogeneities::Vector{T}
     dofmapping::Dict{Int,Int} # global dof -> index into dofs and inhomogeneities
-    bcvalues::Vector{BCValues{T}}
+    bcvalues::Vector{BCValues{Float64}} # BCValues are just the shape functions, not the BCs themselves
     dh::DH
     closed::ScalarWrapper{Bool}
 end
 
-function ConstraintHandler(dh::AbstractDofHandler)
+function ConstraintHandler(dh::AbstractDofHandler;bctype=Float64)
     @assert isclosed(dh)
-    ConstraintHandler(Dirichlet[], AffineConstraint[], Int[], Int[], Float64[], Dict{Int,Int}(), BCValues{Float64}[], dh, ScalarWrapper(false))
+    ConstraintHandler(Dirichlet[], AffineConstraint[], Int[], Int[], bctype[], Dict{Int,Int}(), BCValues{Float64}[], dh, ScalarWrapper(false))
 end
 
 """
@@ -357,7 +357,7 @@ function update!(ch::ConstraintHandler, time::Real=0.0)
 end
 
 # for faces
-function _update!(inhomogeneities::Vector{Float64}, f::Function, faces::Set{<:BoundaryIndex}, field::Symbol, local_face_dofs::Vector{Int}, local_face_dofs_offset::Vector{Int},
+function _update!(inhomogeneities::Vector, f::Function, faces::Set{<:BoundaryIndex}, field::Symbol, local_face_dofs::Vector{Int}, local_face_dofs_offset::Vector{Int},
                   components::Vector{Int}, dh::AbstractDofHandler, facevalues::BCValues,
                   dofmapping::Dict{Int,Int}, time::T) where {T}
 
--- a/docs/src/literate/helmholtz.jl
+++ b/docs/src/literate/helmholtz.jl
@@ -51,6 +51,8 @@ using LinearAlgebra
 const ∇ = Tensors.gradient
 const Δ = Tensors.hessian;
 
+k = 2.0
+
 grid = generate_grid(Quadrilateral, (150, 150))
 
 dim = 2
@@ -60,6 +62,7 @@ qr_face = QuadratureRule{dim-1, RefCube}(2)
 cellvalues = CellScalarValues(qr, ip);
 facevalues = FaceScalarValues(qr_face, ip);
 
+# add Complex?
 dh = DofHandler(grid)
 push!(dh, :u, 1)
 close!(dh)
@@ -68,20 +71,12 @@ close!(dh)
 # This is a good testing strategy for PDE codes and known as the method of manufactured solutions.
 
 function u_ana(x::Vec{2, T}) where {T}
-    xs = (Vec{2}((-0.5,  0.5)),
-          Vec{2}((-0.5, -0.5)),
-          Vec{2}(( 0.5,  -0.5)))
-    σ = 1/8
-    s = zero(eltype(x))
-    for i in 1:3
-        s += exp(- norm(x - xs[i])^2 / σ^2)
-    end
-    return max(1e-15 * one(T), s) # Denormals, be gone
+    return exp(im*k*x[1])
 end;
 
-dbcs = ConstraintHandler(dh)
 # The (strong) Dirichlet boundary condition can be handled automatically by the Ferrite library.
-dbc = Dirichlet(:u, union(getfaceset(grid, "top"), getfaceset(grid, "right")), (x,t) -> u_ana(x))
+dbcs = ConstraintHandler(dh;bctype=ComplexF64)
+dbc = Dirichlet(:u, union(getfaceset(grid, "left"), getfaceset(grid, "bottom"), getfaceset(grid, "top"), getfaceset(grid, "right")), (x,t) -> u_ana(x))
 add!(dbcs, dbc)
 close!(dbcs)
 update!(dbcs, 0.0)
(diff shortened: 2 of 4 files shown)


kimauth:
Hi and welcome to Ferrite.jl!

I am not aware of anyone who tried to use `Ferrite` with complex numbers - most of us use it for solid mechanics where that is not a common use case. From your PR it looks like the main problem is that we restrict the number type of nodes to `<:Real`. Being more permissive on that side doesn't seem like a problem to me and it's probably even possible without breaking changes.

I am not familiar with complex valued problems, could you expand a little on which parts would be complex numbers? E.g. does it make any sense to have complex valued nodal coordinates, but real valued fields (or the other way around)? Or is it save to assume that they always have the same number type?

koehlerson:
> E.g. does it make any sense to have complex valued nodal coordinates, but real valued fields (or the other way around)?

I think the goal should be that the number type is part of the field, i.e. it needs to be specified when pushing a field inside a `DofHandler`. Consider a quantity of an electromagnetic field that is described by some complex value, then your domain is still describable with `Real` but your field isn't. Further, if you have multiple fields, but only one acting on the complex domain, it shouldn't be part of the `DofHandler` but rather only a specification of one field

kimauth:
I don't quite understand what you mean by:
> Further, if you have multiple fields, but only one acting on the complex domain, then it becomes clear why it shouldn't be part of the DofHandler but rather only a specification of one field

Though the `Field` doesn't really need to know its number type, does it? I believe they're basically only used for dof-distribution and a dof of a complex field is still just one dof. From that perspective the `DofHandler` shouldn't care about the number types of the fields. But it stores the grid and thus needs to know the number type of the nodes.

The `ConstraintHandler` on the other hand needs to know the number type of the problem, because it actually stores the prescribed values. If you had some complex valued field and some real valued field on the same domain, you'd thus need several `ConstraintHandler` - but you'd also get problems with the variable vector not being concretely typed, so perhaps that's a little too complicated of an example to start with.

koehlerson:
If you define a field with vectorial values, then you specify the vectorial type in the `push` operation, since you declare there all specifications of that particular field. If the field is complex valued, then it's in the same way just a feature of that field. The number type of the nodes will still be a `Real`. There is, imo, no need to encode the nodes in complex numbers, at least I don't see any advantage in it. Your domain is still just a subset of the one,two or three dimensional physical space and therefore perfectly describable by real values. What I'm arguing for is, that it should be stored associated to a field in a `<:AbstractDofHandler` such that is saved somewhere and other parts, as e.g. the `ConstraintHandler` can access it and conduct the right number type when needed

jondea:
Thank you both for the helpful discussion!

> I am not familiar with complex valued problems, could you expand a little on which parts would be complex numbers? E.g. does it make any sense to have complex valued nodal coordinates, but real valued fields (or the other way around)? Or is it save to assume that they always have the same number type?

Complex valued fields arise naturally when we assume the field is time harmonic like in the Helmholtz equation, [here is an example with the interaction between the Helmholtz and time-harmonic linear elasticity equations](
https://oomph-lib.github.io/oomph-lib/doc/acoustic_fsi/acoustic_fsi_annulus/html/index.html) which might help. In these types of problems the coordinates are real and the field is complex (if you're interested, PMLs make use of complex coordinates, but the complex coordinates are typically hidden from the FEM using a coordinate transformation). To answer your last question, it is not safe to assume they have the same number type.

>  The number type of the nodes will still be a `Real`.

I may have my definition of a node wrong, but I think it would be natural for a complex valued field to be described by complex valued nodes.

>  Your domain is still just a subset of the one,two or three dimensional physical space and therefore perfectly describable by real values.

The domain is describable by real values, but the field is not naturally.

> What I'm arguing for is that it should be stored associated to a field in a `<:AbstractDofHandler` such that is saved somewhere and other parts, as e.g. the `ConstraintHandler` or the vtk export, can access it and conduct the right number type when needed

I don't know enough about the workings of Ferrite to know if that would work, but I think the most important properties would be:
- The assembled matrix has complex entries, otherwise we end up with a Jacobian with duplicated partial derivatives (see [Cauchy–Riemann equations](https://en.wikipedia.org/wiki/Cauchy%E2%80%93Riemann_equations)). 
- The field in all its representations is complex valued, allowing us to operate on it in the natural mathematical/Julian way. For example, writing residuals using complex values rather than juggling two separate real values which happen to interact as if they were components of a complex number. 

> _Edit: I'd actually expect the CellValues to run into problems with complex valued nodes, too. But from a quick glance at your example, the nodes don't actually have complex coordinates. Why do you need the changes to Grid then?_

Good point, I think I accidentally left this in from a previous attempt. However, I think this is only unnecessary in this example because I never build an explicit representation of the actual field using a grid. I just solve the linear system and operate on the resulting vector of unknown directly. I assume there are reasonable cases where you would want such a representation?

koehlerson:
> I may have my definition of a node wrong, but I think it would be natural for a complex valued field to be described by complex valued nodes.

I don't think so, since it would be unnecessary memory usage. A node is just a point inside your physical domain and therefore the imaginary part will always be zero. In your example, you also use Float64 as your Node datatype, see here: 
https://github.com/Ferrite-FEM/Ferrite.jl/pull/451/files#diff-d04e26d2490486e6301ad9db2bccf6f9d1cc2781d89f4106874cff3ec1b5a40dR56

if you check `grid.nodes` the output should be a `Vector{Node{Float64}}`.

> > What I'm arguing for is that it should be stored associated to a field in a `<:AbstractDofHandler` such that is saved somewhere and other parts, as e.g. the `ConstraintHandler` or the vtk export, can access it and conduct the right number type when needed

This is still one valid route. The most important part is that the whole DofHandler shouldn't be constrained as you've done here:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/451/files#diff-8db5ed06971ce1004169862f3e499f55666dd630868843be8a53d30ea385e505R28

but rather only the association of one field should be changed, such that the `ConstraintHandler`, vtk export etc. can derive the right number type of that particular field. You need to think about other use cases where, maybe, not only one field is present, but two or even more and possibly one field is not complex (in a loose sense). 

Another route is to don't save this field specification inside the `DofHandler` because it's not really needed for dof counting/management (which is the ultimate purpose of the DofHandler) and just pass it whenever explicitly needed, so when creating the `ConstraintHandler`, `create_sparsity_pattern`, `Dirichlet`, vtk export. Then, of course, the `bcvalues` inside the `DofHandler` need some tweaking because - at least I think so - the `ConstraintHandler` gets it's number type from this or at least assumes to be of the same type.

jondea:
I see, I did have the wrong definition of a node. I thought it was the name given to any coefficient applied to a shape function. If nodes refer exclusively to the definition of the domain, I completely agree.

I've been playing around with complex dofs a bit more and I think it's a bit clearer. Here's what I'm proposing.

DofHandler stores the dof type which, this allows us to define it once for the problem
- passing it in separately potentially very confusing and adds no new functionality, because if you don't pass in the same type everywhere, you get an error
- I've never seen a problem with both real and complex unknowns. if you have a problem with fields with different types, you have to do something clever with the matrices anyway, so it should be handled in something like the MixedDofHandler. Ive never seen such a problem, but u suspect you'd have to promote everything to complex anyway, or at the very least separate your problem into blocks and up/downcast when they interact.

jondea:
Sorry, I accidentally pressed close! I've been playing around with this for a bit and I'm hoping to submit an improved version soon.

src/Dofs/ConstraintHandler.jl
```diff
@@ -67,14 +67,14 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
...
     inhomogeneities::Vector{T}
     dofmapping::Dict{Int,Int} # global dof -> index into dofs and inhomogeneities
-    bcvalues::Vector{BCValues{T}}
+    bcvalues::Vector{BCValues{Float64}} # BCValues are just the shape functions, not the BCs themselves
     dh::DH
     closed::ScalarWrapper{Bool}
 end
 
-function ConstraintHandler(dh::AbstractDofHandler)
+function ConstraintHandler(dh::AbstractDofHandler;bctype=Float64)
```
fredrikekre, line 75: An interface where you pass the type as the first argument:
```suggestion
function ConstraintHandler(T::Type{<:Number}, dh::AbstractDofHandler)
```
would be more typical I think.

Issue #453, opened 2022-06-04
Self-inclusion in topology information incomplete


termi-official:
The new ExclusiveTopology returns an incomplete neighborhood when asking for self-inclusion of the element (e.g. getneighborhood(top, grid, cellidx; include_self=true) - this case is not covered by our unit tests https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_grid_dofhandler_vtk.jl#L313-L338 .

Also I notices two undocumented convenience functions https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L395-L396 , maybe we could add these to the docs or are they for purely internal use?
 
cc @koehlerson

PR #455, opened 2022-06-06
topology fixes and docs clean up

koehlerson:
fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/518 and https://github.com/Ferrite-FEM/Ferrite.jl/issues/453


docs/src/reference/grid.md
```diff
@@ -41,6 +41,7 @@ getcoordinates!
 Ferrite.ExclusiveTopology
 Ferrite.getneighborhood
 Ferrite.faceskeleton
+Ferrite.toglobal
```
termi-official, line 44: I just want to quickly discuss the naming, because it is ambiguous in distributed memory codes, where global index is the unique index of an entity in the process set. Since we call the element local index "vertex index", how do we call this "global index"?

cc @fredrikekre

src/Grid/grid.jl
```diff
@@ -371,7 +371,12 @@ function getneighborhood(top::ExclusiveTopology, grid::AbstractGrid, vertexidx::
     cell_vertices = vertices(getcells(grid,cellid))
     global_vertexid = cell_vertices[local_vertexid]
     if include_self
-        return [top.vertex_vertex_neighbor[global_vertexid].neighbor_info; vertexidx]
+        self_reference_local = VertexIndex[]
+        for cellid in top.vertex_to_cell[global_vertexid]
+            local_vertex = VertexIndex(cellid,findfirst(x->x==global_vertexid,vertices(getcells(grid,cellid))))
+            push!(self_reference_local,local_vertex)
```
termi-official, line 377: Is there a performance hit in not preallocating the size? (we should have the size in the tables)

src/Grid/grid.jl
```diff
@@ -392,8 +397,13 @@ Returns an iterateable face skeleton. The skeleton consists of `FaceIndex` that
 """
 faceskeleton(top::ExclusiveTopology, grid::AbstractGrid) =  top.face_skeleton
 
-toglobal(grid::Grid,vertexidx::VertexIndex) = vertices(getcells(grid,vertexidx[1]))[vertexidx[2]]
-toglobal(grid::Grid,vertexidx::Vector{VertexIndex}) = unique(toglobal.((grid,),vertexidx))
+"""
+    toglobal(grid::AbstractGrid, vertexidx::VertexIndex)
+    toglobal(grid::AbstractGrid, vertexidx::Vector{VertexIndex}) -> Vector{Int}
+This function takes the local vertex representation (a `VertexIndex`) and looks up the unique global id (an `Int`).
```
termi-official, line 403: See above.

src/Grid/grid.jl
```diff
@@ -392,8 +397,13 @@ Returns an iterateable face skeleton. The skeleton consists of `FaceIndex` that
 """
 faceskeleton(top::ExclusiveTopology, grid::AbstractGrid) =  top.face_skeleton
 
-toglobal(grid::Grid,vertexidx::VertexIndex) = vertices(getcells(grid,vertexidx[1]))[vertexidx[2]]
-toglobal(grid::Grid,vertexidx::Vector{VertexIndex}) = unique(toglobal.((grid,),vertexidx))
+"""
+    toglobal(grid::AbstractGrid, vertexidx::VertexIndex)
```
termi-official, line 401:
```suggestion
    toglobal(grid::AbstractGrid, vertexidx::VertexIndex) -> Int
```

to make it consistent with the other comment?

termi-official (changes requested):
Thanks for the quick fix! The edge and face neighborhood should still have the same issue.

docs/src/reference/grid.md
koehlerson: I think global/local will never be uniquely defined. Just another example would be the p4est concept where you have sort of "global" cells and "local" cells

koehlerson:
> Thanks for the quick fix! The edge and face neighborhood should still have the same issue.

Face neighborhood should not, since it should be 1:1 correspondence for conforming grids. Nonconforming grids cannot encode the face neighborhood as discussed in the previous PR. For edge neighborhood this is true, but it is sort of exotic feature which probably isn't needed that often, but when it someone needs it I can do a follow up PR

koehlerson:
ok nvm, I added the `include_self` functionality for edges. However, still the statement about the face neighborhood should hold true. I think at this point I can't do better than just extending the returned array by the index itself for faces

docs/src/reference/grid.md
termi-official: Let us discuss this one next week, whether we can come up with a clean solution.

termi-official:
> > Thanks for the quick fix! The edge and face neighborhood should still have the same issue.
> 
> Face neighborhood doesn't have this issue, since it should be 1:1 correspondence for conforming grids. Nonconforming grids cannot encode the face neighborhood as discussed in the previous PR. For edge neighborhood this is true, but it is sort of exotic feature which probably isn't needed that often, but when it someone needs it I can do a follow up PR

Indeed, I agree that edge neighborhoods are not that commonly needed.

I think I will open up a follow up PR extending the docs, because something is inconsistent in my head. Maybe I can find what it is with an example.

```
                   7-----8-----9
                   |  3  |  4  |
                   4-----5-----6
                   |  1  |  2  |
                   1-----2-----3
```

For vertex 5 the neighborhood in the exclusive topology is spanned by the global vertices (2,4,6,8) expressed in the element local vertex indices, correct? 

Now with this PR `include_self=true` includes also 5 expressed in the element local vertex indices, right?

For the right face on element 1 the neighborhood is the left face in element two, right? 

So just a part of the face "(2,5)". With `include_self=true`, we now get both halves of the face "(2,5)" in our set. However, this is a semantically different neighborhood. Do you see what I mean? For the vertex we get for one the calls either the vertices connected by the 1-faces or the vertices setminus the vertex for which we query the neighborhood, while for the faces we obtain the opposing local portion or both local portions.

Do you agree that it would be better to rename the method to something different (i think they actually return something like to a vertex-patch, when I think about cell patches) and make the `getneighborhood` method consistent with the face neighborhood or vice versa?

test/test_grid_dofhandler_vtk.jl
```diff
@@ -271,10 +271,11 @@ end
     hexgrid = generate_grid(Hexahedron,(2,2,1)) 
     topology = ExclusiveTopology(hexgrid)
     @test topology.edge_neighbor[1,11] == Ferrite.EntityNeighborhood(EdgeIndex(4,9))
+    @test getneighborhood(topology,hexgrid,EdgeIndex(1,11),true) == [EdgeIndex(4,9),EdgeIndex(2,12),EdgeIndex(3,10),EdgeIndex(1,11)]
     @test topology.edge_neighbor[2,12] == Ferrite.EntityNeighborhood(EdgeIndex(3,10))
+    @test getneighborhood(topology,hexgrid,EdgeIndex(2,12),true) == [EdgeIndex(3,10),EdgeIndex(1,11),EdgeIndex(4,9),EdgeIndex(2,12)]
```
termi-official, line 276: Can we relax these tests to not rely on the ordering of the elements? I.e. comparing sets instead of arrays?

termi-official (changes requested):
I found a few minor issues.

Also, I think test coverage for `getneighborhood` with FaceIndex is missing in 2d and 3d.

src/Grid/grid.jl
koehlerson: > Also I notices two undocumented convenience functions https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L395-L396 , maybe we could add these to the docs or are they for purely internal use?

https://github.com/Ferrite-FEM/Ferrite.jl/issues/453

koehlerson: Since Mischa will use it as well in his yet to be published documentation example, it is not internal per se

koehlerson:
> I found a few minor issues.
> 
> Also, I think test coverage for `getneighborhood` with FaceIndex is missing in 2d and 3d.

Since the FaceIndex `getneighborhood` dispatch does not reconstruct but instead only accesses the neighborhood matrix, I'd say that test coverage is missing is a bit stretched. There are tests which check the matrix entries and thus, the logic of the function. I can change them to use the function instead, but it's not really a benefit

termi-official:
> Since the FaceIndex getneighborhood dispatch does not reconstruct but instead only accesses the neighborhood matrix, I'd say that test coverage is missing is a bit stretched. There are tests which check the matrix entries and thus, the logic of the function. I can change them to use the function instead, but it's not really a benefit

These are two separate things. With one test we check if the data structure is correctly constructed. The next test is to check if the interface to the data structure works correctly. Note that we can make changes to the backend without changing the contract of the interface.

test/test_grid_dofhandler_vtk.jl
termi-official: Ping @koehlerson

koehlerson: should be done in a separate PR to improve the tests in general

termi-official (approved):
LGTM. I think this can be merged @fredrikekre .

fredrikekre:
You two are in charge of this code, so if you are happy with it I am happy with it :)

PR #456, opened 2022-06-19
Fix bug with L2projection of mixed grid

lijas:
We were not using the local index to get the quadrature point data, so it was not working for arbitrary cellsets.


kimauth:
Lgtm. I did confuse myself with the fact that we call the vector of cell indices `set` while reviewing though, perhaps we should change that to something like `cellindices` or `cellids`.

kimauth (approved):

fredrikekre:
> I did confuse myself with the fact that we call the vector of cell indices set while reviewing though

Same!

PR #457, opened 2022-06-30
Add global_dof_range to get the global index range for a field.

fredrikekre:


KristofferC:
Will this always be a dense range? In that case,  maybe returning a range is suitable?

fredrikekre:
No since we distribute element by element, but I planned to add an option to number so you get a global block matrix instead and in that case I suppose it can return a range.

kimauth:
Why only for the `DofHandler`? 🙃

src/Dofs/DofHandler.jl
```diff
@@ -103,6 +103,25 @@ function dof_range(dh::DofHandler, field_name::Symbol)
...
+    field_range = dof_range(dh, field_name)
+    for i in 1:getncells(dh.grid)
+        celldofs!(eldofs, dh, i)
+        for j in field_range
+            @inbounds d = eldofs[j]
+            d in dofs || push!(dofs, d)
+        end
+    end
+    return sort!(collect(Int, dofs))
+end
```
KnutAM, line 123: I needed this with MixedDofHandler to test #532 
```suggestion
function global_dof_range(dh::MixedDofHandler, field_name::Symbol)
    dofs = Set{Int}()
    for fh in dh.fieldhandlers
        if field_name ∈ Ferrite.getfieldnames(fh)
            _global_dof_range!(dofs, dh, fh, field_name, fh.cellset)
        end
    end
    return sort!(collect(Int, dofs))
end

function global_dof_range(dh::DofHandler, field_name::Symbol)
    dofs = Set{Int}()
    _global_dof_range!(dofs, dh, dh, field_name, 1:getncells(dh.grid))
    return sort!(collect(Int, dofs))
end

function _global_dof_range!(dofs, dh, dh_fh, field_name, cellset)
    eldofs = celldofs(dh, first(cellset))
    field_range = dof_range(dh_fh, field_name)
    for i in cellset
        celldofs!(eldofs, dh, i)
        for j in field_range
            @inbounds d = eldofs[j]
            d in dofs || push!(dofs, d)
        end
    end
end
```

KristofferC:
What is this used for?

fredrikekre:
Nothing :) Just rebased an old PR. I think you kinda need something like this when creating blocked sparsity patterns though since you need to compute the block sizes.

KristofferC:
Then it could be `n_global_dof` or something maybe 🤷

PR #458, opened 2022-06-30
Expand manual section on Dirichlet constraints.

fredrikekre:


docs/src/manual/boundary_conditions.md
```diff
@@ -10,20 +10,85 @@ the most common ones, Dirichlet and Neumann boundary conditions, and how to do i
...
+Adding the constraints to the constraint handler is done with [`add!`](@ref):
+
+```julia
+add!(ch, dbc1)
+add!(ch, dbc2)
+```
+
+Finally, just like for the dof handler, we need to use [`close!`](@ref) to finalize the
+constraint handler. Internally this will then compute the degrees-of-freedom that match the
+constraint we added.
```
koehlerson, line 74:
```suggestion
constraints we added.
```

docs/src/manual/boundary_conditions.md
```diff
@@ -10,20 +10,85 @@ the most common ones, Dirichlet and Neumann boundary conditions, and how to do i
...
+
+Since the constraints can in general depend on time we also need to need to call
+[`update!`](@ref) with the current time in order to compute the prescribed values. Note that
+this has to be done even if the constraints does *not* depend on time (as `dbc2` above). The
```
koehlerson, line 78: what do you think about putting this in a warning box? When I started using JuAFEM I didn't notice that this is required

docs/src/manual/boundary_conditions.md
```diff
@@ -10,20 +10,85 @@ the most common ones, Dirichlet and Neumann boundary conditions, and how to do i
...
+```julia
+dbc1 = Dirichlet(
+    :u,                       # Name of the field
+    getfaceset(grid, "left"), # Part of the boundary
+    (x, t) -> 1.0 * t,        # Function mapping coordinate and time to a prescribed value
+)
+```
+
+The field name is given as a symbol, just like when the field was added to the dof handler,
+the part of the boundary where this constraint is active is given as a face set, and the
```
koehlerson, line 35:
```suggestion
the part of the boundary where this constraint is active is given as a `<:BoundaryIndex` set, and the
```
I think we can also do node and edge sets, can't we?

docs/src/manual/boundary_conditions.md
fredrikekre: I was just going to submit a patch that calls `update` implicitly in `close!`, it should be very cheap.

koehlerson: nice

docs/src/manual/boundary_conditions.md
fredrikekre: Yea I suppose, but feels slightly more advanced/unusual. Perhaps it could be a subsection about it or something.

PR #459, opened 2022-06-30
Implicitly call update!(ch) in close!(ch).

### LINKED ISSUE #207: Having to send a time into the function for constraints is a design failure.

fredrikekre:
Fixes #207, closes #213, closes #435.


fredrikekre:
I updated this PR a bit and intend to merge it. It now adds the call to `update!(ch)` in `close!(ch)`, but also allows specifying function of the form `f(x)` for the common case where they don't depend on time.

fredrikekre:
Regarding #435, if one needs something like that it can be managed by using a closure over a mutable state.

PR #463, opened 2022-07-05
More hyperelasticity fixes

termi-official:
Last PR was merged faster than I could comment. :p But I think we start converging with the hyperelasticity example towards a really good example for nonlinear fem (in the solid mechanics context).

I basically simplified the Newton to be more like the in the theoretical description provided in this article. Convergence is the same as before this PR.

### DIFF
--- a/docs/src/literate/hyperelasticity.jl
+++ b/docs/src/literate/hyperelasticity.jl
@@ -78,7 +78,7 @@ using Ferrite, Tensors, TimerOutputs, ProgressMeter, IterativeSolvers
 # and the tangent of ``\mathbf{S}`` as
 #
 # ```math
-# \frac{\partial \mathbf{S}}{\partial \mathbf{C}} = 2 \frac{\partial^2 \Psi}{\partial \mathbf{C}^2}.
+# \frac{\partial \mathbf{S}}{\partial \mathbf{C}} = 4 \frac{\partial^2 \Psi}{\partial \mathbf{C}^2}.
 # ```
 #
 # Finally, for the finite element problem we need ``\mathbf{P}`` and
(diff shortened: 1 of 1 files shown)


koehlerson:
I have to say that I also like the notation with \Delta \Delta more. Not only that it's nice to visualize the update to that increment, but  one can use u + \Delta u as an initial guess for the next incremental solve, which isn't possible if \Delta u is the last newton iteration update (which is then very close to a zero update, since it's the one from the last Newton iteration)

termi-official:
Probably I have a typo in my code or I am misunderstanding something in your arguments. So, let me elaborate. 
## This PR 
The Newton-Raphson method is usually defined as the iterative technique to find roots of a system of nonlinear equations 
$$F(u) = 0$$
with updates of the form
$$ u_{n+1} = u_{n} + \Delta u_{n} $$
where we find the increment $\Delta u_{n}$ by solving the problem 
$$ \frac{d F(u)}{du} \Delta u_{n} = -F(u_n)$$ 
with $u_0 = 0$.

## Master
In the implementation found in the master we have the formulation
$$ u_{n+1} = u_0 + \Delta u_{n+1} $$
where $u_0 = 0$ and 
$$ \Delta u_{n+1} = \Delta u_{n} + \Delta \Delta u_n $$.
Here we find $\Delta \Delta u_n$ by solving the problem 
$$ \frac{d F(u)}{du} \Delta \Delta u_{n} = -F(u_n)$$ 
with $\Delta u_{n} = 0$.

-----

From a numerical perspective nothing changes, because the latter is literally just a rewrite of the former with one extra step. So the only change which I made is to make the algorithm closer to textbook (and the theory section which we have provided). I guess the current implementation is a remnant of a load stepping implementation, from which this example may be derived, in which the $u_n$'s were the load steps and not the Newton solution sequence, which now causes the confusion?

koehlerson:
It's only a rewrite, however, it is not only a rewrite as soon as you have incremental loading, i.e. incremental variational problems, which is likely to be the case. Currently people can read the example and copy paste it, wrap the while loop stuff in a for loop for the quasi-time and have a basic incremental loading newton solver.

PR #466, opened 2022-07-12
Fix PointEvalHandler when first point is missing + add some kwargs

kimauth:
The `PointEvalHandler` could not be constructed when the first point was not found.

Also adds keyword arguments for how many neigboring points to use and whether or not to print warnings for not found points. The former was a random hard-coded choice in the first place (so perhaps better allow the user to change it if necessary) and the latter is useful if many points are expected to not be found.

### DIFF
--- a/src/PointEval/PointEvalHandler.jl
+++ b/src/PointEval/PointEvalHandler.jl
@@ -40,17 +40,17 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
     end
 end
 
-function PointEvalHandler(grid::AbstractGrid, points::AbstractVector{Vec{dim,T}}) where {dim, T}
+function PointEvalHandler(grid::AbstractGrid, points::AbstractVector{Vec{dim,T}}; search_nneighbours=3, warn=true) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
-    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts)
+    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbours, warn)
     return PointEvalHandler(grid, cells, local_coords)
 end
 
-function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}) where {dim, T<:Real, C}
+function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbours, warn) where {dim, T<:Real, C}
 
     # set up tree structure for finding nearest nodes to points
     kdtree = KDTree(reinterpret(Vec{dim,T}, getnodes(grid)))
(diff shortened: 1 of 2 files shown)


src/PointEval/PointEvalHandler.jl
```diff
@@ -40,17 +45,17 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
...
-function PointEvalHandler(grid::AbstractGrid, points::AbstractVector{Vec{dim,T}}) where {dim, T}
+function PointEvalHandler(grid::AbstractGrid, points::AbstractVector{Vec{dim,T}}; search_nneighbors=3, warn=true) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
-    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts)
+    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn)
     return PointEvalHandler(grid, cells, local_coords)
 end
 
-function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}) where {dim, T<:Real, C}
+function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, serach_nneighbors, warn) where {dim, T<:Real, C}
```
fredrikekre, line 54:
```suggestion
function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn) where {dim, T<:Real, C}
```

src/PointEval/PointEvalHandler.jl
```diff
@@ -40,17 +45,17 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
...
 
     # set up tree structure for finding nearest nodes to points
     kdtree = KDTree(reinterpret(Vec{dim,T}, getnodes(grid)))
-    nearest_nodes, _ = knn(kdtree, points, 3, true) #TODO 3 is a random value, it shouldn't matter because likely the nearest node is the one we want
+    nearest_nodes, _ = knn(kdtree, points, serach_nneighbors, true)
```
fredrikekre, line 58:
```suggestion
    nearest_nodes, _ = knn(kdtree, points, search_nneighbors, true) 
```

src/PointEval/PointEvalHandler.jl
```diff
@@ -261,7 +266,7 @@ function _get_point_values!(
     # extract variables
     local_coords = ph.local_coords
     # preallocate some stuff specific to this cellset
-    pv = PointScalarValuesInternal(first(local_coords), ip)
+    pv = PointScalarValuesInternal(local_coords[findfirst(x->!isnothing(x), local_coords)], ip)
```
fredrikekre, line 269: Maybe just  return early if there is no found points?
```suggestion
    idx = findfirst(!isnothing, local_coords)
    idx === nothing && return out_vals
    pv = PointScalarValuesInternal(local_coords[idx], ip)
```

Issue #394, opened 2021-10-16
Grid inconsistencies


koehlerson:
There are a few inconsistencies related to the `Grid` type

- We have in `Grid` the cellsets defined as `Dict{String,Set{Int}}`, but I guess we should use `Dict{String,Set{CellIndex}}` https://github.com/Ferrite-FEM/Ferrite.jl/blob/a8b06904d15946483cc14e3656217ed2c279777e/src/Grid/grid.jl#L105
- `compute_vertex_values` should probably renamed to `compute_nodal_values` since the vertex is defined as the corners of a cell and neglects the "higher order" nodes
https://github.com/Ferrite-FEM/Ferrite.jl/blob/a8b06904d15946483cc14e3656217ed2c279777e/src/Grid/grid.jl#L246-L260 
 I don't know who or what part of Ferrite uses this function, but strictly speaking this would be a breaking change. Although easy to fix for everyone.
 
 @termi-official edit the issue or write me if I forgot something

termi-official:
We also should check for more definition inconsistencies in what edges and faces are.

KnutAM:
I tried to get a bit of an overview of the grid interface and figured I could share my notes/thoughts here. I found a few (probably known) inconsistencies and made some suggestions on how they could be fixed. Also related to the `get*(...)` vs `*(...)` issue... 

## Grid interface
### Coordinates
* `getcoordinates`
  * `Grid, Int` -> Cell coordinates
  * `Node` -> **Node** coordinates
  * `Grid, FaceIndex` -> Cell coordinates
  * `CellCache` -> Cell coordinates

* `getcoordinates!`
  * `x, Grid, Union{Int,Cell,CellIndex,FaceIndex}`: `x` -> Cell coordinates

* `Ferrite.cellcoords!`
  * `x, Union{[Mixed]DofHandler,Grid}, n::Int`: `x`->Cell coordinates

* `Ferrite.get_coordinate_eltype(::Union{Grid,Node})` returns the `T` in `Vec{dim,T}`

### Nodes
* `getnodes`
  * `Grid` -> `::Vector{Node}`
  * `Grid, Int` -> `Node`
  * `Grid, Union{String,Vector{Int}}` -> `Vector{Node}`

* `Ferrite.cellnodes!`
  * `ids::Vector{Int}, Union{Grid,MixedDofHandler}, Int`: `ids` -> The cell's node numbers

### Cells
* `cellid`
[...]

PR #468, opened 2022-07-18
Neumann BC

KnutAM:
Start of a more convenient interface for Neumann BC. 
```julia
nh = NeumannHandler(dh::AbstractDofHandler)
add!(nh, Neumann(:u, ∂Ω, fv::FaceValues, fun)) # fun(x::Vec,t)-> isa(fv, ScalarValued) ? Number : Vec
apply!(f::Vector, nh, time)
```
Still plenty of todos that I will try to fix in the upcoming days/weeks depending on available time
- [ ] Support MixedDofHandler
- [ ] Profiling
- [ ] Documentation
- [ ] More tests


src/neumann.jl
```diff
@@ -0,0 +1,58 @@
+# If face_value is ScalarValued, f should return scalar 
+# If face_value is VectorValued, f should return a vector
+struct Neumann{FV}      # cells and faces are a deconstructed
+    cells::Vector{Int}  # FaceIndex, to allow loop over cells 
+    faces::Vector{Int}  # with CellIterator 
+    face_values::FV     
+    field_name::Symbol
+    fun # f(x::Vec, t::Number, normal::Vec)->Union{<:Number, <:Vec}
```
kimauth, line 8: `fun` will be abstractly typed, as each function has its own type. Perhaps it should be a type parameter?

KnutAM: Yes, I have planned to do that, but right now I let it mirror the Dirichlet (except the restriction of ::Function).

src/neumann.jl
```diff
@@ -0,0 +1,58 @@
...
+        dΓ = getdetJdV(fv, q_point)
+        x = spatial_coordinate(fv, q_point, cell.coords)
+        n = getnormal(fv, q_point)
+        b = fun(x, time, n)
+        for i in 1:getnbasefunctions(fv)
+            δu = shape_value(fv, q_point, i)
+            fe[i] += δu ⋅ b * dΓ
+        end
+    end
+end
```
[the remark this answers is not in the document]
KnutAM: What motivated me to include everything is that Dirichlet boundary conditions are currently very easy, but Neumann is not. 
Perhaps I'm mistaken here, but I thought this one should cover most standard cases (contrary to element routines, which are more specific to the physics). 
If you have time/possibility, could you share how this is not covering your case?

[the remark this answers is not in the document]
kimauth: I can for example have boundary conditions on a diffusion problem. Possibly that could be squeezed in to run with this, but in coupled problems you could also need several fields in the expression for the Neumann bcs, resulting in needing several `Values` objects (I believe this is the case e.g. for phase field fracture). And as Ferrite is more of a general toolbox than a mechanical FE program, we should also consider that people might be solving completely different differential equations - I'm not sure how exactly the boundary conditions that are arising from e.g. electro-magnetics or electro-chemically coupled problems look.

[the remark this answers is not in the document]
KristofferC: Yes, by necessity this has to be tweaked to the particular weak form that you are solving (since a Neumann condition is just a part of the weak form). But depending on how you formulate it, (primal vs dual format) that will look different. So, to me, this starts to look like a "weak form library" but if that is the case, why only have the part of the weak form on the face and not the whole weak form in it?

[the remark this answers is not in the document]
KnutAM: Thanks for the feedback guys!
I'll leave it open for now, working on making Neumann more convenient but independent of the weak form. 
Might turn into something useful, or not - let's see :)

src/iterators.jl
```diff
@@ -143,3 +149,54 @@ function _check_same_celltype(grid::AbstractGrid, cellset::AbstractVector{Int})
         end
     end
 end
+
+struct FaceIterator{CI<:CellIterator} <: AbstractGridIterator
+    faces::Vector{Int}
+    current_faceid::ScalarWrapper{Int}
```
termi-official, line 155: Before we go down the rabbit hole of how we define what exactly a face is - looking the remaining your intent is to define a "boundary iterator", correct, and not an iterator over all faces?

termi-official (commented):
Some of the changes are quite interesting. I left some comments for clarification.

src/iterators.jl
KnutAM: Technically it isn't limited to the boundary, as it iterates over a faceset. It can also be used with internal boundaries. 
When discussing with @fredrikekre (who had the original idea), it seemed like that could be useful for some cases you have (perhaps with some additional features like checking if the face has already been iterated over and if so skip it?)

src/neumann.jl
```diff
@@ -0,0 +1,84 @@
+# If face_value is ScalarValued, f should return scalar 
+# If face_value is VectorValued, f should return a vector
+struct Neumann{FV}      # cells and faces are a deconstructed
```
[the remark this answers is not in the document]
KnutAM: Yes, the way it is implemented now is for weak forms of the types:

VectorValued: E.g. mechanical equilibrium, dual/mixed form of scalar problem (heat flow, diffusion, etc.)
$\int_{\Gamma} \boldsymbol{f}(\boldsymbol{x}, t, \boldsymbol{n}) \cdot \boldsymbol{\delta u} \ \mathrm{d}\Gamma$

ScalarValued: E.g. standard form of  e.g. heat flow/diffusion etc. 
$\int_{\Gamma} f(\boldsymbol{x}, t, \boldsymbol{n})  \delta u \ \mathrm{d}\Gamma$

src/iterators.jl
termi-official: I have indeed several use-cases where it is necessary to iterate over unique entities. Just to give you an idea why such a feature is generally useful. Currently we have the skeleton stored in the `ExplicitTopology` without materializing the faces.
1. Some of my error estimators require the integral over faces (in the sense of codimension 1 entities w.r.t. the element). For example Kelly's error indicator utilizes the relation ||∇(u-uₕ)||ₑ ≅ √( C hₑ ∑ₖ (hₖ ∫ |J[∇uₕ]|²) dS ) . However, my faces are usually non-conforming.
2. For continuum mechanical problems so-called discontinuous Galerkin formulations are extremely useful. The basic idea behind these formulations is simple: Instead of solving some balance in $\nabla \cdot F(u)$ we introduce our flux field F as an independent unknown $\tilde{F} = F(u)$ and solve additionally for some balance in $\nabla \cdot \tilde{F}$. Here most formulations utilize some penalty term on the faces, which again comes in the form of integrals over the faces. However, we can at least assume conforming faces.
3. The current dof assignment is way slower than necessary. If every entity in the mesh is materialized and has a unique identifier, then assigning dofs can be done in constant complexity - which can speed up AMR simulations quite a bit! We can do similar things in meshes with subdomains (i.e. MixedDofHandler).

termi-official: Indeed, I usually need both + some extra information and the `ExclusiveTopology` fixes the problem of having a unique iteration for now (since it extracts the skeleton). Let me leave 1 by the side for now, because it drastically complicates everything and we can go back when I have a working forest mesh. For the remaining cases I usually need matching quadrature points, normal and dof information. Note that, e.g., if you just just put quadrature points on two conforming half faces of a triangle, then their indices are rotated an/or revesed against each other. For completeness, I also have seen that some implementations also have face iterators which give you additionally the full element information for all "local portions of a face", but I do not exactly know when we need these.

Pulling in @koehlerson the only open todo to handle 4 was matching the quadrature points rigorously, right? For 2 it was also just https://github.com/Ferrite-FEM/Ferrite.jl/issues/398 being left to address?

koehlerson: So if a `FaceIterator` is added, it would be very nice to distinguish between a boundary, duplicate or unique iteration over the faces. The boundary iteration seems to be complete in this branch, duplicate is a nobrainer and the unique iteration could be done when accepting a `Vector{FaceIndex}` which you can get from `faceskeleton(top::ExclusiveTopology)`. The problem with the latter is that multiple `FaceValues` need to be involved, e.g., for a FSI problem a displacement `FaceValue` and fluid `FaceValue` because they probably don't share the interpolation type (and definitely they don't share the material dependent stuff and thus they don't share properties at the gauss points). It's likely the same for all other mentioned examples. It would be nice to have a struct that wraps the multiple `FaceValue`s because handling them your own is (at least imho) quite error prone. As an example see this artificial jump evaluation: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_grid_dofhandler_vtk.jl#L355-L393 which goes into the direction of example 1 and 4 

> Pulling in @koehlerson the only open todo to handle 4 was matching the quadrature points rigorously, right? For 2 it was also just https://github.com/Ferrite-FEM/Ferrite.jl/issues/398 being left to address?

yes

KnutAM: Are you talking about renaming the current `FaceValues` object? It seems quite breaking to me?
But the idea of having an `InterfaceValues` type wrapping two `FaceValues` makes a lot of sense to me for automating the matching of integration points! 
From the FaceIterator point of view, the ability to provide an additional `Vector{FaceIndex}` is no problem, and should play nicely with an `InterfaceValues` type if added later.

termi-official: Deprecating first and renaming the `FaceValues` and/or subtypes in a later release is one option. Such breaking changes should be fine as long as we provide a transition path over an acceptable time window, because it will only involve renaming things in their source codes. My reasoning is basically that the current state can be misleading if we start providing infrastructure for acutal face operations - note here that the current `FaceScalar/VectorValues` are designed around the local portion of some face and not the actual face.

Regarding the additional `Vector{FaceIndex}` we might want to think trough this a bit more, because an unclean interface (as in software interface) to handling "faces" can lead to really painful problems later (and obviously also strongly depends on the actual definition of what a face is). Maybe an approach is to have some face materialization interface which is at least specialized on the mesh?

KnutAM: Interesting points! Are you now also talking about the difference between a geometrical face and a cell face (when talking about the local portion vs global)? Or just that the current values are representative of only one side of a face (the one of the cell that they are initialized on)? I've always thought of the current implementation as just `CellFaceValues` with the `Cell` part being implicit. 

I see your point with the additional input. From that point of view, it might make sense to wait with the `InterfaceValues` interface of `FaceIterator`, until the infrastructure for actual face operations is further developed? And then one can see if the same iterator makes sense or not. 
In this case, only the main interface: `FaceIterator(dh::AbstractDofHandler, faceset::Set{FaceIndex}, cellset=nothing)` should be documented now, and it is initially only dealing with `(Cell)FaceValues`.

termi-official: Happy that my feedback is at least a bit useful. :) I am primarily concerned about the possibility that faces are topologically (and hence geometrically) non-conforming, because we never really assume that a face has to be conforming in the implementation. Working with non-conforming problems currently works really good (see e.g. https://github.com/Ljon4ik4/FEMRefinements.jl/blob/main/examples/HeatEquation.jl). The interpretation of faces as half-faces is also a bit problematic. Think about 1d elements in 3d (e.g. trusses), where more than two trusses can join at the same point. This issue can be generalized to higher dimensional elements and spaces. However, most of the time the interpretation of faces as two half-faces is really beneficial from a performance point of view, because we know that we do not have to deal with something weird. :)

I am definitely in favor of starting with designing an interface for faces and edges before moving on with some full-fledged face iterator. However, the main interface makes sense to me. Depending on how the "face interface" will turn out, maybe `FaceIterator(dh::AbstractDofHandler, cellset=nothing)` will be sufficient. If faces will be materialized, than it might bebeneficial to have something more like `FaceIterator(dh::AbstractDofHandler, faceset::Set{Face}, cellset=nothing)`.

Also, another issue may be defining a face is mixed dimensional problems (e.g. 1D-3D coupled problems). Currently a face is defined with respect to the dimension ambient space, which has weird implications. E.g. for a quadrilateral element embedded in 3D the element itself is the face. I do not have an idea yet on how we could address this and the full implication here.

KnutAM: Definitively useful, and very interesting points! And as you say, the definitions of faces and interfaces become much more tricky for mixed and nonconforming cases. 

The more I look at the proposal below, the less I like it. So I think I will close this PR soon, and then open a pure  BoundaryFaceIterator PR to simplify the Neumann boundary conditions (less convenient than here, but gives a much cleaner code in Ferrite.jl). 

My feeling is that solving the interface values problems is a much larger discussion, which I think deserves its separate issue? Or perhaps should be continued under #394?

termi-official: I still really like the idea of having an easy to use iterator interface, not wanting to talk your idea bad. But I agree that this is a larger discussion. Let us move to the mentioned issue for now as you suggested. Still, I think we can leave the draft open, because we can rebase it against the newer version once we have settled with the face/edge issues - or we can open a new PR. :)

KnutAM: No problem - the point of a draft pr is to get feedback on whether some ideas are relevant as far as I'm concerned:)
But closing this PR doesn't affect the option of rebasing later and I think it's nice not to have too many dormant PRs

KnutAM:
Closing this PR as it is out of `Ferrite.jl`'s scope

KnutAM:
Closing this PR as it is out of `Ferrite.jl`'s scope

PR #469, opened 2022-07-20
Throw a more descriptive error message when passing a too long element dof vector.

fredrikekre:


fredrikekre:
This benchmark:
```
using Ferrite, BenchmarkTools

const ip = Lagrange{1,RefCube,1}()
const qr = QuadratureRule{1,RefCube}(1)
const cv = CellScalarValues(qr, ip)
const ue = rand(getnbasefunctions(cv))
const qp = 1

@btime function_value(cv, qp, ue)
```
goes from 4ns to 2ns too.

KnutAM:
Nice! 
Is the plan to fix the same for FaceValues etc. as well, seems to be a few similar `@assert`s.

fredrikekre:
Done.

PR #470, opened 2022-07-25
max(u...) -> maximum(u) in plasticity example

KnutAM:
Seemed like a strange code construct to have in an example...

### DIFF
--- a/docs/src/literate/plasticity.jl
+++ b/docs/src/literate/plasticity.jl
@@ -328,7 +328,7 @@ function solve()
         ## Update the old states with the converged values for next timestep
         states_old .= states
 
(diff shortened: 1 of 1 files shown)


KristofferC (approved):

docs/src/literate/plasticity.jl
```diff
@@ -328,7 +328,7 @@ function solve()
         ## Update the old states with the converged values for next timestep
         states_old .= states
 
-        u_max[timestep] = max(abs.(u)...) # maximum displacement in current timestep
+        u_max[timestep] = maximum(abs.(u)) # maximum displacement in current timestep
```
fredrikekre, line 331: `maximum(abs, u))` would be even better :)

PR #471, opened 2022-08-18
Add assemble function for non-square matrices

lijas:
Convenient for Lagrange multiplier constraints etc.

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -51,6 +51,27 @@ function assemble!(a::Assembler{T}, dofs::AbstractVector{Int}, Ke::AbstractMatri
 end
 
 """
+    assemble!(a::Assembler, rowdofs, coldofs, Ke)
+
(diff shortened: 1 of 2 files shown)


src/assembler.jl
```diff
@@ -50,6 +50,27 @@ function assemble!(a::Assembler{T}, dofs::AbstractVector{Int}, Ke::AbstractMatri
     end
 end
 
+"""
+    assemble!(a::Assembler, rowdofs, coldofs, Ke)
+
+Assembles the matrix `Ke` into `a` according to the dofs specified by `rowdofs` and `coldofs`.
+"""
+function assemble!(a::Ferrite.Assembler{T}, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix{T}) where {T}
```
fredrikekre, line 58: Perhaps some code duplication can be avoided by combining these two methods

fredrikekre (approved):

PR #474, opened 2022-08-30
Remove onboundary from examples

KnutAM:
If removed from Neumann example, probably good to remove from examples as well? Ref #473

Edit: Also adopted an unrelated doc change from @fredrikekre's comment in [#470](https://github.com/Ferrite-FEM/Ferrite.jl/pull/470#discussion_r928557474)


lijas:
Move the face integration outside the cell loop and do:
```
for (cellid, faceid) in union(getfaceset(grid, "left"), getfaceset(grid, "right") )
```
is my prefered way :)

KnutAM:
I like that better too, and it is what would be done by the BoundaryFaceIterator that is under construction!

KnutAM:
Handled in #495

PR #475, opened 2022-09-08
Fix face node ordering for Tetrahedron{3}

fredrikekre:
Fix face node ordering for Tetrahedron{3} s.t. they are counter clockwise from the outside.


fredrikekre:
Concerning that this passes tests :D

lijas:
Where would this cause an issue?We dont have any example that uses the normal, I think.

kimauth:
Add some tests? 🙃

PR #477, opened 2022-09-09
Different way of handling shells

lijas:
@kimauth mentioned that the current way of handling shells in the DofHandler interfears with cohesive zone elements.

This PR adds a wrapper interpolation for shells that does not mess with the DofHandler.


src/interpolations.jl
```diff
@@ -703,3 +699,35 @@ function value(ip::CrouzeixRaviart{2,1}, i::Int, ξ::Vec{2})
...
+
+################################################################
+# Wrapper Interolation for structural elements (shells, beams) #
+################################################################
+"""
+
+"""
+struct ShellInterpolation{dim_p,dim_s,shape,order,I<:Interpolation} <: Interpolation{dim_s,shape,order} 
+    ip::I
+end
```
termi-official, line 711: readoing this, wouldn't something like `StructuralInterpolation` be more precise?

termi-official (commented):
Great PR! However, I think this raises again the question on how faces and edges are defined.

src/Dofs/DofHandler.jl
```diff
@@ -248,7 +247,7 @@ function __close!(dh::DofHandler{dim}) where {dim}
                     end # edge loop
                 end
             end
-            if interpolation_info.nfacedofs > 0 && (interpolation_info.dim == dim)
+            if interpolation_info.nfacedofs > 0
```
termi-official, line 250: Removing this assumption may break mixed-dimensional elements.

termi-official:
Is this obsolete now?

PR #481, opened 2022-09-19
Periodic Nirvana

fredrikekre:
This patch fixes most remaining issues with periodic constraints:

 - collect_periodic_faces now return Vector{PeriodicFacePair} instead of Dict{Int,Int}. This is necessary since, in 3D, we need to keep track of i) the relative rotation between the faces and ii) if they are mirrored (opposing normal vectors) or not
 - Several bugs related to 3D periodicities and rotation/mirroring have been fixed
 - Numbering for quadratic interpolation for 3D cubes have been fixed to match the rest of Ferrite. The added tests for periodic constraints revealed this issue

With this patch, periodic constraints should work fine for linear and quadratic interpolations for and all supported cell types in both 2 and 3 dimensions.


KristofferC:
Are there any examples that use periodic B.C yet?

fredrikekre:
https://ferrite-fem.github.io/Ferrite.jl/stable/examples/computational_homogenization/

PR #486, opened 2022-09-26
Non-overlapping Distributed Assembly

termi-official:
In this PR a prototype for distributed assembly on non-overlapping mesh decompositions is proposed

~~**CAUTION - HIGH RAM USAGE**~~ Should be under control now.

# Open Problems/TODOs

* [x] Non-local interaction entries in distributed matrix missing (part of ghost layer)
* [x] Put distributed code into a abstraction layer
* [x] RHS assembly not working currently (non implemented error in PartitionedArrays when touching the RHS vector)
* [x] Constraint handler integration
* [x] Fix high memory memory usage
* [x] Proper naming of variables 👻
* [x] Manufactured solution test instead of norm
* [x] Generalize to higher order
* [x] Allow partitions to be disconnected from the boundary
* [x] Get vector valued problem running
* [x] Get rid of the just included celldofs/facedofs/vertexdofs/edgedofs dics
* [x] Merge master
* [x] HYPRE assembler
* [x] Fix regression in constraint handler
* [x] Split into 3 extensions (FerriteDistributed, FerriteHypre, FerritePartitionedArrays)
* [x] Proper test coverage
* [ ] Merge master
* [ ] Changelog entry

## Removed TODOs

* [x] ~~Remove unnecessary communication~~ Compression and stuff is future work...
* [x] ~~Ghost layer construction via codim-1 entities (should be close to the lower bound in terms of communication)~~ Wont't work for mixed meshes.
* [x] ~~Optimized exchanger construction (for PartitionedArrays.jl)~~
* [x] ~~Update to PartitionedArrays 0.3.0~~ Should be a follow up PR
* [x] ~~Backport to Julia 1.6 (use https://github.com/SciML/DiffEqBase.jl/commit/c2de920a88a39814ac368f703c40fbc282a9c078 for help)~~ Follow up if we want.

# Out of scope (stuff for future PRs)

* Mixed grids
* We do not fully utilize the potential of PartitionedArrays.jl
* Parallel mesh IO
* CSR/CSC assembly
* Reblance operations
* Space filling curve based partitioning
* Optimized ghost layer construction
* Parallel condensation (matrix product infrastructure missing in PartitionedArrays.jl)
* Optimized `generate_distributed_grid`
* Periodic BC handling
* Distributed Point finder


termi-official:
System matrix assembly for linear ansatz seems to work for all cases now. I will leave higher order interpolations for later, because there are more some fundamental problems which have to be resolved first. Also note that this is not an AMR ready implementation, as distributed AMR meshes usually work differently from the presented one.

Pulling in @fredrikekre @koehlerson for some help on design decisions and simplifying the algorithms.

1. Dof management becomes is currently weird, partly because we do not materialize the mesh. I moved some close code to the dof handler https://github.com/Ferrite-FEM/Ferrite.jl/blob/do/distributed-assembly/src/Dofs/DofHandler.jl#L25-L27 . Also, I have an almost duplicate code for the dof assignment https://github.com/Ferrite-FEM/Ferrite.jl/blob/do/distributed-assembly/docs/src/literate/distributed_assembly.jl#L172-L232 .
2. Ghost layer is literally a nightmare right now. The communication overhead is quite big and the algorithmic complexity seems to be higher than necessary.
3. The grid data structure seems to be a main source for suboptimal algorithms, because nothing is materialized or linked from the elements https://github.com/Ferrite-FEM/Ferrite.jl/blob/do/distributed-assembly/src/Grid/DistributedGrid.jl#L10-L55

termi-official:
Proof of concept for Lagrange interpolations of scalar-valued problems. Still need to integrate tests, integrate docs and simplify the code. Some TODOs are left for future work.

src/Grid/DistributedGrid.jl
```diff
@@ -0,0 +1,370 @@
...
+    grid_topology = ExclusiveTopology(grid_to_distribute)
+    return DistributedGrid(grid_to_distribute, grid_topology, grid_comm; partition_alg=partition_alg)
+end
+
+function create_partitioning(grid::Grid{dim,C,T}, grid_topology::ExclusiveTopology, n_partitions, partition_alg) where {dim,C,T}
+    N = getncells(grid)
+    @assert N > 0
+    
+    if n_partitions == 1
+        return ones(N)
```
fredrikekre, line 95:
```suggestion
        return ones(Metis.idx_t, N)
```

termi-official:
I guess this PR hits a functional state. Left solution, right partitioning.

Screenshot from 2022-10-24 21-29-35

Now begins the hard part. This one is currently dead-slow and has a tons of redundant communication. Please note that this one is fundamentally designed to be non-overlapping, hence communication during ghost layer construction may be unavoidable. After we get this one done I will also implement the two other common distributed assembly techniques (i.e. overlapping grid as in e.g. [ug4](https://link.springer.com/article/10.1007/s00791-014-0232-9) and sharing the full mesh between all processes).

I think the first three things which I would like to discuss are
1. how to improve shared entries
2. how to remove the redundancies between the DoFHandler and the DistributedDoFHandler (there is much copy pasta going on)
3. how to get rid of dimensional-copy-pasta (i.e. big branching code over dimension with minor differences in between, e.g seen [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/do/distributed-assembly/src/assembler.jl#L377-L485))

I will start writing proper tests once we have settled with a better design and some minimal distributed memory interface.

termi-official:
Superseded by https://github.com/Ferrite-FEM/FerriteDistributed.jl/pull/7

PR #489, opened 2022-09-27
improve performance of `apply!` be rewriting the part that zero out rows

KristofferC:
As a consequence the `strategy` kwarg is no longer useful.

Benchmark code:

```julia
grid = generate_grid(Hexahedron, (100, 100, 100));
dim = 3
cellvalues = CellScalarValues(qr, ip);
dh = DofHandler(grid)
push!(dh, :u, 1)
close!(dh);
K = create_sparsity_pattern(dh)
ch = ConstraintHandler(dh);
∂Ω = union(
    getfaceset(grid, "left"),
    getfaceset(grid, "right"),
    getfaceset(grid, "top"),
    getfaceset(grid, "bottom"),
);
add!(ch, dbc);
close!(ch)
update!(ch, 0.0);
f = zeros(size(K, 1))

using BenchmarkTools
@btime apply_zero!(K, f, ch)
```

```julia
julia> @btime apply_zero!(K, f, ch) # master
  692.526 ms (16 allocations: 423.98 MiB)

julia> @btime apply_zero!(K, f, ch)
  687.897 ms (4 allocations: 544 bytes)
```

Time is approx the same but we no longer need to keep two copies of the stiffness matrix in memory at the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+ - Runtime and allocations for application of boundary conditions in `apply!` and
+   `apply_zero!` have been improved. As a result, the `strategy` keyword argument is
+   obsolete and thus ignored. ([#489][github-489])
 
 ## [0.3.9] - 2022-10-19
 ### Added
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -541,8 +541,8 @@ function apply_zero!(K::Union{SparseMatrixCSC,Symmetric}, f::AbstractVector, ch:
     apply!(K, f, ch, true)
 end
 
+# For backwards compatibility, not used anymore
 @enumx ApplyStrategy Transpose Inplace
-# For backwards compatibility
 const APPLY_TRANSPOSE = ApplyStrategy.Transpose
 const APPLY_INPLACE = ApplyStrategy.Inplace
 
@@ -573,16 +573,7 @@ function apply!(KK::Union{SparseMatrixCSC,Symmetric}, f::AbstractVector, ch::Con
 
     # Remove constrained dofs from the matrix
     zero_out_columns!(K, ch.prescribed_dofs)
-    if strategy === ApplyStrategy.Transpose
-        K′ = copy(K)
-        transpose!(K′, K)
-        zero_out_columns!(K′, ch.prescribed_dofs)
-        transpose!(K, K′)
-    elseif strategy === ApplyStrategy.Inplace
-        K[ch.prescribed_dofs, :] .= 0
-    else
-        error("Unknown apply strategy")
-    end
+    zero_out_rows!(K, ch.dofmapping)
 
     # Add meandiag to constraint dofs
     @inbounds for i in 1:length(ch.inhomogeneities)
@@ -805,6 +796,15 @@ function zero_out_columns!(K, dofs::Vector{Int}) # can be removed in 0.7 with #2
     end
 end
 
+function zero_out_rows!(K, dofmapping)
+    rowval = K.rowval
+    nzval = K.nzval
+    @inbounds for i in eachindex(rowval, nzval)
+        if haskey(dofmapping, rowval[i])
+            nzval[i] = 0
+        end
+    end
+end
 
 function meandiag(K::AbstractMatrix)
     z = zero(eltype(K))


fredrikekre (approved):

KristofferC:
Could someone just verify the performance difference.

koehlerson:
```julia
julia> @btime apply_zero!(K, f, ch) #master
  340.741 ms (16 allocations: 423.98 MiB)

julia> @btime apply_zero!(K, f, ch) #kc/row_zero
  408.249 ms (4 allocations: 544 bytes)
```

KnutAM:
```julia
  498.754 ms (16 allocations: 423.98 MiB) # master
   665.802 ms (4 allocations: 544 bytes) # kc/row_zero
145.603 ms (4 allocations: 544 bytes) # kam/both_zero
  ```
Perhaps you already discussed why not storing the indices of`nzvals` directly today, but at least it seems faster :)
  
The downside is that creating the constraint handler is slower, 
even when giving the matrix to `close!` for added performance (it needs K internally to find the indices, also possible directly from dh, but was too much work:))
  ```julia
  138.396 ms (120174 allocations: 77.98 MiB) # master
  143.173 ms (120163 allocations: 80.44 MiB) # kc/row_zero
  713.404 ms (160598 allocations: 107.49 MiB) # kam/both_zero (with `close!(ch, K)`)
  1.824 s (160622 allocations: 2.96 GiB) # kam/both_zero (with `close!(ch)`)
  ```
  (For the close!(ch) about 1 GB of allocations and 0.3 s can be saved by filling with a Singelton type instead of Float64)

**Comment by KristofferC** [comment]
> Perhaps you already discussed why not storing the indices ofnzvals directly today, but at least it seems faster :)

Then you are storing approx the equivalent memory of half a stiffness matrix there which maybe is too much.

**Comment by KristofferC** [comment]
It is interesting we get quite different benchmarks. On my desktop with a pretty beefy CPU I get:

```
julia> @btime apply_zero!(K, f, ch) # master
  407.463 ms (16 allocations: 423.98 MiB)

julia> @btime apply_zero!(K, f, ch) # PR
  369.013 ms (4 allocations: 544 bytes)
```

If I slap a `Threads.@threads` on it and run with 8 cores I get

```
julia> @btime apply_zero!(K, f, ch)
  168.278 ms (53 allocations: 4.89 KiB)
```

**Comment by KnutAM** [comment]
> > Perhaps you already discussed why not storing the indices ofnzvals directly today, but at least it seems faster :)
> 
> Then you are storing approx the equivalent memory of half a stiffness matrix there which maybe is too much.

Perhaps I misunderstand here, but wouldn't it be (on average) `2*num_prescribed_dofs*bandwidth` ?
(But it was more for fun, I suppose this will always be dominated by linear solve and your solution is less invasive)

**Comment by KristofferC** [comment]
Yeah, I'm wrong, you only need the mapping for the constrained dofs (of course). Caching this mapping in the constraint handler makes sense to me I think, it would just move a part of the logic in the new function here to there. I guess the only drawback is that we typically do not send in the stiffness matrix to the constraint handler when it is created...

**Comment by koehlerson** [comment]
> It is interesting we get quite different benchmarks. On my desktop with a pretty beefy CPU (i9-12900K) I get:
> 
> ```
> julia> @btime apply_zero!(K, f, ch) # master
>   407.463 ms (16 allocations: 423.98 MiB)
> 
> julia> @btime apply_zero!(K, f, ch) # PR
>   369.013 ms (4 allocations: 544 bytes)
> ```
> 
> If I slap a `Threads.@threads` on it and run with 8 cores I get
> 
> ```
> julia> @btime apply_zero!(K, f, ch)
>   168.278 ms (53 allocations: 4.89 KiB)
> ```

I used my laptop with 12th Gen i5-1240p but I can retry with threads If you want

**Comment by termi-official** [comment]
Probably stupid question, but if the performance of applying (affine) constraints is of concern, why don't we apply the constraints on element-level as in deal.ii? (see e.g. https://www.dealii.org/current/doxygen/deal.II/classAffineConstraints.html#a373fbdacd8c486e675b8d2bff8943192 and https://www.dealii.org/current/doxygen/deal.II/step_27.html#Creatingthesparsitypattern)

**Comment by lijas** [comment]
> Probably stupid question, but if the performance of applying (affine) constraints is of concern, why don't we apply the constraints on element-level as in deal.ii?

Why not both :D 

I think this PR can be merged, even though it seems to be a bit slower on @koehlerson  benchmarks.

**Comment by fredrikekre** [comment]
`master`:
```
@btime apply_zero!($K, $f, $ch)
794.046 ms (16 allocations: 423.98 MiB)
```
`kc/row_zero`:
```
@btime apply_zero!($K, $f, $ch)
643.115 ms (4 allocations: 544 bytes)
```

KristofferC:
I should probably be get rid of the identity hashing thing. It is a bit unclear if that is optimizing for the specific benchmark.

Issue #513, opened 2022-10-09
Document custom quadrature rules


fredrikekre:
From  https://github.com/Ferrite-FEM/Ferrite.jl/pull/482#issuecomment-1272555611 (cc @edljk)

> it could perhaps be an easier way to extend quadrature rules in Ferrite.

It is actually pretty easy to do this already -- you can construct your own quadrature rule by:

```julia
julia> weights = [2.0, 2.0];

julia> points = [Vec{2}((-0.5, 0.0)), Vec{2}((0.5, 0.0))];

julia> qr = QuadratureRule{2,RefTetrahedron,Float64}(weights, points)
QuadratureRule{2, RefTetrahedron, Float64}([2.0, 2.0], Vec{2, Float64}[[-0.5, 0.0], [0.5, 0.0]])
```

It would be nice to i) document this and ii) add a constructor `QuadratureRule{dim,shape}(weights, points)` that doesn't require specifying the element type (`Float64` in the example above).

00krishna:
I can start to work on this.

termi-official:
I think https://github.com/JuliaApproximation/FastGaussQuadrature.jl is a solid foundation for tensor product elements

fredrikekre:
Ferrite used to use that package directly in the past (https://github.com/Ferrite-FEM/Ferrite.jl/pull/136) but it was removed since compile times where extremely bad back then.

PR #514, opened 2022-10-09
Extend 2D Gauss Legendre quadrature for triangles up to order 15

edljk:
Should be more tested. Seams to be ok on the eigenvalues test of order 4&5.


src/Quadrature/gaussquad_tri_table.jl
```diff
@@ -72,6 +72,584 @@ function _get_gauss_tridata(n::Int)
...
+    0.1417072194148800 0.5503529418209990 0.0727579168454201 / 2.0
+    0.0250035347626864 0.7283239045974108 0.0283272425310575 / 2.0
+    0.0095408154002995 0.9236559335875002 0.0094216669637328 / 2.0];
+    elseif n == 11
+    xw=[0.5346110482707582 0.5346110482707582 0.0009270063289607 / 2.0
+    0.3989693029658552 0.3989693029658552 0.0771495349148131 / 2.0
+    0.2033099004312824 0.2033099004312824 0.0593229773807741 / 2.0
+    0.1193509122825813 0.1193509122825813 0.0361845405034180 / 2.0
+    0.0323649481112759 0.0323649481112759 0.0136597310026779 / 2.0
+    0.5346110482707582 -0.0692220965415165 0.0009270063289607 / 2.0
```
fredrikekre, line 127: Some points like this are located outside the reference triangle, is that correct? I have never seen that before.

src/Quadrature/gaussquad_tri_table.jl
```diff
@@ -72,6 +72,584 @@ function _get_gauss_tridata(n::Int)
     0.72849239295540 0.26311282963464 0.02723031417443 / 2.0
     0.26311282963464 0.00839477740996 0.02723031417443 / 2.0
     0.00839477740996 0.72849239295540 0.02723031417443 / 2.0];
+    elseif n == 9
+    xw=[0.3333333333333333 0.3333333333333333 0.0971357962827987 / 2.0
```
KristofferC, line 76: The first line in this file has a reference to where the numbers were taken from. I guess these are not taken from the same place? In that case, they should also have a reference how they got computed or where they were taken from.

edljk: Thank you @KristofferC. I add a reference at the begining of the file. Not sure this is the right way to do it.

@fredrikekre you are perfectly right. This is very suspicious and you can observe the same surprising data points in the original paper of D.A. Dunavant.
I am not an expert of the wild world of quadrature formula that I just discovered.
At this point I do not know exactly what to do. Empirically test the quadrature orders obtained with these data points/weights (when the quadrature points are admissible)? Focus on other quadrature approaches which are used in well established libraries? 

It would be great to have high precision quadrature routines available (Float64, BigFloat..).

edljk: A list of options with references [quadrature formula](https://www.math.unipd.it/~alvise/SETS_CUBATURE_TRIANGLE/rules_triangle.html)

fredrikekre:
`basix` (which is backend for FEniCS) doesn't give any points outside at least, but I couldn't find out the reference or anything like that:
```julia
using PyCall, Printf

basix = pyimport("basix")

for order in 1:11
    n = 2order - 1
    points, weights = basix.make_quadrature(basix.CellType.triangle, n)
    for k in eachindex(weights)
        if k == 1
            println("elseif n == $(order)")
            print("xw=[")
        end
        for l in 1:2
            @printf("%1.16f ", points[k, l])
        end
        @printf("%1.16f", weights[k])
        if k == length(weights)
            print("]")
        end
        println()
    end
end
```

https://docs.fenicsproject.org/basix/main/python/_autosummary/basix.html?highlight=quadraturetype#basix.make_quadrature

edljk:
That seems more reasonable!
No problem detected in my naive tests

```julia
using Ferrite, SpecialFunctions

function test_quadorder(;degmax::Int64 = 10, nbtest::Int64 = 10, 
                         ordermax::Int64 = 20)
    f(x, y, p, q) = x ^ p * y ^ q
    for k = 1:nbtest
        for degree = 1:degmax
            p = rand(0:degree)
            q = degree - p 
            @show (p, q)
            Ipq = gamma(BigFloat("$(p + 1)")) * gamma(BigFloat("$(q + 1)")) / 
                  gamma(BigFloat("$(p + q + 3)"))        
            for order = degree:ordermax 
                xw = Ferrite._get_gauss_tridata(order)
                Inum = sum(f(xw[k, 1], xw[k, 2], p, q) * xw[k, 3] for k = 1:size(xw, 1))
                #@show (p, q, order, abs(Inum - Ipq))
                if false && abs(Inum - Ipq) > 1e-14
                    @show Ipq
                    @show (p, q, order, abs(Inum - Ipq))
                end
                @show (p, q, order, abs(Inum - Ipq))
            end
        end
    end
    nothing
end
```

src/Quadrature/gaussquad_tri_table.jl
```diff
@@ -1,77 +1,2567 @@
...
-    0.89890554336594 0.05054722831703 0.03245849762320 / 2.0
-    0.26311282963464 0.72849239295540 0.02723031417443 / 2.0
-    0.72849239295540 0.00839477740996 0.02723031417443 / 2.0
-    0.00839477740996 0.26311282963464 0.02723031417443 / 2.0
-    0.72849239295540 0.26311282963464 0.02723031417443 / 2.0
-    0.26311282963464 0.00839477740996 0.02723031417443 / 2.0
-    0.00839477740996 0.72849239295540 0.02723031417443 / 2.0];
+    if n == 1
+        xw=[0.3333333333333333 0.3333333333333333 0.5000000000000000]
+    elseif n == 2
```
fredrikekre, line 12: Hmm, didn't notice in my tests that this now requires many more points (6 vs 3), that is quite unfortunate.

termi-official: Fenics utilizes a different type of quadrature points which are not minimal (see e.g. https://www.osti.gov/pages/servlets/purl/1595014).

edljk: .. probably more stable..
Why not keeping Dunavant samplings up to order 8 and use `basix` quadrature points for higher orders?

termi-official: I think the stability highly depends on what kind of function you want to integrate.

fredrikekre: > Why not keeping Dunavant samplings up to order 8 and use basix quadrature points for higher orders?

Yes, let's do this. Then we don't change any existing rules, and if we find better rules for the highers order we can always update later.

fredrikekre:
Can you update the tests: https://github.com/Ferrite-FEM/Ferrite.jl/blob/f6c7e239d30bae7c1276bc26d0c5a33fd4544c18/test/test_quadrules.jl#L36 and add a note at the top of the CHANGELOG.md file?

fredrikekre:
Thanks (I just moved the note under the same `Added` header). Do you think it is useful to have up to order 20? The function and file becomes very large like this :) For example, the highest interpolation we have now is 5, which, if you assemble the mass matrix, requires `2order-1 = 25` -> `order = 13`.

edljk:
Higher order may be useful for instance  when computing mass matrix with a density (also discretized as a fem).
Btw, feel free to cut the file if you feel more comfortable with a lighter data file.

fredrikekre:
Thanks for the PR. I limited it to 15 in the end, and we will add better documentation on how to use your own quadrature rules if you need more fine grained control over it (#513).

Issue #515, opened 2022-10-10
Requesting help with the error "det(J) is not positive" when using custom Delaunay grids


zkk960317:
I need to use a random triangle mesh. But when I define my own mesh, and replace the original mesh in the "_heat_equation.jl_" example, I get the following error in function _reinit!( )_ : 
_ERROR: ArgumentError: det(J) is not positive: det(J) = -1.0736838778944306_

The elements of the Delaunay grid is shown in the figure by Paraview, where green num is cellnum and yellow num is nodenum.
![image](https://user-images.githubusercontent.com/105714520/194836285-0ca61c30-d4c4-4cbe-8100-e2a6561610a1.png)

Following is the code:
```
using Ferrite, SparseArrays
using VoronoiDelaunay
function generate_delaunay2D_grid(dims, esize)
    r0=0.6 #Random factor
    ne = (Int(round(dims[1] / esize)), Int(round(dims[2] / esize)))
    ne_t1 = (ne[1] - 1) * (ne[2] - 1)
    ne_t2 = ne_t1 + 2 * (ne[1] + ne[2])
    rdx = rand(ne[1] - 1, ne[2] - 1) .* esize .* r0 .- esize * r0 * 0.5
    rdy = rand(ne[1] - 1, ne[2] - 1) .* esize .* r0 .- esize * r0 * 0.5
    for i = 1:ne[1]-1, j = 1:ne[2]-1
        rdx[i, j] += i * esize
        rdy[i, j] += j * esize
    end
    width = max_coord - min_coord
    ratio = 1 / max(dims[1], dims[2]) * width
    rdx = rdx * ratio
    rdy = rdy * ratio
    nodes = Point2D[Point(min_coord + rdx[i], min_coord + rdy[i]) for i in 1:ne_t1]
    ### add points at boundaries
    nodes = [nodes; [Point2D(min_coord, min_coord + esize * i * ratio) for i = 0:ne[2]];
        [Point2D(max_coord, min_coord + esize * i * ratio) for i = 0:ne[2]];
        [Point2D(min_coord + esize * i * ratio, min_coord) for i = 1:ne[1]-1];
        [Point2D(min_coord + esize * i * ratio, min_coord + width * dims[2] / dims[1]) for i = 1:ne[1]-1]]
    tess = DelaunayTessellation()
    sizehint!(tess, ne_t2)
    @time push!(tess, nodes)
[...]

PR #517, opened 2022-10-10
add a note in the grid section about clockwise ordering

koehlerson:
something like this maybe? https://github.com/Ferrite-FEM/Ferrite.jl/issues/515

### DIFF
--- a/docs/src/manual/grid.md
+++ b/docs/src/manual/grid.md
@@ -43,7 +43,13 @@ FerriteMeshParser.get_ferrite_grid
 ```
 
 If you are missing the translation of an Abaqus element that is equivalent to a `Ferrite.Cell`,
-consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request. 
+consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request.
+
+!!! note "Common pitfalls for reading a custom mesh"
+    In case you write your own custom mesh reader, take care of the Ferrite.jl's anti-clockwise cell node ordering.
+    For an illustrative example consider the quadrilateral node ordering in the [`Grid` Datastructure](@ref) example.
+    If you encounter the error message `det(J) is not positive`, but your mesh looks fine in a vtk export, it is very likely
+    that the node ordering is clockwise instead of anti-clockwise.
 
 ## `Grid` Datastructure
 


termi-official:
Can the error message contain a link to this?

KnutAM:
Nice! 
At least I haven't found it, so it would be nice to document Ferrite's ordering scheme. 
Then, as @termi-official says but instead refer to the ordering scheme in the docs in the error message. 
And then I would say that the note is no longer necessary. But until then, this would be good to have!

docs/src/manual/grid.md
```diff
@@ -43,7 +43,13 @@ FerriteMeshParser.get_ferrite_grid
 ```
 
 If you are missing the translation of an Abaqus element that is equivalent to a `Ferrite.Cell`,
-consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request. 
+consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request.
+
+!!! note "Common pitfalls for reading a custom mesh"
+    In case you write your own custom mesh reader, take care of the Ferrite.jl's anti-clockwise cell node ordering.
```
KnutAM, line 49:
```suggestion
    In case you read custom cells or write your own custom mesh reader, take care of the Ferrite.jl's anti-clockwise cell node ordering.
```

koehlerson:
> Nice! At least I haven't found it, so it would be nice to document Ferrite's ordering scheme. 

In the docs the quadrilateral ordering is introduced in anti-clockwise ordering, but I agree that could be written down explicitly with a reference to `reference_coordinates`

fredrikekre:
Would be good with some plots for every element type (and interpolation) with numbering. Perhaps this can be done with Makie or FerriteViz?

koehlerson:
there is at least an issue about it :D https://github.com/Ferrite-FEM/FerriteViz.jl/issues/15

koehlerson:
grafik
So I would do something like this? Will include the local face and edge number and maybe I can make some markers for the dofs, but I'm not sure about it, will probably be a bit harder

fredrikekre:
That looks great. I don't think dof markers are necessary, they follow the same order

termi-official:
While they follow the same ordering the exact dof numbering can be different if we have different interpolations for geometry and the solution. Also note that dofs can also be associated to higher order entities than only nodes.

termi-official:
Maybe as followup we can provide an example to find the flipped elements in FerriteVis.jl. Should be very straight forward (i.e. plot L2 element of order 0 with value of det(J)).

termi-official:
AFAIK we currently only have a **guarantee** that elements are not flipped if we utilize the provided grid generators, yes. But I think it might be helpful to plot the determinants and hence detect flipped elements.

docs/src/manual/grid.md
```diff
@@ -43,7 +43,13 @@ FerriteMeshParser.get_ferrite_grid
 ```
 
 If you are missing the translation of an Abaqus element that is equivalent to a `Ferrite.Cell`,
-consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request. 
+consider to open an [issue](https://github.com/Ferrite-FEM/FerriteMeshParser.jl/issues/new) or a pull request.
+
```
KnutAM, line 47:
```suggestion
### Reading custom meshes
```
When I checked the docs, it looks like this only applies to FerriteMeshParser otherwise... 
Perhaps you have a better idea for the header naming though...

koehlerson: Either yours or double the `pitfalls` , so

```suggestion
### Pitfalls when reading custom meshes
```

docs/src/manual/grid.md
```diff
@@ -71,7 +77,48 @@ julia> cells = [
...
-Additionally, the data structure `Grid` can hold node-, face- and cellsets. 
+Each `<: AbstractCell` defines, just as in the example above, the nodes in anti-clockwise ordering.
+This is a convention in Ferrite, which is used to ensure that the transformation from reference space to physical space yields a positive determinant of the Jacobi.
+The function `reference_coordinates` (in `src/interpolations.jl`) defines the expected node ordering by the index of the returned `Vector{Vec{dim,T}}` and is dispatched for each `Interpolation`.
+### Anti-Clockwise Ordering
+Ferrite.jl's ordering w.r.t. nodes, edges and faces of each `<:AbstractCell` or `<:Interpolation`, respectively, can be visualized by `elementinfo` of [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl) with the Makie backend of your choice.
+Below, the anti-clockwise ordering of nodes and faces are shown for different two-dimensional interpolations.
+
+```@example ordering
+import CairoMakie
```
KnutAM, line 88: Would it make sense to generate this figure statically, to avoid depending on Makie to build the docs.
Looks like the CI-time for docs went up by 3 minutes...

docs/src/manual/grid.md
KnutAM: Yours is more concrete and better!

KnutAM:
LGTM otherwise, but I don't feel experienced enough to press the approve button :)

docs/src/manual/grid.md
koehlerson: I included it not statically to have the ability later to create more figures on the fly, since they bloat quite a bit the git repository. However, if the majority wants them statically tracked by git, I can include it. I thought it's maybe a better approach for the future, since its likely that there will be more and more small pictures in the docs whether its about examples or general docs. What do you prefer @fredrikekre

lijas:
Should we add a note about note ordering here aswell, and maybe a link to the docs? :P 

https://github.com/Ferrite-FEM/Ferrite.jl/blob/2f8655970f29a7d37d41a5e721cdf10d8b21ddbc/src/FEValues/common_values.jl#L17

docs/src/manual/grid.md
KnutAM: I like Fredrik's approach of the assets folder in the ghpages branch...
That shouldn't bloat the git history...

KnutAM:
I think it is a nice PR that would be nice to get merged! I've closed my previous concern above. 

> Should we add a note about note ordering here aswell, and maybe a link to the docs? :P

My suggestion would be to change: 
`"det(J) is not positive: det(J) = $(detJ)"`
to 
`"det(J) is not positive: det(J) = $(detJ) (hint: check node ordering)"`

A merge with master would be needed as well

@koehlerson: Let me know if you want me to do anything!

koehlerson:
I can finalize it, I wasn't sure how to deal with the `gh-page` branch upload, so if this should be merged with a ref to the gh-pages upload, let me know

KnutAM:
> I wasn't sure how to deal with the gh-page branch upload, so if this should be merged with a ref to the gh-pages upload, let me know

I think it can be merged in any way as I think including FerriteViz in the doc is possible now with Julia 1.9

Previously, I've uploaded figures to the [gh-branch assets folder](https://github.com/Ferrite-FEM/Ferrite.jl/tree/gh-pages/assets) and added the figures to the [list of downloads](https://github.com/Ferrite-FEM/Ferrite.jl/blob/ee6e26d3d0bf3fac62b6d77ebe7eaeb07cf8b51a/docs/download_resources.jl#L11)

termi-official:
The JSServe bug does not affect this PR, because on the current FerrizeViz release we have an older Makie version, right?

koehlerson:
> The JSServe bug does not affect this PR, because on the current FerriteViz release we have an older Makie version, right?

No it's not related, since it uses CairoMakie and you only need JSServe for WGLMakie and interactivity stuff of the plot

termi-official:
Ah, you are right. Thanks.

koehlerson:
need https://github.com/Ferrite-FEM/FerriteViz.jl/pull/78 before rebasing and merging

Issue #520, opened 2022-10-15
inconsistent higher order node between RefCube and RefShape


koehlerson:
the ordering seems not consistent between second order tet and cube

https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/interpolations.jl#L548
https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/interpolations.jl#L460

Tetrahedron goes first order - second order - first order
Cube edge is defined as first order - first order - second order

kimauth:
This is only true for 3D, right? At least the 2D quadratic tetrahedron goes first order - first order - second order: https://github.com/Ferrite-FEM/Ferrite.jl/blob/084a7c9d967047efa986dcb3ac0668033e023084/src/interpolations.jl#L310

koehlerson:
yes, this is only in 3D inconsistent :)

PR #523, opened 2022-10-18
resolve inconsistency between refcube and refshape interpolation in 3D

koehlerson:
https://github.com/Ferrite-FEM/Ferrite.jl/issues/520


kimauth (approved):

termi-official:
Can we include some regression test to catch future problems with wrong dispatches on these functions?

koehlerson:
second order Hexahedron has inconsistent edges as well :(
```
highorderedge = (1, 2, 9)
firstorderedge = (1, 2)
highorderedge = (2, 3, 10)
firstorderedge = (2, 3)
highorderedge = (3, 4, 11)
firstorderedge = (3, 4)
highorderedge = (4, 1, 12)
firstorderedge = (4, 1)
highorderedge = (5, 6, 13)
firstorderedge = (1, 5)
highorderedge = (6, 7, 14)
firstorderedge = (2, 6)
highorderedge = (7, 8, 15)
firstorderedge = (3, 7)
highorderedge = (8, 5, 16)
firstorderedge = (4, 8)
highorderedge = (1, 5, 17)
firstorderedge = (5, 6)
highorderedge = (2, 6, 18)
firstorderedge = (6, 7)
highorderedge = (3, 7, 19)
firstorderedge = (7, 8)
highorderedge = (4, 8, 20)
firstorderedge = (8, 5)
```

I guess the points match but are numbered differently

test/test_interpolations.jl
```diff
@@ -80,6 +80,23 @@ for interpolation in (
             @test __outward_normal(coords, facenodes) ≈ normal
         end
     end
+    # regression for https://github.com/Ferrite-FEM/Ferrite.jl/issues/520
+    interpolation_type = typeof(interpolation).name.wrapper
```
koehlerson, line 84: is there a better way to extract the type of the interpolation, e.g. `Lagrange`, `Serendipity` etc.?

fredrikekre: You can check `interpolation isa Lagrange` instead

koehlerson: ideally this should be checked for non-lagrangian interpolations, too. Since in the codebase  `DiscontinuousLagrange`, `CrouzeixRaviart`, `BubbleEnrichedLagrange` is already introduced

termi-official: Also I will include H(div) and H(curl) interpolation soon, because our group need them.

lijas:
I dont remember where I noticed that... but for hexahedron it looks like we use the same numbering
https://vtk.org/doc/nightly/html/classvtkQuadraticHexahedron.html

koehlerson:
> I dont remember where I noticed that... but for hexahedron it looks like we use the same numbering https://vtk.org/doc/nightly/html/classvtkQuadraticHexahedron.html

I had this conversation in mind: https://github.com/Ferrite-FEM/Ferrite.jl/pull/353#issuecomment-888338148

lijas:
Actually, it seems like they order their edges for hexahedron differently:
VTK: (0,1), (1,2), (2,3), (3,0), (4,5), (5,6), (6,7), (7,4), (0,4), (1,5), (2,6), (3,7) (0-based indexing)
Ferrite. (1,2), (2,3), (3,4), (4,1), (1,5), (2,6), (3,7), (4,8), (5,6), (6,7), (7,8), (8,5)

So they number their vertical edges 9,10,11, and 12, and for us they are 5,6,7 and 8

koehlerson:
can you tell me where we need to change code in the writevtk file? I'm also fine if you want to push the changes on this branch in case you find some time in the coming days :)

lijas:
It should matter if we output the results on a quadratic hexahedron mesh, right?

I guess we should change the ordering of the edges 
here: https://github.com/Ferrite-FEM/Ferrite.jl/blob/f3057f380d0452968f7414fad6fbe0ad9cc2117e/src/Grid/grid.jl#L732
and here: https://github.com/Ferrite-FEM/Ferrite.jl/blob/f3057f380d0452968f7414fad6fbe0ad9cc2117e/src/interpolations.jl#L501
to be consistent with that of VTK.

Or we can implement nodes_to_vtkorder for QuadraticHexahedrons.

koehlerson:
ye but `nodes_to_vtkorder` only gives the nodes in the vtk order and not the edges. So, we can still pass a hexahedron with matching node order but with incorrect edge order? I'm also okay to revert the last commit and change the serendipity edge and linear hexahedron edge definition the be align with vtk. 

What do you think @fredrikekre @KristofferC @kimauth ?

termi-official:
We really should decouple the vtk representation from our internal representation. If I find some time I will make a better VTK export based on interpolation of data to the nodes instead of some reordering magic (which does not work e.g. for CR elements).

lijas:
> So, we can still pass a hexahedron with matching node order but with incorrect edge order?

Yes, this is true, I did not think of that. So our edge ordering could be different I guess :thinking:

koehlerson:
> > So, we can still pass a hexahedron with matching node order but with incorrect edge order?
> 
> Yes, this is true, I did not think of that. So our edge ordering can be different I guess thinking

Actually it shouldn't matter that we order them differently, because apparently all of us use the whole time the linear hexahedron export and the linear hexahedron shares the "wrong" edge definition.

test/test_interpolations.jl
KnutAM: > is there a better way to extract the type of the interpolation, e.g. Lagrange, Serendipity etc.?

@koehlerson : I noted down this trick with ref to this thread (used it in testing as well), but recently I saw 
```julia
@inline basetypeof(x::T) where T = basetypeof(T)
@generated function basetypeof(::Type{T}) where T
    if T isa Union
        T
    else
        getfield(parentmodule(T), nameof(T))
    end
end
```
from https://github.com/rafaqz/DimensionalData.jl/blob/3087e3e445ddb57e3f02630d6077cff585a132ac/src/LookupArrays/utils.jl#L73-L80

PR #527, opened 2022-10-21
Add doc for option all in add*set!

KnutAM:


src/Grid/grid.jl
```diff
@@ -555,7 +555,9 @@ _warn_emptyset(set, name) = length(set) == 0 && @warn("no entities added to the
 
 Adds a cellset to the grid with key `name`.
 Cellsets are typically used to define subdomains of the problem, e.g. two materials in the computational domain.
-The `MixedDofHandler` can construct different fields which live not on the whole domain, but rather on a cellset. 
+The `MixedDofHandler` can construct different fields which live not on the whole domain, but rather on a cellset.
+`all=true` (default) implies that `f(x)` must return `true` for all nodal coordinates `x` in the cell if the cell
```
kimauth, line 559: Saying that `all=true` is the default seems a little redundant to me, as that is clear from the function signature.

kimauth (approved):
Thanks, this is a good point!

src/Grid/grid.jl
KnutAM: True!

PR #528, opened 2022-10-25
Local condensation of constraints

fredrikekre:
This patch adds functionality for applying constraints locally on the
element matrix/vector before assembling them into the global
matrix/vector. When doing local condensation there is no need to apply
constraints on the global system after the assembly. The main motivation
for this functionality is when using "write-only" global arrays (for
example some array type in an external library such as HYPRE) where it
is expensive or impractical to modify the array after assembling.

Local constraint condensation can be done with `apply_local!` or
`apply_assemble!`:

    apply_local!(local_matrix, local_vector, global_dofs, ch)

`apply_local!` can be used when all constraints are local to the element
(e.g. Dirichlet boundary constraints) since the constraints do not
couple to dofs outside the dofs of the element itself.

    apply_assemble!(assembler, global_dofs, local_matrix, local_vector)

`apply_assemble!` condensates constraints and assembles the result into
the global system, i.e. performs `apply_local!` + `assemble!`. The advantage
with `apply_assemble!` is that it handles general constraints since it has
access to the global matrix/vector and can write to entries outside the
local matrix when necessary.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1634,3 +1634,178 @@ function __check_periodic_faces_f(grid::Grid, fi::FaceIndex, fj::FaceIndex, xmi,
 
     return PeriodicFacePair(fi, fj, node_rot, mirror)
 end
+
+
+######################################
+## Local application of constraints ##
+######################################
+
+"""
+    apply_local!(
+        local_matrix::AbstractMatrix, local_vector::AbstractVector,
+        global_dofs::AbstractVector, ch::ConstraintHandler;
+        apply_zero::Bool = false
+    )
+
+Similar to [`apply!`](@ref) but perform condensation of constrained degrees-of-freedom
+locally in `local_matrix` and `local_vector` *before* they are to be assembled into the
+global system.
+
+When the keyword argument `apply_zero` is `true` all inhomogeneities are set to `0` (cf.
+[`apply!`](@ref) vs [`apply_zero!`](@ref)).
+
+This method can only be used if all constraints are "local", i.e. no constraint couples with
+dofs outside of the element dofs (`global_dofs`) since condensation of such constraints
+requires writing to entries in the global matrix/vector. For such a case,
+[`apply_assemble!`](@ref) can be used instead.
+
+Note that this method is destructive since it, by definition, modifies `local_matrix` and
+`local_vector`.
+"""
+function apply_local!(local_matrix::AbstractMatrix, local_vector::AbstractVector,
+                      global_dofs::AbstractVector, ch::ConstraintHandler;
+                      apply_zero::Bool = false)
+    return _apply_local!(local_matrix, local_vector, global_dofs, ch, apply_zero,
+                         #=global_matrix=# nothing, #=global_vector=# nothing)
+end
+
+# Element local application of boundary conditions. Global matrix and vectors are necessary
+# if there are affine constraints that connect dofs from different elements.
+function _apply_local!(local_matrix::AbstractMatrix, local_vector::AbstractVector,
+                       global_dofs::AbstractVector, ch::ConstraintHandler, apply_zero::Bool,
+                       global_matrix, global_vector)
+    # TODO: With apply_zero it shouldn't be required to pass the vector.
+    length(global_dofs) == size(local_matrix, 1) == size(local_matrix, 2) == length(local_vector) || error("?")
+    # First pass over the dofs check whether there are any constrained dofs at all
+    has_constraints = false
+    has_nontrivial_affine_constraints = false
+    # 1. Adjust local vector
+    @inbounds for (local_dof, global_dof) in pairs(global_dofs)
+        # Check if this dof is constrained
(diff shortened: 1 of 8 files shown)


termi-official:
That would be really great for the distributed assembly!

src/assembler.jl
```diff
@@ -199,6 +199,20 @@ end
 end
 
 
+## assemble! with local condensation ##
+
+# TODO: assemble_zero! instead of last Bool arg?
+# TODO: Call this apply_assemble! instead?
```
fredrikekre, line 205: Thoughts? I don't see any real reason for making this a method of `assemble!`, in particular it isn't so obvious that something quite different is happening under the hood when passing the constraint handler as the second argument. Might as well make it clear by using a different function name altogether.

termi-official: `assemble_eliminated!` or `assemble_constrained!` maybe?

fredrikekre: It does `apply!` + `assemble!` which is what these operations are called in isolation in Ferrite so :shrug:

KnutAM: 👍 for not using `assemble!` for the combined function.
But wouldn't it be clearer to first call `apply_local*!` (mirroring the global versions) and then `assemble!`, instead of having one function that does both?

fredrikekre: The PR includes `apply_local!` too, but that only works if all constraints are local (e.g. Dirichlet), otherwise you need to write to the global matrix at positions outside the local matrix.

KnutAM: Do you mean that exporting this call might be a bit misleading to export as it also modifies `Kglobal` and `fglobal`:
`_apply_local!(local_matrix, local_vector, global_dofs, ch, apply_zero, Kglobal, fglobal)`?
(I suppose it would be simplified to `apply[_zero]_local!(local_matrix, local_vector, global_dofs, ch, assembler)`, but it still is not purely local?)

fredrikekre: I feel it is a bit weird to have parts of the assembly happen in `apply_local!` and parts in the following `assemble!` call.

KnutAM: Right, sorry! I didn't think of what happens in `apply_local!` as a type of assembly, but it is and then it makes perfect sense to combine it into one! 
I would then vote for removing the bool + using the `apply_assemble!`/`apply_zero_assemble` versions.

src/Dofs/ConstraintHandler.jl
```diff
@@ -1587,3 +1587,186 @@ function __check_periodic_faces_f(grid::Grid, fi::FaceIndex, fj::FaceIndex, xmi,
...
+        v = ch.inhomogeneities[pdofs_index]
+        if !apply_zero && v != 0
+            for j in axes(local_matrix, 1)
+                local_vector[j] -= v * local_matrix[j, local_dof]
+            end
+        end
+        # Check if this is an affine constraint
+        if has_nontrivial_affine_constraints || (
+           coeffs = ch.dofcoefficients[pdofs_index];
+           !(coeffs === nothing || isempty(coeffs))
```
lijas, line 1651: Difficult to read when everything is in the if statement? :P 
Also, maybe we should check that coeffs is not empty in the close!(ch), so we don't have to check for that here?

src/assembler.jl
```diff
@@ -199,6 +199,43 @@ end
...
+This is similar to using [`apply_local!`](@ref) followed by [`assemble!`](@ref) with the
+advantage that non-local constraints can be handled, since this method can write to entries
+of the global matrix and vector outside of the indices in `global_dofs`.
+
+When the keyword argument `apply_zero` is `true` all inhomogeneities are set to `0` (cf.
+[`apply!`](@ref) vs [`apply_zero!`](@ref)).
+
+Note that this method is destructive since it modifies `local_matrix` and `local_vector`.
+"""
+function apply_assemble!(
```
lijas, line 224: I think `assemble!(a, ch, ..)` is a better name. It clear that it will assemble and take the constraints into account. `apply_assemble!` sounds like it could do something else?

lijas (approved):
Nice. It seems to follow the same structure as in apply!(K, f, ch).

I guess this will this work for any other `Assembler{T}`?

src/assembler.jl
termi-official: I think it could be mistaken for some matrix-free SpMV under constraints.

KristofferC:
Is there any guidance to give when one should use the local vs the global version?

termi-official:
I think a few benchmarks can be used to compare the performance of both approaches and on this basis we can give recommendations.

src/assembler.jl
fredrikekre: But then `apply!` could be mistaken for matrix-free SpMV without constraints :p I think nobody would use this function without reading the docs, and I think it makes sense to reuse the same verbs we already have.

fredrikekre:
For the heat equation example with 2000 x 2000 elements:
```
# assemble! + apply!
  1.821 s (4000013 allocations: 152.62 MiB)

# apply_local! + assemble!
1.618 s (4000013 allocations: 152.62 MiB)
```

PR #529, opened 2022-10-26
Unify Dirichlets and affine constraints

lijas:
Currently, it is possible for Dirichlet and affine constraints to lock the same dof. This PR unifies Dirichlets and Affine constraints and makes it impossible for a dof to be locked by both affine and Dirichlet constraints.

If two conflicting constraints are added (prescribing the same dof), the most recently added constraint will be used.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -54,6 +54,7 @@ function __to_components(c)
     return components
 end
 
+const DofCoefficients{T} = Vector{Pair{Int,T}}
 """
     AffineConstraint(constrained_dofs::Int, master_dofs::Vector{Int}, coeffs::Vector{T}, b::T) where T
 
@@ -61,7 +62,7 @@ Define an affine/linear constraint to constrain dofs of the form `u_i = ∑(u[j]
 """
 struct AffineConstraint{T}
     constrained_dof::Int
-    entries::Vector{Pair{Int, T}} # masterdofs and factors
+    entries::DofCoefficients{T} # masterdofs and factors
     b::T # inhomogeneity
 end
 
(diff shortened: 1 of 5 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -239,17 +236,32 @@ end
...
-        (ac.constrained_dof == newac.constrained_dof) &&
-            error("Constraint already exist for dof $(ac.constrained_dof)")
-        any(x -> x.first == newac.constrained_dof, ac.entries) &&
-            error("New constrained dof $(newac.constrained_dof) is already used as a master dof.")
-    end
-    push!(ch.acs, newac)
-    push!(ch.prescribed_dofs, newac.constrained_dof)
-    push!(ch.inhomogeneities, newac.b)
+function add!(ch::ConstraintHandler, ac::AffineConstraint)
+    add_prescribed_dof!(ch, ac.constrained_dof, ac.b, isempty(ac.entries) ? nothing : ac.entries)
```
fredrikekre, line 240: Here we lose information about this coming from an affine constraint. That is probably mostly fine, but in `update!` we use this to see if the inhomogeneity should be updated or not...
```suggestion
    add_prescribed_dof!(ch, ac.constrained_dof, ac.b, ac.entries)
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -239,17 +236,35 @@ end
...
-            error("New constrained dof $(newac.constrained_dof) is already used as a master dof.")
-    end
-    push!(ch.acs, newac)
-    push!(ch.prescribed_dofs, newac.constrained_dof)
-    push!(ch.inhomogeneities, newac.b)
+function add!(ch::ConstraintHandler, ac::AffineConstraint)
+    # TODO: Would be nice to pass nothing if ac.entries is empty, but then we lose the fact
+    #       that this constraint is an AffineConstraint which is currently needed in update!
+    #       in order to not update inhomogeneities for affine constraints
+    add_prescribed_dof!(ch, ac.constrained_dof, ac.b, #=isempty(ac.entries) ? nothing : =# ac.entries)
```
lijas, line 243: If ac.entries is empty then it is a pure dirichlet, so note sure such a AffineConstraint should/would ever be added

lijas (commented):
I approve

PR #531, opened 2022-10-28
overview of examples / topology optimization example

blaszm:
I tried to summarize the main topics of each example / tutorial in two to three short sentences.
Keywords should help to make it easier to navigate between the examples, though my choices are somewhat random / unrelated.
Some examples are written in passive voice, some use first person (“we”), maybe a unique style should be employed? Keywords are included in some examples (cf. example 5, 14), while missing in most others. I looked up the tutorial section of [deal.ii](https://www.dealii.org/current/doxygen/deal.II/Tutorial.html). I feel the graph is too complicated for our purposes but I tried to adopt the different lists.

Note that some parts are still under construction (missing links, part 2 with lists sorting the examples by topic)

Please feel free to comment on any suggestions or improvements.


KristofferC:
Good idea!

docs/src/overview.md
```diff
@@ -0,0 +1,102 @@
...
+
+ This program combines ideas from the examples Incompressible Elasticity and Hyperelasticity to construct a mixed element solving three-dimensional displacement-pressure equations.
+	
+ Keywords: Non-linear problem, hyperelasticity, finite strain, large deformations, Newton's method, automatic differentiation, coupled problem, mixed elements, displacement-pressure equation, incompressibility
+
+12.	[Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)  
+
+ The weak form of the semi-discretized incompressible Navier-Stokes equations are derived from the strong form and implemented in the finite element code. Then, the time-dependent solution is calculated by using the DifferentialEquations.jl package. For this purpose, the PDE is required in a specific form.  
+	
+ Keywords: Fluid dynamics, weak and strong form, solver, incompressibility, time-dependent problem
```
termi-official, line 84: Can we move this into a separate list below the "pure" Ferrite examples? This is technically an example of how to integrate Ferrite.jl with DifferentialEquations.jl and qualitatively different from the other examples.

blaszm: Yes!

blaszm: changed in current version

blaszm:
I also added now an example for topology optimization, which is also included in the overview. Big thanks to Dustin and Maxi for the help! Please feel free to review/propose changes.

termi-official:
Can you compress the gif? I usually opt for https://gif.ski/ .

Also @fredrikekre how have we settled regarding binary files? git-lfs or ordinary git?

DRollin (approved):

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+mutable struct MaterialState{T, S <: AbstractArray{SymmetricTensor{2, 2, T}, 1}} 
+    χ::T ## density
+    ε::S ## strain in each quadrature point
+end
+
+function MaterialState(ρ, n_qp)
+    return MaterialState(ρ, Array{SymmetricTensor{2,2,Float64},1}(undef, n_qp))
+end
+
+function UpdateMaterialStates!(χn1, states, dh)
```
DRollin, line 157: This is capitalized like a constructor, but it is a regular function.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+        state.χ = χn1[cellid(element)]
+    end
+end
+#md nothing # hide
+
+# Next, we define a function to calculate the driving forces for all elements.
+# For this purpose, we iterate through all elements and calculate the average strain in each
+# element. Then, we compute the driving force from the formula introduced at the beginning.
+
+function DrivingForces(states, mp, dh)
```
DRollin, line 168: This is capitalized like a constructor, but it is a regular function.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+end
+#md nothing # hide
+
+# Now we calculate the Laplacian. For this purpose, we will later create the grid topology of 
+# the grid by using the function `ExclusiveTopology`. Then we iterate through each face of each element,
+# obtaining the neighboring element by using the `getneighborhood` function. For boundary faces,
+# the function call will return an empty object. In that case we use the dictionary to instead find the opposite
+# face, as discussed in the introduction. Then, the approximation of the Laplacian reduces to the sum below. 
+
+function Laplacian(dh, topology, χn, Δh)
```
DRollin, line 187: This is capitalized like a constructor, but it is a regular function.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+
+    Δχ = pΨ + mp.β*p_Ω*∇²χ
+    
+    χn1 = Bisection(χn, Δχ, ρ, mp.η*p_Ω, mp.χ_min)
+    
+    return χn1
+end
+#md nothing # hide
+    
+# Now, we move on the Finite Element part of the program. We use the following function to assemble our linear system.
```
DRollin, line 266:
```suggestion
# Now, we move on to the Finite Element part of the program. We use the following function to assemble our linear system.
```

docs/src/literate/topology_optimization.jl
DRollin, line 157: This is capitalized like a constructor, but it is a regular function.
```suggestion
function update_material_states!(χn1, states, dh)
```

docs/src/literate/topology_optimization.jl
DRollin, line 168:
```suggestion
function compute_driving_forces(states, mp, dh)
```

docs/src/literate/topology_optimization.jl
DRollin, line 187:
```suggestion
function approximate_laplacian(dh, topology, χn, Δh)
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+        
+        if(ρ_trial > ρ)
+            λ_lower = λ_trial
+        elseif(ρ_trial < ρ)
+            λ_upper = λ_trial
+        end
+        λ_trial = 1/2*(λ_upper+λ_lower)
+    end
+
+    return χ_trial
```
DRollin, line 244: The comment before the function states that λ is being calculated, but in the end χ_trial is returned.
Later, the function is used to compute χn1, so I guess that this is not a mistake.
However, to me it is a bit confusing, maybe a comment on that could be useful.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+
+    return ∇²χ
+end
+#md nothing # hide
+
+# To ensure that the mass is kept constant, we have to calculate the constraint
+# parameter $\lambda$, which we do via the bisection method. We repeat the calculation
+# until the difference between the current density and target density nearly vanishes. 
+
+function Bisection(χn, Δχ, ρ, ηs, χ_min)
```
DRollin, line 215:
```suggestion
function compute_χn1(χn, Δχ, ρ, ηs, χ_min) 
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+    return χ_trial
+end
+#md nothing # hide
+
+# Finally, we put everything together to update the density. We weight our numerical parameters 
+# $\eta$ and $\beta$ by the average driving forces to make the choice of the parameters independent
+# from the specific problem. 
+
+function UpdateDensity(dh, states, mp, ρ, topology, Δh)   
+    pΨ, χn = DrivingForces(states, mp, dh) ## driving forces, old density field
```
DRollin, line 253:
```suggestion
    pΨ, χn = compute_driving_forces(states, mp, dh) ## driving forces, old density field
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+    ∇²χ = Laplacian(dh, topology, χn, Δh) ## Laplacian
```
DRollin, line 254:
```suggestion
    ∇²χ = approximate_laplacian(dh, topology, χn, Δh) ## Laplacian
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+
+    p_Ω = sum(pΨ)/length(pΨ) ## average driving force
+
+    Δχ = pΨ + mp.β*p_Ω*∇²χ
+    
+    χn1 = Bisection(χn, Δχ, ρ, mp.η*p_Ω, mp.χ_min)
```
DRollin, line 260:
```suggestion
    χn1 = compute_χn1(χn, Δχ, ρ, mp.η*p_Ω, mp.χ_min)
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+
+    return χ_trial
+end
+#md nothing # hide
+
+# Finally, we put everything together to update the density. We weight our numerical parameters 
+# $\eta$ and $\beta$ by the average driving forces to make the choice of the parameters independent
+# from the specific problem. 
+
+function UpdateDensity(dh, states, mp, ρ, topology, Δh)
```
DRollin, line 252:
```suggestion
function update_density(dh, states, mp, ρ, topology, Δh)   
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+        assemble!(assembler, celldofs(element), re, Ke)
+    end
+
+    return K, r
+end
+#md nothing # hide
+
+# The element routine is used to calculate the elementwise stiffness matrix and the residual. In contrast to a purely
+# elastomechanic problem, for topology optimization we additionally use our material state to receive the density value of
+# the element and to store the strain of each quadrature point.
```
DRollin, line 293:
```suggestion
# the element and to store the strain at each quadrature point.
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+                break
+            else
+                conv = :true
+            end
+        else
+            conv = :false
+        end
+  
+        ## update density
+        χ = UpdateDensity(dh, states, mp, ρ, topology, Δh)
```
DRollin, line 447:
```suggestion
        χ = update_density(dh, states, mp, ρ, topology, Δh)
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+        
+        ## update old displacement, density and compliance
+        un .= u
+        Δu .= 0.0
+        UpdateMaterialStates!(χ, states, dh)
```
DRollin, line 452:
```suggestion
        update_material_states!(χ, states, dh)
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+On this page, you find a documentation for each example currently available for `Ferrite`. The collection contains programs displaying different aspects and techniques of the `Ferrite` toolbox. 
+The programs are sorted as
+
+* A complete list shortly summarizing the topic, what they teach and which keywords related to the task they contain
+* Grouped by topic 
+
+## List of examples (chronological order)
+1.	[Heat Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/heat_equation/)  
+
+ This example is the easiest way for getting started. The heat equation is solved on a unit square. For this purpose, a program is set up containing all the important aspects of finite element computation, including a short introduction to weak and strong forms, trial and test functions, degrees of freedom, boundary conditions, element assembly, global assembly, solving the system and visualizing the results.
```
DRollin, line 16:
```suggestion
 This example presents the easiest way to get started. The heat equation is solved on a unit square. For this purpose, a program is set up containing the core aspects of finite element computation: weak and strong forms, trial and test functions, degrees of freedom, boundary conditions, element routine, global assembly, solving the system and visualizing the results.
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+ This example is the easiest way for getting started. The heat equation is solved on a unit square. For this purpose, a program is set up containing all the important aspects of finite element computation, including a short introduction to weak and strong forms, trial and test functions, degrees of freedom, boundary conditions, element assembly, global assembly, solving the system and visualizing the results.  
+ 
+ Keywords: Fundamentals, heat equation, weak and strong form, Dirichlet boundary condition, assembly
+
+2.	[Postprocessing](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/postprocessing/)  
+
+ Based on example 1, visualization of the flux variable is done via the L2Projector. Additionally, the point evaluation along a line is shown.  
+	
+ Keywords: Fundamentals, postprocessing, heat equation, flux, L2projection, point Evaluation
```
DRollin, line 24:
```suggestion
 Keywords: Fundamentals, postprocessing, heat equation, flux, L2-projection, point evaluation
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+3.	[Helmholtz equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/helmholtz/)  
+
+ The Helmholtz equation is solved on a unit square. Dirichlet and Neumann boundary conditions are applied at different parts of the boundary. A known analytical solution is approximately reproduced to verify the finite element solution.  
+	
+ Keywords: Fundamentals, Helmholtz equation, Dirichlet boundary condition, Neumann boundary condition, verification
```
DRollin, line 30:
```suggestion
 Keywords: Fundamentals, Helmholtz equation, Dirichlet boundary condition, Neumann boundary condition, verification, boundary integrals
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+4.	[Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/)  
+
+ A mixed element for solving unidirectional displacement-pressure coupling is constructed. The solution for nearly incompressible materials is compared for different interpolations.  
+	
+ Keywords: Mixed elements, unidirectional coupling, displacement-pressure equation, incompressibility, stability
```
DRollin, line 36:
```suggestion
 Keywords: Mixed elements, unidirectional coupling, displacement-pressure equation, incompressibility, stability, exploiting system symmetry
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+5.	[Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/)  
+
+ A hyperelastic material model is solved in a finite strain setting. Stress and material tangent are recovered by using automatic differentiation. Newton’s method is used to iteratively solve the resulting non-linear system.
```
DRollin, line 40:
```suggestion
 A hyperelastic material model is solved in a finite strain setting. Stress and material tangents are recovered by using automatic differentiation. Newton’s method is used to iteratively solve the resulting non-linear system.
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+  
+ A basic Ginzburg-Landau model is solved by combining `Ferrite` with `ForwardDiff`. Using threads, the calculation time is optimized.  
+	
+ Keywords: Gizburg-Landau, ForwardDiff.jl, parallelization, optimization, performance
+
+10.	[Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/)  
+
+ A program for the use of linear shell elements is set up. The theoretical background is introduced as well.  
+	
+ Keywords: Shell elements, displacements, rotations
```
DRollin, line 72:
```suggestion
 Keywords: Shell elements, displacements, rotations, ForwardDiff.jl, under integration
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+ A two-dimensional representative volume element (RVE) is loaded in shear. Dirichlet and periodic boundary conditions are applied and the results from the homogenization are compared to the Voigt and Reuss bounds.  
+	
+ Keywords: homogenization, periodic boundary conditions, representative volume element (RVE), microscale, Voigt and Reuss bound
+
+13.	[Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)  
+
+ Stokes flow on a quarter circle is solved by applying periodic boundary conditions. The weak and strong form of the problem are discussed including constrains. Mesh generation is done directly via the Gmsh API.  
+	
+ Keywords: Periodic boundary conditions, multiple fields, mean value constraint, mesh generation with Gmsh, coupled problem, weak and strong form
```
DRollin, line 90:
```suggestion
 Keywords: Periodic boundary conditions, mean value constraint, mesh generation with Gmsh, coupled problem, weak and strong form
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Other examples
+1.	[Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)  
+
+ The weak form of the semi-discretized incompressible Navier-Stokes equations are derived from the strong form and implemented in the finite element code. Then, the time-dependent solution is calculated by using the DifferentialEquations.jl package. For this purpose, the PDE is required in a specific form.  
+	
+ Keywords: Fluid dynamics, weak and strong form, solver, incompressibility, time-dependent problem
+
+## Grouped by topic
+* Fundamentals: [Heat Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/heat_equation/), [Postprocessing](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/postprocessing/), [Helmholtz Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/helmholtz/)
```
DRollin, line 106:
```suggestion
* Fundamentals: [Heat Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/heat_equation/), [Postprocessing](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/postprocessing/), [Helmholtz Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/helmholtz/), [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/), [Computational homogenization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/computational_homogenization/)
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+* Non-linear Problems: [Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/), [von Mises Plasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/plasticity/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Topology Optimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/topology_optimization/)
+* Time dependent problems: [Time Dependent Problems](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/transient_heat_equation/), [Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)
+* Advanced: [Threaded Assembly](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/threaded_assembly/), [Ginzburg-Landau model energy minimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/landau/)
```
DRollin, line 109:
```suggestion
* Advanced: [Threaded Assembly](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/threaded_assembly/), [Ginzburg-Landau model energy minimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/landau/), [Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/), [Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)
```

DRollin (commented):
Hi,
since I am not working on topology optimization, I focused manly on language an things like that.
Overall it was nice to read and and understandable, even without knowing much about topology optimization.
I suggested to rename some functions to avoid regular functions looking like constructors.
(To me that would be confusing.)
Apart from that I added only minor suggestions.

Since this is my first review, feel free to give me some advice how to improve future reviews.

Thanks a lot!

blaszm:
Thank you for your very extensive review! I appreciate the feedback and especially the comments regarding readability.

I was not aware of conventions regarding function names and constructors, but we can change these names for sure. I will resolve most issues, but there are some points where a more detailed discussion may be necessary. I will reply individually.

docs/src/overview.md
blaszm: In my opinion, the Threaded Assembly example is the only truely "advanced" example in the sense that it only improves the speed of the program, making it optional for someone who does not care about performance too much. In contrast, all other examples showcase something new and different.

My stance on this is of course subjective and I would ask for addtional opinions.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+    n_el = length(χn)
+    
+    χ_trial = zeros(n_el)
+    ρ_trial = 0.0
+    
+    λ_lower = minimum(Δχ) - ηs
+    λ_upper = maximum(Δχ) + ηs
+    λ_trial = 0.0
+    
+    while(abs(ρ-ρ_trial)>1e-7)
```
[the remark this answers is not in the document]
blaszm: In general I would agree. In this case, to my knowledge there should not be any instabilities in the mathematical method to prevent the loop from exiting eventually. I can understand, that it might be nice to implement such a feature, but the question remains how to choose a suitable maximum number (the only possibility I see is using a very large number just in case or aborting the program with an error message to prevent the user from obtaining wrong results).

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+  
+ A basic Ginzburg-Landau model is solved by combining `Ferrite` with `ForwardDiff`. Using threads, the calculation time is optimized.  
+	
+ Keywords: Gizburg-Landau, ForwardDiff.jl, parallelization, optimization, performance
+
+10.	[Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/)  
+
+ A program for the use of linear shell elements is set up. The theoretical background is introduced as well.  
+	
+ Keywords: Shell elements, displacements, rotations, ForwardDiff.jl, under integration
```
termi-official, line 72: Just a quick question - isn't the correct translation here "reduced integration" or am I simply misunderstanding the wording?

docs/src/overview.md
DRollin: I just thought that all examples should be listed in some category.
That's why I added the ones that were not listed in any category yet.
I think that it would be nice to have an examples under "Fundamentals" which shows how to treat multiple fields.
(Even if it is pretty straight forward.) And "Incompressible Elasticity" would be a good choice for that, I think.

Regarding "Computational Homogenization" you are probably right: Its too specific for "Fundamentals".

docs/src/overview.md
blaszm: In this example I would consider the resulting stiffness matrix as a single material tangent, as it is not a coupled problem.

docs/src/overview.md
termi-official: I think that, from a technical point of view, the Ginzburg-Landau model is an advanced modeling example that combines some of the techniques of other examples to speed up the computation. However, the threaded assembly is an example showing advanced algorithms. Not sure if the distinction is worth to make at the moment, because we just have a few examples for now. Stokes is the basic example for periodic boundary condtitions and mean value constraint, while the linear shell tells you how to deal with shells. Would not consider these two advanced either.

docs/src/overview.md
DRollin: I copied that term from the example.
Personally, I would call it "selectively reduced integration" or "reduced integration"

blaszm: I think you are correct if I am not mistaken, however the same wording is used in the documentation of the example.

docs/src/overview.md
blaszm: I would like to keep the keyword "multiple fields" to show that the DofHandler here consists of different parts.

docs/src/literate/topology_optimization.jl
blaszm: I agree on this part. The reason we calculate it this way is the following: 

In order to compute χn1 we need to calculate λ, which at that moment is unknown. Therefore we use the bisection algorithm to compute λ. The way this works is we choose a λ_trial from either starting conditions or the last iteration and check, whether our exit condition in the while loop is satisfied. If this is the case, we know that our field χ_trial is the correct χn1 (before the check we only "hope" that this is the case and if the condition is not fulfilled, we need to start the next iteration). 

Maybe you could add 2-3 lines of comments before the function to summarize my thoughts in a way someone not familiar with the code can understand it?

docs/src/literate/topology_optimization.jl
blaszm: Maybe call the function instead compute_bisection to highlight what the function does/what is happening inside?

docs/src/literate/topology_optimization.jl
blaszm: See above, call the function compute_bisection(..)?

docs/src/literate/topology_optimization.jl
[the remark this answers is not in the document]
DRollin: You know the algorithm better than I do, so if you say it's not needed, I am fine with that.

docs/src/literate/topology_optimization.jl
DRollin: That would also be fine.
I usually name functions according to what they are supposed to achieve and not the way they do it.
This way, one could change the algorithm inside without changing the name and the general functionality.
But here that would not be so important, so choose what ever you prefer.

docs/src/overview.md
DRollin: Makes sense, just leave it as it is.
(I am just used to the plural form, that's it.)

docs/src/overview.md
DRollin: And should we add the same key word for other examples with multiple fields?

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,502 @@
...
+        ∇²χ[i] = (χn[nbg[1]]+χn[nbg[2]]+χn[nbg[3]]+χn[nbg[4]]-4*χn[i])/(Δh^2)
+    end
+
+    return ∇²χ
+end
+#md nothing # hide
+
+# To ensure that the mass is kept constant, we have to calculate the constraint
+# parameter $\lambda$, which we do via the bisection method. We repeat the calculation
+# until the difference between the current density and target density nearly vanishes.
```
DRollin, line 213:
```suggestion
# For the iterative computation of the solution, a function is needed to update the local densities.
# To ensure that the total mass is kept constant, we have to calculate the constraint
# parameter $\lambda$, which we do via the bisection method. We repeat the calculation
# until the difference between the current global density and target density nearly vanishes. 
```

I added a sentence in the beginning to make clear, that the goal of the function is to compute the densities and that the constraint parameter is only needed to do the computation.
To me it would be clear now.
I also added a distinction between "local" and "global" density, so the reader can better understand what ρ in the function is and why it is used as criterion for the iteration.
However, I am not sure if that is the appropriate wording to make that distinction.
I just didn't come up with a better idea.

blaszm:
```suggestion
# For the iterative computation of the solution, a function is needed to update the densities in each element.
# To ensure that the mass is kept constant, we have to calculate the constraint
# parameter $\lambda$, which we do via the bisection method. We repeat the calculation
# until the difference between the average density (calculated from the elementwise trial densities) and the target density nearly vanishes. 
```

blaszm: I think your suggestion is already quite good. For someone familiar with the topic the notion of local/global densities should make sene. I slightly changed your proposition to make it a bit easier for newcomers to the topic to understand what is going on.

docs/src/overview.md
blaszm: I don't think it is necessary to introduce it as a keyword considering the trick is done in many examples and only improves performance very slightly

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+ The Helmholtz equation is solved on a unit square. Dirichlet and Neumann boundary conditions are applied at different parts of the boundary. A known analytical solution is approximately reproduced to verify the finite element solution.  
+	
+ Keywords: Fundamentals, Helmholtz equation, Dirichlet boundary condition, Neumann boundary condition, verification, boundary integrals
+
+4.	[Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/)  
+
+ A mixed element for solving unidirectional displacement-pressure coupling is constructed. The solution for nearly incompressible materials is compared for different interpolations.  
+	
+ Keywords: Mixed elements, unidirectional coupling, displacement-pressure equation, incompressibility, stability
```
blaszm, line 36:
```suggestion
 Keywords: Mixed elements, multiple fields, unidirectional coupling, displacement-pressure equation, incompressibility, stability
```

docs/src/overview.md
blaszm:
```suggestion
 Keywords: Shell elements, displacements, rotations, multiple fields, ForwardDiff.jl, reduced integration
```

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+
+ A program for the use of linear shell elements is set up. The theoretical background is introduced as well.  
+	
+ Keywords: Shell elements, displacements, rotations, ForwardDiff.jl, under integration
+ 
+11.	[Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/)  
+
+ This program combines ideas from the examples Incompressible Elasticity and Hyperelasticity to construct a mixed element solving three-dimensional displacement-pressure equations.
+	
+ Keywords: Non-linear problem, hyperelasticity, finite strain, large deformations, Newton's method, automatic differentiation, coupled problem, mixed elements, displacement-pressure equation, incompressibility
```
blaszm, line 78:
```suggestion
 Keywords: Non-linear problem, hyperelasticity, finite strain, large deformations, Newton's method, automatic differentiation, coupled problem, mixed elements, multiple fields, displacement-pressure equation, incompressibility
```

docs/src/overview.md
blaszm: Yes, done!

docs/src/overview.md
blaszm: I agree about the part that every example should be listed at least in one category. If not category is a good fit, we should add a category "other examples".

Personally, I would consider using FEM with multiple fields and coupling to be already too advanced for "Fundamentals". That does not mean it has to be a very hard problem, but most students / beginners should start with simpler problems in my opinion, since there are already a lot of additional steps to consider for coupled problems.

docs/src/literate/topology_optimization.jl
blaszm: I agree with your reasoning.

docs/src/overview.md
DRollin: I was thinking of this example as a fundamental one introducing this aspect.
Thinking of it in another category, I would agree that this key word is not needed.

docs/src/overview.md
DRollin: Maybe we could have a category "Coupled Problems" and list this example as the first one?

blaszm: I think I will add two categories: "Multiple fields" to account for all problems which are either fully coupled, unidirectional coupled or contain multiple fields for the calculation (shells) and "Periodic problems" for the examples Comp. homogenization and stokes flow. This way, each example is present at least once in the Grouped by topic section.

docs/src/overview.md
```diff
@@ -0,0 +1,109 @@
...
+* Non-linear Problems: [Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/), [von Mises Plasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/plasticity/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Topology Optimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/topology_optimization/)
+* Time dependent problems: [Time Dependent Problems](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/transient_heat_equation/), [Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)
+* Advanced: [Threaded Assembly](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/threaded_assembly/), [Ginzburg-Landau model energy minimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/landau/)
```
blaszm, line 109:
```suggestion
* Time dependent problems: [Time Dependent Problems](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/transient_heat_equation/), [Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)
* Multiple fields: [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/), [Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)    
* Periodic problems: [Computational homogenization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/computational_homogenization/), [Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)
* Advanced: [Threaded Assembly](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/threaded_assembly/), [Ginzburg-Landau model energy minimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/landau/)
```

blaszm: (reasoning see above)

docs/src/overview.md
termi-official: I think we should be a bit careful regarding the wording. I would classify "coupled problems" as a subsection of problems containing multiple fields, where, loosely speaking, different sets of fields have different underlying physics (e.g. fluid-stucture interacton). Incompressibility on the other hand acts as a constraint on your PDE. Certainly, incompressibility is advanced from a purely theoretical perspective, but apart from this is the algorithmic treatment of incompressible linear elasticity rather straight forward. 

However, I agree hat we might want to have a simpler first example on multiple fields that does not have this large theoretical building (for users without background in mechanics). I can set up some reaction-diffusion system as a simple first example for multiple fields here for the future, if this is what we need.

blaszm:
```suggestion
* Fundamentals: [Heat Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/heat_equation/), [Postprocessing](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/postprocessing/), [Helmholtz Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/helmholtz/), [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/)
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+# We start by introducing the local, elementwise density $\chi \in [\chi_{\text{min}}, 1]$ of the material, where we choose
+# $\chi_{\text{min}}$ slightly above zero to prevent numerical instabilities. Here, $\chi = \chi_{\text{min}}$ means void and $\chi=1$
+# means bulk material. Then, we use a SIMP ansatz (solid isotropic material with penalization) for the stiffness tensor
+# $C(\chi) = \chi^p C_0$, where $C_0$ is the stiffness of the bulk material. The SIMP exponent $p>1$ ensures that the
+# model prefers the density values void and bulk before the intermediate values. The variational formulation then yields
+# the modified Gibbs energy 
+# ```math
+# G = \int_{\Omega} \frac{1}{2} \chi^p \varepsilon : C : \varepsilon \; \text{d}V - \int_{\Omega} \boldsymbol{f} \cdot \boldsymbol{u} \; \text{d}V - \int_{\partial\Omega} \boldsymbol{t} \cdot \boldsymbol{u} \; \text{d}A.
+# ```
+# which we will solve with the finite element method using Ferrite. Furthermore, we receive the evolution equation of the density
```
termi-official, line 33: Wording. We do not "solve" the energy with the finite element method.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+# ```
+# ```math
+# \beta \nabla \chi \cdot \textbf{n} = 0 \quad \forall \textbf{x} \in \partial \Omega, 
+# ```
+# with the thermodynamic driving force
+# ```math
+# p_\chi = \frac{1}{2} p \chi^{p-1} \varepsilon : C : \varepsilon.
+# ```
+# The evolution equation consists of the driving force, the damping paramter $\eta$, the regularization parameter $\beta$ times the Laplacian,
+# which is necessary to avoid numerical issues like mesh dependence or checkerboarding, and the constraint parameters $\lambda$, to keep the mass constant,
```
termi-official, line 46: "and"..."and".

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+# and $\gamma$, to avoid leaving the set $[\chi_{\text{min}}, 1]$. By including gradient regularization, it becomes necessary to calculate the Laplacian.
+# The Finite Difference Method for square meshes with the edge length $\Delta h$ approximates the Laplacian as follows:
```
termi-official, line 48: Can you quickly comment on why we use finite differences over finite elements here, to elaborate it for readers not proficient in topology optimization?

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+    dbc = ConstraintHandler(dh)
+    add!(dbc, Dirichlet(:u, getnodeset(dh.grid, "clamped"), (x,t) -> zero(Vec{2}), [1,2]))
+    close!(dbc)
+    t = 0.0
+    update!(dbc, t)
+    return dbc
+end
+#md nothing # hide
+
+# Now, we define a struct to store all necessary material parameters (stiffnes tensor of the bulk material
```
termi-official, line 118:
```suggestion
# Now, we define a struct to store all necessary material parameters (stiffness tensor of the bulk material
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+
+function MaterialParameters(E, ν, χ_min, p, β, η) 
+    δ(i,j) = i == j ? 1.0 : 0.0 ## helper function
+
+    G = E / 2(1 + ν) ## =μ
+    λ = E*ν/(1-ν^2) ## correction for plane stress included
+
+    temp(i,j,k,l) = λ * δ(i,j)*δ(k,l) + G* (δ(i,k)*δ(j,l) + δ(i,l)*δ(j,k)) 
+    
+    C = SymmetricTensor{4, 2}(temp)
```
termi-official, line 139: We can merge these lines

docs/src/literate/topology_optimization.jl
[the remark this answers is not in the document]
termi-official: Sorry to bother again, but can you add a comment here in the example about the convergence guarantee?

docs/src/overview.md
```diff
@@ -0,0 +1,112 @@
...
+* Non-linear Problems: [Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/), [von Mises Plasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/plasticity/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Topology Optimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/topology_optimization/)
+* Time dependent problems: [Time Dependent Problems](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/transient_heat_equation/), [Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)
+* Multiple fields: [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/), [Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/)
+* Coupled problems: [Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)
```
termi-official, line 110: Stokes flow is not a coupled problem, but a "multiple field" problem.

docs/src/overview.md
```diff
@@ -0,0 +1,112 @@
...
+	
+ Keywords: Fluid dynamics, weak and strong form, solver, incompressibility, time-dependent problem
+
+## Grouped by topic
+* Fundamentals: [Heat Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/heat_equation/), [Postprocessing](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/postprocessing/), [Helmholtz Equation](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/helmholtz/), [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/)
+* Non-linear Problems: [Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/), [von Mises Plasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/plasticity/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Topology Optimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/topology_optimization/)
```
termi-official, line 107: Computational homogenization is missing.

docs/src/overview.md
```diff
@@ -0,0 +1,112 @@
...
+* Time dependent problems: [Time Dependent Problems](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/transient_heat_equation/), [Incompressible Navier-Stokes Equations via DifferentialEquations.jl](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/ns_vs_diffeq/)
```
termi-official, line 108: Actually the first problem should be "transient heat equation" right? @koehlerson

termi-official (changes requested):
Thanks for the great improvements Mischa! I think some more minor adjustments and we should be fine.

docs/src/overview.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Documentation of examples
+
+On this page, you find a documentation for each example currently available for `Ferrite`. The collection contains programs displaying different aspects and techniques of the `Ferrite` toolbox. 
+The programs are sorted as
+
+* A complete list shortly summarizing the topic, what they teach and which keywords related to the task they contain
+* Grouped by topic 
+
+## List of examples (chronological order)
```
KnutAM, line 13:
```suggestion
## List of examples
```
I think the list order looks good, but e.g. postprocessing was added after the von Mises example.
Super nice work @blaszm - I need to study that topology example closer - the method does look very neat!
(Maybe it could make sense to move "Nearly Incompressible Hyperelasticity" after "Hyperelasticity" since it follows from "Incompressible elasticity" and "Hyperelasticity"?)

termi-official: Ah right, the  "Nearly Incompressible Hyperelasticity" example should also be advanced, right?

blaszm, line 13:
```suggestion
## List of examples (sequential order)
```
My intention was to create a list with the same order as on the sidebar of the homepage. I changed the desciption to match that, though it may be possible examples later in the list were added earlier (in time).

docs/src/overview.md
blaszm: I would agree that this would be more discriptive, I just used the titles used in the sidebar to match and avoid confusion.

docs/src/overview.md
blaszm, line 110:
```suggestion
* Multiple fields: [Incompressible Elasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/incompressible_elasticity/), [Linear shell](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/linear_shell/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Stokes flow](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/stokes-flow/)     
```

blaszm: Then we currently don't have a "true" coupled problem and the coupled problem list is obsolete for now.

docs/src/overview.md
blaszm:
```suggestion
* Non-linear Problems: [Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/hyperelasticity/), [von Mises Plasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/plasticity/), [Nearly Incompressible Hyperelasticity](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/quasi_incompressible_hyperelasticity/), [Computational homogenization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/computational_homogenization/), [Topology Optimization](https://ferrite-fem.github.io/Ferrite.jl/dev/examples/topology_optimization/)
```

docs/src/literate/topology_optimization.jl
blaszm:
```suggestion
# Furthermore, we receive the evolution equation of the density in the strong form as
```

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+# $\chi_{\text{min}}$ slightly above zero to prevent numerical instabilities. Here, $\chi = \chi_{\text{min}}$ means void and $\chi=1$
+# means bulk material. Then, we use a SIMP ansatz (solid isotropic material with penalization) for the stiffness tensor
+# $C(\chi) = \chi^p C_0$, where $C_0$ is the stiffness of the bulk material. The SIMP exponent $p>1$ ensures that the
+# model prefers the density values void and bulk before the intermediate values. The variational formulation then yields
+# the modified Gibbs energy 
+# ```math
+# G = \int_{\Omega} \frac{1}{2} \chi^p \varepsilon : C : \varepsilon \; \text{d}V - \int_{\Omega} \boldsymbol{f} \cdot \boldsymbol{u} \; \text{d}V - \int_{\partial\Omega} \boldsymbol{t} \cdot \boldsymbol{u} \; \text{d}A.
+# ```
+# which we will solve with the finite element method using Ferrite. Furthermore, we receive the evolution equation of the density 
+# and the additional Neumann boundary condition
```
blaszm, line 34:
```suggestion
# and the additional Neumann boundary condition in the strong form
```

docs/src/literate/topology_optimization.jl
blaszm: Wording is fine here I think. "Constaint parameters" is the last item in the list and the second "and" refers to the second parameter within the list of constaint parameters.

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,503 @@
...
+# p_\chi + \eta \dot{\chi} + \lambda + \gamma - \beta \nabla^2 \chi \ni 0 \quad \forall \textbf{x} \in \Omega, 
+# ```
+# ```math
+# \beta \nabla \chi \cdot \textbf{n} = 0 \quad \forall \textbf{x} \in \partial \Omega, 
+# ```
+# with the thermodynamic driving force
+# ```math
+# p_\chi = \frac{1}{2} p \chi^{p-1} \varepsilon : C : \varepsilon.
+# ```
+# The evolution equation consists of the driving force, the damping paramter $\eta$, the regularization parameter $\beta$ times the Laplacian,
```
blaszm, line 45:
```suggestion
# We obtain the mechanical displacement field by applying the Finite Element Method to the weak form
# of the Gibbs energy using Ferrite. In contrast, we use the evolution equation (i.e. the strong form) to calculate
# the value of the density field $\chi$. The advantage of this "split" approach is the very high computation speed.
# The evolution equation consists of the driving force, the damping paramter $\eta$, the regularization parameter $\beta$ times the Laplacian,
```

docs/src/literate/topology_optimization.jl
blaszm, line 139:
```suggestion
    C = SymmetricTensor{4, 2}((i,j,k,l) -> λ * δ(i,j)*δ(k,l) + G* (δ(i,k)*δ(j,l) + δ(i,l)*δ(j,k)))
```

docs/src/literate/topology_optimization.jl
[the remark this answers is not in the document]
blaszm: (See above)

docs/src/literate/topology_optimization.jl
```diff
@@ -0,0 +1,505 @@
...
+    
+    return
+end
+#md nothing # hide
+
+# Lastly, we call our main function and compare the results. To create the
+# complete output with all iteration steps, it is possible to set the output 
+# parameter to `true`.
+
+topopt(0.5*1e-3, 0.5, 60, "betasmall"; output=false);
```
blaszm, line 485:
```suggestion
topopt(0.5*1e-3, 0.5, 60, "betasmall"; output=:false);
```

KristofferC:
Is this good to go? I think it is fine to merge this even if there are some small outstanding things to address, they can always be polished in future  PRs.

termi-official (approved):
Sure, you are right Kristoffer. I think we started going a bit over board with fine tuning. 

Thanks again @blaszm for the new example and getting an overview!

fredrikekre:
Now that we have quite many examples it might be time to go through them and split into "Tutorials" and "Code gallery". This way we can ensure that each tutorial introduces and teaches some new concept, and come in increasing order of complexity. Then the code gallery can just be cool stuff you can do with Ferrite.

PR #532, opened 2022-10-29
Add apply_analytical! function

KnutAM:
Seems like more users may find this convenient, e.g.  #434 
In contrast to the solution in  #434, this solution also supports different function and geometric interpolations. 

Remaining 
- [x] Tests
- [x] Add to transient heat flow example


fredrikekre:
I think a more general name would be better since this is not only useful for initial conditions. For example in the homogenization example: https://github.com/Ferrite-FEM/Ferrite.jl/blob/8e447e26128481fd50ac291b2370b84d74ef8a32/docs/src/literate/computational_homogenization.jl#L519-L522

This seems to be simply to transfer an analytical function to FE space (without doing the L2 projection or solving the pertinent `t = 0` equation). I don't have a good suggestion though.

KnutAM:
Good point! `apply_analytical!` then perhaps, follows in line with the `apply!` functions already existing for modifying the degree of freedom vector?

termi-official:
This seems to only work correctly for pure ODE systems. Algebraic constraints in e.g. DAEs may be a problem, especially enforcing the correct compatibilities between the fields. We should at least document this if you want to merge this.

KnutAM:
@termi-official: do you mean that it is difficult for the user to define the correct functions `f(x)` in the case of a complicated DAE? In that case, isn't it clear that the user must supply a valid analytical solution? If I misunderstand you here, do you have a concrete example?

I can write a short section about using this for initial conditions, and add a warning there about the compatibility. Would that work?

termi-official:
Indeed, that is my point - writing down these equations for the algebraic portion can be super painful. Usually it is simpler to just provide the solution to the ODE portion of the problem and let the initializer figure out what the compatible solution to the algebraic equations are. At least that is how it is usually handled by DAE solvers. Not sure what the best approach would be here, though.
Also as a side note: It happens quite often that users are not aware that they are solving DAEs and not ODEs.

koehlerson:
Maybe worth it to compress the gif a little, it's almost three times larger than before

termi-official:
Ah sorry forgot to mention. We usually utilize gifsky for compressing the gifs. From our experience the file size is small and visual is still great.

fredrikekre:
deal.ii: https://www.dealii.org/current/doxygen/deal.II/namespaceVectorTools.html#a761f008bdeb7d94a69205ae824deefad

KnutAM:
> deal.ii: https://www.dealii.org/current/doxygen/deal.II/namespaceVectorTools.html#a761f008bdeb7d94a69205ae824deefad

To me, "interpolate" sounds less clear - I then think of the `get_point_values` function. 
In my head, interpolation is "discrete points -> continuous function" while this PR implements "continuous function -> discrete points". But maybe I'm thinking wrong here?
 
If `apply_analytical!` is not good, what about `setdofvalues!`?
Of course, if you prefer `interpolate!` (or `interpolate_to_dofs!`), it is also fine - just let me know what you prefer!

KnutAM:
@termi-official: Does [this text](https://ferrite-fem.github.io/Ferrite.jl/previews/PR532/manual/boundary_conditions/#Initial-Conditions) suffice as the warning for DAE?

termi-official:
If you do not like interpolation, then you could also call it projection, nodal projection, proiective interpolation, or similar.

docs/src/manual/boundary_conditions.md
```diff
@@ -260,3 +260,29 @@ pdbc = PeriodicDirichlet(
...
+x = zeros(ndofs(dh))
+apply_analytical!(x, dh, :p, x->ρ*g*x[2])
+```
+
+See also [Time Dependent Problems](@ref) for one example. 
+
+*Note about solving DAE:* 
+A Differential Algebraic Equations (DAE) is an equation of the form
+``\boldsymbol{r}(\boldsymbol{x}(t),\boldsymbol{x}'(t),t)=\boldsymbol{0}``,
+which cannot be expressed as an ODE.
```
termi-official, line 285: Maybe we can be a bit more precise here and give a quick example. My idea is as follows.

```suggestion
A Differential Algebraic Equations (DAE) is an equation of the form
``\boldsymbol{r}(\boldsymbol{x}(t),\boldsymbol{x}'(t),t)=\boldsymbol{0}``,
which usually cannot be expressed as a true ODE. They occur often,
but not always, in forms where some time derivatives are missing
``
x_1'(t) = f(\boldsymbol{x}(t),t)
0 = g(\boldsymbol{x}(t),t)`
``
```

docs/src/manual/boundary_conditions.md
```diff
@@ -260,3 +260,29 @@ pdbc = PeriodicDirichlet(
...
+In for such equations, it is usually necessary to specify initial conditions 
+for both ``\boldsymbol{x}(0)`` and ``\boldsymbol{x}'(0)``, and these must be consistent,
+i.e. ``\boldsymbol{r}(\boldsymbol{x}(0),\boldsymbol{x}'(0),0)=\boldsymbol{0}``.
```
termi-official, line 288: This is only true for index 1 DAEs. For index>1 you get hidden constraints which must be fulfilled too. This makes the initialization so tricky.

termi-official (commented):
Some additional comments.

KnutAM:
> If you do not like interpolation, then you could also call it projection, nodal projection, proiective interpolation, or similar.

Thanks, yes, that's a good idea - the function does a projection of the analytical function onto the FE-space! 

We already have the non-mutating `project` function for the L2-projector, whose output is a vector with one item per "dof location". Could defining `project!` with an output vector with one item per dof be a problem?

docs/src/manual/boundary_conditions.md
KnutAM: Perhaps something like this then?
```suggestion
For such equations, it is usually necessary to specify initial conditions 
for both ``\boldsymbol{x}(0)`` and ``\boldsymbol{x}'(0)``. 
These functions must be consistent, i.e. fulfilling both ``\boldsymbol{r}(\boldsymbol{x}(0),\boldsymbol{x}'(0),0)=\boldsymbol{0}`` and all additional constraints.
```
(I think I'll change from `x` to `a` for the unknowns later, just to avoid confusion with the coordinate)

KnutAM:
Checked the deal.ii history, and seems like their `interpolate` function stems from interpolating a solution field from one mesh to another (https://github.com/dealii/dealii/commit/ee3c4c3e5e39e15c4a5e90236b706c25110e237e), and I guess that the analytical "interpolation" then just followed as a natural consequence. 
Mirroring that does perhaps make sense and could be useful for (adaptive) re-meshing?
Then this PR would give: `interpolate!(a, dh, f, [field], [cellset])`
A future PR could give: `interpolate!(anew, new_dh, aold, old_dh)`?

*I've changed the order of arguments and made `field` optional (default to the first field in the dof handler and error if there is more than one field)*

docs/src/literate/transient_heat_equation.jl
```diff
@@ -180,16 +183,24 @@ A = (Δt .* K) + M;
 # by `get_rhs_data`. The function returns a `RHSData` struct, which contains all needed informations to apply
 # the boundary conditions solely on the right-hand-side vector of the problem.
 rhsdata = get_rhs_data(ch, A);
-# We set the initial time step, denoted by uₙ,  to $\mathbf{0}$.
+# We set the values at initial time step, denoted by uₙ, to a bubble-shape described by 
+# $(x_1^2-1)(x_2^2-1)$, such that it is zero at the boundaries and half the max temperature in the center.
```
KnutAM, line 187:
```suggestion
# $(x_1^2-1)(x_2^2-1)$, such that it is zero at the boundaries and the maximum temperature in the center.
```

docs/src/literate/computational_homogenization.jl
```diff
@@ -516,16 +516,12 @@ round.(ev; digits=-8)
...
-add!(chM, Dirichlet(:u, Set(1:getnnodes(grid)), (x, t) -> εᴹ[Int(t)] ⋅ x, [1, 2]))
-close!(chM)
 uM = zeros(ndofs(dh))
 
 vtk_grid("homogenization", dh) do vtk
     for i in 1:3
         ## Compute macroscopic solution
-        update!(chM, i)
-        apply!(uM, chM)
+        apply_analytical!(uM, dh, x->εᴹ[i] ⋅ x)
```
fredrikekre, line 524:
```suggestion
        apply_analytical!(uM, dh, x -> εᴹ[i] ⋅ x)
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -91,15 +92,17 @@ f = zeros(ndofs(dh));
...
+t_rise = 100
 ch = ConstraintHandler(dh);
 
 # Here, we define the boundary condition related to $\partial \Omega_1$.
 ∂Ω₁ = union(getfaceset.((grid, ), ["left", "right"])...)
 dbc = Dirichlet(:u, ∂Ω₁, (x, t) -> 0)
 add!(ch, dbc);
-# While the next code block corresponds to the linearly increasing temperature description on $\partial \Omega_2$.
+# While the next code block corresponds to the linearly increasing temperature description on $\partial \Omega_2$
+# untill `t=t_rise`, and then keep constant
```
fredrikekre, line 103:
```suggestion
# until `t=t_rise`, and then keep constant
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -91,15 +92,17 @@ f = zeros(ndofs(dh));
...
 ∂Ω₂ = union(getfaceset.((grid, ), ["top", "bottom"])...)
-dbc = Dirichlet(:u, ∂Ω₂, (x, t) -> t*(max_temp/T))
+dbc = Dirichlet(:u, ∂Ω₂, (x, t) -> max_temp*clamp(t/t_rise, 0, 1))
```
fredrikekre, line 105:
```suggestion
dbc = Dirichlet(:u, ∂Ω₂, (x, t) -> max_temp * clamp(t / t_rise, 0, 1))
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -127,10 +130,10 @@ function doassemble_K!(K::SparseMatrixCSC, f::Vector, cellvalues::CellScalarValu
             for i in 1:n_basefuncs
                 v  = shape_value(cellvalues, q_point, i)
                 ∇v = shape_gradient(cellvalues, q_point, i)
-                fe[i] += v * dΩ
+                fe[i] += 0.1*v * dΩ
```
fredrikekre, line 133:
```suggestion
                fe[i] += 0.1 * v * dΩ
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -127,10 +130,10 @@ function doassemble_K!(K::SparseMatrixCSC, f::Vector, cellvalues::CellScalarValu
...
                 for j in 1:n_basefuncs
                     ∇u = shape_gradient(cellvalues, q_point, j)
-                    Ke[i, j] += (∇v ⋅ ∇u) * dΩ
+                    Ke[i, j] += 1.e-3*(∇v ⋅ ∇u) * dΩ
```
fredrikekre, line 136:
```suggestion
                    Ke[i, j] += 1e-3 * (∇v ⋅ ∇u) * dΩ
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -180,16 +183,24 @@ A = (Δt .* K) + M;
 # by `get_rhs_data`. The function returns a `RHSData` struct, which contains all needed informations to apply
 # the boundary conditions solely on the right-hand-side vector of the problem.
 rhsdata = get_rhs_data(ch, A);
-# We set the initial time step, denoted by uₙ,  to $\mathbf{0}$.
+# We set the values at initial time step, denoted by uₙ, to a bubble-shape described by 
+# $(x_1^2-1)(x_2^2-1)$, such that it is zero at the boundaries and the maximum temperature in the center.
 uₙ = zeros(length(f));
+apply_analytical!(uₙ, dh, x->(x[1]^2-1)*(x[2]^2-1)*max_temp);
```
fredrikekre, line 189:
```suggestion
apply_analytical!(uₙ, dh, x -> (x[1]^2 - 1) * (x[2]^2 - 1) * max_temp);
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -180,16 +183,24 @@ A = (Δt .* K) + M;
...
 # Here, we apply **once** the boundary conditions to the system matrix `A`.
 apply!(A, ch);
 
 # To store the solution, we initialize a `paraview_collection` (.pvd) file.
 pvd = paraview_collection("transient-heat.pvd");
+t=0
```
fredrikekre, line 195:
```suggestion
t = 0
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -180,16 +183,24 @@ A = (Δt .* K) + M;
...
+vtk_grid("transient-heat-$t", dh) do vtk
+    vtk_point_data(vtk, dh, uₙ)
+    vtk_save(vtk)
+    pvd[t] = vtk
```
fredrikekre, line 199: I think other examples use https://github.com/jipolanco/WriteVTK.jl/blob/6e528bd6a016efb84f5cb50f2dc43bcdc59ab8d4/src/gridtypes/ParaviewCollection.jl#L59 but since I don't see that in the WriteVTK documentation maybe it is deprecated or discouraged? Edit: I see that this approach was also used later in this file, so let's at least have it consistent (if you change here, also change below).

docs/src/manual/boundary_conditions.md
```diff
@@ -260,3 +260,34 @@ pdbc = PeriodicDirichlet(
...
+# Initial Conditions
+
+When solving time-dependent problems, initial conditions, different from zero, may be required. 
+For finite element formulations of ODE-type, 
+i.e. ``\boldsymbol{u}'(t) = \boldsymbol{f}(\boldsymbol{u}(t),t)``, 
+where ``\boldsymbol{u}(t)`` are the degrees of freedom,
+initial conditions can be specified by the [`apply_analytical!`](@ref) function.
+For example, specify the initial pressure as a function of the y-coordinate
+```julia
+ρ = 1000; g=9.81    # density [kg/m³] and gravity [N/kg]
```
fredrikekre, line 273:
```suggestion
ρ = 1000; g = 9.81    # density [kg/m³] and gravity [N/kg]
```

docs/src/manual/boundary_conditions.md
```diff
@@ -260,3 +260,34 @@ pdbc = PeriodicDirichlet(
...
+grid = generate_grid(Quadrilateral, (10,10))
+dh = DofHandler(grid); push!(dh, :u, 2); push!(dh, :p, 1); close!(dh)
+u = zeros(ndofs(dh))
+apply_analytical!(u, dh, x->ρ*g*x[2], :p)
```
fredrikekre, line 277:
```suggestion
apply_analytical!(u, dh, x -> ρ * g * x[2], :p)
```

docs/src/manual/boundary_conditions.md
```diff
@@ -260,3 +260,34 @@ pdbc = PeriodicDirichlet(
...
+```
+
+See also [Time Dependent Problems](@ref) for one example. 
+
+*Note about solving DAE:* 
+A Differential Algebraic Equations (DAE) is an equation of the form
+``\boldsymbol{r}(\boldsymbol{u}(t),\boldsymbol{u}'(t),t)=\boldsymbol{0}``,
+which usually cannot be expressed as a true ODE. They occur often,
+but not always, in forms where some time derivatives are missing
+``
```
fredrikekre, line 287:
````suggestion
```math
````

File: docs/src/manual/boundary_conditions.md [hunk 15]
```diff
@@ -260,3 +260,34 @@ pdbc = PeriodicDirichlet(
...
+u_1'(t) = f(\boldsymbol{u}(t),t)
+0 = g(\boldsymbol{u}(t),t)`
+``
```
fredrikekre, line 290:
````suggestion
```
````

File: src/Dofs/apply_analytical.jl [hunk 16]
```diff
@@ -0,0 +1,114 @@
...
+    ntuple(i->default_interpolation(getcelltype(i)), length(fhs))
+end
+
+function _default_interpolation(dh::DofHandler)
+    return default_interpolation(typeof(getcells(dh.grid, 1)))
+end
+
+# If fieldname==nothing, check that dh only has one field and return that fieldname 
+# If a fieldname is given, check that it exist in the dofhandler
+function _check_fieldname(dh, fieldname)
```
fredrikekre, line 14: Might be easier to always require to specify? Same as e.g. Dirichlet BC etc

src/Dofs/apply_analytical.jl
```diff
@@ -0,0 +1,114 @@
...
+    f_dofs = zeros(Int, length(celldofinds))
+
+    # Check f before looping
+    length(f(first(coords))) == field_dim || error("length(f(x)) must be equal to dimension of the field ($field_dim)")
+
+    for cellnr in cellset
+        cellcoords!(coords, dh, cellnr)
+        celldofs!(c_dofs, dh, cellnr)
+        # f_dofs .= c_dofs[celldofinds]
+        foreach(i->(f_dofs[i] = c_dofs[celldofinds[i]]), 1:length(celldofinds))
```
fredrikekre, line 92: `for`-loop look nicer IMO.

src/Dofs/apply_analytical.jl
```diff
@@ -0,0 +1,114 @@
...
+
+    getnbasefunctions(ip_geo) == length(coords) || error("coords=$coords not compatible with ip_ge=$ip_geo")
+    ref_coords = reference_coordinates(ip_fun)
+    length(ref_coords)*field_dim == length(dofs) || error("$ip_fun must have length(dofs)=$(length(dofs)) reference coords")
+
+    for (i_dof, refpoint) in enumerate(ref_coords)
+        x_dof = zero(XT)
+        for (i,x_node) in enumerate(coords)
+            x_dof += value(ip_geo, i, refpoint) * x_node
+        end
```
fredrikekre, line 108: Feels like this exist somewhere, maybe `BCValues`?

fredrikekre (commented):
Left some comments :)

src/Dofs/apply_analytical.jl
KnutAM: Yes,
https://github.com/Ferrite-FEM/Ferrite.jl/blob/2f8655970f29a7d37d41a5e721cdf10d8b21ddbc/src/FEValues/face_values.jl#L264
and 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/2f8655970f29a7d37d41a5e721cdf10d8b21ddbc/src/FEValues/face_values.jl#L278
but this seems specialized for faces? In that case, we would call `f(x)` even more times?
(I can create a similar construct pre-caching the geometric shape values at the reference points if desired, but I feel the code will become complicated then. )

src/Dofs/apply_analytical.jl
KnutAM: Makes sense!

docs/src/literate/transient_heat_equation.jl
KnutAM: I haven't seen `collection_add_timestep` before. What example did you think of?

fredrikekre: My research code I guess :D I thought we had some other time dependent problems. It used to be the only advertised way before, but maybe that has changed.

src/Dofs/apply_analytical.jl
KnutAM: Introduced `CellDofValues` now. 
I'm not sure about the name (as usual), but it is internal.

src/Dofs/apply_analytical.jl
```diff
@@ -0,0 +1,102 @@
...
+end
+
+function _default_interpolation(dh::DofHandler)
+    return default_interpolation(typeof(getcells(dh.grid, 1)))
+end
+
+"""
+    apply_analytical!(
+        a::AbstractVector, dh::AbstractDofHandler, f::Function, 
+        fieldname::Symbol, cellset=1:getncells(dh.grid))
```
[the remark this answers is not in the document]
KnutAM: True!

src/Dofs/apply_analytical.jl
```diff
@@ -0,0 +1,95 @@
...
+    end
+    return a
+end
+
+function _apply_analytical!(
+    a::Vector, dh::AbstractDofHandler, celldofinds, field_dim,
+    ip_fun::Interpolation, ip_geo::Interpolation, f::Function, cellset)
+
+    coords = getcoordinates(dh.grid, first(cellset))
+    cdv = CellDofValues(ip_fun, ip_geo)
```
fredrikekre, line 69: Seems like this could simply be `CellScalarValues` with a custom quadrature rule with points in the reference coordinates?

fredrikekre (commented):
LGTM, but looks like `CellDofValues` could be removed?

src/Dofs/apply_analytical.jl
KnutAM: OK, I'll look into that when I have time. 
I guess this should be a form of 1st order lobatto quadrature then. But since this is not implemented for RefTet I have to think a bit how to fix that. 
Will ping once fixed!

fredrikekre: You can just plug whatever coordinates you want in a "fake" quadrature rule, and think of them as "points for evaluation" instead of quadrature points, for example
```
julia> using Ferrite

julia> ip = Lagrange{2,RefTetrahedron,1}();

julia> points_for_evaluation = Ferrite.reference_coordinates(ip);

julia> fake_quadrature_rule = QuadratureRule{2,RefTetrahedron,Float64}(fill(NaN, 3), points_for_evaluation);

julia> cellvalues = CellScalarValues(fake_quadrature_rule, ip)
CellScalarValues{2, Float64, RefTetrahedron} with 3 shape functions and 3 quadrature points
```

You can add a constructor like
```
function QuadratureRule{D,R}(weights::AbstractVector{T}, points::AbstractVector{Vec{D,T}}) where {D, R, T}
    QuadratureRule{D,R,T}(weights, points)
end
```
to make it easier (no need to specify `T`) (see also https://github.com/Ferrite-FEM/Ferrite.jl/issues/513).

KnutAM: Thanks for those tips - now I also learned how to do custom quad rules!
I reduced away the dim input in the constructor as well - hope that is ok!

src/Quadrature/quadrature.jl
```diff
@@ -40,6 +40,11 @@ struct QuadratureRule{dim,shape,T}
     points::Vector{Vec{dim,T}}
 end
 
+function QuadratureRule{shape}(weights::AbstractVector{Tw}, points::AbstractVector{Vec{dim,Tp}}) where {dim, shape, Tw, Tp}
```
fredrikekre, line 43: I think it makes sense to require `dim` here (even though it can be inferred from `points`) since otherwise the first "parameter" of `QuadratureRule` in this constructor does not match the first parameter of the returned value:
```suggestion
function QuadratureRule{dim, shape}(weights::AbstractVector{Tw}, points::AbstractVector{Vec{dim,Tp}}) where {dim, shape, Tw, Tp}
```

docs/src/literate/transient_heat_equation.jl
fredrikekre: https://github.com/jipolanco/WriteVTK.jl/pull/113

PR #534, opened 2022-11-07
Quick fix for threaded assembly example on Julia 1.8

kimauth:
This guarantees the threaded assembly example to work correctly even on Julia 1.8. I am not really sure though what's the best practice pattern to use now - a channel with scratch values I guess?

### DIFF
--- a/docs/src/literate/threaded_assembly.jl
+++ b/docs/src/literate/threaded_assembly.jl
@@ -127,7 +127,7 @@ function doassemble(K::SparseMatrixCSC, colors, grid::Grid, dh::DofHandler, C::S
 
     for color in colors
         ## Each color is safe to assemble threaded
-        Threads.@threads for i in 1:length(color)
+        Threads.@threads :static for i in 1:length(color)
             assemble_cell!(scratches[Threads.threadid()], color[i], K, grid, dh, C, b)
         end
     end


fredrikekre:
> I am not really sure though what's the best practice pattern to use now - a channel with scratch values I guess?

Probably something like https://github.com/Ferrite-FEM/Ferrite.jl/blob/0e3f4d20faec741e5da05a2f9e60c67910b8ef3a/docs/src/literate/threaded_assembly_spawn.jl#L108-L121, but `:static` should be fine too.

KristofferC:
Have someone benchmarked the Channel approach?

lijas:
This is what I get:

Channels:
```
    for color in colors
        @sync begin
            workqueue = Channel{Int}(Inf)
            foreach(x -> put!(workqueue, x), color)
            close(workqueue)
            for _ in 1:Threads.nthreads()
                Threads.@spawn begin
                    local scratch = create_scratchbuffer(K, f, dh)
                    for ci in workqueue
                        # println("Now I am doing color $(ci)")
                        assemble_cell!(scratch, ci, grid, dh, C)
                    end
                end
            end
        end
    end
```
```
$ julia --project -t 1 docs/src/literate/threaded_assembly.jl 
  0.775692 seconds (250.56 k allocations: 11.765 MiB, 6.53% gc time)
$ julia --project -t 4 docs/src/literate/threaded_assembly.jl 
  0.240576 seconds (1.00 M allocations: 44.171 MiB, 25.31% gc time)
```

@ threads:
```
for color in colors
        ## Each color is safe to assemble threaded
        Threads.@threads :static for i in 1:length(color)
            assemble_cell!(scratches[Threads.threadid()], color[i], grid, dh, C)
        end
    end
```
```
$ julia --project -t 1 docs/src/literate/threaded_assembly.jl 
  0.535967 seconds (14.03 k allocations: 632.516 KiB)
$ julia --project -t 4 docs/src/literate/threaded_assembly.jl 
  0.170042 seconds (55.96 k allocations: 2.461 MiB)
```

But here create_scratchvalues(..) is also included in the time

PR #535, opened 2022-11-07
Handle prescribed dofs in RHS of affine constraints

### LINKED ISSUE #530: Affine constraint between two DBC constraints should work
```
dh = DofHandler(grid)
push!(dh, :u, 1)
close!(dh)

ch_affine_first = ConstraintHandler(dh)
add!(ch_affine_first, AffineConstraint(2, [3 => 1.0], 0.0))
add!(ch_affine_first, Dirichlet(:u, getfaceset(grid, "right"), (x, t) -> 1)) # Works without this inhomogeneity
close!(ch_affine_first)
update!(ch_affine_first, 0)

ch_dbc_first = ConstraintHandler(dh)
add!(ch_dbc_first, Dirichlet(:u, getfaceset(grid, "right"), (x, t) -> 1)) # Works without this inhomogeneity
add!(ch_dbc_first, AffineConstraint(2, [3 => 1.0], 0.0))
close!(ch_dbc_first)
update!(ch_dbc_first, 0)

K_affine_first = create_sparsity_pattern(dh, ch_affine_first)
f_affine_first = zeros(size(K_affine_first, 1))
fill!(K_affine_first.nzval, 1)

K_dbc_first = create_sparsity_pattern(dh, ch_dbc_first)
f_dbc_first = zeros(size(K_dbc_first, 1))
fill!(K_dbc_first.nzval, 1)

apply!(K_affine_first, f_affine_first, ch_affine_first)
apply!(K_dbc_first, f_dbc_first, ch_dbc_first)
```

`K_affine_first` and `K_dbc_first` are equal, but not `f_affine_first` and `f_dbc_first`.

fredrikekre:
This patch fixes affine constraints with prescribed dofs in the RHS. In particular, we allow dofs that are prescribed by just an inhomogeneity (i.e. DBC) but disallow "nesting" affine constraints.

Concretely, consider e.g. the following two constraints:

    u2 = f(t)
    u3 = u2 + b3

Before this patch this was not handled correctly since the inhomogeneity for `u3` was taken as `b3`, but it should really be `b3 + f(t)` by substituting `u2` for `f(t)`. Since we allow for time-dependent inhomogeneities this substitution has to be done at runtime and not during `close!(::ConstraintHandler)`.

Nested constraints, e.g.

    u2 = u3
    u3 = u5

are still not allowed but in the future this can be resolved in `close!(::ConstraintHandler)` to

    u2 = u5
    u3 = u5

However, this patch checks for such nesting and raises an error instead of resulting in incorrect answers as is the case before.

Fixes #530.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -76,6 +76,9 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     prescribed_dofs::Vector{Int}
     free_dofs::Vector{Int}
     inhomogeneities::Vector{T}
+    # Store the original constant inhomogeneities for affine constraints used to compute
+    # "effective" inhomogeneities in `update!` and then stored in .inhomogeneities.
+    affine_inhomogeneities::Vector{Union{Nothing,T}}
     # `nothing` for pure DBC constraint, otherwise affine constraint
     dofcoefficients::Vector{Union{Nothing, DofCoefficients{T}}}
     # global dof -> index into dofs and inhomogeneities and dofcoefficients
@@ -88,7 +91,7 @@ end
 function ConstraintHandler(dh::AbstractDofHandler)
     @assert isclosed(dh)
     ConstraintHandler(
-        Dirichlet[], Int[], Int[], Float64[], Union{Nothing,DofCoefficients{Float64}}[],
+        Dirichlet[], Int[], Int[], Float64[], Union{Nothing,Float64}[], Union{Nothing,DofCoefficients{Float64}}[],
         Dict{Int,Int}(), BCValues{Float64}[], dh, ScalarWrapper(false),
     )
 end
@@ -181,6 +184,7 @@ function close!(ch::ConstraintHandler)
     I = sortperm(ch.prescribed_dofs)
     ch.prescribed_dofs .= ch.prescribed_dofs[I]
     ch.inhomogeneities .= ch.inhomogeneities[I]
+    ch.affine_inhomogeneities .= ch.affine_inhomogeneities[I]
     ch.dofcoefficients .= ch.dofcoefficients[I]
 
     copy!(ch.free_dofs, setdiff(1:ndofs(ch.dh), ch.prescribed_dofs))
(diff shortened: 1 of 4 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -377,6 +404,25 @@ function update!(ch::ConstraintHandler, time::Real=0.0)
         _update!(ch.inhomogeneities, dbc.f, dbc.faces, dbc.field_name, dbc.local_face_dofs, dbc.local_face_dofs_offset,
                  dbc.components, ch.dh, ch.bcvalues[i], ch.dofmapping, ch.dofcoefficients, convert(Float64, time))
     end
+    # Compute effective inhomogeneity for affine constraints with prescribed dofs in the
+    # RHS. For example, in u2 = w3 * u3 + w4 * u4 + b2 we allow e.g. u3 to be prescribed by
+    # a trivial constraint with just an inhomogeneity (e.g. DBC), for example u3 = f(t).
+    # This value have now been computed in _update! and we can compute the effective
+    # inhomogeneity h2 for u2 which becomes h2 = w3 * u3 + b2 = w3 * f3(t) + b2.
+    for i in eachindex(ch.prescribed_dofs, ch.dofcoefficients, ch.inhomogeneities)
```
lijas, line 413: `for (dof, coeffs, h, ac_h) for zip(ch.prescribed_dofs, ch.dofcoefficients, ch.inhomogeneities, ch.affine_inhomogeneities)
` instead ?
(If we make `ch.affine_inhomogeneities` in to a vector instead.)

src/Dofs/ConstraintHandler.jl
```diff
@@ -76,6 +76,7 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     prescribed_dofs::Vector{Int}
     free_dofs::Vector{Int}
     inhomogeneities::Vector{T}
+    affine_inhomogeneities::Dict{Int,T}
```
lijas, line 79: Is it better to make it  a vector instead (same size as `inhomogeneities`), since we will only acces it when looping over the precribed_dofs?

src/Dofs/ConstraintHandler.jl
fredrikekre: I need the index to insert the result so thought this was cleaner.

lijas: Ok, I missed that.

PR #436, opened 2022-03-08
More efficient way of creating condensed sparsity pattern

lijas:
Creating the sparsity pattern with affine constraints was extremely slow. 

```
grid = generate_grid(Quadrilateral, (300,300))

dh = DofHandler(grid)
push!(dh, :u, 2)
close!(dh);

ch_periodic = ConstraintHandler(dh);
periodic = PeriodicDirichlet(
    :u,
    ["left" => "right", "bottom" => "top"],
    [1, 2]
)
add!(ch_periodic, periodic)
close!(ch_periodic)
update!(ch_periodic, 0.0)

@time K = create_sparsity_pattern(dh, ch_periodic)
```

```
Before:  11.439089 seconds (239.59 k allocations: 355.780 MiB, 0.22% gc time)
Now:       0.244780 seconds (181.27 k allocations: 411.271 MiB, 10.16% gc time)
```

A lot of %gc time still which I dont understand

PR #537, opened 2022-11-25
Remove some unnecessary allocations in affine constraint condensation

fredrikekre:
This patch removes some unnecessary allocations of a sparse matrix column which was needed before when the sparse matrix changed in-place. Since #436 we create a new matrix for the new entries and then add them in the end instead.

Comparison for the Stoke example in the docs:
- Master:
  ```julia
  julia> @time create_sparsity_pattern(dh, ch);
    0.009940 seconds (2.91 k allocations: 13.387 MiB)
  ```
- PR:
  ```julia
  julia> @time create_sparsity_pattern(dh, ch);
    0.004960 seconds (66 allocations: 12.581 MiB)
  ```

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -686,7 +686,7 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
         col_coeffs = coefficients_for_dof(dofmapping, dofcoefficients, col)
         if col_coeffs === nothing
             for ri in nzrange(K, col)
-                row = @inbounds K.rowval[ri]
+                row = K.rowval[ri]
                 row_coeffs = coefficients_for_dof(dofmapping, dofcoefficients, row)
                 row_coeffs === nothing && continue
                 for (d, _) in row_coeffs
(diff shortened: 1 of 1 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -683,13 +683,10 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
...
     for col in 1:ndofs
-        # Since we will possibly be pushing new entries to K, the field K.rowval will grow.
-        # Therefor we must extract this before iterating over K
-        range = nzrange(K, col)
-        _rows = K.rowval[range]
         col_coeffs = coefficients_for_dof(dofmapping, dofcoefficients, col)
         if col_coeffs === nothing
-            for row in _rows
+            for ri in nzrange(K, col)
+                row = @inbounds K.rowval[ri]
```
KristofferC, line 689: This `@inbounds` feel kind of silly.

fredrikekre:
```suggestion
                row = K.rowval[ri]
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -700,7 +697,8 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
                 end
             end
         else
-            for row in _rows
+            for ri in nzrange(K, col)
+                row = @inbounds K.rowval[ri]
```
fredrikekre, line 701:
```suggestion
                row = K.rowval[ri]
```

lijas (approved):

PR #538, opened 2022-11-25
Remove `create_sparsity_pattern(::MixedDofHandler)`

fredrikekre:
This patch removes the special method for creating a sparsity pattern of a `MixedDofHandler` and slightly generalizes the one for `DofHandler`. The removed method was almost identical to the one for `DofHandler` except that it allocated a new global dof vector with the correct size for every element. The only necessary generalization is thus to call `resize!` on the global dof vector for every element. This should result in fewer allocations when creating a sparsity pattern for a `MixedDofHandler`.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -333,7 +333,7 @@ with stored values in the correct places.
 
 See the [Sparsity Pattern](@ref) section of the manual.
 """
-create_sparsity_pattern(dh::DofHandler) = _create_sparsity_pattern(dh, nothing, false)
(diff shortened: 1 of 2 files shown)


kimauth (approved):
Cool, looking into these extra allocations that  `create_sparsity_pattern(dh::MixedDofHandler)` had was still on my to-do list for joining the dofhandlers! 🙂 
On the same `grid` both dofhandlers allocate equally little now. 

I don't understand what happened on the windows CI though, seems to be unrelated to this PR?

KristofferC:
Some regression in win 1.8.3

PR #539, opened 2022-11-28
Sparsity patterns without constrained dofs

fredrikekre:
This patch adds a new keyword argument `keep_constrained::Bool=true` to
`create_sparsity_pattern` which, when set to `false`, do not create
entries in the global matrix for rows/values that correspond to
constrained degrees of freedom. Naturally this can only be used with
local condensation of constraints (cf. https://github.com/Ferrite-FEM/Ferrite.jl/pull/528).

Another necessary change included in this patch is that assembly now
ignores entries in the local matrix that are zero (in particular
rows/columns in the local matrix which have been zeroed out in the local
condensation).


lijas:
Would be nice to see some benchmarks on how much memory is saved (in for example the heat-example :) )

fredrikekre:
Basically none :D The number of constrained dofs are usually negligible compared to free dofs.

PR #541, opened 2022-11-28
Skip further lookup work in constraint condensation if current matrix entry is zero.

fredrikekre:


src/Dofs/ConstraintHandler.jl
```diff
@@ -745,11 +745,12 @@ function _condense!(K::SparseMatrixCSC, f::AbstractVector, dofcoefficients::Vect
         col_coeffs = coefficients_for_dof(dofmapping, dofcoefficients, col)
         if col_coeffs === nothing
             for a in nzrange(K, col)
+                Kval = K.nzval[a]
+                iszero(Kval) && continue
```
lijas, line 749: Kval will most likely not be zero, right? So maybe it is an unnecessary check?

fredrikekre: I don't think it is that uncommon? I.e. if you have no coupling between some dofs?

lijas: From the heat_equation tutorial, the stiffness matrix has no zero values (before we call apply!) i.e. `count(i-> i==0, K.nzval) == 0` . But idk, maybe it has many zeros it in other cases?

KristofferC: That has only one field. But if you have multiple fields it's often that some fields don't couple in the weak form.

PR #545, opened 2022-12-04
Implement dof renumbering by fields or components

fredrikekre:
This patch adds functionality to renumber dofs in a DofHandler by field
or by components, resulting in globally blocked numbering.

After this, perhaps we could require renumbering by field before calling
`global_dof_range` (#457) since it can then return a range? The current
implementation in #457 could still be useful perhaps, but then in some
kind of DofTools namespace?

For review: First commit just moves code around, second commit is new stuff.


KnutAM:
One use case for #457 that I'm using is residual scaling. 
I also used it for testing in #532 
For these cases, I think it would be nice not to require renumbering first.

PR #546, opened 2022-12-05
Rework CellIterator by extracting out the caching part

fredrikekre:
This patch reworks the CellIterator by extracting out the caching parts to its own CellCache struct that can be used independently. Elements of a CellIterator are now CellCache instead of the CellIterator itself.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -306,16 +306,16 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcfaces::Set{Index}, inter
 
     # loop over all the faces in the set and add the global dofs to `constrained_dofs`
     constrained_dofs = Int[]
+    cc = CellCache(ch.dh, UpdateFlags(; nodes=false, coords=false, dofs=true))
     for (cellidx, faceidx) in bcfaces
         if cellidx ∉ cellset
             delete!(dbc.faces, Index(cellidx, faceidx))
             continue # skip faces that are not part of the cellset
         end
-        _celldofs = fill(0, ndofs_per_cell(ch.dh, cellidx))
-        celldofs!(_celldofs, ch.dh, cellidx) # extract the dofs for this cell
+        reinit!(cc, cellidx)
         r = local_face_dofs_offset[faceidx]:(local_face_dofs_offset[faceidx+1]-1)
-        append!(constrained_dofs, _celldofs[local_face_dofs[r]]) # TODO: for-loop over r and simply push! to ch.prescribed_dofs
-        @debug println("adding dofs $(_celldofs[local_face_dofs[r]]) to dbc")
+        append!(constrained_dofs, cc.dofs[local_face_dofs[r]]) # TODO: for-loop over r and simply push! to ch.prescribed_dofs
+        @debug println("adding dofs $(cc.dofs[local_face_dofs[r]]) to dbc")
     end
 
     # save it to the ConstraintHandler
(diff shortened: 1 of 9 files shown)


lijas:
Nice!

src/Dofs/ConstraintHandler.jl
```diff
@@ -352,18 +352,16 @@ function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpo
     ncomps = length(dbc.components)
     nnodes = getnnodes(ch.dh.grid)
     interpol_points = getnbasefunctions(interpolation)
-    _celldofs = fill(0, ndofs_per_cell(ch.dh, first(cellset)))
     node_dofs = zeros(Int, ncomps, nnodes)
     visited = falses(nnodes)
-    for cell in CellIterator(ch.dh, collect(cellset)) # only go over cells that belong to current FieldHandler
-        celldofs!(_celldofs, cell) # update the dofs for this cell
+    for cell in CellIterator(ch.dh, cellset) # only go over cells that belong to current FieldHandler
```
lijas, line 357: Should we use the CellChache internally? In that case you could do 

```
cc = CellCache(dh, UpdateFlags=(..))
for cellid in cellset
   reinit!(cc, cellid)
    ....
```

src/iterators.jl
```diff
@@ -1,145 +1,190 @@
...
 
-function check_compatible_geointerpolation(cv::Union{CellValues, FaceValues}, ci::CellIterator)
-    if length(getnodes(ci)) != getngeobasefunctions(cv)
-        msg = """The given CellValues and CellIterator are incompatiblet.
-        Likely an appropriate geometry interpolate must be passed when constructing the CellValues.
-        See also issue #265: https://github.com/Ferrite-FEM/Ferrite.jl/issues/265"""
-        throw(ArgumentError(msg))
-    end
+# reinit! FEValues with CellCache
+reinit!(cv::CellValues, cc::CellCache) = reinit!(cv, cc.coords)
```
lijas, line 103: Personally i never liked this method...I think it hides to much information to new users.
Using `reinit!(cv, getcoordinates(cc))` is clearer i think.

src/iterators.jl
```diff
@@ -1,145 +1,190 @@
...
-    else
-        return (reinit!(ci, state), state+1)
+# TODO: Can always resize and combine the two reinit! methods maybe?
+function reinit!(cc::CellCache, i::Int)
+    cc.cellid[] = i
+    if cc.flags.nodes
+        cellnodes!(cc.nodes, cc.grid, i)
+    end
+    if cc.flags.coords
+        cellcoords!(cc.coords, cc.grid, i)
```
lijas, line 76: Here you can change to `getcoordinates`, since `dh isa DofHandler` and has concrete cell type? And we will probably soon deprecate `cellcoords!`

src/iterators.jl
```diff
@@ -1,145 +1,190 @@
...
-        if ci.dh !== nothing && ci.dh isa MixedDofHandler
-            cellnodes!(ci.nodes, ci.dh, ci.current_cellid[])
-        else
-            cellnodes!(ci.nodes, ci.grid, ci.current_cellid[])
-        end
+function reinit!(cc::CellCache{<:Any,<:AbstractGrid,<:MixedDofHandler}, i::Int)
+    @assert cc.dh isa MixedDofHandler
+    cc.cellid[] = i
+    if cc.flags.nodes
+        resize!(cc.nodes, nnodes_per_cell(cc.dh, i))
```
lijas, line 88: This is new behavior, since previously we would have gotten an error if we loop over a cellset with different cell types. Maybe this is better :thinking: @kimauth ?

lijas (commented):
Nice

src/iterators.jl
fredrikekre: Sure, but could be in a cleanup PR, this is just replicating the functionality we already have :)

src/iterators.jl
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L668-L685 ?

PR #552, opened 2022-12-12
Introduce getgrid(dh::AbstractDofHandler) function

KnutAM:
I find myself quite often using `dh.grid`, and that is also the case in Ferrite's source code. 
While this is fine for both `DofHandler` and `MixedDofHandler`, it feels "dangerous" for `AbstractDofHandler`, and I think it would be nice to have an access function here.

If you think this is a good idea, like [and subscribe]:)
Then I will go through the rest of the source code and update it. 
For now, only DofHandler.jl is updated as a demo


fredrikekre:
I am at least a sucker for field access, and disliker of `get*` functions :)

termi-official:
Note: This is related to #486 . I propagated getgrid there through most of the code as a compatibility layer against non-distributed assembly.

KnutAM:
> Note: This is related to #486 . I propagated getgrid there through most of the code as a compatibility layer against non-distributed assembly.

Ah, great 👍  - I missed that you had already done that there!
What's the status/plan with that one? 
Will that be merged as a whole, or should such small changes be pulled out as separate prs and merged individually?

termi-official:
No problem. I have to admit that the PR is really hard to read, which is why I am trying to break the PR into several smaller ones, fixing one open issue after another. Two major things which I want to consider first is 
1. propagating the abstraction layer through the codebase and addressing some coupling issues regarding mixed-dimensional and mixed-element meshes (see #519)
2. merge the dof handlers, because otherwise I have to write two separate dof handlers for the distributed assembly, or possibly a third, incompatible one, none of which is satisfactory for me.

However, since I am currently not having time to address these issues properly, I am fine if you cherry pick some changes, polish them, and merge them.

KnutAM:
Closing as handled in #486

PR #553, opened 2022-12-13
fix DiscontinuousLagrange value

koehlerson:
someone forgot to test :disguised_face:  @termi-official


termi-official:
Derp. Not sure how that happened.

PR #556, opened 2022-12-14
RCF: JuliaFormatter

KnutAM:
Tested to use JuliaFormatter on the grid generators (which should be a pretty "hard" example).
* ac9ba30b34183e748a8293598dcda919ef0c0fea: Automatic formatting
* 60640efc990cd7ba0b9d85a8af8bc4ed180e9762: After manual tuning

It doesn't seem to work that well, so I don't propose that we should have it as CI.
But it might be nice to have a recommendation in the contribution section to check formatting (on the changes) before requesting a review. (It also highlights what is the desired formatting for white spaces etc.)
Probably, the settings should be tuned a bit - it is especially bad at keeping manual alignments for multiple lines which there were many of in grid_generators...


KnutAM:
Closing as this doesn't seem to be that useful

PR #557, opened 2022-12-14
fix triangle grid generator and correction in discontinuous lagrange reference coordinate for order 0

koehlerson:
Found this with @termi-official 
## Triangle Grid Generator
Before this the `grid_generator` for `Triangle` constructed something like this:

```
4----5
|\   |
|  \ |
1----2

Triangle((1,2,4))
Triangle((2,5,4))
```

Now the following triangles are returned
```
Triangle((2,4,1))
Triangle((4,2,5))
```

which is in line with the definition of the interpolation
image

## Discontinuous Lagrange Reference coordinate

Dispatch was just wrong. (0,0) is not the middle node of the triangle

### DIFF
--- a/src/Grid/grid_generators.jl
+++ b/src/Grid/grid_generators.jl
@@ -311,16 +311,16 @@ function generate_grid(::Type{Triangle}, nel::NTuple{2,Int}, LL::Vec{2,T}, LR::V
     node_array = reshape(collect(1:n_nodes), (n_nodes_x, n_nodes_y))
     cells = Triangle[]
     for j in 1:nel_y, i in 1:nel_x
-        push!(cells, Triangle((node_array[i,j], node_array[i+1,j], node_array[i,j+1]))) # ◺
-        push!(cells, Triangle((node_array[i+1,j], node_array[i+1,j+1], node_array[i,j+1]))) # ◹
+        push!(cells, Triangle((node_array[i+1,j], node_array[i,j+1], node_array[i,j]))) # ◺
+        push!(cells, Triangle((node_array[i,j+1], node_array[i+1,j], node_array[i+1,j+1]))) # ◹
     end
 
     # Cell faces
     cell_array = reshape(collect(1:nel_tot),(2, nel_x, nel_y))
-    boundary = FaceIndex[[FaceIndex(cl, 1) for cl in cell_array[1,:,1]];
-                               [FaceIndex(cl, 1) for cl in cell_array[2,end,:]];
-                               [FaceIndex(cl, 2) for cl in cell_array[2,:,end]];
-                               [FaceIndex(cl, 3) for cl in cell_array[1,1,:]]]
+    boundary = FaceIndex[[FaceIndex(cl, 3) for cl in cell_array[1,:,1]];
+                               [FaceIndex(cl, 2) for cl in cell_array[2,end,:]];
+                               [FaceIndex(cl, 3) for cl in cell_array[2,:,end]];
+                               [FaceIndex(cl, 2) for cl in cell_array[1,1,:]]]
 
     boundary_matrix = boundaries_to_sparse(boundary)
 
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -153,9 +153,18 @@ function value(ip::DiscontinuousLagrange{dim,shape,order}, i::Int, ξ::Vec{dim})
 end
 
 # Excepting the L0 element.
-function reference_coordinates(ip::DiscontinuousLagrange{dim,shape,0}) where {dim,shape}
+function reference_coordinates(ip::DiscontinuousLagrange{dim,RefCube,0}) where dim
     return [Vec{dim, Float64}(ntuple(x->0.0, dim))]
 end
+
+function reference_coordinates(ip::DiscontinuousLagrange{2,RefTetrahedron,0})
+    return [Vec{2,Float64}((1/3,1/3))]
+end
+
+function reference_coordinates(ip::DiscontinuousLagrange{3,RefTetrahedron,0})
+   return [Vec{3,Float64}((1/4,1/4,1/4))]
+end
+
 function value(ip::DiscontinuousLagrange{dim,shape,0}, i::Int, ξ::Vec{dim}) where {dim,shape}
     return 1.0
 end


termi-official:
Ping @fredrikekre at least the triangle/tet generators do generate grids with wrong nodal assignments (just the orientation is correct). Will take a look into the hex generator if I find time.

KnutAM:
How can either the old or new definitions be "wrong"? 
Cannot the changes in the definition of, e.g. cell 1, be obtained by shearing and rotating?
(Asking because if I misunderstand, I should probably add more checks to the FerriteMeshParser)

koehlerson:
Hm so I'm not totally sure if its per se wrong, but I'm also not sure how to handle this as a user when you have a solution field. We took this example:
grafik

which is now based on the old definition of the `Triangle` generator, i.e.
```
julia> grid.cells
8-element Vector{Triangle}:
 Triangle((1, 2, 4))
 Triangle((2, 5, 4))
...
```
If you extract the dofs of the second cell, you get 
```
julia> celldofs(dh,2)
3-element Vector{Int64}:
 2
 4
 3
```
Accessing with this the dof vector gives you 
```
julia> u[[2,4,3]]
3-element Vector{Float64}:
 0.0
 0.25
 0.0
```
which is also correct.

However, the problem is, if I'm locally on an element and I want to compute now based on this `u[[2,4,3]]` anything, lets say the gradient or function value, then I'd assume that `u[[2,4,3]]` is in line with the ordering of the interpolation, i.e. the first local dof is associated with the first shape function and so on, but this is not the case. The value `0.25` is associated with the the locally third shape function but assuming the order is correct would then give an association with the second shape function.

I'm not sure if this makes the point clearer. Maybe there is another way to extract the permutation based on the local dof range `[2,4,3]`, but it feels a bit strange that the provided `grid_generator` return rotated triangles

koehlerson:
> But I think what you ask for is exactly what the cellvalues are doing? It is `reinit!`:ed with the cell coordinates which maps it back to the reference element. Can't you just create a cellvalues object (potentially with a different integration rule if required) to do those operations? How would one otherwise handle elements that are 45 degrees rotated (without just reimplementing the functionality of the cellvalues)?

Then my counter question is, why is nobody using cellvalues when exporting to paraview :D Or how can the paraview export be rotation invariant

KnutAM:
I guess you can do that since you can take the interpolation from the dofhandler?

koehlerson:
Yes the interpolation is within the dofhandler. Looking at `reinit!` I don't really see where the code lies that takes care of such rotations in terms of geometric/algebraic interpolation permutation

I tried to get the permutation based on 

```
local_idx = [2,4,3]
_perm = sortperm(local_idx)
u^e = u[local_idx[_perm]]
```

but I think this doesnt work for some special cases.

All of this manifests basically in such a behavior:
grafik
which should look symmetric like this:
grafik


For quads it works ootb, because they don't rotate in the grid_generator

KnutAM:
> Yes the interpolation is within the dofhandler. Looking at `reinit!` I don't really see where the code lies that takes care of such rotations in terms of geometric/algebraic interpolation permutation

Should be this part, this should be the isoparametric mapping: https://github.com/Ferrite-FEM/Ferrite.jl/blob/570b3b5fa69d0ca56addf82dd9c8bbf790eec860/src/FEValues/cell_values.jl#L167

termi-official:
> How can either the old or new definitions be "wrong"? Cannot the changes in the definition of, e.g. cell 1, be obtained by shearing and rotating? (Asking because if I misunderstand, I should probably add more checks to the FerriteMeshParser)

It's not the definition being wrong. The generator gives you a grid with triangles, that are not consistent with how triangles are defined topologically in Ferrite.jl. 

> But I think what you ask for is exactly what the cellvalues are doing?

I started unwrapping everything because I had issues figuring out what was going wrong in the gradient visualization code which I wrote with Maxi. But to answer your question, yes, we could use cellvalues here, too, and no, it won't fix the issue.

> It is `reinit!`:ed with the cell coordinates which maps it back to the reference element. Can't you just create a cellvalues object (potentially with a different integration rule if required) to do those operations? How would one otherwise handle elements that are 45 degrees rotated (without just reimplementing the functionality of the cellvalues)?

I think we might be able to do something like this. However, the mapping you talk about is the geometric transformation from the grid triangle to a reference. This is a different matter than the topological mapping. I am not sure how exactly it happens that the dofs gets associated to the wrong entity. I think the assembly with test and trial space equally is not affected due to the symmetry of the functionals in this example.

KnutAM:
> I think we might be able to do something like this. However, the mapping you talk about is the geometric transformation from the grid triangle to a reference. This is a different matter than the topological mapping. I am not sure how exactly it happens that the dofs gets associated to the wrong entity. I think the assembly with test and trial space equally is not affected due to the symmetry of the functionals in this example.

Perhaps I misunderstand, but I thought that all calculations, including dof-ordering, are only defined on the reference element. 
Consequently, it is necessary to first do the geometric transformation to the reference and then do any operations. 
I.e. in the above examples, the dof-ordering on the reference element is always the same, but the mapping back is different as the element nodal coordinates are different? I.e. `celldofs` will always give the dof ordering for the reference element.

koehlerson:
I'm a bit confused, isn't then the coupling between the trial and test functions off? I'm asking because they do not change their order after reinit (which is totally fine)

```
julia> cellvalues.dNdξ
3×3 Matrix{Vec{2, Float64}}:
 [1.0, 0.0]    [1.0, 0.0]    [1.0, 0.0]
 [0.0, 1.0]    [0.0, 1.0]    [0.0, 1.0]
 [-1.0, -1.0]  [-1.0, -1.0]  [-1.0, -1.0]

julia> cellvalues.dNdx
3×3 Matrix{Vec{2, Float64}}:
 [0.0, -1.0]  [0.0, -1.0]  [0.0, -1.0]
 [1.0, 1.0]   [1.0, 1.0]   [1.0, 1.0]
 [-1.0, 0.0]  [-1.0, 0.0]  [-1.0, 0.0]

julia> Ferrite.derivative(cellvalues.func_interp,Vec{2}((1.0,1.0)))
3-element Vector{Vec{2, Float64}}:
 [1.0, 0.0]
 [0.0, 1.0]
 [-1.0, -1.0]
```

The above is reinited for cell 2

koehlerson:
Just tried it, the `Jinv` part in the `reinit!` does take care about rotations according to the `reinit!` ed values in cell 2

```
julia> cellvalues.dNdx
3×3 Matrix{Vec{2, Float64}}:
 [0.0, -1.0]  [0.0, -1.0]  [0.0, -1.0]
 [1.0, 1.0]   [1.0, 1.0]   [1.0, 1.0]
 [-1.0, 0.0]  [-1.0, 0.0]  [-1.0, 0.0]

julia> Ferrite.derivative(cellvalues.func_interp,Vec{2}((1.0,1.0)))
3-element Vector{Vec{2, Float64}}:
 [1.0, 0.0]
 [0.0, 1.0]
 [-1.0, -1.0]
```

but now I'm wondering how the function values is then computed correctly, since only `dNdx` is rotated and not 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/570b3b5fa69d0ca56addf82dd9c8bbf790eec860/src/FEValues/common_values.jl#L148

At this point I'm very confused :D

fredrikekre:
I am confused here, just like Knut. What would you define as the correct order for a random triangle like
```
                    /-   
                  /- |   
                /-   |   
             /--     \   
           /-         |  
         /-           |  
      /--             |  
    /-                |  
  /-                  \  
--                     | 
  ---\                 | 
      ---\             | 
          ---\         \ 
              ---\      |
                  ---\  |
                      --|
```
?

koehlerson:
redefine the mesh in terms of quads :disguised_face: 

I think its not about saying that there is an ordering in any unstructured case. I was confused by two reasons

1. I expected that `celldofs(dh,cellid)` returns the `i`-th dof associated with the `i`-th base function, which is just a wrong assumption as it seems
2. The grid generator triangles are rotated. This is not problematic, but I would think that this is confusing for beginners if you look up the interpolation definitions and then you take a look at `grid.cells` and you see its permuted. Especially having in mind the `detJ negative` issue with external grids. 

So it seems that our code that computes the gradient/fluxes is just not rotation invariant (because the dofhandler isn't?) and I currently don't see a way without cellvalues/reimplementing cellvalues functionality

fredrikekre:
> I expected that celldofs(dh,cellid) returns the i-th dof associated with the i-th base function, which is just a wrong assumption as it seems

This should be true, do you have an example where it is not?

koehlerson:
> > I expected that celldofs(dh,cellid) returns the i-th dof associated with the i-th base function, which is just a wrong assumption as it seems
> 
> This should be true, do you have an example where it is not?

grafik

Yes my introduction example yields 
```
julia> grid.cells
8-element Vector{Triangle}:
 Triangle((1, 2, 4))
 Triangle((2, 5, 4))
...

julia> celldofs(dh,2)
3-element Vector{Int64}:
 2
 4
 3

julia> u[[2,4,3]]
3-element Vector{Float64}:
 0.0
 0.25
 0.0
```

but the 0.25 is associated with the third basis function, or am I missing something?

termi-official:
Can we cherry pick the fix for Lagrange 0 so we do not forget this again?

PR #559, opened 2022-12-15
Fixes the discontinuous lagrange reference coords dispatch

koehlerson:


src/interpolations.jl
```diff
@@ -153,9 +153,18 @@ function value(ip::DiscontinuousLagrange{dim,shape,order}, i::Int, ξ::Vec{dim})
...
     return [Vec{dim, Float64}(ntuple(x->0.0, dim))]
 end
+
+function reference_coordinates(ip::DiscontinuousLagrange{2,RefTetrahedron,0})
+    return [Vec{2,Float64}((1/3,1/3))]
+end
+
+function reference_coordinates(ip::DiscontinuousLagrange{3,RefTetrahedron,0})
+   return [Vec{3,Float64}((1/4,1/4,1/4))]
+end
```
termi-official, line 166: We had a general solution via ntuple, right?

termi-official (changes requested):

src/interpolations.jl
koehlerson: no :D that's 1:1 the piece of code of the other branch

termi-official (approved):

fredrikekre:
Add a test? :)

termi-official:
I already tried a bit of stuff to test interpolations, but I cannot find a good angle. Do you have something in mind how we can improve interpolation testing @fredrikekre ?

PR #566, opened 2022-12-18
Make Assembler a more complete assembler

fredrikekre:


fredrikekre:
Superseded by #1058 (more or less).

PR #567, opened 2022-12-19
Add package extension for BlockArrays.jl.

fredrikekre:
This patch adds support for global block matrices. A global block matrix
can be obtained by `create_sparsity_pattern(BlockMatrix, dh, ch)` or by
`create_sparsity_pattern(A::BlockMatrix, dh, ch)`. The former infers the
block sizes by looking at the fields of the DofHandler, and the latters
uses the blocksizes of the (partially) initilized BlockMatrix (i.e. by
`undef_blocks`).

For global assembly, `start_assemble(::BlockMatrix, ...)` returns a
`BlockAssembler` which pre allocates a vector for pre-computing the
(block number, local index) pairs for each DoF. BlockArrays only support
local condensation of constraints currently and you have to use
`apply_assemble!` instead of `assemble!` + `apply!`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,9 +13,11 @@ Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
 [weakdeps]
+BlockArrays = "8e7c35d0-a365-5155-bbbb-fb81a777f24e"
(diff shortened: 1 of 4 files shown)


KnutAM:
Nice! 
Small question, what does `dof_range(dh, :u)` return in this case?

fredrikekre:
No change, the local matrix is always blocked even when not using a globally blocked matrix.

PR #570, opened 2022-12-21
Make showgrid work for all cells

KnutAM:
Ref #420 
Hit this problem again, would be nice to resolve this if possible!
Making new PR from a new branch as history got messed up when merging with master...

*Side note: Came up when checking the discussion of how to parameterize the cell type, as the standard quadratic quad from Abaqus is not supported as this has `Serendipity`-type interpolation (8 nodes)...*

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -36,22 +36,25 @@ nnodes(c::C) where {C<:AbstractCell} = nnodes(typeof(c))
 nnodes(::Type{<:AbstractCell{dim,N,M}}) where {dim,N,M} = N
 
 # Typealias for commonly used cells
-const Line  = Cell{1,2,2}
-const Line2D = Cell{2,2,1}
-const Line3D = Cell{3,2,0}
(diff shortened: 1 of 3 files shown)


src/Grid/grid.jl
```diff
@@ -708,25 +711,11 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", grid::Grid)
     print(io, "$(typeof(grid)) with $(getncells(grid)) ")
-    typestrs = sort!(collect(Set(celltypes[typeof(x)] for x in grid.cells)))
-    str = join(io, typestrs, '/')
+    typestrs = sort!(collect(Set(repr(typeof(x)) for x in grid.cells)))
```
fredrikekre, line 714: This seems to be really slow, try showing something like `generate_grid(Triangle, (1000, 1000))` in the REPL.

KnutAM: Sorry, missed that performance regression. Fixed in #559.

PR #575, opened 2023-01-04
AffineConstraint for nonzero b and nonlinear problem + docfixes

KnutAM:
This PR fixes an issue recently discussed on Slack. The test code and derivations are included here for reference.
In addition, it adds further docstrings relevant when using AffineConstraints. 

<details>
<summary>Test code</summary>
Without this PR, the nonlinear problem doesn't converge. 
With this PR, the code shows how the wrong results are obtained if using the `a=-K\r` instead of `a=K\f` for the linear solution method, which this PR documents. 

```julia
using Ferrite

params = (k=1.0, f=1.0, a=1.0, b=0.2, tol=1e-10, maxiter=2)

grid = generate_grid(Line, (2,))
addfaceset!(grid, "center", x->x[1]≈0.0)

dh = DofHandler(grid)
push!(dh, :u, 1)
close!(dh)

eval_norm(r, ch) = sqrt(sum(i->r[i]^2, Ferrite.free_dofs(ch)))

function doassemble!(K, r, dh, a, params)
    # Spring elements 
    k = params.k
    Ke = [k -k; -k k]
    # Quick and dirty assem
    assembler = start_assemble(K, r)
    for cell in CellIterator(dh)
        ae = a[celldofs(cell)]
        re = Ke*ae
        assemble!(assembler, celldofs(cell), Ke, re)
    end
    r[3] -= params.f # Force on the main dof (right side)
end

ch = ConstraintHandler(dh)
add!(ch, Dirichlet(:u, getfaceset(grid, "center"), (x,t)->Vec{1}((0.0,))))
add!(ch, AffineConstraint(1, [3=>params.a], params.b))
close!(ch)

r = zeros(ndofs(dh))

a = zeros(ndofs(dh))
update!(ch, 0.0)

# Nonlinear solution
apply!(a, ch) 
for niter = 0:params.maxiter
    doassemble!(K, r, dh, a, params)
    apply_zero!(K, r, ch)
    rnorm = eval_norm(r, ch)
    println("$niter: $rnorm, $r, $a")
    rnorm < params.tol && break
    Δa = -K\r 
    apply_zero!(Δa, ch)
    a .+= Δa
end
eval_norm(r, ch) > params.tol && @warn "Did not converge"
a_nonlinear = copy(a)

# Linear solution (r)
a_linear_r = zero(a)
doassemble!(K, r, dh, a_linear_r, params)
apply!(K, r, ch)
a_linear_r = -K\r 
apply!(a_linear_r, ch)

# Linear solution (f)
a_linear_f = zero(a)
doassemble!(K, r, dh, a_linear_f, params)
f = -r
apply!(K, f, ch)
a_linear_f = K\f
apply!(a_linear_f, ch)

# Analytical comparison
a3 = (params.f/params.k - params.b)/(1 + params.a)
a_analytical = [params.a*a3+params.b 0.0 a3]
@show a_linear_r
@show a_linear_f
@show a_nonlinear
@show a_analytical
```
</details>

<details>
<summary>Analytical solution</summary>

# Spring
   |------ $c$ ------| 

   |-- $k$ --|-- $k$ --|--> $f$

   $u_1$ -> $u_2$ -> $u_3$ ->

 **Kinematical constraints**
 - Dirichlet: $u_2 = 0$
 - Affine: $u_1 = au_3 + b$
  
 **Free body diagrams**
 
 <- $ku_1$ -- (1) -- $f_c$ ->
 
 <- $(f_c+ku_3)$ -- (3) -- $f$ -> 

where $f_c$ is the affine constraint force and $f$ is the external force

**Equilibrium**
- $0 = f_1 = -k u_1 + f_c$
- $0 = f_3 = -k u_3 - f_c + f$
- $0 = f_1+f_3 = -k (u_1 + u_3)$
- $f = k u_1 + k u_3$
- $f = k (u_3 (a+1) + b)$
- $u_3 = (f/k - b)/(1+a)$ 

</details>

### DIFF
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -7,10 +7,10 @@ DocTestSetup = :(using Ferrite)
 PDEs can in general be subjected to a number of constraints, 
 
 ```math
-g_I(\boldsymbol{a}) = 0, \quad I = 1 \text{ to } n_c
+g_I(\underline{a}) = 0, \quad I = 1 \text{ to } n_c
 ```
 
-where $g$ are (non-linear) constraint equations, $\boldsymbol{a}$ is a vector of the
+where $g$ are (non-linear) constraint equations, $\underline{a}$ is a vector of the
 degrees of freedom, and $n_c$ is the number of constraints. There are many ways to
 enforce these constraints, e.g. penalty methods and Lagrange multiplier methods. 
 
@@ -43,7 +43,7 @@ K = create_sparsity_pattern(dh, ch)
 ```
 
 ### Solving linear problems
-To solve the system ``\boldsymbol{K}\boldsymbol{a}=\boldsymbol{f}``, account for affine constraints the same way as for 
+To solve the system ``\underline{\underline{K}}\underline{a}=\underline{f}``, account for affine constraints the same way as for 
 `Dirichlet` boundary conditions; first call `apply!(K, f, ch)`. This will condense `K` and `f` inplace (i.e
 no new matrix will be created). Note however that we must also call `apply!` on the solution vector after 
 solving the system to enforce the affine constraints:
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -43,8 +43,8 @@ K = create_sparsity_pattern(dh, ch)
 ```
 
 ### Solving linear problems
-When solving the system `K⋅a=f`, we account for the affine constraints in the same way as we account for 
-`Dirichlet` boundary conditions; by first calling `apply!(K, f, ch)`. This will condense `K` and `f` inplace (i.e
+To solve the system ``\boldsymbol{K}\boldsymbol{a}=\boldsymbol{f}``, account for affine constraints the same way as for 
+`Dirichlet` boundary conditions; first call `apply!(K, f, ch)`. This will condense `K` and `f` inplace (i.e
 no new matrix will be created). Note however that we must also call `apply!` on the solution vector after 
 solving the system to enforce the affine constraints:
 
@@ -63,15 +63,16 @@ It is important to check the residual **after** applying boundary conditions whe
 solving nonlinear problems with affine constraints. 
 `apply_zero!(K, r, ch)` modifies the residual entries for dofs that are involved 
 in constraints to account for constraint forces. 
-The following pseudo-code shows the typical pattern for solving a non-linear problem with Newtons method:
+The following pseudo-code shows a typical pattern for solving a non-linear problem with Newtons method:
 ```julia
 a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
+Δa = similar(a)         # Preallocate Δa
 apply!(a, ch)           # Make the guess fulfill all constraints in `ch`
 for iter in 1:maxiter
     doassemble!(K, r, ...)
     apply_zero!(K, r, ch)   # Modify `K` and `r` to account for the constraints. 
     check_convergence(r, ...) && break # Only check convergence after `apply_zero!(K, r, ch)`
-    Δa=-K\r                 # Calculate update
+    Δa .-= K \ r            # Calculate update
     apply_zero!(Δa, ch)     # Change constrained values such that `a+Δa` fullfills constraints provided that `a` does.
     a .+= Δa
 end
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -58,13 +58,12 @@ apply!(a, ch) # enforces affine constraints
 
 ```
 
-Note that replacing `f` with `r=-f` and solving `a = -K\r` in the example above will not work correctly for affine constraints.
-In this case, use the procedure for nonlinear problems below. 
+Note: Replacing `f` with `r=-f` and solving `a = -K\r` in the example above may give the wrong result.
+Instead, use the procedure for nonlinear problems when applying affine constraints to such problem formulations.
 
 ### Solving nonlinear problems
-When solving the nonlinear system `r(a)=0` using a Newton-type of method
-the updated guess is given by `Δa=-K\r` 
-(where `K=∂r/∂a` if the standard Newton's method is used).
+A Newton-type solution method for the nonlinear system `r(a)=0` uses the 
+update `Δa=-K\r` (where `K=∂r/∂a` if the standard Newton's method is used).
 ```julia
 a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
 apply!(a, ch)           # Make the guess fulfill all constraints in `ch`
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -63,7 +63,7 @@ It is important to check the residual **after** applying boundary conditions whe
 solving nonlinear problems with affine constraints. 
 `apply_zero!(K, r, ch)` modifies the residual entries for dofs that are involved 
 in constraints to account for constraint forces. 
-The following pseudo-code shows a typical pattern for solving a non-linear problem with Newtons method:
+The following pseudo-code shows a typical pattern for solving a non-linear problem with Newton's method:
 ```julia
 a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
 Δa = similar(a)         # Preallocate Δa
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -72,7 +72,8 @@ for iter in 1:maxiter
     apply_zero!(K, r, ch)   # Modify `K` and `r` to account for the constraints. 
     check_convergence(r, ...) && break # Only check convergence after `apply_zero!(K, r, ch)`
     Δa = K \ r              # Calculate the (negative) update
-    apply_zero!(Δa, ch)     # Change constrained values such that `a+Δa` fullfills constraints provided that `a` does.
+    apply_zero!(Δa, ch)     # Change the constrained values in `Δa` such that `a-Δa`
+                            # fullfills constraints if `a` did.
     a .-= Δa
 end
 ```
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -58,9 +58,6 @@ apply!(a, ch) # enforces affine constraints
 
 ```
 
-Note: Replacing `f` with `r=-f` and solving `a = -K\r` in the example above may give the wrong result.
-Instead, use the procedure for nonlinear problems when applying affine constraints to such problem formulations.
-
 ### Solving nonlinear problems
 It is important to check the residual **after** applying boundary conditions when 
 solving nonlinear problems with affine constraints. 
--- a/docs/src/manual/constraints.md
+++ b/docs/src/manual/constraints.md
@@ -66,14 +66,13 @@ in constraints to account for constraint forces.
 The following pseudo-code shows a typical pattern for solving a non-linear problem with Newton's method:
 ```julia
 a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
-Δa = similar(a)         # Preallocate Δa
 apply!(a, ch)           # Make the guess fulfill all constraints in `ch`
 for iter in 1:maxiter
     doassemble!(K, r, ...)  # Assemble the residual, r, and stiffness, K=∂r/∂a.
     apply_zero!(K, r, ch)   # Modify `K` and `r` to account for the constraints. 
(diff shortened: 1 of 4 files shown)


docs/src/manual/constraints.md
```diff
@@ -35,14 +35,15 @@ add!(ch, lc1)
...
+Affine constraints will affect the sparsity pattern of the stiffness matrix, and as such, it is important to also include 
 the `ConstraintHandler` as an argument when creating the sparsity pattern:
 
 ```julia
 ```
 
-When solving the system, we account for the affine constraints in the same way as we account for 
+### Solving linear problems
+When solving the system `K⋅a=f`, we account for the affine constraints in the same way as we account for
```
fredrikekre, line 46: Maybe use math mode here instead?
```suggestion
### Solving linear problems

When solving the system ``\underline{\underline{K}} \underline{a} = \underline{f}`, we account for the affine constraints in the same way as we account for 
```

docs/src/manual/constraints.md
```diff
@@ -55,4 +56,23 @@ apply!(K, f, ch)
 a = K\f
 apply!(a, ch) # enforces affine constraints
 
-```
\ No newline at end of file
+```
+
+Note: Replacing `f` with `r=-f` and solving `a = -K\r` in the example above may give the wrong result.
+Instead, use the procedure for nonlinear problems when applying affine constraints to such problem formulations.
```
fredrikekre, line 62: I don't really understand this part, and I don't think it is specific to affine constraints but inhomogeneities in general?

`apply!` essentially modifies `f` as: `f[free_dofs] = f[free_dofs] - K[free_dofs, constrained_dofs] * a[constrained]` where `a[constrained]` is the inhomogeneities. Note the `-` there. If you switch the sign of your right hand side, then that `-` should be a `+`. It feels a bit strange to modify the system your are solving with a `-` and then still using the same constraints (although I see why one would think this gives the same result).

docs/src/manual/constraints.md
```diff
@@ -55,4 +56,23 @@ apply!(K, f, ch)
...
+
+### Solving nonlinear problems
+A Newton-type solution method for the nonlinear system `r(a)=0` uses the 
+update `Δa=-K\r` (where `K=∂r/∂a` if the standard Newton's method is used).
```
fredrikekre, line 66: Maybe math mode here too?

docs/src/manual/constraints.md
```diff
@@ -55,4 +56,23 @@ apply!(K, f, ch)
...
+```julia
+a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
+apply!(a, ch)           # Make the guess fulfill all constraints in `ch`
+for iter in 1:maxiter
+    doassemble!(K, r, ...)
+    apply_zero!(K, r, ch)   # Modify `K` and `r` to account for the constraints. 
+    check_convergence(r, ...) && break # Only check convergence after `apply_zero!(K, r, ch)`
+    Δa=-K\r                 # Calculate update
```
fredrikekre, line 74:
```suggestion
    Δa = K \ r             # Calculate update
```

docs/src/manual/constraints.md
```diff
@@ -55,4 +56,23 @@ apply!(K, f, ch)
...
+    apply_zero!(Δa, ch)     # Change prescribed values such that `a+Δa` fullfills constraints provided that `a` does.
+    a .+= Δa
```
fredrikekre, line 76:
```suggestion
    a .-= Δa
```

docs/src/manual/constraints.md
fredrikekre, line 66: I would add explicitly here something like
```
It is important to check the residual **after** applying boundary conditions so that residual entries for constrained dofs are handled properly. The following pseudo-code shows the typical pattern for solving a non-lienar problem with Newtons method:

File: src/Dofs/ConstraintHandler.jl [hunk 6]
```diff
@@ -923,10 +925,22 @@ function _check_cellset_dirichlet(grid::AbstractGrid, cellset::Set{Int}, nodeset
     end
 end
 
+"""
+    create_symmetric_sparsity_pattern(dh::AbstractDofHandler, ch::ConstraintHandler, coupling)
```
fredrikekre, line 929: Off topic, but I think perhaps this functionality should be removed.

docs/src/manual/constraints.md
```diff
@@ -16,7 +16,7 @@ enforce these constraints, e.g. penalty methods and Lagrange multiplier methods.
 
 ## Affine constraints
 
-Affine constraints can be handled directly in Ferrite. Affine (or linear) constraints can typically
+Affine constraints can be handled directly in Ferrite. Affine (first) and linear (second) constraints can typically
```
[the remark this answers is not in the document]
KnutAM: I wasn't sure if it is worth highlighting the difference between linear and affine constraints, but that is more theory.

docs/src/manual/constraints.md
KnutAM: Yes, of course - sorry! 
But shouldn't we use boldface instead of underlines? Seems to be used in most other places.

docs/src/manual/constraints.md
KnutAM: From a performance point of view to avoid allocations?
It seems a bit counter-intuitive to me to have `Δa` imply the negative update.

docs/src/manual/constraints.md
KnutAM: You're right! 
Perhaps this should rather be described in the doc-string of `apply!`?

src/Dofs/ConstraintHandler.jl
KnutAM: 👍 
I don't think I've ever used it

docs/src/manual/constraints.md
termi-official: I think we use bold for continuous and underline for discrete variables (see https://ferrite-fem.github.io/Ferrite.jl/stable/examples/hyperelasticity/)

KnutAM: 👍 , I've homogenized the constraint page to use underlines for discrete variables

src/Dofs/ConstraintHandler.jl
termi-official: Why?

KnutAM: I wrongly thought that it was just a wrapper for `Symmetric`, but since it doesn't store all entries, it is faster. Maybe I should use it after all... 
On the other hand, just using `Symmetric(K)` gives a similar performance when solving
```julia
julia> dh = DofHandler(generate_grid(Quadrilateral, (100,100))); push!(dh, :u, 2); close!(dh);

julia> K = create_sparsity_pattern(dh);
julia> K.nzval .= rand(length(K.nzval));
julia> KS = create_symmetric_sparsity_pattern(dh);
julia> KS.data.nzval .= rand(length(KS.data.nzval));

julia> linsol(K,a) = K\a;
julia> f = rand(ndofs(dh));

julia> @btime linsol(Symmetric($K), $f);
  115.710 ms (90 allocations: 68.00 MiB)
julia> @btime linsol($KS, $f);
  117.943 ms (90 allocations: 54.95 MiB)
julia> @btime linsol($K, $f);
  148.528 ms (74 allocations: 65.69 MiB)

julia> @btime create_symmetric_sparsity_pattern($dh);
  5.400 ms (24 allocations: 17.90 MiB)
julia> @btime create_sparsity_pattern($dh);
  7.909 ms (24 allocations: 31.19 MiB)
  ```

Comment (termi-official) [reply src/Dofs/ConstraintHandler.jl:929]: Fair point. Operations regarding the symmetry should not become bottlenecks during assembly, because we can symmetrize at the element level as in the incompressible linear elasticity example.

File: docs/src/manual/constraints.md (hunk 2 above)
Comment (fredrikekre) [reply docs/src/manual/constraints.md]: Yea, I think that would be better.

File: docs/src/manual/constraints.md (hunk 4 above)
Comment (fredrikekre) [reply docs/src/manual/constraints.md]: Yea, avoids allocation of `K \r ` (`tmp = K \ r; Δa = -tmp`).

>  It seems a bit counter-intuitive to me to have Δa imply the negative update.

I guess it depends on how you state the problem, this is how it is put up in e.g. https://ferrite-fem.github.io/Ferrite.jl/stable/examples/hyperelasticity/#Newton's-Method

File: src/Dofs/ConstraintHandler.jl (hunk 6 above)
Comment (fredrikekre) [reply src/Dofs/ConstraintHandler.jl:929]: I'll open a separate issue for discussing this.

File: docs/src/manual/constraints.md [hunk 8]
```diff
@@ -55,4 +56,24 @@ apply!(K, f, ch)
...
+The following pseudo-code shows a typical pattern for solving a non-linear problem with Newton's method:
+```julia
+a = initial_guess(...)  # Make any initial guess for a here, e.g. `a=zeros(ndofs(dh))`
+Δa = similar(a)         # Preallocate Δa
+apply!(a, ch)           # Make the guess fulfill all constraints in `ch`
+for iter in 1:maxiter
+    doassemble!(K, r, ...)  # Assemble the residual, r, and stiffness, K=∂r/∂a.
+    apply_zero!(K, r, ch)   # Modify `K` and `r` to account for the constraints. 
+    check_convergence(r, ...) && break # Only check convergence after `apply_zero!(K, r, ch)`
+    Δa .= .- (K \ r)        # Calculate update
```
fredrikekre, line 75:
```suggestion
    Δa .-= K \ r            # Calculate update
```

fredrikekre: I misunderstood the code I think. It doesn't seem necessary to shuffle the data to a preallocated Δa here, just call `Δa = K \ r` and then `a .-= Δa`?

docs/src/manual/constraints.md
KnutAM: Fair point, let's just go with the negative update - it is clear in any case!
I'll update https://ferrite-fem.github.io/Ferrite.jl/stable/reference/boundary_conditions/#Ferrite.apply_zero! to be consistent also then.

src/Dofs/ConstraintHandler.jl
```diff
@@ -568,12 +579,14 @@ apply!
...
-for all prescribed degrees of freedom).
+boundary conditions and affine constraints such that `du = K \\ rhs` gives the expected 
+result (e.g. `du` zero for all degrees of freedom prescribed by Dirichlet conditions).
 
     apply_zero!(v::AbstractVector, ch::ConstraintHandler)
 
-Zero-out values in `v` corresponding to prescribed degrees of freedom.
+Zero-out values in `v` corresponding to prescribed degrees of freedom and update values 
+prescribed by affine constraints, such that if `a` fullfilled the affine constraints, 
+`a±v` also will.
```
fredrikekre, line 589:
```suggestion
prescribed by affine constraints, such that if `a` fullfills the constraints,
`a ± v` also will.
```

fredrikekre (approved):
LGMT, just some nits.

test/test_constraints.jl
```diff
@@ -261,6 +261,63 @@ end
         @test a ≈ aa ≈ al ≈ a_rhs1 ≈ a_rhs2
     end
 
+    # Test nonlinear solution procedure (on linear problem) with affine constraints 
+    # using standard assembly (i.e. not local condensation)
+    @testset "nonlinear" begin
+        params = (k=1.0, f=1.0, a=1.0, b=0.2, tol=1e-10, maxiter=2)
+        grid = generate_grid(Line, (2,)); addfaceset!(grid, "center", x->x[1]≈0.0)
+        dh = DofHandler(grid); push!(dh, :u, 1); close!(dh)
```
fredrikekre, line 269: Merged https://github.com/Ferrite-FEM/Ferrite.jl/pull/578 so:
```suggestion
        dh = DofHandler(grid); add!(dh, :u, 1); close!(dh)
```
(Maybe you need to rebase on master to get CI working after that change)

src/Dofs/ConstraintHandler.jl
```diff
@@ -568,12 +579,14 @@ apply!
     apply_zero!(K::SparseMatrixCSC, rhs::AbstractVector, ch::ConstraintHandler)
 
 Adjust the matrix `K` and the right hand side `rhs` to account for prescribed Dirichlet
-boundary conditions such that `du = K \\ rhs` give the expected result (e.g. with `du` zero
-for all prescribed degrees of freedom).
+boundary conditions and affine constraints such that `du = K \\ rhs` gives the expected 
+result (e.g. `du` zero for all degrees of freedom prescribed by Dirichlet conditions).
```
KnutAM, line 583:
```suggestion
result (e.g. `du` zero for all prescribed degrees of freedom).
```

KnutAM:
*Windows test failure is unrelated to this PR, it also occurs on master.*

PR #536, opened 2022-11-13
Improve `get_coordinates!`

kimauth:
Small improvements on the definition of `getcoordinates!`. In particular:
-  Compatible with `AbstractGrid` interface . See the `SmallGrid` example from the docs, that doesn't need to define `cellcoords!` anymore
-  Reduce code duplication. There is `getcoordinates!` working on grids and `cellcoords!` working on dofhandlers (but also on grids...). For `DofHandler` both fall back to one definition with this PR. We should deprecate one of them in the future. 
- Slightly better performance for coordinate queries on mixed grids. See micro-benchmarks below.

# Coordinate query benchmarks
This PR is part of making an effort to join the two different dofhandlers. Therefore, here are some interesting performance investigations on querying coordinates:
## Regular grid
The difference on a standard non-mixed grid is minor:
Regular Grid | #PR | #master
--- | --- | ---                                                      
`getcoordinates!(xe, grid, cellid)` |  4.291 ns (0 allocations: 0 bytes) | 4.584 ns (0 allocations: 0 bytes)                                                                     
`Ferrite.cellcoords!(xe, grid, cellid)` |   4.250 ns (0 allocations: 0 bytes) | 4.875 ns (0 allocations: 0 bytes)                                                     
`Ferrite.cellcoords!(xe, dh, cellid)` |   4.291 ns (0 allocations: 0 bytes) | 5.208 ns (0 allocations: 0 bytes) 

## Mixed grid
The PR performs much better than master on querying coordinates directly from the grid for a cell vector of type `AbstractCell`. However, that is still about twice as slow compared to querying coordinates from the `MixedDofHandler`.

`AbstractCell` | #PR | #master
--- | --- | ---                                                      
`getcoordinates!(xe, grid, cellid)` | 44.147 ns (0 allocations: 0 bytes) | 1.262 μs (20 allocations: 704 bytes)                                                   
`Ferrite.cellcoords!(xe, grid, cellid)` | 43.875 ns (0 allocations: 0 bytes) | 1.2 μs (14 allocations: 464 bytes)                                    
`Ferrite.cellcoords!(xe, dh, cellid)` | 20.687 ns (0 allocations: 0 bytes) | 20.687 ns (0 allocations: 0 bytes) 

_Notice:
For mixed grids (and thus `MixedDofHandler`) `cellcoords!(xe, dh, cellid)` is unchanged between the PR and master._

[...]

PR #576, opened 2023-01-07
Improve cellnodes!

kimauth:
Improve `cellnodes!` in the same way as #536 did for `getcoordinates!` and move it to `grid.jl`.

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -709,6 +709,18 @@ end
 @inline getcoordinates(grid::AbstractGrid, cell::CellIndex) = getcoordinates(grid, cell.idx)
 @inline getcoordinates(grid::AbstractGrid, face::FaceIndex) = getcoordinates(grid, face.idx[1])
 
+function cellnodes!(global_nodes::Vector{Int}, grid::AbstractGrid, i::Int)
+    cell = getcells(grid, i)
+    cellnodes!(global_nodes, grid, cell)
+end
+function cellnodes!(global_nodes::Vector{Int}, grid::AbstractGrid, cell::AbstractCell)
+    @assert length(global_nodes) == nnodes(cell)
+    @inbounds for i in 1:length(global_nodes)
+        global_nodes[i] = cell.nodes[i]
+    end
+    return global_nodes
+end
+
 function Base.show(io::IO, ::MIME"text/plain", grid::Grid)
     print(io, "$(typeof(grid)) with $(getncells(grid)) ")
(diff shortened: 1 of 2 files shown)


src/Grid/grid.jl
```diff
@@ -709,6 +709,18 @@ end
 @inline getcoordinates(grid::AbstractGrid, cell::CellIndex) = getcoordinates(grid, cell.idx)
 @inline getcoordinates(grid::AbstractGrid, face::FaceIndex) = getcoordinates(grid, face.idx[1])
 
+function cellnodes!(global_nodes::Vector{Int}, grid::AbstractGrid, i::Int)
+    cell = getcells(grid, i)
+    cellnodes!(global_nodes, grid, cell)
+end
+function cellnodes!(global_nodes::Vector{Int}, grid::AbstractGrid, cell::AbstractCell)
```
fredrikekre, line 716: Where is this method used? It didn't exist before.

kimauth: In the function just above, it is a function barrier around the `cell` object. This has an impact for mixed grids, though mostly when `grid.cells` is poorly typed (i.e. `Vector{AbstractCell}`).

fredrikekre: Okay, perhaps make it internal and add a comment then?

fredrikekre: Isn't that only something you would use internally then though? I see it can be used e.g. here: https://github.com/Ferrite-FEM/Ferrite.jl/blob/9de9677fb3fb0e78601271a39c123b04ee2636f6/src/iterators.jl#L87-L94 but that is internal use.

Also no need to pass the grid in this case.

KnutAM: I think it would be ok to have it external (if the grid input is removed) since not everyone is using the iterator interfaces. To avoid ambiguities for e.g. `cellnodes!(global_nodeids::AbstractVector, cell::MyCell)` , perhaps 
```suggestion
cellnodes!(global_nodes::AbstractVector, cell::AbstractCell) = _cellnodes!(global_coords, cell)
function _cellnodes!(global_nodes::Vector{Int}, cell::AbstractCell)
```

src/Grid/grid.jl
```diff
@@ -709,6 +709,18 @@ end
...
+    @assert length(global_nodes) == nnodes(cell)
+    @inbounds for i in 1:length(global_nodes)
+        global_nodes[i] = cell.nodes[i]
+    end
```
KnutAM, line 720: And to simplify, perhaps:
```suggestion
    copyto!(global_nodes, cell.nodes)
```

PR #577, opened 2023-01-07
Remove grid representation from `MixedDofHandler`

kimauth:
After #536 and #576, there is no need for the extra grid representation in the `MixedDofHandler` anymore (compare especially the benchmarks in #536).


termi-official:
After this PR, what else is missing to merge the dof handlers? A convenience wrapper aroung add which internally adds the field handler and then?

fredrikekre (approved):

PR #579, opened 2023-01-09
Simplify extension of AbstractGrid

KnutAM:
I went through the codebase and replaced direct field access when reading with already defined get functions. 
No usage change, but reduces the number of functions required to define an `AbstractGrid`, 
see the updated example and test.


KnutAM:
I started looking at changing in other files too, but I feel like we should have something like  
`get_node_coordinate(grid, nodeid)` as I started calling `getcoordinates(getnodes(grid, nodeid))` a lot. 
*And `getcoordinates` is a bit tricky as it returns node coordinates if given a `Node`, but the cell coordinates for all other cases, even `getcoordinates(grid, fi::FaceIndex)`, as was pointed out on slack some time ago...*

termi-official:
The last commit is a bit problematic, because it potentially changes the complexity of the call.

KnutAM:
> The last commit is a bit problematic, because it potentially changes the complexity of the call.

Do you mean for some special abstract grid type?
I first changed, e.g. `Tx = typeof(first(grid.nodes).x)` to `Tx = typeof(getcoordinates(getnodes(grid, 1)))`, but then went back to using `first` to be fully compatible (in case of offset arrays): `Tx = typeof(getcoordinates(first(getnodes(grid))))`

termi-official:
`getnodes(grid)` potentially allocates some big array (e.g. for some AMR grids) while `getnodes(grid, 1)` may not.

KnutAM:
Fair point, and could be fixed by this PR if you think that is better for distributed grids. 
I can only see a use for offset arrays with distributed grids anyways (but wouldn't indexing by 1 give a potential problem there. Or are the nodes on each core indexed by a local node index?)

A more general approach (outside the scope of this pr) could be to use `EndpointRanges.jl` and do 
`Tx = typeof(getcoordinates(getnodes(grid, ibegin)))` which I find quite cool that it works:)

termi-official:
Indeed, a separate function `getcoordinatetype` definitely makes most sense. Lets do it like this.

To quickly comment on the distributed assembly, in the current design we have 3 (not yet well-separated) layers. First, there is the local layer, where local manipulations work locally as expected and we have process local indices. Then we have some transitional layer which maps between local and global entities and we have the synchronization layer, which synchronizes global entities. This e.g. allows to switch between distributed assembly and the current version by just changing three lines of code (exchange the grid, dof handler and assembler).

src/Dofs/ConstraintHandler.jl
```diff
@@ -1059,7 +1059,7 @@ function _add!(ch::ConstraintHandler, pdbc::PeriodicDirichlet, interpolation::In
                field_dim::Int, offset::Int, is_legacy::Bool, rotation_matrix::Union{Matrix{T},Nothing}, ::Type{dof_map_t}, iterator_f::F) where {T, dof_map_t, F <: Function}
     grid = ch.dh.grid
     face_map = pdbc.face_map
-    Tx = typeof(first(ch.dh.grid.nodes).x) # Vec{D,T}
+    Tx = typeof(getcoordinates(first(getnodes(ch.dh.grid)))) # Vec{D,T}
```
termi-official, line 1062:
```suggestion
    Tx = get_coordinate_eltype(ch.dh.grid) # Vec{D,T}
```
?

KnutAM: That returns the element type of the coordinate

Issue #580, opened 2023-01-10
Support for Wedge&Pyramid elements


termi-official:
We are currently missing support for two rather common elements for complex geometries, namely 
* [x] wedges #581
* [ ] pyramids #623

However, this might not be straight forward, because these elements have different face geometries, breaking some of the internal assumptions on having the same face geometry on each face.

This could also pave the way for more generalized formulations (e.g. virtual elements).

KnutAM:
That would be nice!
When this issue is solved, please open an issue for FerriteMeshParser.jl to change the example: https://ferrite-fem.github.io/FerriteMeshParser.jl/stable/examples/user_element/

Issue #569, opened 2022-12-21
Support for H(curl) and H(div) elements


termi-official:
I just open an issue on this, because it came up again on slack and I think we should keep track of this.

Currently it is not straight forward to implement H(curl) or H(div) elements, but it is a feature that is asked for rather often. Canonical examples are Nedelec (for H(curl)) and Raviart-Thomas (for H(div)). To support these elements, I think there are two assumptions made in the code that have to be investigated.
1. I think it is assumed that one of the "dim" parameters controls the vector length for the evaluated, while for the actual evaluation of basis functions on a reference element is seems to be assumed that a scalar is returned (https://github.com/Ferrite-FEM/Ferrite.jl/blob/4787e3ddf3dc41f77c24e863429a1827d7f3b4f6/src/FEValues/cell_values.jl#L129-L138). This is also reflected in the dof distribution (e.g. here https://github.com/Ferrite-FEM/Ferrite.jl/blob/4787e3ddf3dc41f77c24e863429a1827d7f3b4f6/src/Dofs/DofHandler.jl#L214-L218), which leads to the wrong number of dofs for the mentioned elements.
2. The mapping of basis functions is "broken" for such elements. They require special Piola mappings to conserve special properties (see e.g. https://defelement.com/ciarlet.html). We should be able to fix this rather easily by equipping each interpolation with a trait, which dispatches to the correct mapping if the basis function is evaluated. The trait system will also be useful for more exotic problems, as for example when the problem itself is matrix valued.

termi-official:
In https://github.com/juhanikataja/afemproblem/blob/main/runme.jl is a prototype for a RT element which can help in designing a better interface.

PR #581, opened 2023-01-10
Wedge Support

termi-official:
This PR contains a fully functional code to support linear and quadratic wedges (see #580) for both dof handlers. Addresses #272 as a side-effect.


termi-official:
Btw the fail in the CI is unrelated to this PR, but I found it when improving the tests. The way the quintic triangle is evaluated is not really stable (numerically).

KristofferC:
Cool, this is an important case for improved generality.

termi-official:
CI is broken due to the deprecation. Refinement leads to decrease in the L_infty norm on Poisson problems with manufactured solutions, indicating convergence.

src/FEValues/face_values.jl
```diff
@@ -255,20 +256,22 @@ function BCValues(::Type{T}, func_interpol::Interpolation{dim,refshape}, geom_in
     end
 
     n_faces = length(qrs)
-    n_qpoints = n_faces == 0 ? 0 : length(getweights(qrs[1])) # assume same in all
+    n_qpoints = n_faces == 0 ? 0 : maximum(length.(getweights.(qrs))) # Bound number of qps correctly.
```
KnutAM, line 259:
```suggestion
    n_qpoints = n_faces == 0 ? 0 : maximum(qr->length(getweights(qr)), qrs) # Bound number of qps correctly.
```

src/Dofs/DofHandler.jl
```diff
@@ -9,13 +9,13 @@ Operates slightly faster than [`MixedDofHandler`](@docs). Supports:
 - `Grid`s with a single concrete cell type.
 - One or several fields on the whole domaine.
 """
-struct DofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
+# struct DofHandler{dim,G<:AbstractGrid{dim}} <: AbstractDofHandler
```
KnutAM, line 12:
```suggestion
```

KnutAM (commented):
Cool! 
I looked through, out of interest, and spotted some minor things.

src/FEValues/face_values.jl
termi-official: Sorry, I am not seeing the error. Can you elaborate?

KnutAM: No error and nothing major, it just avoids allocating to temp arrays. 
```julia
using Ferrite, BenchmarkTools
julia> T=Float64; dim=3; refshape=RefCube; func_interpol=Lagrange{dim,refshape,1}(); geom_interpol=func_interpol;
julia> interpolation_coords = Ferrite.reference_coordinates(func_interpol);
julia> for face in Ferrite.faces(func_interpol)
               dofcoords = Vec{dim,T}[]
               for facedof in face
                   push!(dofcoords, interpolation_coords[facedof])
               end
               qrf = QuadratureRule{dim,refshape,T}(fill(T(NaN), length(dofcoords)), dofcoords) # weights will not be used
               push!(qrs, qrf)
           end
julia> getqpoints1(qrs) = maximum(length.(Ferrite.getweights.(qrs)));
julia> getqpoints2(qrs) = maximum(qr->length(Ferrite.getweights(qr)), qrs);
julia> @btime getqpoints1($qrs)
  29.246 ns (1 allocation: 160 bytes)
4
julia> @btime getqpoints2($qrs)
  5.600 ns (0 allocations: 0 bytes)
4
```

src/interpolations.jl
```diff
@@ -72,53 +115,58 @@ Return the number of base functions for an [`Interpolation`](@ref) or `Values` o
...
 
-# Fallbacks for the interpolations which are used to distribute the dofs correctly
 """
-Number of dofs per vertex
+    vertexdof_indices(::Interpolation)
+
+Tuple of tuples containing the dof indices associated with the respective vertex.
+The indices must be consistent with the vertex indices of the reference element.
+"""
+vertexdof_indices(ip::Interpolation) = ntuple(_ -> (), nvertices(ip))
```
KnutAM, line 130: As noted in #394, 👍  for giving these different names and documenting them.
To be clearer for vector-valued problems, could these be named, e.g., `vertex_refcoord_indices` (alternatively `vertex_node_indices`) since they don't refer to the local dof number, but the reference coordinate number (or dof-location).

src/interpolations.jl
```diff
@@ -72,53 +115,58 @@ Return the number of base functions for an [`Interpolation`](@ref) or `Values` o
...
+The indices must be consistent with the indices of the edges in the reference element.
 """
-nedgedofs(::Interpolation)   = 0
+edgedof_interior_indices(ip::Interpolation) = ntuple(_ -> (), nedges(ip))
+
 """
-Number of dofs per face
+    facedof_indices(::Interpolation)
+
+Tuple of tuples containing all dof indices associated with the respective edges.
```
KnutAM, line 151:
```suggestion
Tuple of tuples containing all dof indices associated with the respective faces.
```
?

src/interpolations.jl
termi-official: I am not understanding. It actually is the local dof number. Do you refer to how Ferrite implements vectorization of Lagrange interpolations?

KnutAM: Yes, maybe I'm completely missing the point here?

KnutAM: This (note, from master) is confusing to me: 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/1fad1afbbe5658f30ba367ffca1dc6eef12e5507/src/Dofs/DofHandler.jl#L212-L219

I.e. for each vertex dof, add `length(dh.field_dims[fi])` dofs

termi-official: Not completely missing the point, but we have a not fully functional implementation for vector-valued elements. Currently, we can only construct vector-valued elements by attaching scalar-valued bases to each dimension - which is unfortunately only part of the story. For the user-facing dof-extraction API we somehow need to compensate for this.
Further, the assumption above is in general faulty, because there are elements which are vector-valued, but cannot be constructed in such a way (e.g. Nedelec and Raviart-Thomas). I will try to fix and document this with Maxi if I find time. I think it is easier to explain when we fix the faulty assumptions first and have functional Nedelec and Raviart-Thomas elements.

KnutAM: Yes, that makes sense. I think that strengthens the point regarding the function name, as what is currently extracted is the index in the `reference_coordinates(::Interpolation)` which corresponds to the dof location (with the current implementation). 
So a third option could be `vertex_dof_location_indices` (but that is getting long). If it is stated in the doc-string that it refers to the dof-location index and not the actual dof number, it is also fine. It is anyways a huge improvement that these functions are documented at all!

termi-official: Well, it *is* the local of the dof, just not for our implementation of the vector-valued Lagrange polynomials, because we vectorize the basis in at different locations in the code base (which is problematic). However, note that it is not the index of the dof location! Think about e.g. Hermite polynomials, where we have two dofs per vertex.

KnutAM: I keep learning about finite elements:) - thanks! 
But since it is not in the current implementation, I would still vote for having a note about that in the docstring (until the implementation is changed).

lijas:
I think #477 handles shell interpolations better. Maybe merging that PR will makes life easier in this PR?

termi-official:
It would definitely make things easier, but the underlying issue that the semantics of edge and face is different in different sections of the codebase stays.
CI works again, so this PR should be ready.

test/test_interpolations.jl
```diff
@@ -9,13 +9,15 @@ for interpolation in (
                       Lagrange{2, RefTetrahedron, 2}(),
                       Lagrange{2, RefTetrahedron, 3}(),
                       Lagrange{2, RefTetrahedron, 4}(),
-                      Lagrange{2, RefTetrahedron, 5}(),
+                      #Lagrange{2, RefTetrahedron, 5}(),
```
termi-official, line 12: Broken (see issues).

test/test_interpolations.jl
```diff
@@ -34,35 +36,79 @@ for interpolation in (
...
+
+    # Check for dirac delta property of interpolation
+    @testset "dirac delta property of dof $dof" for dof in 1:n_basefuncs
+        N_dof = Ferrite.value(interpolation, coords[dof])
+        for k in 1:n_basefuncs
+            if k == dof
+                @test N_dof[k] ≈ 1.0
             else
-                @test N_node[k] ≈ 0.0 atol=4eps(Float64)
+                @test N_dof[k] ≈ 0.0 atol=4eps(Float64) #broken=typeof(interpolation)==Lagrange{2, RefTetrahedron, 5}&&dof==4&&k==18
```
termi-official, line 106: Left this for reproduction purposes.

termi-official (commented):
This should summarize the remaining issues and clarify some stuff @fredrikekre

docs/src/devdocs/interpolations.md
```diff
@@ -19,13 +19,12 @@ Ferrite.derivative(::Interpolation{dim}, ::Vec{dim}) where {dim}
...
-Ferrite.nfacedofs(::Interpolation)
-Ferrite.edges(::Interpolation)
-Ferrite.nedgedofs(::Interpolation)
-Ferrite.ncelldofs(::Interpolation)
+Ferrite.vertexdof_indices(::Interpolation)
+Ferrite.edgedof_indices(::Interpolation)
+Ferrite.edgedof_interior_indices(::Interpolation)
+Ferrite.facedof_indices(::Interpolation)
+Ferrite.facedof_interior_indices(::Interpolation)
+Ferrite.celldof_interior_indices(::Interpolation)
```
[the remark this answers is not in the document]
lijas: I dont see the need for "cell" "local" or "dof"...seems excessive?
Why not just `faceindices`, `edgeindices`, `edgeindices_interior` etc..

src/interpolations.jl
```diff
@@ -212,17 +269,11 @@ struct DiscontinuousLagrange{dim,shape,order} <: Interpolation{dim,shape,order}
...
-# ncelldofs(::DiscontinuousLagrange{3,RefTetrahedron,order}) where {order} = (order+1)*(order+2)/2
 getnbasefunctions(::DiscontinuousLagrange{dim,shape,order}) where {dim,shape,order} = getnbasefunctions(Lagrange{dim,shape,order}())
-ncelldofs(::DiscontinuousLagrange{dim,shape,order}) where {dim,shape,order} = getnbasefunctions(DiscontinuousLagrange{dim,shape,order}())
-
 getnbasefunctions(::DiscontinuousLagrange{dim,shape,0}) where {dim,shape} = 1
-ncelldofs(::DiscontinuousLagrange{dim,shape,0}) where {dim,shape} = 1
 
-faces(::DiscontinuousLagrange{dim,shape,order}) where {dim,shape,order} = ()
+# This just moves all dofs into the interior of the element.
+celldof_interior_indices(ip::DiscontinuousLagrange{dim,shape,order}) where {dim,shape,order} = (collect(1:getnbasefunctions(ip))...,)
```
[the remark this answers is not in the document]
lijas: i think `ntuple(i->i, getnbasefunctions(ip))`

docs/src/devdocs/interpolations.md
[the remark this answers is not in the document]
termi-official: `faceindices` is ambiguous with `FaceIndex`, hence I included dof. `celllocal` is ment to indicate that it is an index into the dof range of a single cell, in contrast to the dof index in e.g. the "global" dof range of the equation system (be it on a single process or in distributed to the index in the distributed sparse matrix range.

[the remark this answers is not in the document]
lijas: But the `faceindices(::Interpolation)` would be "operating" on a `::Interpolation`, so I dont see why anyone would think they would get the global dofs. Btw, doesn't "dof" and "indices" describe the same thing here :P ?

I kind of get your first point though.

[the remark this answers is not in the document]
termi-official: I always assume that (new) users don't have very technical knowledge about internals of a package. 

For me a dof is also an index represented as an index, but here I want to refer to a (local) index, which you need to obtain the dof (index) locally on a cell. So the dof index is the index into the solution vector and system matrix, while this index is the index locally in a cell, to get the dof index. Does that make sense?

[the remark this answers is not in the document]
lijas: Ok i see your point. But I still think your suggestion is a bit long and clunky. I guess it does not matter to much since it a internal method. But lets see if the other ferrite devs have any suggestions/opinions.

src/interpolations.jl
KnutAM: Sorry, I missed this comment @termi-official 
Not sure how to best write it, but a note along the lines of 

"Vector-valued interpolations are obtained by applying the same scalar `Interpolation` instance multiple times in the current implementation. For such scalar interpolations, only one degree of freedom exists per vertex. This may change in the future." 

might help clear up some confusion as to why a tuple of tuples is currently returned and what is implied by dof_index. I'm still not sure if I follow 100 % how e.g. `facedof_indices` would change if we would have a "true" vector interpolation, but I trust you know that part; so take my suggestion with a grain of salt if I've misunderstood something again. (Btw. those (facedof etc. ) devdocs are stated as missing in the preview)

I also noted that some `vertexdof_indices` returns `NTuple{N,Int}` instead of `NTuple{N,NTuple{1,Int}}`, (although that difference was not introduced by this pr):
- https://github.com/Ferrite-FEM/Ferrite.jl/blob/400ffbab36891c446bde7a6b55a5a78f614cb769/src/interpolations.jl#L424
- https://github.com/Ferrite-FEM/Ferrite.jl/blob/400ffbab36891c446bde7a6b55a5a78f614cb769/src/interpolations.jl#L361

termi-official: No problem Knut! :) Might not be super easy to explain. Please note that you can also have a scalar interpolation, but multiple dofs per vertex. My favorite example is Hermite polynomials on line elements - 4 dofs in total, 2 per vertex, each scalar-valued. I think detailed documentation should come with the PR that fixes the problem, because I do not know all code sections that will break by doing this. I think I will leave the elaboration to a future PR when the finite element spaces are better separated (see e.g. #616 and #569).

You can see this PR as the foundation to fix some problematic assumptions regarding the elements - namely that each entity has exactly the same number of dofs. Here the tuples allow e.g. to distribute 4 dofs on quad faces of a wedge and 3 dofs on triangle faces of a wedge. For a true vector-valued element, e.g. [Nedelec 1 on a triangle](https://defelement.com/elements/examples/triangle-N1curl-1.html), we would have 1 dof per face, but each call to value should return a Vec{2} and not a scalar that is vectorized internally with the help of `field_dim`. However, we would have `field_dim=2` for Nedelec 1, which currently causes several internal issues. Does this make sense? 
Further this PR unlocks the foundation for native space-time elements, which is really huge! (I am not aware of another framework that can handle this).

Also good catch, I messed up merging master. Will fix.

src/FEValues/face_values.jl
```diff
@@ -232,48 +232,50 @@ for each dof-position determined by the `func_interpol`. Used mainly by the `Con
 """
 struct BCValues{T}
     M::Array{T,3}
```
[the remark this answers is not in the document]
fredrikekre: You can use a vector + a pointer vector like in the dofhandler for the dofs.

src/interpolations.jl
```diff
@@ -38,6 +40,68 @@ julia> getnbasefunctions(ip)
...
+            [length(i) for i ∈ facedof_interior_indices(interpolation)],
+            length(celldof_interior_indices(interpolation)),
+            2,
+        )
+    end
+    function InterpolationInfo(interpolation::Interpolation{1})
+        new(
+            [length(i) for i ∈ vertexdof_indices(interpolation)],
+            Int[],
+            [0, 0], # Well. Apparently vertices are also faces. :)
```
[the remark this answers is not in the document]
fredrikekre: I guess it is needed to add bc on a face which is also a vertex in 1d or osmething

src/Dofs/MixedDofHandler.jl
```diff
@@ -285,22 +281,25 @@ function _close!(dh::MixedDofHandler{dim}, cellnumbers, global_field_names, fiel
...
 
-            if ip_info.nedgedofs > 0 && dim == 3 #Edges only in 3d
-                nextdof = add_edge_dofs(cell_dofs, cell, edgedicts[fi], field_dims[local_num], ip_info.nedgedofs, nextdof)
+            # Then the edge dofs
+            if dim == 3
+                if ip_info.dim == 3 # Regular 3D element
+                    nextdof = add_edge_dofs(cell_dofs, cell, edgedicts[fi], field_dims[local_num], ip_info.nedgedofs, nextdof)
+                elseif ip_info.dim == 2 # 2D embedded element in 3D
+                    nextdof = add_edge_dofs(cell_dofs, cell, edgedicts[fi], field_dims[local_num], ip_info.nfacedofs, nextdof)
+                end
```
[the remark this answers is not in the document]
fredrikekre: Issue?

src/Dofs/DofHandler.jl
```diff
@@ -9,13 +9,12 @@ Operates slightly faster than [`MixedDofHandler`](@ref). Supports:
...
 - One or several fields on the whole domaine.
 """
-struct DofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
+struct DofHandler{dim,G<:AbstractGrid{dim}} <: AbstractDofHandler
     field_names::Vector{Symbol}
     field_dims::Vector{Int}
     # TODO: field_interpolations can probably be better typed: We should at least require
     #       all the interpolations to have the same dimension and reference shape
     field_interpolations::Vector{Interpolation}
-    bc_values::Vector{BCValues{T}} # TODO: BcValues is created/handeld by the constrainthandler, so this can be removed
```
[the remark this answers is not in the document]
termi-official: It has been removed in past PRs and functionality has been fully moved to the `ConstraintHandler`. Still breaking e.g. for binary IO.

src/Dofs/MixedDofHandler.jl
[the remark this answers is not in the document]
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/issues/619

src/interpolations.jl
```diff
@@ -281,16 +334,17 @@ function value(ip::Lagrange{1,RefCube,1}, i::Int, ξ::Vec{1})
     ξ_x = ξ[1]
     i == 1 && return (1 - ξ_x) * 0.5
     i == 2 && return (1 + ξ_x) * 0.5
-    throw(ArgumentError("no shape function $i for interpolation $ip"))
+    throw(BoundsError("no shape function $i for interpolation $ip"))
```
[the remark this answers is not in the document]
termi-official: If you think so then just revert the change.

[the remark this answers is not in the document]
termi-official: Not having a strong opinion here. I just thought this is similar to a out of bounds access, basically. E.g. we could also implement some generic Lagrange elements which has some internal arrays to manage the computation or smth.

termi-official:
Thanks for working through this !

PR #583, opened 2023-01-15
Fix broken links in docs

kimauth:


src/PointEval/PointEvalHandler.jl
```diff
@@ -314,7 +314,7 @@ Create an iterator over the points in the [`PointEvalHandler`](@ref).
 The elements of the iterator are either a [`PointLocation`](@ref), if the corresponding
 point could be found in the grid, or `nothing`, if the point was not found.
 
-A `PointLocation` can be used to query the cell ID with [`cellid`](@ref), and can be used
+A `PointLocation` can be used to query the cell ID with `cellid`, and can be used
```
KnutAM, line 317:
```suggestion
A `PointLocation` can be used to query the cell ID with the `cellid` function, and can be used
```
Only a suggestion to make it clearer until the docs for `cellid` are available.

KnutAM (approved):
Nice!
(Would be nice to publish the docstring for AffineConstraint, but that is a separate issue)

src/PointEval/PointEvalHandler.jl
kimauth: Yeah, not sure if `cellid` should be an explicit part of the API with docs. Perhaps if it gets a different name.

kimauth:
I didn't add the doctoring for `AffineConstraint` to the docs because I wasn't sure if it is meant to be internal.

PR #591, opened 2023-02-01
Minor simplifications and changes to the hyper-elasticity example.

fredrikekre:


docs/src/literate/hyperelasticity.jl
```diff
@@ -372,20 +373,23 @@ function solve()
     prog = ProgressMeter.ProgressThresh(NEWTON_TOL, "Solving:")
 
     while true; newton_itr += 1
-        u .= un .+ Δu # Current guess
+        ## Construct the current guess
+        u .= un .+ Δu
+        ## Compute residual and tanget for current guess
```
fredrikekre, line 378:
```suggestion
        ## Compute residual and tangent for current guess
```

PR #592, opened 2023-02-02
Fix constraint condensation for `Symmetric` matrices.

fredrikekre:
This patch fixes an issue where `f` would be incorrectly modified in `apply!(K, f, ch)` when `K::Symmetric` and when using inhomogeneous constraints. Specifically this patch make sure that entries below the diagonal of the matrix backing the `Symmetric` aren't used, and that the corresponding entries from above the diagonal are used in their place.


termi-official:
Small suggestion here, maybe it makes sense to split up the function into two (one for Symmetric one for SparseMatrixCSC) instead of putting the branches in the existing function.

fredrikekre:
The rest of the method is the same though, so I think that would give more code duplication. I don't think the branch is too bad.

PR #596, opened 2023-02-03
Introduce "Control dofs"

KnutAM:
Idea to introduce dofs that are not bound to a cell. Currently, I see two typical usages for this
1. AffineConstraints (see simple example below). For research, used for more flexible periodic boundary conditions. 
2. Lagrange multipliers for e.g. contact problems

but there might be more. 
Therefore, I think it would be good to have a discussion about a suitable interface. Currently, the dofs are added at the end of the list, which plays nicely with `close!(dh)` and `create_sparsity_pattern(dh, [ch])`
The current implementation is mostly thought of for point 1. For nr 2, perhaps it could be nice to be able to reuse a previously closed dofhandler to recreate a new one with a different number of "control dofs". 

<details>
<summary> Beam bending with prescribed torque </summary>

```julia
# # Control dofs in Ferrite.jl
# In this example, we introduce control dofs. 
# In e.g. Abaqus, these are implemented 
# as so-called reference points and can have
# 6 degrees of freedoms (3 translations 
# + 3 rotations) in 3d. 
# 
# The standard case will be to use linear constraints
# (affine constraints with `b=0`) for which the control 
# dofs act as main dofs 
# 
# ## Specific problem 
# Solve beam bending of a cantilever beam subjected to 
# a bending moment at the free end. 
setup_env = true
if setup_env    
    using Pkg
    Pkg.add(url="https://github.com/Ferrite-FEM/Ferrite.jl"; rev="kam/CtrlDofs")
    Pkg.add(url="https://github.com/KnutAM/FerriteNeumann.jl") # Required by FerriteProblems
    Pkg.add(url="https://github.com/KnutAM/MaterialModelsBase.jl")
    Pkg.add(url="https://github.com/KnutAM/FerriteAssembly.jl")
    Pkg.add(url="https://github.com/KnutAM/FerriteProblems.jl")
end

using Ferrite 
using MaterialModelsBase # material interface: AbstractMaterial
using FerriteAssembly    # Defines doassemble!, setup_cellbuffer, create_states
using FerriteProblems    # Defines FerriteAssembly.element_routine! for AbstractMaterial

# Define material::MaterialModelsBase.AbstractMaterial
struct PlaneStress{T} <: AbstractMaterial
    E4::SymmetricTensor{4,2,T,9}
end
function PlaneStress(E::Number=1.0e3; ν=0.3) 
    E4_voigt = (E/(1-ν^2))*[1 ν 0; ν 1 0; 0 0 (1-ν)/2]
    E4 = fromvoigt(SymmetricTensor{4,2}, E4_voigt)
    return PlaneStress(E4)
end
function MaterialModelsBase.material_response(m::PlaneStress, ϵ, args...)
    σ=m.E4⊡ϵ
    return σ, m.E4, NoMaterialState()
end

# Utility function for adding affine constraints 
function add_affine_constraints!(ch, field, component, faceset, entries_function) 
    # entries_function(x) = entries, b for use in AffineConstraint
    dh = ch.dh
    field_idx = Ferrite.find_field(dh, field)
    ip = Ferrite.getfieldinterpolation(dh, field_idx)
    dim = Ferrite.getfielddim(dh, field_idx)
    offset = Ferrite.field_offset(ch.dh, field)

    local_face_dofs, local_face_dofs_offset =
        Ferrite._local_face_dofs_for_bc(ip, dim, [component], offset)

    fv = Ferrite.BCValues(ip, ip, FaceIndex)
    completed_dofs = Set{Int}()
    cc = CellCache(dh)
    for (cellnr, facenr) in faceset
        fv.current_face[] = facenr
        reinit!(cc, cellnr)
        for location in 1:getnquadpoints(fv)
            local_dof = local_face_dofs[local_face_dofs_offset[facenr]+(location-1)]
            dofnr = celldofs(cc)[local_dof]
            dofnr in completed_dofs && continue
            x = spatial_coordinate(fv, location, getcoordinates(cc))
            entries, b = entries_function(x)
            ac = AffineConstraint(dofnr, entries, b)
            add!(ch, ac)
            push!(completed_dofs, dofnr)
        end
    end
end

hl_ratio = 10
grid = generate_grid(Quadrilateral, 4 .*(hl_ratio,1), 
                        Vec((0.0, -0.5)), Vec((hl_ratio/1, 0.5)))
dh = DofHandler(grid)
push!(dh, :u, 2)
Ferrite.add_control_dof!(dh, :a, 1)
close!(dh)

ch = ConstraintHandler(dh)
add!(ch, Dirichlet(:u, getfaceset(grid,"left"), Returns(zero(Vec{2}))))
mdof = first(Ferrite.get_control_dofrange(dh, :a))
add_affine_constraints!(ch, :u, 1, getfaceset(grid,"right"), x->([mdof=>x[2],],0.0))
close!(ch)

r = zeros(ndofs(dh))
ip = Ferrite.getfieldinterpolation(dh, 1)
qr = QuadratureRule{2, RefCube}(2)

material = PlaneStress()
cv = CellVectorValues(qr, ip)
cb = setup_cellbuffer(dh, cv, material)
states = create_states(dh, material, cv)

update!(ch, 0.0)
assembler = start_assemble(K, r)
doassemble!(assembler, cb, states, dh, zeros(ndofs(dh)))

# With the control dof we can now prescribe the torque=1 Nm,
# while still requiring that the end of the beam remains flat. 
r[mdof] = 1 
apply!(K,r,ch)

a = K\r 
apply!(a,ch)

vtk_grid("affinetest", dh) do vtkfile
    vtk_point_data(vtkfile, dh, a) # displacement field
end
```

Side-note: It would be nice to have some functionality to simplify getting the dofs on a face, which would simplify `add_affine_constraints!`.

</details>

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -20,12 +20,13 @@ struct DofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
     cell_dofs_offset::Vector{Int}
     closed::ScalarWrapper{Bool}
     grid::G
+    ctrl_dofs::Vector{Tuple{Symbol,Int}} # Could also be OrderedDict...
     ndofs::ScalarWrapper{Int}
 end
 
 function DofHandler(grid::AbstractGrid)
     isconcretetype(getcelltype(grid)) || error("Grid includes different celltypes. Use MixedDofHandler instead of DofHandler")
-    DofHandler(Symbol[], Int[], Interpolation[], BCValues{Float64}[], Int[], Int[], ScalarWrapper(false), grid, Ferrite.ScalarWrapper(-1))
+    DofHandler(Symbol[], Int[], Interpolation[], BCValues{Float64}[], Int[], Int[], ScalarWrapper(false), grid, Tuple{Symbol,Int}[], Ferrite.ScalarWrapper(-1))
 end
 
 function Base.show(io::IO, ::MIME"text/plain", dh::DofHandler)
@@ -58,6 +59,17 @@ function find_field(dh::DofHandler, field_name::Symbol)
     j === nothing && error("could not find field :$field_name in DofHandler (existing fields: $(getfieldnames(dh)))")
     return j
 end
+
+get_num_control_dofs(dh::DofHandler) = sum(last, dh.ctrl_dofs; init=0)
+function get_control_dofrange(dh::DofHandler, name::Symbol)
+    @assert isclosed(dh)
+    ctrl_start = ndofs(dh)-get_num_control_dofs(dh)
+    nr = findfirst(ctrldof->first(ctrldof)==name, dh.ctrl_dofs)
+    isnothing(nr) && throw(KeyError("$name not a control dof in the DofHandler"))
+    d1 = nr==1 ? 1 : (sum(i->last(dh.ctrl_dofs[i]), 1:(nr-1)) + 1)
+    return ctrl_start .+ (d1:(d1 + last(dh.ctrl_dofs[nr]) - 1))
+end
+add_control_dof!(dh::DofHandler, name::Symbol, number::Int) = push!(dh.ctrl_dofs, (name,number))
 
 # Calculate the offset to the first local dof of a field
 function field_offset(dh::DofHandler, field_name::Symbol)
@@ -288,7 +300,7 @@ function __close!(dh::DofHandler{dim}) where {dim}
         # push! the first index of the next cell to the offset vector
         push!(dh.cell_dofs_offset, length(dh.cell_dofs)+1)
     end # cell loop
-    dh.ndofs[] = maximum(dh.cell_dofs)
+    dh.ndofs[] = maximum(dh.cell_dofs) + get_num_control_dofs(dh)
     dh.closed[] = true
 
     return dh, vertexdicts, edgedicts, facedicts


termi-official:
I would like to propose an alternative design here. We can also make control dofs via subdomains and special elements. For example, we could add a 0D element which holds a vector of indices which is exactly the number of other elements it is "tied" to. With the new sparsity pattern algorithm we can then easily add the new entries. What do you think?

KnutAM:
I did some tests for that [here](https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...kam/GlobalDofs) where I introduce a `GlobalDofInterpolation` which can be added to a `SubDofHandler` (or `DofHandler`). 

For some reason, it seems like `assemble!(...)` becomes extremely slow (went from 3% of the simulation time to over 50 %) when I tried with a single global dof. But for some cases, this might be the only reasonable way, and it might just require a smarter `assemble!` function, I haven't had time to investigate further. (In that test case, it was a test to add an arc-length constraint equation).

termi-official:
> I did some tests for [here](https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...kam/GlobalDofs) where I introduce a `GlobalDofInterpolation` which can be added to a `SubDofHandler` (or `DofHandler`).
> 
:tada: I mean another take would be to introduce different types of subdofhandlers. Since the dof handler is not type stable anyway I do not think this would hurt.

> But a type of super-element is also nice! So this would then be put in its own subdofhandler I guess? You probably have to recreate the grid struct, since it would change the eltype of the `cells` Vector, but that is a minor inconvenience.
> 
Indeed, thats the idea.

> For some reason, it seems like `assemble!(...)` becomes extremely slow (went from 3% of the simulation time to over 50 %) when I tried with a single global dof. But for some cases, this might be the only reasonable way, and it might just require a smarter `assemble!` function, I haven't had time to investigate further. (In that test case, it was a test to add an arc-length constraint equation).

In this PR or the branch?

KnutAM:
> In this PR or the branch?

The branch

termi-official:
I guess this is also resolved in #1482 ?

PR #599, opened 2023-02-10
Fix slow show(grid)

KnutAM:
The use of `repr` on every cell in #570 was very slow.
With this PR, there is still performance regression for mixed grids (factor 3), but still ok (in my opinion).

@fredrikekre: Is there a better way to get each type in the union? In that case the suggestion below might make sense, even though I don't think it matters that much...

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -726,7 +726,11 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", grid::Grid)
     print(io, "$(typeof(grid)) with $(getncells(grid)) ")
-    typestrs = sort!(collect(Set(repr(typeof(x)) for x in grid.cells)))
+    if isconcretetype(eltype(grid.cells))
+        typestrs = [repr(eltype(grid.cells))]
+    else
+        typestrs = sort!(repr.(Set(typeof(x) for x in grid.cells)))
+    end
     join(io, typestrs, '/')
     print(io, " cells and $(getnnodes(grid)) nodes")
 end


src/Grid/grid.jl
```diff
@@ -726,7 +726,11 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", grid::Grid)
     print(io, "$(typeof(grid)) with $(getncells(grid)) ")
-    typestrs = sort!(collect(Set(repr(typeof(x)) for x in grid.cells)))
+    if isconcretetype(eltype(grid.cells))
+        typestrs = [repr(eltype(grid.cells))]
+    else
+        typestrs = sort!(repr.(Set(typeof(x) for x in grid.cells)))
+    end
```
KnutAM, line 733: Can do something like this too, but feels a bit hacky..
```suggestion
    uniontypes(T::Union) = Set(getfield(T, key) for key in fieldnames(typeof(T)))
    cells = getcells(grid) # getcelltype(grid) N/A as it returns type of first cell
    CT = eltype(cells)
    if isa(CT, Union)
        typestrs = sort!(repr.(uniontypes(CT)))
    elseif isconcretetype(CT) || length(cells) > 10_000_000
        typestrs = [repr(CT)]
    else
        typestrs = sort!(repr.(Set(typeof(x) for x in cells)))
    end
```

fredrikekre:
If you create the `Set` with types instead of strings it seems pretty ok:
```
julia> @time sort!(collect(Set(repr(typeof(x)) for x in grid.cells)))
110.462024 seconds (44.03 M allocations: 17.203 GiB, 1.71% gc time, 0.03% compilation time)
1-element Vector{String}:
 "Triangle"

julia> @time collect(Set(typeof(x) for x in grid.cells))
  0.378749 seconds (247.76 k allocations: 51.872 MiB, 5.09% gc time, 35.21% compilation time)
1-element Vector{DataType}:
 Triangle (alias for Cell{2, 3, 3})
```

KnutAM:
Exactly, that's where I got the 3x (actually 5x for this timing) reduction from though:
```julia
oldversion(grid) = sort!(collect(Set(celltypes[typeof(x)] for x in grid.cells)))
newversion(grid) = sort!(repr.(Set(typeof(x) for x in grid.cells)))
masterversion(grid) = sort!(collect(Set(repr(typeof(x)) for x in grid.cells)))
grid = generate_grid(Triangle, (100, 100));
@time newversion(grid);
  0.002467 seconds (34 allocations: 298.586 KiB)
 @time oldversion(grid);
  0.000490 seconds (9 allocations: 288.680 KiB)
@time masterversion(grid);
  0.921047 seconds (480.01 k allocations: 192.543 MiB, 3.66% gc time
 ```
 (Timing after a first run for compilation and I have defined the const celltypes=Dict from the previous Ferrite version)

PR #600, opened 2023-02-20
Fix coloring for edge case with one or zero cells

lijas:


src/Grid/coloring.jl
```diff
@@ -189,6 +190,13 @@ The resulting colors can be visualized using [`vtk_cell_data_colors`](@ref).
     ```
 """
 function create_coloring(g::Grid, cellset=1:getncells(g); alg::ColoringAlgorithm.T=ColoringAlgorithm.WorkStream)
+    
+    if length(cellset) == 0
+        return Vector{Int}[Int[]]
+    elseif length(cellset) == 1
+        return Vector{Int}[Int[first(cellset)]]
+    end
```
fredrikekre, line 198: I think this can be done inside of workstream_coloring instead

lijas: ` if length(cellset) == 0` is also needed for GREEDY so maybe better to do it outside for both?

fredrikekre: You sure? Works for me:
```
julia> create_coloring(generate_grid(Quadrilateral, (2, 2)), Int[]; alg=ColoringAlgorithm.Greedy)
Vector{Int64}[]
```

PR #601, opened 2023-02-23
Remove incorrect statements about initial conditions

KnutAM:
Just noticed a small formatting error, and as far as I understand (@termi-official ?), an extra quote that shouldn't have been there.

### DIFF
--- a/docs/src/manual/boundary_conditions.md
+++ b/docs/src/manual/boundary_conditions.md
@@ -272,17 +272,4 @@ u = zeros(ndofs(dh))
 apply_analytical!(u, dh, :p, x -> ρ * g * x[2])
 ```
 
-See also [Time Dependent Problems](@ref) for one example. 
-
-*Note about solving DAE:* 
-A Differential Algebraic Equations (DAE) is an equation of the form
-``\boldsymbol{r}(\boldsymbol{u}(t),\boldsymbol{u}'(t),t)=\boldsymbol{0}``,
-which usually cannot be expressed as a true ODE. They occur often,
-but not always, in forms where some time derivatives are missing
-```math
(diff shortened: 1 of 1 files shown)


termi-official:
I also noticed that some error slipped in here. You do not need to specify u'(0) for first order DAEs. You just need a u which is consistent with your residual. You can think about this one this way: If your system has n equations, then the corresponding u has n entries which have to be determined to be consistent with the residuals. If you have some second derivative, then you get u' as an additional unknown.

docs/src/manual/boundary_conditions.md
```diff
@@ -280,8 +280,8 @@ A Differential Algebraic Equations (DAE) is an equation of the form
 which usually cannot be expressed as a true ODE. They occur often,
 but not always, in forms where some time derivatives are missing
 ```math
-u_1'(t) = f(\boldsymbol{u}(t),t)
-0 = g(\boldsymbol{u}(t),t)`
+u_1'(t) = f(\boldsymbol{u}(t),t)\\
+0 = g(\boldsymbol{u}(t),t)
```
KnutAM, line 284:
```suggestion
0 = g(\boldsymbol{u}(t),\boldsymbol'{u}(t),t)
```
Do you mean something like this? 
Feel free to edit the branch as you see fit - this is really your area of expertise!

termi-official: I will address this in a separate PR as I need to think a bit about how to formulate the issue in an understandable way.

KnutAM: OK, in that case, I think it is better to remove this temporarily as no docs are better than wrong docs.

PR #602, opened 2023-02-23
fix vtk docs export

argeht:
In the documentation for the export as VTK file the description of the method vtk_cell_data was missing and the description of the method vtk_point_data was duplicated. 
The description of the method vtk_cell_data was added and the duplication of the method vtk_point_data was removed.

### DIFF
--- a/docs/src/reference/export.md
+++ b/docs/src/reference/export.md
@@ -27,8 +27,8 @@ reshape_to_nodes
 
 ```@docs
 vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}; compress::Bool) where {dim,C,T} 
-vtk_point_data(vtk::WriteVTK.DatasetFile, data::Union{Vector{SymmetricTensor{2,dim,T,M}}},name::AbstractString) where {dim,T,M}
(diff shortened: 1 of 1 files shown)


fredrikekre:
> Opened this pull request (their first ever)

Very cool, welcome :rocket:

docs/src/reference/export.md
```diff
@@ -27,8 +27,8 @@ reshape_to_nodes
 
 ```@docs
 vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}; compress::Bool) where {dim,C,T} 
-vtk_point_data(vtk::WriteVTK.DatasetFile, data::Union{Vector{SymmetricTensor{2,dim,T,M}}},name::AbstractString) where {dim,T,M}
-vtk_point_data(vtk::WriteVTK.DatasetFile, data::Union{ Vector{Tensor{order,dim,T,M}}, Vector{SymmetricTensor{order,dim,T,M}}}, name::AbstractString) where {order,dim,T,M}
+vtk_point_data
+vtk_cell_data
```
fredrikekre, line 31: Looks like we don't have documentation for this method, so this currently errors when included like this. Ferrite doesn't actually define any methods for that, it calls directly to `WriteVTK.vtk_cell_data`, but perhaps it would be good to add documentation to it in Ferrite anyway?
```suggestion
```

Issue #603, opened 2023-02-23
Describe requirements for correct initial conditions


KnutAM:
Ref discussion in #601 (and also #532)

PR #604, opened 2023-02-24
DAE initialization note

termi-official:
Trying to address https://github.com/Ferrite-FEM/Ferrite.jl/issues/603 .


docs/src/manual/initial_and_boundary_conditions.md
```diff
@@ -272,4 +272,32 @@ u = zeros(ndofs(dh))
...
+    of initial conditions.
+    What the exact hidden constraints are have to be worked out on the specific DAE.
+    For the example above it is sometimes possible to just solve the second equation for ``s``, i.e.
+    ```math
+    0 = g(u(t,\boldsymbol{x}), \nabla u(t,\boldsymbol{x}), s(t,\boldsymbol{x}))
+    ```
+    given some fixed choice of ``u``.
+    We refer to the paper ["Consistent Initial Condition Calculation for Differential-Algebraic Systems" 
+    by Brown et al.](dx.doi.org/10.1137/S1064827595289996) for more details on this matter and some possible solutions.
+
```
KnutAM, line 302: Thanks for providing such a detailed description Dennis, it is very interesting!
But this does indeed seem very difficult to explain in simple terms for the general case to not confuse (new) users. 
Thinking more about this, could something like the following, even if over-simplistic, work?
```suggestion
!!! note "Initialization of Differential-Algebraic Systems"
    Note that `apply_analytical!` does not ensure consistency between the applied field 
    and the specific problem being solved. As a simple example, consider a mechanical problem: 
    `apply_analytical!` doesn't ensure that equilibrium is fulfilled in the initial state. 
    In general for advanced, so-called "Differential-Algebraic Systems", 
    consistent initialization is a research field on its own, see e.g. 
    ["Consistent Initial Condition Calculation for Differential-Algebraic Systems" by Brown et al.](dx.doi.org/10.1137/S1064827595289996)
    
```
or do you think this misses the point since in many cases a non-equilibrium initial condition is perfectly fine (and we even demo that in the transient heat flow:))

I am, however, sad not to have all such knowledge/insights available. From that perspective, it might have been nice to have a separate webpage containing "Ferrite: Theory and examples" (or something) to supplement the documentation, but that's a different discussion

docs/src/manual/initial_and_boundary_conditions.md
```diff
@@ -272,4 +272,32 @@ u = zeros(ndofs(dh))
...
 See also [Time Dependent Problems](@ref) for one example.
```
KnutAM, line 303: Should we put this above the note?

docs/src/manual/initial_and_boundary_conditions.md
termi-official: Thanks for looking into this and the suggestion. I am happy and open for ideas on how to improve/simplify this part. However, the point here is not about *equilibrium*, but about *consistency* of the solution. This is especially problematic in differential-algebraic problems. 

Think about it this way: The algebraic equations act as constraints on the time evolution of your problem, where the constraints also have some "implicit" or "hidden evolution". However, the constraint might not be directly visible in the equation and hence messing it up can be very easy (for example if a zero initialization of some internal variable could be inconsistent with your evolving field variable).

Don't you think the (theory) manual we have is sufficient in combination with the examples?

KnutAM: Yes, I agree that equilibrium is not the point. I meant it more as a warning that one can not supply whatever initial condition but have to ensure that it is consistent with the problem to be solved. 

(The additional site was more a long term idea that would be useful for e.g. teaching and reference since you, we and maybe more use Ferrite in teaching, for which I think the current state is not meant to be, nor detailed enough, as course material)

KnutAM (approved):
But I like the new commit! 
If the example link is moved above the note, I think it good to go!

docs/src/manual/initial_and_boundary_conditions.md
termi-official: Quick question here, why would you move it above the note?

KnutAM: It is easy to miss that reference when reading the docs, but if switched, the note is still hard to miss. And it follows (in my mind) more logically as DAEs are not discussed in that example.

docs/src/manual/initial_and_boundary_conditions.md
```diff
@@ -272,4 +272,12 @@ u = zeros(ndofs(dh))
...
 
+We refer to the paper ["Consistent Initial Condition Calculation for Differential-Algebraic Systems"  by Brown et al.](dx.doi.org/10.1137/S1064827595289996) 
+for some details on the initialization of DAEs.
+
+!!! note "Consistency"
+    `apply_analytical!` does not enforce consistency of the applied solution with the system of 
+    equations. Some problems, like for example differential-algebraic systems of equations (DAEs)
+    need extra care during initialization, as noted above.
+
 See also [Time Dependent Problems](@ref) for one example.
```
KnutAM, line 283:
```suggestion
See also [Time Dependent Problems](@ref) for one example.

!!! note "Consistency"
    `apply_analytical!` does not enforce consistency of the applied solution with the system of 
    equations. Some problems, like for example differential-algebraic systems of equations (DAEs)
    need extra care during initialization. We refer to the paper ["Consistent Initial Condition Calculation for Differential-Algebraic Systems"  by Brown et al.](dx.doi.org/10.1137/S1064827595289996) for more details on this matter.

```

fredrikekre (commented):
I suggest keeping the filename (which translates to the URL) even though the header changes so that old links will still work.

docs/src/manual/initial_and_boundary_conditions.md
```diff
@@ -273,3 +273,9 @@ apply_analytical!(u, dh, :p, x -> ρ * g * x[2])
 ```
 
 See also [Time Dependent Problems](@ref) for one example.
+
+!!! note "Consistency"
+    `apply_analytical!` does not enforce consistency of the applied solution with the system of 
+    equations. Some problems, like for example differential-algebraic systems of equations (DAEs)
+    need extra care during initialization. We refer to the paper ["Consistent Initial Condition Calculation for Differential-Algebraic Systems"  by Brown et al.](dx.doi.org/10.1137/S1064827595289996) for more details on this matter.
```
fredrikekre, line 280: Probably you should use the full https URL instead, this gives:
```
┌ Warning: invalid local link: unresolved path in manual/boundary_conditions.md
│   link_url = "dx.doi.org/10.1137/S1064827595289996"
└ @ Documenter.HTMLWriter ~/.julia/packages/Documenter/ZRco4/src/html/HTMLWriter.jl:2195
```

PR #614, opened 2023-03-09
Fix stokes flow.

termi-official:


docs/src/literate/stokes-flow.jl
```diff
@@ -461,7 +462,7 @@ function main()
     vtk_grid("stokes-flow", grid) do vtk
         vtk_point_data(vtk, dh, u)
     end
-    Sys.isapple() || @test norm(u) ≈ 0.32254330524111213 #src
+    Sys.isapple() || @test norm(u) ≈ 0.322804206709586 #src
```
fredrikekre, line 465: Good catch. I wonder if this fixes the example on macOS:
```suggestion
    @test norm(u) ≈ 0.322804206709586 #src
```

termi-official: Btw, do you know under which circumstances the mean value constraint is valid? It seems like we require no slip in a Stokes problem to make it work, but I cannot find material on the theory.

termi-official:
Is it valid to have periodic velocity but non-periodic pressure?

termi-official:
I think gmsh generates a slightly different mesh on OSX than on windows and linux, which caused CI to fail.

fredrikekre:
> I think gmsh generates a slightly different mesh on OSX than on windows and linux, which caused CI to fail.

Ah, that would definitely explain the small changes in the result, never thought of that as an option.

I made another bugfix to this example: https://github.com/Ferrite-FEM/Ferrite.jl/commit/d9446c2bfae305c859fce071000bcd1056887e48 but with the previous definition of `constrained_dof` that was actually correct before :)

PR #618, opened 2023-03-13
Set up actions/cache for doc builds.

fredrikekre:


.github/workflows/ci.yml
```diff
@@ -41,6 +41,13 @@ jobs:
       - uses: julia-actions/setup-julia@v1
         with:
           version: '1'
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.julia/artifacts
+            ~/.julia/packages
+            ~/.julia/registries
```
KristofferC, line 49: `~/.julia/compiled`?

fredrikekre: Yea that would save lots of time but I don't think it works: https://julialang.slack.com/archives/C67EFTEF3/p1678482871336759

PR #623, opened 2023-03-20
Pyramids

lijas:
Adds RefPyramid, Pyramid cells, Lagrange{RefPyramid, 1&2}


src/interpolations.jl
```diff
@@ -915,6 +920,37 @@ function value(ip::Lagrange{3,RefPrism,2}, i::Int, ξ::Vec{3})
...
+            Vec{3, Float64}((1.0, 0.0, 0.0)),
+            Vec{3, Float64}((1.0, 1.0, 0.0)),
+            Vec{3, Float64}((0.0, 1.0, .0)),
+            Vec{3, Float64}((0.0, 0.0, 1.0))]
+end
+
+function value(ip::Lagrange{3,RefPyramid,1}, i::Int, ξ::Vec{3})
+    (x,y,z) = ξ
+    zzero = z ≈ 1.0
+    i == 1 && return zzero ? 0.0 : (-x*y+(z-1)*(-x-y-z+1))/(z-1)
```
fredrikekre, line 946: This means the derivative can't be evaluated at `z = 1.0` (truncating the dual part of the input, so it will have 0 derivative).

KristofferC: At least not until ForwardDiff 0.16.

But this is discontinuous around z = 1? Is that intended?

lijas: I used this as ref https://defelement.com/elements/examples/pyramid-Lagrange-1.html
I think that the pyramid is degenerate at the fifth node. But note that N_I = 1 at node I, and zero at the other nodes is still true for pyramids

Maybe @termi-official has some insight regarding pyramid elements

termi-official: One of the basis functions is $-xy/(z-1)$. So yes, you have a singularity at $z=1$ (note $x=y=0$). IIRC this is a problem with the way how these pyramidal elements are constructed. If you want the derivative at this specific point you can formulate y and x as a function of z and apply l'Hospital.

termi-official: Wieners, C. (1997). Conforming discretizations on tetrahedrons, pyramids, prisms and hexahedrons. Preprint, University of Stuttgart. has a proof that there exists no continuous differentiable (bi-)linear function on pyramids.

src/interpolations.jl
```diff
@@ -915,6 +920,37 @@ function value(ip::Lagrange{3,RefPrism,2}, i::Int, ξ::Vec{3})
...
+    i == 2 && return zzero ? 0.0 : x*(y+z-1)/(z-1)
+    i == 3 && return zzero ? 0.0 : -x*y/(z-1)
+    i == 4 && return zzero ? 0.0 : y*(x+z-1)/(z-1)
+    i == 5 && return zzero ? 1.0 : z
```
fredrikekre, line 950: This one is at least not problematic:
```suggestion
    i == 5 && return z
```

For the other ones, if we are sufficiently close to z = 1, can we replac them with expressions that have the same value and derivative in the limit perhaps?

termi-official: There is no derivative at and close to z=1 with x=y=0. See e.g. l'Hospital on the basis function `i=3` with the coordinates x and y depending on z gives you $-(x'*y + y'*x)/(z'-1) = 0/0$ again. Maybe we should throw an error in this case for the derivative?

src/docs.jl
```diff
@@ -183,3 +183,43 @@ Face numbers:                            | Face identifiers:
...
+-----------------------------------------+----------------------------
+Vertex numbers:                                  | Vertex coordinates:
+                                                 | v1: 𝛏 = (-1.0, -1.0, -1.0)
+         5 -. _               5 -. _             | v2: 𝛏 = ( 1.0, -1.0, -1.0)
+         | \    \._           |\     \._         | v3: 𝛏 = ( 1.0,  1.0, -1.0)
+  ^ ξ₃   |   \      \ _       | \        \ _     | v4: 𝛏 = (-1.0,  1.0, -1.0)
+  |  ξ₂  |     \       3      |  4--------- 3    | v5: 𝛏 = (-1.0, -1.0,  1.0)
+  | /    |       \    /       | /          /     | v6: 𝛏 = ( 1.0, -1.0,  1.0)
+  |/_ _  |         \ /        |/          /      | v7: 𝛏 = ( 1.0,  1.0,  1.0)
+      ξ₁ 1----------2         1----------2       | v8: 𝛏 = (-1.0,  1.0,  1.0)
```
termi-official, line 202: I think the coordinates are wrong.

src/docs.jl
```diff
@@ -183,3 +183,43 @@ Face numbers:                            | Face identifiers:
...
+-----------------------------------------+-----------------------------
+Edge numbers:                            | Edge identifiers:
+            +----8---+        +----8---+ |
+          5/        /|      5/|        | |  e1: (v1, v2),  e2: (v2, v3)
+          /       7/ |12    / |9     12| |  e3: (v3, v4),  e4: (v4, v1)
+         +----6---+  |     +  |        | |  e5: (v5, v6),  e6: (v6, v7)
+         |        |  +     |  +---4----+ |  e7: (v7, v8),  e8: (v8, v5)
+       10|      11| /    10| /1       /  |  e9: (v1, v5), e10: (v2, v6)
+         |        |/3      |/        /3  | e11: (v3, v7), e12: (v4, v8)
+         +---2----+        +---2----+    |
```
termi-official, line 212: I think the edge numbering and picture is wrong.

lijas:
1. I have checked the heat_equation.jl with pyramid elements and it converges to the analytical solution (notably a bit slower than with for example Hexahedrons). It seems like PR #640 will check this

2. PR #689 will update the docs for all reference shapes/cells, so perhaps there is no need to add docs about the RefPyramid in this PR?

3. Regarding the singularity at the 5th node. This does not seems to be an issue in practice, because the derivative of the shape functions never evaluated at the nodes, right? Perhaps if you have some :lobatto integration scheme...

lijas:
I added quadratic interpolations aswell. I also noticed that I had not used defelements numberings scheme, so I updated that aswell. (Note however that did not match VTK numbering scheme for pyramids)

The quadratic elements does not coverge for the heat equations problem, so there is something wrong...

lijas:
The quadratic interpolation for pyramids does not converge for some reason, even though in passes all the tests in test_interpolation.jl. I have double checked with [defelements](https://defelement.com/elements/examples/pyramid-Lagrange-2.html), and I am using the same numbering scheme. 

@termi-official  Any idea what the issue could be? :thinking:

termi-official:
The quadrature rule seems to be broken. Choose N=20 and qr order=3 for the linear element and it also diverges.

lijas:
Ok thanks for checking. Is N=20 the number of elements in each direction? Maybe qr order=3 is simply to few points?
Currently the pyramids are using the same type of quadrature rule as the wedges (:polyquad). [Here is another paper deriving quadrature rule for pyramids](https://www.sciencedirect.com/science/article/pii/S0168874X1200203X?via%3Dihub#s0065). I will try to use these points instead to see if it works better.

termi-official:
> Is N=20 the number of elements in each direction? 

Yes.

> Maybe qr order=3 is simply to few points?

As stated in your edit, there is no exact quadrature rule for pyramids, because some basis functions are fractional polynomials. This was just an example. If you take the quadratic pyramids and increase the quadrature, then the error should not approximately double, even if you are not exactly integrating. The solution to this heat problem is actually super simple and we use a regular discretization, so there is no reason that any of your basis functions should heavily oscillate on such a fine mesh on any element.

> Currently the pyramids are using the same type of quadrature rule as the wedges (:polyquad). [Here is another paper deriving quadrature rule for pyramids](https://www.sciencedirect.com/science/article/pii/S0168874X1200203X?via%3Dihub#s0065). I will try to use these points instead to see if it works better.
> 

I do not think that polyquad is per-se the issue, because it is known to work well. However, I do not fully understand your transformation in the quadrature rule. You seem to move some of the integration points *outside* the integration domain, which breaks the numerical integration.

```
julia> qr = QuadratureRule{RefPyramid}(2)
QuadratureRule{RefPyramid, Float64, 3}([0.6028728035309393, 0.5159484657839318, 0.5159484657839318, 0.5159484657839318, 0.5159484657839318], Vec{3, Float64}[[0.39170896144022116, 0.39170896144022116, 0.6082910385597788], [1.5735845722389405, 0.8546635164271443, 0.14533648357285572], [0.8546635164271443, 1.5735845722389405, 0.14533648357285572], [0.13574246061534806, 0.8546635164271443, 0.14533648357285572], [0.8546635164271443, 0.13574246061534806, 0.14533648357285572]])

julia> qr.points
5-element Vector{Vec{3, Float64}}:
 [0.39170896144022116, 0.39170896144022116, 0.6082910385597788]
 [1.5735845722389405, 0.8546635164271443, 0.14533648357285572]
 [0.8546635164271443, 1.5735845722389405, 0.14533648357285572]
 [0.13574246061534806, 0.8546635164271443, 0.14533648357285572]
 [0.8546635164271443, 0.13574246061534806, 0.14533648357285572]
```

lijas:
Thanks @termi-official . There was in issue with the mapping of the quadrature points. Now it seems like the both the linear and quadratic interpolations converge.

Do you have any more feedback on this PR? Do you think we should add some docs in this PR or #689 ?

termi-official (approved):
> Do you have any more feedback on this PR? Do you think we should add some docs in this PR or #689 ?

Since the linked PR will take care of the docs (and seems to have already for Pyramids) lets leave it separate.

src/interpolations.jl
```diff
@@ -1090,6 +1096,115 @@ function shape_value(ip::Lagrange{RefPrism, 2}, ξ::Vec{3}, i::Int)
...
+            Vec{3, Float64}((1.0, 0.0, 0.0)),
+            Vec{3, Float64}((0.0, 1.0, 0.0)),
+            Vec{3, Float64}((1.0, 1.0, 0.0)),
+            Vec{3, Float64}((0.0, 0.0, 1.0))]
+end
+
+function shape_value(ip::Lagrange{RefPyramid,1}, ξ::Vec{3}, i::Int)
+    (x,y,z) = ξ
+    zzero = z ≈ 1.0
+    i == 1 && return zzero ? 0.0 : (-x*y+(z-1)*(-x-y-z+1))/(z-1)
```
KnutAM, line 1118:
```suggestion
    i == 1 && return zzero ? zero(x) : (-x*y+(z-1)*(-x-y-z+1))/(z-1)
```
To make type-stable for non-Float64 input?

lijas:
Docs will be fixed in another PR, but in case we want ascii-documentation in the future, I paste possible representations here:

alt 1
```
Vertex numbers:                                  | Vertex coordinates:
                                                 |
         5 -. _               5 -. _             |
         | \    \._           |\     \._         | 
  ^ ξ₃   |   \      \ _       | \        \ _     | 
  |  ξ₂  |     \       3      |  4--------- 3    | 
  | /    |       \    /       | /          /     | 
  |/_ _  |         \ /        |/          /      |
      ξ₁ 1----------2         1----------2       | 
```

Alt 2:
```
Vertex numbers:                                  | Vertex coordinates:
         5 _                 5 _           
        / \  \ _            /|   \ _         
       .   |     \         . |       \        
^ ξ₃    |    |      4      | 1--------4      
|      .       |   /      . /        /        
+-> ξ₂ |        \ /       |/        /
/      2---------3        2--------3           
ξ₁     
```

PR #637, opened 2023-03-24
Minor performance improvements for `close!(::MixedDofHandler)`

fredrikekre:
This patch tweaks the dof distribution for `MixedDofHandler` slightly:
 - `push!` cell dofs directly into `dh.cell_dofs` instead of first pushing to an element celldofs vector and then pushing to the global one. This removes one `append!` per cell, but does not affect the benchmark that much (dict lookup is still dominating).
 - Use the computed dict token for extraction in `get_or_create_dofs!`
 - Change keyword arguments in `get_or_create_dofs!` to positional. This does not matter for performance, but gives nicer profile stackframes.

I used this benchmark code:

```julia
using Ferrite
using BenchmarkTools
function dofhandler(::Type{DHT}, grid) where DHT
    dh = DHT(grid)
    add!(dh, :v, 2, Lagrange{2,RefCube,2}())
    add!(dh, :s, 1, Lagrange{2,RefCube,1}())
    close!(dh)
    return dh
end
@btime dofhandler(MixedDofHandler, $grid)
```

with the following (disappointing) results:

```
1.605 s (319 allocations: 640.42 MiB) # master
1.573 s (316 allocations: 639.34 MiB) # patch
```

For reference, `DofHandler` gives 1.3s, 405MiB for this benchmark.

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -268,12 +268,12 @@ function _close!(dh::MixedDofHandler{dim}, cellnumbers, global_field_names, fiel
     end
 
     # loop over all the cells, and distribute dofs for all the fields
-    cell_dofs = Int[]  # list of global dofs for each cell
     for ci in cellnumbers
+        @debug "Creating dofs for cell #$ci"
 
         cell = dh.grid.cells[ci]
-        empty!(cell_dofs)
-        @debug "Creating dofs for cell #$ci"
+        len_cell_dofs = length(dh.cell_dofs)
+        dh.cell_dofs_offset[ci] = len_cell_dofs + 1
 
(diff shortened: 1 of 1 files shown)


kimauth (approved):

KristofferC:
For posterity (although this is with a much stronger CPU):

```julia
  using Ferrite
  using BenchmarkTools
  function dofhandler(::Type{DHT}, grid) where DHT
      dh = DHT(grid)
      add!(dh, :v, Lagrange{RefQuadrilateral, 2}()^2)
      add!(dh, :s, Lagrange{RefQuadrilateral, 1}())
      close!(dh)
      return dh
  end
  @btime dofhandler(DofHandler, $grid)
#  156.550 ms (182 allocations: 401.86 MiB)

Issue #558, opened 2022-12-14
Improve test coverage of foundational code and document fundamental assumptions


termi-official:
https://github.com/Ferrite-FEM/FerriteViz.jl/pull/51 revealed another problem with something seemingly simple (i.e. generating triangle grids). With the help of https://github.com/Ferrite-FEM/Ferrite.jl/pull/517 @koehlerson  and I were able to track down issues in the build in grid generators. Some more of the recent PRs (e.g. https://github.com/Ferrite-FEM/Ferrite.jl/pull/553 and https://github.com/Ferrite-FEM/Ferrite.jl/pull/523) fix symptoms of something more fundamental. Technically speaking we have test coverage (in the sense of path coverage) on fundamental code, but we do not test the implementation assumptions. For example, the interpolation tests (https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_interpolations.jl) cover some mathematical properties, which should be fulfilled by the implementation, but they do not cover some important implementation details. This is different from https://github.com/Ferrite-FEM/Ferrite.jl/issues/399, which suggest a guideline for implementing new elements.

Testing the foundation will be quite tricky, because it is neither a real unit nor an integration test. Let us discuss possibilities in this thread with the goal to avoid regressions in the future (and maybe to uncover faulty assumptions between different portions of the code).

termi-official:
Related issue https://github.com/Ferrite-FEM/Ferrite.jl/issues/817

PR #640, opened 2023-03-24
Helper script for convergence rates for Poisson problems

termi-official:
I often use such scripts to test if my implementation converges correctly. We likely want to keep full convergence testing for all elements out of default CI since this can take an awful amount of time. Can be seen as integration level testing towards #558 .


termi-official:
Alternatively I can also port over this script https://github.com/Ferrite-FEM/FerriteViz.jl/blob/master/docs/src/ferrite-examples/heat-equation.jl

KnutAM:
I find such an integration test to be a good idea. In order to catch a few more potential errors, perhaps it would be beneficial to have a case with fixed known solution which does not depend on calculating the rhs (e.g. errors in the numerical integration weights would not be caught by this test). Could perhaps be included in a github action to be run before each new release?

Did you intend to cast it into something like this?
<details>
<summary> Copied your code and played around a bit </summary>

The main point is the loop at the bottom... 
```julia
using Ferrite, SparseArrays, ForwardDiff, Test
import LinearAlgebra: diag

struct TestCase{CT,O}
    qe_order::Int
    num_elems::Int
    point_tol::Float64
    L2_tol::Float64
    L∞_tol::Float64
end
function TestCase(;element, ip_order, qe_order, num_elems, point_tol=1e-2, L2_tol=1e-2, L∞_tol=1e-2)
    return TestCase{element, ip_order}(qe_order, num_elems, point_tol, L2_tol, L∞_tol)
end

# Solution parameters
analytical_solution(x) = prod(cos, x*π/2)
# Get the RHS analytically
analytical_rhs(x) = -sum(diag(ForwardDiff.hessian(analytical_solution,x)))

function setup_problem(case::TestCase{element,ip_order}) where {element, ip_order}
    qe_order = case.qe_order
    N = case.num_elems

    # Construct Ferrite stuff
    ip_geo = Ferrite.default_interpolation(element)
    dim = Ferrite.getdim(ip_geo)
    grid = generate_grid(element, ntuple(x->N, dim));
    refshape = Ferrite.getrefshape(ip_geo)
    ip = Lagrange{dim, refshape, ip_order}()
    qr = QuadratureRule{dim, refshape}(qe_order)
    cellvalues = CellScalarValues(qr, ip, ip_geo);
    dh = DofHandler(grid)
    add!(dh, :u, 1, ip)
    close!(dh);
    ch = ConstraintHandler(dh);
    ∂Ω = union(
        values(getfacesets(grid))...
    );
    dbc = Dirichlet(:u, ∂Ω, (x, t) -> analytical_solution(x))
    add!(ch, dbc);
    close!(ch)
    return dh, ch, cellvalues 
end

# Standard assembly copy pasta for Poisson problem
function assemble_element!(Ke::Matrix, fe::Vector, cellvalues::CellScalarValues, coords)
    ## Reset to 0
    fill!(Ke, 0)
    fill!(fe, 0)
    ## Loop over quadrature points
        ## Get the quadrature weight
        x = spatial_coordinate(cellvalues, q_point, coords)
        for i in 1:n_basefuncs
            δu  = shape_value(cellvalues, q_point, i)
            ∇δu = shape_gradient(cellvalues, q_point, i)
            ## Add contribution to fe
            fe[i] += analytical_rhs(x) * δu * dΩ
            for j in 1:n_basefuncs
                ## Add contribution to Ke
                Ke[i, j] += (∇δu ⋅ ∇u) * dΩ
            end
        end
    end
    return Ke, fe
end

# Standard assembly copy pasta for Poisson problem
function assemble_global(cellvalues::CellScalarValues, K::SparseMatrixCSC, dh::DofHandler)
    ## Allocate the element stiffness matrix and element force vector
    fe = zeros(n_basefuncs)
    ## Allocate global force vector f
    f = zeros(ndofs(dh))
    ## Create an assembler
    ## Loop over all cels
    for cell in CellIterator(dh)
        ## Reinitialize cellvalues for this cell
        reinit!(cellvalues, cell)
        coords = getcoordinates(cell)
        ## Compute element contribution
        assemble_element!(Ke, fe, cellvalues, coords)
        ## Assemble Ke and fe into K and f
        assemble!(assembler, celldofs(cell), Ke, fe)
    end
    return K, f
end

# Assemble and solve
function solve(dh, ch, cellvalues)
    K, f = assemble_global(cellvalues, create_sparsity_pattern(dh), dh);
    apply!(K, f, ch)
    u = K \ f;
end

# Check L2 convergence
function check_and_compute_convergence(dh, u, cellvalues; testatol)
    L2norm = 0.0
    L∞norm = 0.0
    local_test_passed = true
    for cell in CellIterator(dh)
        reinit!(cellvalues, cell)
        coords = getcoordinates(cell)
        uₑ = u[celldofs(cell)]
            x = spatial_coordinate(cellvalues, q_point, coords)
            uₐₙₐ    = prod(cos, x*π/2)
            uₐₚₚᵣₒₓ = function_value(cellvalues, q_point, uₑ)
            L2norm += norm(uₐₙₐ-uₐₚₚᵣₒₓ)*dΩ
            L∞norm = max(L∞norm, norm(uₐₙₐ-uₐₚₚᵣₒₓ))
            if local_test_passed && !isapprox(uₐₙₐ, uₐₚₚᵣₒₓ; atol=testatol)
                local_test_passed = false
                # Potentially print something
            end
        end
    end
    L2norm, L∞norm, local_test_passed
end

function check_solution(case::TestCase)
    dh, ch, cellvalues = setup_problem(case)
    u = solve(dh, ch, cellvalues)
    L2norm, L∞norm, local_test_passed = check_and_compute_convergence(dh, u, cellvalues; testatol=case.point_tol)
    L2_passed = (L2norm < case.L2_tol)
    L∞_passed = (L∞norm < case.L∞_tol)
    L2_passed || @show L2norm
    L∞_passed || @show L∞norm
    # Potentially print warnings...
    return (local_test_passed, L2_passed, L∞_passed)
end


# Finite element approximation parameters
cases = Dict(
    "Wedge, linear ip" => TestCase(element=Wedge, ip_order=1, qe_order=3, num_elems=25),
    "Triangle, linear ip" => TestCase(element=Triangle, ip_order=1, qe_order=2, num_elems=25))

for (name, case) in cases
    results = check_solution(case)
    if all(results)
        printstyled("$name: Passed \n"; color=:green, bold=true)
    else
        printstyled("$name: Failed: $results \n", color=:red, bold=true)
    end
end
```
</details>

termi-official:
Looking at the current CI times for the examples in the documentation I do not think github runners have enough power to make such integration tests. We will likely just bump into a timeout before finishing such tests.

I had nothing specific in mind. With such scripts we can catch major regressions and actually compute convergence curves towards analytical solutions, which is super useful to see if everything works correctly together (because only then the convergence curve is correct).

KnutAM:
But for the check, I suppose a single point (or maybe a few) on the curve would suffice (i.e. as I do in the modified script) - it would catch any regression in accuracy. And then this should be quite fast (I think docs etc. is taking a lot extra due to additional packages, but this case would only need Ferrite and its dependencies). Then for manual testing, one can still use the same functions to create the full convergence curves? 
Or do you believe we need to check for quite fine meshes as well?

test/heat_equation_analytical.jl
```diff
@@ -0,0 +1,168 @@
...
+    for q_point in 1:getnquadpoints(cellvalues)
+        ## Get the quadrature weight
+        dΩ = getdetJdV(cellvalues, q_point)
+        x = spatial_coordinate(cellvalues, q_point, coords)
+        ## Loop over test shape functions
+        for i in 1:n_basefuncs
+            δu  = shape_value(cellvalues, q_point, i)
+            ∇δu = shape_gradient(cellvalues, q_point, i)
+            ## Add contribution to fe
+            fe[i] += analytical_rhs(x) * δu * dΩ
```
AbdAlazezAhmed, line 44: Looks like `forwardDiff` doesn't play work that well with `Vec`
```julia
julia> @benchmark ConvergenceTestHelper.analytical_rhs(t) setup=(t = Vec(1.0,2.0) ;)
BenchmarkTools.Trial: 10000 samples with 15 evaluations.
 Range (min … max):  973.067 ns … 236.959 μs  ┊ GC (min … max): 0.00% … 98.26%
 Time  (median):       1.036 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.196 μs ±   5.064 μs  ┊ GC (mean ± σ):  9.35% ±  2.20%

   ▅█                                                            
  ▄███▄▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁
  973 ns           Histogram: frequency by time         2.53 μs <

 Memory estimate: 1.78 KiB, allocs estimate: 20.

julia> using StaticArrays

julia> @benchmark ConvergenceTestHelper.analytical_rhs(t) setup=(t = SVector{2}(Vec(1.0,2.0)) ;)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  1.266 ns … 30.051 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.272 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.293 ns ±  0.586 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▃█▇█▆▆▂▃▂▃▁                   ▁                           ▂
  ▆████████████▄▃▁▃▁▁▁▁▁▃▆█▇█▇▇█▇█▇▇▁▃▁▁▁▁▃▁▁▄▁▄▁▇▆▇▅▆▄▅▄▆▁▆ █
  1.27 ns      Histogram: log(frequency) by time     1.35 ns <

 ```
 
 Timing the assembly for `Lagrange{RefTriangle, 5}()`, N = 20:
 ```julia
   2.177157 seconds (32.24 M allocations: 2.280 GiB, 11.30% gc time, 25.05% compilation time)
```
wrapping `x` in `SVector`:
```julia
  0.097493 seconds (841 allocations: 18.978 MiB)
```

KnutAM: Interesting! Defining `analytical_rhs(x) = -tr(hessian(analytical_solution,x))` is another way to solve that, which uses `Tensors`'s `hessian` function (in the end probably compiles to the same code as SVector). Otherwise, regular `Array`s are created as intermediate results.

termi-official: Yea, I should just use laplace here directly.

termi-official:
This new test increases CI times by ~30 seconds. We can drive down the timings by removing some of the interpolations from the tests, but I am not sure which.

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.3.14"
 
 [deps]
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
```
fredrikekre, line 7: Mistake?

termi-official: Yes. Not sure how this happened.

src/exports.jl
```diff
@@ -17,6 +17,7 @@ export
     DiscontinuousLagrange,
     Serendipity,
     getnbasefunctions,
+    getrefshape,
```
termi-official, line 20: I use this regularly, so I decided to export it here. If you think that this requires more discussion then I will revert it and adjust the test.

PR #644, opened 2023-03-25
RFC: New syntax for constructing `MixedDofHandler`

kimauth:
Each of the 3 commits represents a possible new syntax for constructing a `MixedDofHandler` on a mixed grid / with subdomains. No renaming here yet, as I think that should be a separate discussion. 

**Version 1:**
```Julia
fh1 = FieldHandler(set1)
add!(fh1, :u, 2, ip_u1)
add!(fh1, :p, 1, ip_p1)

fh2 = FieldHandler(set2)
add!(fh2, :u, 2, ip_u2)

dh = MixedDofHandler(grid, fh1, fh2)
```
_Does not allow to check if the cellset and interpolation are sound for each `FieldHandler` before `MixedDofHandler(grid, fh1, fh2)`. See the comments in the first commit._

**Version 2:**
```Julia
fh1 = FieldHandler(grid, set1)
add!(fh1, :u, 2, ip_u1)
add!(fh1, :p, 1, ip_p1)

fh2 = FieldHandler(grid, set2)
add!(fh2, :u, 2, ip_u2)

dh = MixedDofHandler(grid, fh1, fh2)
```
_Allows a lot better error messages during the construction steps, but grid has to be repeated as an argument in several places. Introduces the cell type as a type parameter to `FieldHandler`._

**Version 3:**
```Julia
fh1 = FieldHandler(grid, set1)
add!(fh1, :u, 2, ip_u1)
add!(fh1, :p, 1, ip_p1)

fh2 = FieldHandler(grid, set2)
add!(fh2, :u, 2, ip_u2)

dh = MixedDofHandler(fh1, fh2)
```
_Stores a reference to the grid in each `FieldHandler` and adds the grid type as additional type parameter to `FieldHandler`. Nicer syntax for the last line. (We could remove the cell type parameter in this case if we wanted.)_

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -28,6 +28,12 @@ mutable struct FieldHandler
     cellset::Set{Int}
 end
 
+function FieldHandler(cellset::Set{Int})
+    # if grid was an argument, could _check_same_celltype,
+    # this way the check happens only during MixedDofHandler construction
+    return FieldHandler(Field[], cellset)
+end
+
 """
     MixedDofHandler(grid::Grid)
 
@@ -51,9 +57,22 @@ struct MixedDofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
     ndofs::ScalarWrapper{Int}
 end
 
-function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
+function MixedDofHandler(grid::Grid{dim, C, T}, fieldhandlers...) where {dim, C, T}
+    @assert isa(fieldhandlers, NTuple{N, FieldHandler} where N) # TODO: can that be specified in the function signature?
+    
+    # construct MixedDofHandler 
     ncells = getncells(grid)
-    MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], Int[], zeros(Int, ncells), zeros(Int, ncells), ScalarWrapper(false), grid, ScalarWrapper(-1))
+    dh = MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], Int[], zeros(Int, ncells), zeros(Int, ncells), ScalarWrapper(false), grid, ScalarWrapper(-1))
+
+    # add fields
+    for fh in fieldhandlers
+        add!(dh, fh) # Checks celltypes, disjoint cellsets, interpolation refshapes
+    end
+
+    # old constructor if no fieldhandlers are given, new constructor (including close!) with fieldhandlers
+    isempty(fieldhandlers) || close!(dh) 
+
+    return dh
 end
 
 function Base.show(io::IO, ::MIME"text/plain", dh::MixedDofHandler)
@@ -185,6 +204,16 @@ function _check_cellset_intersections(dh::MixedDofHandler, fh::FieldHandler)
     for _fh in dh.fieldhandlers
         isdisjoint(_fh.cellset, fh.cellset) || error("Each cell can only belong to a single FieldHandler.")
     end
+end
+
+function add!(fh::FieldHandler, name::Symbol, dim::Int, ip::Interpolation)
+    # can't check if ip is compatible with celltype here without type parameter on FieldHandler
+    # can't support add!(fh, name, dim) without type parameter 
+    
+    field = Field(name, ip, dim)
+    push!(fh.fields, field)
+    
+    return fh
 end
 
 function add!(dh::MixedDofHandler, name::Symbol, dim::Int)
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -23,15 +23,16 @@ A `FieldHandler` must fullfill the following requirements:
 
 Notice that a `FieldHandler` can hold several fields.
 """
-mutable struct FieldHandler{C}
+mutable struct FieldHandler{C, G}
     fields::Vector{Field}
     cellset::Set{Int}
+    grid::G
 end
 
-function FieldHandler(grid::Grid, cellset::Set{Int})
+function FieldHandler(grid::G, cellset::Set{Int}) where G<:AbstractGrid
     _check_same_celltype(grid, cellset)
     celltype = typeof(getcells(grid, first(cellset)))
-    return FieldHandler{celltype}(Field[], cellset)
+    return FieldHandler{celltype, G}(Field[], cellset, grid)
 end
 
 """
@@ -57,20 +58,28 @@ struct MixedDofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
     ndofs::ScalarWrapper{Int}
 end
 
-function MixedDofHandler(grid::Grid{dim, C, T}, fieldhandlers...) where {dim, C, T}
-    @assert isa(fieldhandlers, NTuple{N, FieldHandler} where N) # TODO: can that be specified in the function signature?
-    
-    # construct MixedDofHandler 
+function MixedDofHandler(grid::Grid{dim, C, T}) where {dim, C, T}
     ncells = getncells(grid)
     dh = MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], Int[], zeros(Int, ncells), zeros(Int, ncells), ScalarWrapper(false), grid, ScalarWrapper(-1))
+    return dh
+end
 
+function MixedDofHandler(fieldhandlers...)     
+    @assert isa(fieldhandlers, NTuple{N, FieldHandler} where N) # TODO: can that be specified in the function signature?
+    @assert length(fieldhandlers) >= 1 "No fieldhandlers given."
+
+    grid = first(fieldhandlers).grid
+    for (i, fh) in enumerate(fieldhandlers)
+        i == 1 && continue
+        fh.grid === grid || error("All FieldHandlers must be based on the same grid.")
+    end
+
+    dh = MixedDofHandler(grid)
     # add fields
     for fh in fieldhandlers
         add!(dh, fh) # Checks celltypes, disjoint cellsets, interpolation refshapes
     end
-
-    # old constructor if no fieldhandlers are given, new constructor (including close!) with fieldhandlers
-    isempty(fieldhandlers) || close!(dh) 
+    close!(dh) 
 
     return dh
 end
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -23,15 +23,15 @@ A `FieldHandler` must fullfill the following requirements:
 
 Notice that a `FieldHandler` can hold several fields.
 """
-mutable struct FieldHandler
+mutable struct FieldHandler{C}
     fields::Vector{Field}
     cellset::Set{Int}
 end
 
-function FieldHandler(cellset::Set{Int})
-    # if grid was an argument, could _check_same_celltype,
-    # this way the check happens only during MixedDofHandler construction
-    return FieldHandler(Field[], cellset)
+function FieldHandler(grid::Grid, cellset::Set{Int})
+    _check_same_celltype(grid, cellset)
+    celltype = typeof(getcells(grid, first(cellset)))
+    return FieldHandler{celltype}(Field[], cellset)
 end
 
 """
@@ -185,17 +185,7 @@ Add all fields of the [`FieldHandler`](@ref) `fh` to `dh`.
 function add!(dh::MixedDofHandler, fh::FieldHandler)
     # TODO: perhaps check that a field with the same name is the same field?
     @assert !isclosed(dh)
-    _check_same_celltype(dh.grid, collect(fh.cellset))
-    _check_cellset_intersections(dh, fh)
-    # the field interpolations should have the same refshape as the cells they are applied to
-    refshapes_fh = getrefshape.(getfieldinterpolations(fh))
-    # extract the celltype from the first cell as the celltypes are all equal
-    cell_type = typeof(dh.grid.cells[first(fh.cellset)])
-    refshape_cellset = getrefshape(default_interpolation(cell_type))
-    for refshape in refshapes_fh
-        refshape_cellset == refshape || error("The RefShapes of the fieldhandlers interpolations must correspond to the RefShape of the cells it is applied to.")
-    end
-
+    _check_cellset_intersections(dh, fh) 
     push!(dh.fieldhandlers, fh)
     return dh
 end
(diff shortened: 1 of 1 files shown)


termi-official:
Version 3 suggests that we can have different grids for the fields handlers. Is this an intended feature?

kimauth:
In fact it is the other way around, version 3 is the only one that is able to check if the same grid underlies all field handlers. 
So no, there is no intention to allow different grids in one `MixedDofHandler`.

lijas:
How about version 1 with also specifying cell type?

```
fh1 = FieldHandler{Hexahedron}(set1)
add!(fh1, :u, 2, ip_u1)
add!(fh1, :p, 1, ip_p1)

fh2 = FieldHandler{Tetrahedron}(set2)
add!(fh2, :u, 2, ip_u3)

dh = MixedDofHandler(grid, fh1, fh2)
```
This way we can check that the interpolation added to the field-handler matches the celltype. One drawback is that we cant check that the set matches the cell-type (that can be done in the close!(::MixedDofhandler) though).

I think this syntax feels kind of "Ferrite-y" :) I mean, we usually have to be pretty explicit with the exact types, e.g. `Lagrange{3,RefCube,2}()`.

src/Dofs/MixedDofHandler.jl
```diff
@@ -51,9 +58,30 @@ struct MixedDofHandler{dim,T,G<:AbstractGrid{dim}} <: AbstractDofHandler
...
-function MixedDofHandler(grid::Grid{dim,C,T}) where {dim,C,T}
+function MixedDofHandler(grid::Grid{dim, C, T}) where {dim, C, T}
     ncells = getncells(grid)
-    MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], Int[], zeros(Int, ncells), zeros(Int, ncells), ScalarWrapper(false), grid, ScalarWrapper(-1))
+    dh = MixedDofHandler{dim,T,typeof(grid)}(FieldHandler[], Int[], zeros(Int, ncells), zeros(Int, ncells), ScalarWrapper(false), grid, ScalarWrapper(-1))
+    return dh
+end
+
+function MixedDofHandler(fieldhandlers...)     
+    @assert isa(fieldhandlers, NTuple{N, FieldHandler} where N) # TODO: can that be specified in the function signature?
```
KnutAM, line 68: Drive-by comment:
```suggestion
function MixedDofHandler(fieldhandlers::FieldHandler...)
```

KnutAM:
I like this direction on the syntax simplifications - nice ideas!

The major advantage that I see with 2 and 3, is the ability to provide a default interpolation (In a nice way, perhaps possible with 1 also as the FieldHandler is mutable and Field is nonconcrete?)

For convenience, would the idea also be to support the following syntaxes?
Only one FieldHandler with a single field (in prep for replacing DofHandler)
```julia
dh = MixedDofHandler(grid, :u, 2, [ip_u])
```
Single field on one FieldHandler
```julia
fh2 = FieldHandler([grid], set2, :u, 2, [ip_u2]) # Here I guess grid would be required if ip_u2 not given 
```

kimauth:
Thanks for your input!

After thinking about it a bit more, I think I'm actually against introducing the cell type as a type parameter, because 
1. I don't see where to use that, we don't seem to need it in Ferrite itself so far.
2. It introduces an inherent type instability in looping over the field handlers, which is a pattern that is needed in user-code for e.g. mixed grids:
```Julia
for fh in dh.fieldhandlers
    # user code
end
```
Handling this type instability well seems like a common pitfall, especially for new Julia users, to me. As long as there is no concrete advantage to having the cell type _as a type parameter_, I therefore suggest to store it in the `FieldHandler` instead, which still allows all 3 suggested syntax alternatives above.

fredrikekre:
Me and @kimauth discussed elsewhere and we think that this would instead be the best alternative:
```julia
dh = MixedDofHandler(grid)

fh1 = FieldHandler(dh, set1)
add!(fh1, :u, 2, ip_u1)
add!(fh1, :p, 1, ip_p1)

fh2 = FieldHandler(dh, set2)
add!(fh2, :u, 2, ip_u2)

close!(dh)
```
which after renames (`MixedDofHandler` -> `DofHandler`, and `FieldHandler` -> `SubDofHandler`) would look like this
```julia
dh = DofHandler(grid)

sdh1 = SubDofHandler(dh, set1)
add!(sdh1, :u, 2, ip_u1)
add!(sdh1, :p, 1, ip_p1)

sdh2 = SubDofHandler(dh, set2)
add!(sdh2, :u, 2, ip_u2)

close!(dh)
```

I initially thought this looked a bit backwards, since it feels like you should be able to pass the smaller building blocks (`SubDofHandler`s) to the larger building block (`DofHandler`). However, this now instead looks a bit like how you create views in Julia, for example a vector view
```
x = ...
y = view(x, subset)
```
which is basically what `SubDofHandler`s are: a "view" of the full `DofHandler` on some sub domain.

This syntax also make it much easier to directly store the reference to the `DofHandler` in the `SubDofHandler` struct since it is passed to the constructor. In the proposals from OP we would have to insert this reference in `close!` or similar, which is a bit backwards.

termi-official:
While I appreciate the idea I still think `SubDofHandler` is misleading from a user-perspective (if it really is just the view). Maybe `SubdomainHandler` or even `SubdomainView` is more descriptive?

fredrikekre:
But it is sub to the dofhandler, not sub to the grid, as subdomainhandler suggests. The plan is to make  it possible to use the SubDofHandler like a dofhandler (i.e. you will be able to query celldofs etc), so then I think the name is apt.

KnutAM:
> `sdh1 = SubDofHandler(dh, set1)`

In one way I like the beauty of this, but I also worry that implicitly adding  `sdh1` to `dh` in this way adds complexity making it harder to understand, especially for new users. 

A thought I've aired to a few of you, is the following. 
In `FerriteNeumann.jl` I've opted to separate the input and what is actually stored, which I've found quite flexible: [link](https://github.com/KnutAM/FerriteNeumann.jl/blob/db0313946a836b8567073b5d1654666d97718c8c/src/Neumann.jl#L31-L45)

For this case, that would give a syntax like
```julia
d1 = Domain(set1)
add!(d1, :u, 2, [ip_u1])
add!(d1, :p, 1, [ip_p1])
d2 = Domain(set2)
add!(d2, :u, 2, [ip_u2])
dh = DofHandler(grid, sd1, sd2);
```
where `dh = DofHandler(grid, d1, d2)` would create `SubDofHandler`s and if needed insert default interpolations. In this case, we can make vectors with eltype `Union` which might even make some previously very slow code patterns acceptable in performance. 

For the single domain case, it would be possible to do (equally many lines as before:))
```julia
d = Domain()
add!(d, :u, 2, [ip_u1])
add!(d, :p, 1, [ip_p1])
dh = DofHandler(grid, d)
```
(The old way can be retained, at least in a transition period, but might give lower performance due to more abstract types)

fredrikekre:
> In one way I like the beauty of this, but I also worry that implicitly adding sdh1 to dh in this way adds complexity making it harder to understand, especially for new users.

You don't necessarily have to think of it like adding sdh1 to dh, instead you have your sdh1 for assembly of that domain, and sdh2 for assembly of that domain.

> where dh = DofHandler(grid, d1, d2) would create SubDofHandlers and if needed insert default interpolations.

But don't you want handles to the sdhs rather than to something like the domains? How do you get those handles?

KnutAM:
> You don't necessarily have to think of it like adding sdh1 to dh, instead you have your sdh1 for assembly of that domain, and sdh2 for assembly of that domain.

But in practice, wouldn't you always pass the full `DofHandler` to an assembly function, which then would assemble each `SubDofHandler`?

> But don't you want handles to the sdhs rather than to something like the domains? How do you get those handles?

Related to above: By looping over the `SubDofHandler`s of the `DofHandler`?

termi-official:
> But it is sub to the dofhandler, not sub to the grid, as subdomainhandler suggests. The plan is to make it possible to use the SubDofHandler like a dofhandler (i.e. you will be able to query celldofs etc), so then I think the name is apt.

Sorry, maybe I misunderstand something, but isn't the `SubDofHandler` intended to be a rename of the `FieldHandler` with some convenience features to check correct construction? And isn't the `FieldHander` just a description of a field on a geometric subdomain rather than a full-fledged dof handler on a geometric (or algebraic) subdomain?

> > where dh = DofHandler(grid, d1, d2) would create SubDofHandlers and if needed insert default interpolations.
> 
> But don't you want handles to the sdhs rather than to something like the domains? How do you get those handles?

I am not really seeing the advantage of this over iterating over a subdomain. Could you elaborate more?

fredrikekre:
After the rework you could pass just the subdofhandler instead.

I would never structure my code like that I think, because there are a bunch of other things that woud be different between the domains anyway, right? I.e., I would be explicit about the different subdomains instead of looping over them

```julia
assemble_domain1(sdh1, cellvalues_u, cellvalues_p, ...)
assemble_domain2(sdh1, cellvalues_u, ...)
```

KnutAM:
> I would never structure my code like that I think

OK, if that is a common pattern, I see the reason and your suggestion is better. 
I have always tried to make one function do that, and rather have different inputs depending on the fieldhandler, see e.g. https://github.com/KnutAM/FerriteAssembly.jl/blob/a05511c881e33ef2eee1eb550fee2059da212e45/src/assembly.jl#L94-L103

termi-official:
Depends on what you want to do. E.g. if you mix element types like hexahedra, tetrahedra, wedges and pyramids, but have the same PDE (e.g. heat), then only the inputs to assemble_element!(...) change (or to be specifiy, buffer lengths..). However, if you have different physics per subdomain, as for example in FSI, then it is closer to what you just describe where the `assemble_element!` also changes. As the current design allows it I think it makes sense to support both. I had something similar to what Knut proposes in mind.

fredrikekre:
>  I have always tried to make one function do that, [...]

You can still do that with the new suggestion too, but with a loop over the fieldhandlers you kind of need to have the same interface to all subdomains or group all cell-related stuff together as in that link. With my suggestion it just is a bit easier perhaps to be explicit about this already before calling the global assembly routine.

KnutAM:
Just to have an example how it would change for a more complicated problem: From the PorousMedia mixed dof example, #448, I think would change as follows, right? (See latest version [here](https://knutam.github.io/FerriteProblems.jl/stable/examples/porous_media/#Problem-setup))
```julia
dh = MixedDofHandler(grid)
add!(dh, FieldHandler([Field(:u, ip3_lin, dim)], getcellset(grid,"solid3")))
add!(dh, FieldHandler([Field(:u, ip4_lin, dim)], getcellset(grid,"solid4")))
add!(dh, FieldHandler([Field(:u, ip3_quad, dim), Field(:p, ip3_lin, 1)], getcellset(grid,"porous3")))
add!(dh, FieldHandler([Field(:u, ip4_quad, dim), Field(:p, ip4_lin, 1)], getcellset(grid,"porous4")))
close!(dh)
```
to
```julia
dh = MixedDofHandler(grid)
sdh1 = SubDofHandler(dh, getcellset(grid,"solid3"))
add!(sdh1, :u, dim, ip3_lin)
sdh2 = SubDofHandler(dh, getcellset(grid,"solid4"))
add!(sdh2, :u, ip4_lin, dim)
sdh3 = SubDofHandler(dh, getcellset(grid,"porous3"))
add!(sdh3, :u, dim, ip3_quad), 
add!(sdh3, :p, 1, ip3_lin)
sdh4 = SubDofHandler(dh, getcellset(grid,"porous4"))
add!(sdh4, :u, dim, ip4_quad)
add!(sdh4, :p, 1, ip4_lin)
close!(dh)
```
A bit more verbose, but this isn't necessarily bad.

fredrikekre:
I guess, but a more "fair" comparison would be
```
dh = MixedDofHandler(grid)
fh1 = FieldHandler([Field(:u, ip3_lin, dim)], getcellset(grid,"solid3"))
add!(dh, fh1)
fh2 = FieldHandler([Field(:u, ip4_lin, dim)], getcellset(grid,"solid4"))
add!(dh, fh2)
fh3 = FieldHandler([Field(:u, ip3_quad, dim), Field(:p, ip3_lin, 1)], getcellset(grid,"porous3"))
add!(dh, fh3fh3)
fh4 = FieldHandler([Field(:u, ip4_quad, dim), Field(:p, ip4_lin, 1)], getcellset(grid,"porous4"))
add!(dh, fh4)
close!(dh)
```
which gives you the `fh*` handles.

KnutAM:
> which gives you the fh* handles.

Depends on whether you need them ;-) (But as said, I think the new syntax is also nice, and I don't mind the extra lines here, and I can now see a few advantages with having them!)

kimauth:
Also become a lot more readable imo if you give it a few paragraphs
```Julia
dh = MixedDofHandler(grid)

sdh1 = SubDofHandler(dh, getcellset(grid,"solid3"))
add!(sdh1, :u, dim, ip3_lin)

sdh2 = SubDofHandler(dh, getcellset(grid,"solid4"))
add!(sdh2, :u, dim, ip4_lin)

sdh3 = SubDofHandler(dh, getcellset(grid,"porous3"))
add!(sdh3, :u, dim, ip3_quad), 
add!(sdh3, :p, 1, ip3_lin)

sdh4 = SubDofHandler(dh, getcellset(grid,"porous4"))
add!(sdh4, :u, dim, ip4_quad)
add!(sdh4, :p, 1, ip4_lin)

close!(dh)
```
These are somewhat complex problems to put together after all.

lijas:
Maybe we should try to avoid creating an "intermidiate" struct all togheter... the goal is just to define all fields in the problem, and perhaps something this could work:

```
dh = DofHandler(grid)

add!(dh, :region1 , cellset) do 
    addfield!(dh, :u, dim, ip1)
    addfield!(dh, :p, dim, ip2)
end
   
addregion!(dh, :region2, cellset2) do 
    addfield!(dh, :u, dim, ip3)
end

close!(dh)
   
sdh = SubDofHandler(dh, :region1) #Maybe...
```

The "region-names" is just a suggestion because it looked a bit nice (dont know if it ever would be used).

A FieldHandler would probobly still be needed in the internals of the DofHandler

KnutAM:
> The "region-names" is just a suggestion because it looked a bit nice (dont know if it ever would be used).

One potential use would be setting up cellvalues and different materials to ensure they are linked to the correct region

fredrikekre:
What is the advantage of something like
```
region1_dh = dh[:region1]
```
over
```
region1_dh = SubDofHandler(dh)
```
?
Typically you need to bundle other things too for every domain (cellvalues, ...) anyway, so why not also bundle the sub dofhandler too?

fredrikekre:
It wouldn't be difficult to support
```
SubDofHandler(dh, cellset, name)
```
though, and have a field in the dofhandler `sdh_names :: Vector{Union{Symbol,Nothing}}`, just wondering about the usecase.

KnutAM:
This syntax would have been nice: `SubDofHandler(dh, name) = SubDofHandler(dh, getcellset(dh.grid, name), name)` :)

But the more I think about this, the more I agree that while having names could be nice, it doesn't bring enough advantages to motivate having it in the DofHandler struct.

fredrikekre:
It there was a way to encode subdomains of the grid that would be a good option though. Then `DofHandler(grid)` could create subdofhandlers directly perhaps, with the same subdomain tags as the grid

PR #645, opened 2023-03-25
Add FieldWise and ComponentWise renumbering for mixed grids

kimauth:


fredrikekre (commented):
Would be nice to join the other methods too, if possible.

src/Dofs/DofRenumbering.jl
```diff
@@ -217,6 +232,57 @@ function compute_renumber_permutation(dh::DofHandler, _, order::DofOrder.Compone
...
+        end
+        order.target_blocks
+    end
+    @assert length(target_blocks) == sum(field_dims)
+    @assert sort!(unique(target_blocks)) == 1:maximum(target_blocks)
+    # Collect all dofs into the corresponding block according to target_blocks
+    nblocks = maximum(target_blocks)
+    dofs_for_blocks = [Set{Int}() for _ in 1:nblocks]
+    component_offsets = pushfirst!(cumsum(field_dims), 0)
+    for fh in dh.fieldhandlers
```
fredrikekre, line 252: Maybe this, with some adjustmens, would work:
```suggestion
    for fh in (dh isa DofHandler ? (dh, ) : dh.fieldhandlers)
```

kimauth: There could sure be some refactoring for better code reuse. I thought it's not really worth it though with the effort to merge the dofhandlers going well.

fredrikekre: Just thought it would give a smaller diff both now and then, but not worth it if it isn't trivial like in the other method

kimauth: I don't think it's worth it here.

fredrikekre: Couldn't help myself :p

PR #646, opened 2023-03-27
Store field names in MixedDofHandler

fredrikekre:
This patch adds a field `field_names::Vector{Symbol}` to `MixedDofHandler` that caches the computation of global field names, i.e. `getfieldnames(dh)`. This mirrors how fields are stored in the DofHandler. The usage of the two similar functions, `getfieldinterpolations` and `getfielddims`, should be investigated and possibly it might make sense to store these directly in the (Mixed)DofHandler too.


kimauth (approved):

kimauth:
The mapping fieldname=>fieldinterpolation is not necessarily globally unique (a field might have different interpolations on different element types) and `getfieldinterpolations` could also often be replaced by `getfieldinterpolation`.
`getfielddims` is a bit more of a greyzone. A field should have the same dimension on any domain, so a mapping filedname=>fielddim should be unique globally. We however do not check this during construction, perhaps we should.

PR #648, opened 2023-03-27
Add method hasfieldname

KnutAM:
I was using the `getfieldnames(::FieldHandler)` antipattern in dependent packages. 
With this gone, the "simple" option is to check if `_find_field(fh, name)` returns `nothing`, but this method is clearly marked as internal. Hence, I propose adding a simple way for checking if a fieldhandler/dofhandler has a given field. 

(I also added a CHANGELOG entry for the removed methods in #647 in case someone else also was using these.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Added
+ - `hasfieldname` method added for checking if a fieldname exists in an `AbstractDofHandler`/`FieldHandler`. 
+
+### Internal changes
+ - `getfielddims(::FieldHandler)`, `getfieldinterpolations(::FieldHandler)`, and `getfieldnames(::FieldHandler)` removed ([#647][github-647])
 
 ## [0.3.13] - 2023-03-23
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
- - `hasfieldname` method added for checking if a fieldname exists in an `AbstractDofHandler`/`FieldHandler`. 
+ - `hasfieldname` method added for checking if a fieldname exists in an `AbstractDofHandler`/`FieldHandler` ([#648][github-648])
 
 ### Internal changes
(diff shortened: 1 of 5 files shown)


kimauth:
The main reason that I added an underscore was that I needed another function name for the version that doesn't error if it doesn't find a field and didn't think this was needed externally. 

Perhaps exposing the behavior of `_find_field` would be the more natural choice to expose compared to `find_field`, as it matches `findfirst` and friends.

KnutAM:
> Perhaps exposing the behavior of `_find_field` would be the more natural choice to expose compared to `find_field`

That is a good idea, and nice to do these renaming breaking changes with the 0.4 release. 
One naming option could be `get_field_index` for the erroring version, but I'm not fully decided if I think this is good considering that it involves a search (on the other hand that is an implementation detail).

Currently, many of my packages define functions that take a `FieldHandler` or a `DofHandler`, making it convenient to have the same functions defined for these (e.g. find_field). But if the MixedDofHandler replaces the DofHandler, this is no longer important from my side.

kimauth:
But that allocates, which is not necessary for that query. And we have all the needed behaviour, we just maybe don't expose the right part of it / enough of it.

KnutAM:
From my side I don't need `getfieldnames`. But if needed, one could define a `getfieldnames` that returns an iterator over the fields, and the user can collect if desired to really have a vector.

fredrikekre:
> But that allocates, which is not necessary for that query.

I think we should store fields in the FieldHandler like in the DofHandler https://github.com/Ferrite-FEM/Ferrite.jl/blob/3baab3ef01f0db7d3b5bb9242ccc32e3acb68f55/src/Dofs/DofHandler.jl#L11-L12 instead. That seems more useful.

kimauth:
I have been thinking of that too, but more in terms of being non-breaking where possible when merging the dofhandlers. 

Iterating over the fields of a `FieldHandler` is a valid request and right now the solution is to iterate over `fh.fields`. That is different from iterating over the fields of a `DofHandler` where you should iterate over `dh.fieldnames`. It would be good if this was the same for `DofHandler` and `FieldHandler`. I don't think `Field` is strictly better than the field representation in `DofHandler`, so changing the field storage in `FieldHandler` seems reasonable to me.

fredrikekre:
Can I suggest putting back the non-optimized `getfieldnames(fh)` until we rework the FieldHandler/SubDofHandler? I think it would be nice to have `getfieldnames(::MixedDofHandler)` for getting all fields, and `getfieldnames(::SubDofHandler)` to get the local fields.

fredrikekre:
I think it is fine to keep it internal. As a user you know what fields you have added, so you don't need to request it, right?

kimauth:
But for a `MixedDofHandler` with subdomains the global order of the fields is not necessarily intuitive. I think it needs to be exposed for renumbering and computing the sparsity pattern, so that a user can know how to specify a field order or a coupling.

koehlerson:
> I think it is fine to keep it internal. As a user you know what fields you have added, so you don't need to request it, right?

In FerriteViz we query them

KnutAM:
See #659 instead

PR #627, opened 2023-03-22
Improve dof distribution interface

termi-official:
This enables e.g. order 3 Serendipity and arbitrary order Lagrange in 2D with the MixedDofHandler. Closes #661.

## TODOs
* [x] Basic fix to distribute high order vertex dofs and high order edge dofs 
* [x] Some test coverage
* [x] Fix mixed tests (faulty semantics?)
* [x] Devdocs

PR #295, opened 2020-09-30
Separate reference domain dimension and nodal coordinate dimension

yijiangh:
### Summary

This PR tries to split the `dim` type parameter to `ξdim` and `xdim`. Before, the dimension of the reference domain (`\xi`) is coupled with the dimension of the nodal coordinate domain.

This PR starts with changing `Values{dim,T,refshape}` to `Values{ξdim,xdim,T,refshape}`, and other changes are made to make things compatible and non-breaking. The `dim` in function argument are renamed to `ξdim` or `xdim` to clarify the context.
This PR is particularly useful for modeling non-standard cell in JuAFEM, e.g. truss elements (1D `Line` cell with nodes in 2D or 3D).

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [x] New feature in a **backwards-compatible** manner.
- [ ] Breaking change: bug fix or new feature that involve incompatible API changes.
- [x] Other (e.g. doc update, configuration, etc)

### Checklist

- [x] I ran all tests on my computer and it's all green (i.e. `] test`).
   - All local tests (julia 1.5.0 on Windows) passed except the ones involve file sha comparison. I'm not sure if the CI tests would pass or not 
```julia
Test Summary:         | Fail  Total
Grid, DofHandler, vtk |   16     16
```
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [x] I have added necessary documentation (if appropriate)

PR #651, opened 2023-03-28
[BREAKING] Embedded elements

### LINKED ISSUE #619: Fix embedded elements
Embedded elements are currently broken in several ways and require several workarounds to function properly.

First workaround to make the constraint elimination work is the assumption that "faces = edges" and "cell = face" for 2D elements in 3D. The dof handlers already work without this assumption, however the constraint handler still requires this. The new hotfix for this has been introduced here https://github.com/Ferrite-FEM/Ferrite.jl/pull/581/files#diff-3e76902e9a74000d5b26953f9acefca30977e4844e04df825a59ebe4ab77469eR43-R47 . Embedded 1D elements in 3D are still fully broken. This issue has been around for a while and is related to #394 .

The next point of failure is in the Cell*Values, where the transformation between reference element and physical space is not implemented. The linear shell example has a hotfix, but this hotfix just works for the specific formulation. Here it might make sense to have a dispatchable function which computes the transformations (and weights) correctly for the specific formulation. It should be noted that this might need changes in Tensors.jl (see https://github.com/Ferrite-FEM/Tensors.jl/pull/188 ) but other approaches should also be feasible.

termi-official:
Cleaner integration of embedded elements. Decouples the reference from the spatial dimension. Closes #619 . 

cc @lijas may need your help regarding the shell example.

## TODO

1. [x] Make things functional
2. [x] Fix performance regression
3. [x] Merge individual `reinit!` functions into a single one to enforce DRY.
4. [x] ~~Migrate point values to the new interface~~ After going through the related code I think this is a PR by itself.
5. [x] Improved doc strings
6. [ ] ~~Update shell example~~ Not sure how.


fredrikekre:
Perhaps we can change the storage in CellValues to `SVector`s and `SMatrix` unconditionally, and then cast to `Tensors.Vec` and `Tensors.Tensor` when possible? Since the storage of StaticArrays and Tensors is equivalent this should be a no-op. Something like this:
```julia
using Tensors
using Tensors.StaticArrays

struct CellVectorValues{rdim,sdim,vdim,T,M1,M2}
    N::Matrix{SVector{vdim,T}}
    dNdx::Matrix{SMatrix{vdim,sdim,T,M1}}
    dNdξ::Matrix{SMatrix{vdim,rdim,T,M2}}
end

function shape_value(cv::CellVectorValues, qp::Int, i::Int)
    return tensor_cast(cv.N[qp, i])
end
function shape_derivative(cv::CellVectorValues, qp::Int, i::Int)
    return tensor_cast(cv.dNdx[qp, i])
end


# Cast to Tensors.(Vec|Tensor) when possible
for dim in 1:3 @eval begin
    @inline function tensor_cast(v::SVector{$dim,T}) where T
        return Vec{$dim,T}(Tuple(v))
    end
    @inline function tensor_cast(v::SMatrix{$dim,$dim,T,M}) where {T,M}
        return Tensor{2,$dim,T,M}(Tuple(v))
    end
end end
# Fallback
@inline tensor_cast(x) = x

function make_values(rdim,sdim,vdim)
    n_qp = 2
    n_shapefuncs = 2
    N = rand(SVector{vdim,Float64}, n_qp, n_shapefuncs)
    dNdx = rand(SMatrix{vdim,sdim,Float64}, n_qp, n_shapefuncs)
    dNdξ = rand(SMatrix{vdim,rdim,Float64}, n_qp, n_shapefuncs)
    return CellVectorValues{rdim,sdim,vdim,Float64,vdim*sdim,vdim*rdim}(N, dNdx, dNdξ)
end

cv_standard = make_values(3, 3, 3)
cv_embedded = make_values(2, 3, 4)

shape_value(cv_standard, 1, 1)        # -> Vec{3}
shape_derivative(cv_standard, 1, 1)   # -> Tensor{2, 3}

shape_value(cv_embedded, 1, 1)        # -> SVector{4}
shape_derivative(cv_embedded, 1, 1)   # -> SMatrix{4, 3}
```

termi-official:
Okay, I think I have broken the judgement of the benchmarks when building the makefile in https://github.com/Ferrite-FEM/Ferrite.jl/pull/388 . Will fix this. I follow up with a quick manual benchmark.

```julia
using Ferrite
using BenchmarkTools

function setup_dhclosed(grid, T)
    dh = T(grid)
    add!(dh, :v, 2, Lagrange{2,RefCube,2}()) 
    add!(dh, :s, 1, Lagrange{2,RefCube,1}())
    close!(dh)
    return dh
end

dim = 2
cellvalues = CellScalarValues(qr, ip);
dh = setup_dhclosed(grid, DofHandler);

@btime CellScalarValues($qr, $ip)

cell = first(CellIterator(dh))
@btime reinit!(cellvalues, cell)
```

Master
* `ctor`:  608.819 ns (6 allocations: 1.45 KiB)
* `reinit!`: 104.871 ns (0 allocations: 0 bytes)

This PR
* `ctor`: 3.104 μs (18 allocations: 2.11 KiB)
* `reinit!`: 106.253 ns (0 allocations: 0 bytes)

I think the regression on the ctor is manageable.

Project.toml
```diff
@@ -31,6 +24,10 @@ Tensors = "1"
 WriteVTK = "1.13"
 julia = "1.6"
 
+[extensions]
```
KristofferC, line 27: I guess this is from using 1.8. A bit unfortunate that it reorganizes the sections. But anyway, should probably put it back to the way it was.

KristofferC:
Would be good to document in what way this is breaking and how to update (e.g. in the changelog).

Project.toml
termi-official: So I guess the ordering of the sections matter in 1.9?

termi-official:
I will add more docs once the design settles. We probably want to fix the vectorization of scalar interpolations first, so I can remove the remaining hacks from this PR.

src/FEValues/cell_values.jl
```diff
@@ -34,137 +34,244 @@ utilizes scalar shape functions and `CellVectorValues` utilizes vectorial shape
...
+    CellVectorValues(Float64, quad_rule, func_interpol, geo_interpol, sdim, vdim)
 end
 
-function CellVectorValues(::Type{T}, quad_rule::QuadratureRule{dim,shape}, func_interpol::Interpolation,
-        geom_interpol::Interpolation=func_interpol) where {dim,T,shape<:AbstractRefShape}
-
-    @assert getdim(func_interpol) == getdim(geom_interpol)
-    @assert getrefshape(func_interpol) == getrefshape(geom_interpol) == shape
+# FIXME sdim should be something like `length(value(geo_interpol))`
+# FIXME vdim should be something like `length(value(func_interpol))`
```
KnutAM, line 112: Would this fix most of the constructor speed regression?

termi-official: If I can find a way to query `sdim` and `vdim` on compile time, likely yes. Currently I introduced the kwargs to pass the information manually (because it is not in the interpolations yet).

KnutAM: Ah, right. We only know `rdim` from the interpolation! 
If a fast track is desired, maybe you could allow passing the kwargs as `Val`'s ?

termi-official:
Converted back to draft as this one has some weird performance regression which I do not understand (docs build times went drastically up).

src/FEValues/cell_values.jl
termi-official: Well, I can't think of a scenario where the ctor will be the bottleneck (3µs is still fast enough, I think).

termi-official:
Is it intended that all data types for N,M, the derivatives and the quadrature should be of type `T`?

src/FEValues/cell_values.jl
termi-official: Resolved it anyway. :)

termi-official:
Maybe we want to fix https://github.com/Ferrite-FEM/Ferrite.jl/issues/676 first such that we can eliminate the introduced extra parameters sdim and vdim in the `CellScalarValues` and `CellVectorValues` constructor.

fredrikekre:
> Maybe we want to fix https://github.com/Ferrite-FEM/Ferrite.jl/issues/676 first such that we can eliminate the introduced extra parameters sdim and vdim in the CellScalarValues and CellVectorValues constructor.

That would let you query `vdim` from the interpolation, but not `sdim`, unless the way to opt-in to having `sdim > rdim` is to vectorize the geometrical interpolation.

src/Ferrite.jl
```diff
@@ -23,11 +24,13 @@ struct RefCube <: AbstractRefShape end
...
+Abstract type which has `CellValues` and `FaceValues` as subtypes. Each Value type is
+equipped with the spatial dimension `sdim` of the grid, the reference elements' dimension
+`rdim`, the data type for the values `T` and the reference element's geometry `refshape`.
 """
-abstract type Values{dim,T,refshape} end
-abstract type CellValues{dim,T,refshape} <: Values{dim,T,refshape} end
-abstract type FaceValues{dim,T,refshape} <: Values{dim,T,refshape} end
+abstract type Values{sdim,rdim,T,refshape} end
+abstract type CellValues{sdim,rdim,T,refshape} <: Values{sdim,rdim,T,refshape} end
+abstract type FaceValues{sdim,rdim,T,refshape} <: Values{sdim,rdim,T,refshape} end
```
fredrikekre, line 33: We should think about which of the parameters make sense to have on the abstract types here, it would be good to only keep the ones that will ever be used by dispatch.

src/FEValues/cell_values.jl
```diff
@@ -34,137 +34,222 @@ utilizes scalar shape functions and `CellVectorValues` utilizes vectorial shape
...
-struct CellVectorValues{dim,T<:Real,refshape<:AbstractRefShape,M} <: CellValues{dim,T,refshape}
-    N::Matrix{Vec{dim,T}}
-    dNdx::Matrix{Tensor{2,dim,T,M}}
-    dNdξ::Matrix{Tensor{2,dim,T,M}}
+#   rdim = reference element dimension
+#   sdim = spatial dimension
+#   vdim = vector dimension (i.e. dimension of evaluation of what `value` should return)
+#   M1   = number of elements in the matrix dNdx (should be vdim × sdim)
+#   M2   = number of elements in the matrix dNdξ (should be vdim × rdim)
+struct CellVectorValues{sdim,rdim,T<:Real,refshape<:AbstractRefShape,vdim,M1,M2} <: CellValues{sdim,rdim,T,refshape}
```
fredrikekre, line 101: Parameter ordering here should also be thought about I think. All of them are needed here, but the ones you would dispatch on should move to the front so you don't have to write `CellVectorValues{<:Any, <:Any, <:Any, <:Any, TheThingIReallyWant}`.

fredrikekre:
> Is it intended that all data types for N,M, the derivatives and the quadrature should be of type T?

I don't think it makes sense to mix number precision here, you will multiple them all in the end so the type with lowest accuracy will "win" anyway. (If that was your question)

Project.toml
fredrikekre: The ordering shouldn't matter, just that 1.8 don't know how to sort this section so it puts it at the end.

src/FEValues/cell_values.jl
fredrikekre: Perhaps `CellVectorValues{vdim, sdim, rdim, T, refshape, M1, M2}` makes sense? Then the first parameter the dimension for  `function_value`, the first two parameters for `function_gradient`. Probably `T` is more useful than `rdim`, but looks nicer to group all the dims together :)

We should also remove `refshape` and replace it with the interpolation(s) as I suggested in https://github.com/Ferrite-FEM/Ferrite.jl/pull/674#discussion_r1161686800 so that you can dispatch on that for selecting the mapping, for example (this part can/should be done in a separate PR though).

fredrikekre:
With the suggested parameter order from https://github.com/Ferrite-FEM/Ferrite.jl/pull/651#discussion_r1165093551 an alternative to https://github.com/Ferrite-FEM/Ferrite.jl/pull/651#issuecomment-1506425261 could be to use `CellVectorValues{vdim,sdim}(...)` as a constructor:
```julia
# Not necessary to explicitly vectorize if specified in the constructor perhaps.
interpolation_fun = VectorizedInterpolation{vdim=4, Lagrange{refdim=2,RefCube,1}()}() 
interpolation_geo = Lagrange{refdim=2, RefCube, 1}()
quad_rule = QuadratureRule{refdim=2, RefCube}(...)

cv = CellVectorValues{vdim=4,sdim=3}(quad_rule, interpolation_fun, interpolation_geo)
```

src/FEValues/cell_values.jl
termi-official: I mean we should be able to query reference dimension and the reference geometry via `Interpolation` anyway. Will fix.

To elaborate on the ordering, I have put vdim on the end because this keeps the parametrization order the same between the values. But I think it also makes sense. To order them by importance as you suggest.

termi-official:
> > Is it intended that all data types for N,M, the derivatives and the quadrature should be of type T?
> 
> I don't think it makes sense to mix number precision here, you will multiple them all in the end so the type with lowest accuracy will "win" anyway. (If that was your question)

What about e.g. complex-valued approximation spaces on real-valued spatial domains?

src/FEValues/cell_values.jl
fredrikekre: If you write a method for both of them you will dispatch on `CellValues` though, so then the order of the subtypes doesn't matter.

src/Ferrite.jl
termi-official: I think from the user side of the interface it makes sense to have access to these parameters to control dispatch on a higher level (e.g. one dispatch for assembly on triangles).

fredrikekre:
> Sorry I am a bit lost why the geometric interpolation should not be vectorized. My idea was that it is vectorized and that we can grab the spatial dimension from the geometric interpolations vdim.

It would be pretty annoying since you would _always_ have to do this. We are also not using a vectorized interpolation internally here, instead the vectorization is that we have vector valued dofs (i.e. the coordinates).

CHANGELOG.md
```diff
@@ -8,10 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Support for vectorized interpolations with dimension different from the spatial 
+  dimension in `CellValues`. ([#651][github-651])
+
 ### Removed
 - **BREAKING**: `MixedDofHandler` has been renamed to `DofHandler`. Update by replacing 
 `MixedDofHandler` to `DofHandler`. ([#667][github-667])
-
+- **BREAKING**: All `CellValues` have a finer granularity on the different types of
+  dimensions used. Custom `CellValues` must be updated to respect the new type 
+  parameters. ([#651][github-651])
```
KnutAM, line 20: Should it also mention that the parameterization is changed, and dispatches must be updated if dispatching on e.g. `T` or `RefShape`?

termi-official: Let's leave the discussion until https://github.com/Ferrite-FEM/Ferrite.jl/pull/627 is merged (+ I merged new master).

termi-official:
Supersedes #295

src/Ferrite.jl
termi-official: Maybe a different angle then. What would a user likely use to dispatch on?

fredrikekre: I have never really used the parameters. Just like in the examples, I only add `::CellValues` to not pass the wrong argument at the wrong slot.

I suppose that it could sometimes be useful to have dim to make sure multiple values have the same ome (`::CellValues{dim}, ::FaceValues{dim}`) but usually you dont have multiple dims in flight in the same script anyway.

termi-official:
So I have updated the parts which I am confident about. I was not able to fully update the shell example, because I do not fully understand all assumptions taken here. Embedded problems work now. I can also provide an example if requested.

src/FEValues/cell_values.jl
```diff
@@ -1,17 +1,20 @@
 """
-    CellValues([::Type{T}], quad_rule::QuadratureRule, func_interpol::Interpolation, [geom_interpol::Interpolation])
+    CellValues([sdim::Int,] [::Type{T},] quad_rule::QuadratureRule, func_interpol::Interpolation, [geom_interpol::Interpolation])
```
fredrikekre, line 2: This is not a very satisfying way of specifying the space dimension, but I am not sure what would be better. Perhaps requiring passing the vectorized geometric interpolation would be an idea?

```
qr = QuadratureRule{RefTriangle}(1)
ip = Lagrange{RefTriange, 1}()
ipg = Lagrange{RefTriangle, 1}() ^ sdim

CellValues(qr, ip, ipg)
```
and extracting `sdim` from `ipg` while unwrapping it to use the scalar versions, still, internally. Thoughts?

fredrikekre:
Looks like construction is still broken for these two cases with vectorized interpolations:
```julia
qr = QuadratureRule{2,RefQuadrilateral}(1)
ip = Lagrange{RefQuadrilateral,1}()

CellValues(qr, ip^1)
CellValues(1, qr, ip^1)
```

fredrikekre:
I pushed a proposed fix https://github.com/Ferrite-FEM/Ferrite.jl/pull/651/commits/291fafbb0b08bf7f062c090f65f5c8ef6b979e51

src/FEValues/cell_values.jl
termi-official: That is what I wanted to do in the first place, but I also wanted to avoid having these weird inconsistencies that we pass the scalar geometric interpolation at some points and the vectorized at other points.

fredrikekre:
Very cool! You can now have e.g. a 2D vector field on a 1D line embedded in 3D space :)

KristofferC:
> You can now have e.g. a 2D vector field on a 1D line embedded in 3D space :)

Examples needed! :D

termi-official:
> > You can now have e.g. a 2D vector field on a 1D line embedded in 3D space :)
> 
> Examples needed! :D

As you wish #715 :P

PR #652, opened 2023-03-29
Move shared dofhandler functionality out of DofHandler.jl

kimauth:
No functional change. This PR moves all functionality that is shared between the dofhandlers out of `Dofs/DofHandler.jl` in order to prepare for removing that file in the future.

### DIFF
--- a/src/Ferrite.jl
+++ b/src/Ferrite.jl
@@ -86,8 +86,8 @@ include("Grid/grid_generators.jl")
 include("Grid/coloring.jl")
 
 # Dofs
-include("Dofs/DofHandler.jl")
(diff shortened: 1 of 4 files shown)


src/Ferrite.jl
```diff
@@ -86,8 +86,8 @@ include("Grid/grid_generators.jl")
 include("Grid/coloring.jl")
 
 # Dofs
-include("Dofs/DofHandler.jl")
 include("Dofs/MixedDofHandler.jl")
+include("Dofs/DofHandler.jl")
```
fredrikekre, line 90: The student becomes the master.

PR #653, opened 2023-03-29
Remove and deprecate nfields

### LINKED ISSUE #444: `Ferrite.nfields` should be called differently

termi-official:
Closes #444 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Internal changes
 - `Ferrite.ndim(dh, fieldname)` has been removed, use `Ferrite.getfielddim(dh,
   fieldname)` instead. ([#658][github-658])
+- `Ferrite.nfields(dh)` has been removed, use `length(Ferrite.getfieldnames(dh))` instead.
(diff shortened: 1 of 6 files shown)


kimauth:
I had been thinking if we should entirely remove `nfields(dh)`. The only place it seems to be used is the soon-to-be-gone `DofHandler`. 
The docstrings for renumbering and the sparsity pattern should probably point to `getfieldnames(dh)`, as the order of the fields matters for them.

I have a hard time coming up with a use case where you'd only want to know the number of fields and nothing more. Usually you'd want to iterate over them and that's what `getfieldnames(dh)` should be used for I believe. 

What do you think?

fredrikekre:
I was about to suggest the same.

termi-official:
Thanks for the feedback on this. I have no strong opinion here. Just thought that it might be a nice convenience function. But I can also remove it if that is what we want to do.

fredrikekre:
Looking at the small diff, it looks like we don't even use it much internally.

Issue #629, opened 2023-03-22
Tracking progress for merging the dofhandlers


kimauth:
Before merging the dofhandlers, the `MixedDofHandler` must be able to do everything that the `DofHandler` does and should ideally be equally fast. This issue keeps track of how far we've come along that way. 

# Progress tracking
The focus here is to compare how a `MixedDofHandler` performs on a **concrete grid** with all fields on the full domain, i.e. how well `MixedDofHandler` works as a drop-in replacement of `DofHandler`. 

Benchmarking code should go in the comments, ideally one category at a time. However, if a method does not work / does not perform well, open a separate issue about it and reference this one. That way it will be easier to keep an overview and do small reviewable PRs to fix issues.

*Syntax*: Given syntax works and yields correct results for `MixedDofHandler`
*Performance*: performance with `MixedDofHandler` is comparable to `DofHandler` + there is a benchmark of it!

|     | Syntax | Performance |
| --- | --- | --- | 
| | | |
|   **Construction**  |  | ✅  | 
|`DofHandler(grid)` | ✅ |   | 
|`add!(dh, name, dim[, ip])` | ✅  |   | 
|`close!(dh)` | ✅ |  |
| | | |
|   **Basic functionality**  |  |  | 
|`ndofs(dh)` | ✅  | ✅ |
|`ndofs_per_cell(dh[, cell])` | ✅ | ✅ |
|`dof_range(dh, field_name)` | ✅  | ❓ |
|`celldofs(dh, i)` | ✅  | ✅ |
|`celldofs!(dofs, dh, i)` | ✅  | ✅   |
| | | |
|   **Renumbering**  |  |  | 
|`renumber!(dh, order)` | ✅  | ✅  |
|`renumber!(dh, DofOrder.FieldWise())`| ✅  | ✅  |
|`renumber!(dh, DofOrder.ComponentWise())`| ✅   | ✅ |
| | | |
|   **Sparsity pattern**  |  |  | 
|`create_sparsity_pattern(dh)`| ✅  | ✅  |
|`create_sparsity_pattern(dh; coupling)`| ✅  | ✅  |
|`create_symmetric_sparsity_pattern(dh)`| ✅  | ✅ |
|`create_symmetric_sparsity_pattern(dh; coupling)`| ✅ | ✅  |
| | | |
|   **Constraints**  |  |  | 
|`ConstraintHandler(dh)`| ✅ | ✅ |
[...]

PR #654, opened 2023-03-29
Use sorted OrderedSet in FieldHandler

fredrikekre:
This changes `FieldHandler.cellset` to be a sorted `OrderedSet` instead of a `Set`. This ensures that loops over sub-domains are done in ascending cell order.

Since e.g. cells, node coordinates, and dofs are stored in ascending cell order this gives a significant performance boost to loops over sub-domains, i.e. assembly-style loops. In particular, this removes the performance gap between `MixedDofHandler` and `DofHandler` in the `create_sparsity_pattern` benchmark in #629.

This is a minimal/initial step towards #625 that can be done before the `DofHandler` merge and rework of `FieldHandler`/`SubDofHandler`.


fredrikekre:
Perhaps `BitSet` would be an alternative here, actually. That would store at maximum `getncells(grid) / 64` (if the cell IDs in a subdomain are randomly distributed in `1:getncells(grid)`) and at minimum `length(set) / 64` (if each subset contain consecutively numbered cells, 1:10, 11:20, ...) `UInt64` per sub-domain. Compared to `Set`, which stores roughly `length(set) * 3 / 2` (I believe dict buffers should be filled up to at most 2/3), this would become worse if your set is randomly numbered and you have more than 64 * 3 / 2 ≈ 100 subdomains. I don't think this matters much though, this storage is nothing compared to dof storage, for example. Maybe it is easier to use BitSet since that is built in :)

KnutAM:
Realized today that at least for threading, the ordered set must be collected: https://julialang.slack.com/archives/C6SMTHQ3T/p1683401329225109. 
Copied example from that thread:
```julia
using OrderedCollections
vec = [1,3,5];
v = zeros(Int, 5);
oset = OrderedSet(vec);
Threads.@threads :static for i in vec
    v[i] = Threads.threadid()
end
Threads.@threads :static for i in oset
    v[i] = Threads.threadid()
end
```
```julia
ERROR: TaskFailedException
Stacktrace:
 [1] wait
   @ .\task.jl:345 [inlined]
 [2] threading_run(fun::var"#45#threadsfor_fun#5"{var"#45#threadsfor_fun#4#6"{OrderedSet{Int64}}}, static::Bool)
   @ Base.Threads .\threadingconstructs.jl:38
 [3] top-level scope
   @ .\threadingconstructs.jl:93

    nested task error: MethodError: no method matching firstindex(::OrderedSet{Int64})
    Closest candidates are:
      firstindex(::Any, ::Any) at abstractarray.jl:402
      firstindex(::Tuple) at tuple.jl:25
      firstindex(::Pair) at pair.jl:50
      ...
    Stacktrace:
     [1] #45#threadsfor_fun#4
       @ .\threadingconstructs.jl:69 [inlined]
     [2] #45#threadsfor_fun
       @ .\threadingconstructs.jl:51 [inlined]
     [3] (::Base.Threads.var"#1#2"{var"#45#threadsfor_fun#5"{var"#45#threadsfor_fun#4#6"{OrderedSet{Int64}}}, Int64})()
       @ Base.Threads .\threadingconstructs.jl:30
```

PR #655, opened 2023-03-29
Refactor to use grid and dofhandler interface internally

termi-official:
First cherry-pick from #486 . Basically uses the interface for the types instead of accessing members of them directly. This is used to allow local views of distributed grids and distributed dof handlers to be used (non-overlapping case!).


src/Grid/grid.jl
```diff
@@ -13,10 +13,23 @@ struct Node{dim,T}
...
+"""
 getcoordinates(n::Node) = n.x
 
 """
-    Ferrite.get_coordinate_eltype(::Node)
+    get_coordinate_type(::Node)
+
+Get the data type of the the node coordinate.
+"""
+get_coordinate_type(::Node{dim,T}) where {dim,T}  = Vec{dim,T}
```
KnutAM, line 29: Perhaps this should be in the AbstractGrid interface, as we discussed in #579 ?

termi-official: Good point.

src/Dofs/ConstraintHandler.jl
```diff
@@ -352,13 +352,14 @@ function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, b
     return local_face_dofs, local_face_dofs_offset
 end
 
-function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpolation::Interpolation, field_dim::Int, offset::Int, bcvalue::BCValues, cellset::Set{Int}=Set{Int}(1:getncells(ch.dh.grid)))
-    if interpolation !== default_interpolation(typeof(ch.dh.grid.cells[first(cellset)]))
+function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpolation::Interpolation, field_dim::Int, offset::Int, bcvalue::BCValues, cellset::Set{Int}=Set{Int}(1:getncells(getgrid(ch.dh))))
+    grid = getgrid(ch.dh)
+    if interpolation !== default_interpolation(typeof(getcells(grid, first(cellset))))
```
KnutAM, line 357:
```suggestion
    if interpolation !== default_interpolation(getcelltype(grid, first(cellset)))
```

src/Dofs/apply_analytical.jl
```diff
@@ -1,18 +1,18 @@
...
 function _default_interpolations(dh::MixedDofHandler)
     fhs = dh.fieldhandlers
-    getcelltype(i) = typeof(getcells(dh.grid, first(fhs[i].cellset)))
+    getcelltype(i) = typeof(getcells(getgrid(dh), first(fhs[i].cellset)))
     ntuple(i -> default_interpolation(getcelltype(i)), length(fhs))
 end
 
 function _default_interpolation(dh::DofHandler)
-    return default_interpolation(typeof(getcells(dh.grid, 1)))
+    return default_interpolation(typeof(getcells(getgrid(dh), 1)))
```
KnutAM, line 9:
```suggestion
    return default_interpolation(getcelltype(getgrid(dh), 1))
```

KnutAM (approved):
LGTM
(But I think it would make sense to use `getcelltype` in a few places)

src/Dofs/apply_analytical.jl
```diff
@@ -1,18 +1,18 @@
 
 function _default_interpolations(dh::MixedDofHandler)
     fhs = dh.fieldhandlers
-    getcelltype(i) = typeof(getcells(dh.grid, first(fhs[i].cellset)))
+    getcelltype(i) = typeof(getcells(getgrid(dh), first(fhs[i].cellset)))
```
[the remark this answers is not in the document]
KnutAM: Oops, sorry - bad code from old pr. Guess local `getcelltype` method should be renamed...

termi-official:
Seems like I forgot some lines. Your suggestion pointed me to them. Thanks!

src/iterators.jl
```diff
@@ -113,7 +113,7 @@ cellid(cc::CellCache) = cc.cellid[]
 celldofs!(v::Vector, cc::CellCache) = copyto!(v, cc.dofs) # celldofs!(v, cc.dh, cc.cellid[])
 
 # TODO: These should really be replaced with something better...
-nfaces(cc::CellCache) = nfaces(eltype(cc.grid.cells))
+nfaces(cc::CellCache) = nfaces(getcelltype(cc.grid))
```
KnutAM, line 116: This fails for non-concrete grids, but unrelated to this pr...
But should be relevant for merging of dofhandlers, cc: @kimauth

termi-official: But should be easy to fix.
```suggestion
nfaces(cc::CellCache) = nfaces(getcelltype(cc.grid, first(cc.cellid.x))
```

kimauth: Yes, I'm aware of this. Not an issue for merging, as this will still work as long as the grid is concrete (so it works for people who are using `DofHandler` now.
Should benchmark that solution, it has the usual problem of extracting an unconcrete cell, which might or might not be an issue.
I think `nfaces(fv::FaceValues)` should be used instead in most cases.

koehlerson:
> . Methods that are typed with `::Grid` or `::DofHandler` are not generic anyway

I disagree with this, you can assemble and solve the heat equation by subtyping `AbstractGrid` https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_abstractgrid.jl

termi-official:
Actually I do not think that this is a question of readability here (because readability of a user often comes from what he is used to read and write).

However, I agree that some methods are specific on our `Grid`. Most of them could be easily generalized to other grid types by simply enforcing the interface (e.g. a grid which does not use `Node` but something else, or a grid holding a grid as in the distributed assembler), which is the reason why I tried to enforce the interface internally to the best of my skills.

koehlerson:
> > I disagree with this, you can assemble and solve the heat equation by subtyping AbstractGrid
> 
> Yes, but you are not hitting any of the code paths that (currently) uses `::Grid` then. But for example, if you have a method with `::ConstraintHandler`, that will always have a `dh` field pointing to the dofhandler, just like `::DofHandler` will always have a field `grid` pointing to the grid (and I ought to be allowed to use it :D )


I think it's the other way around. We are hitting code paths that used to use `::Grid` which incrementally changed to `::AbstractGrid`, e.g. https://github.com/Ferrite-FEM/Ferrite.jl/pull/379 Here the main changes are to periodic BCs in the `ConstraintHandler` that has been added without `<:AbstractGrid` support, which is not problematic, but which should definitely be changed, because the periodic bc are then simply richer in features, aren't they?

For `AbstractDofHandler` I agree that the discussion is different since there isn't a collection of interface functions which make a `AbstractDofHandler` work and is tested (afaik). Nonetheless it should be discussed if this should be the goal if we think of distributed or adaptive things

termi-official:
But you are hitting code paths that uses `<:AbstractGrid`. Here I am not seeing why we should artificially restrict these code paths by enforcing specific variable names (and types) in functions like `reshape_to_nodes` or even `add!`, when they should just work out fine if some `OtherGrid <: AbstractGrid` fullfills the interface. I mean, yes, we can just copy paste the codes and change for example `grid.nodes` to `grid.mynodes` for `OtherGrid`, but I am completely lost to see why this is beneficial over just dispatching `getnodes`. May you elaborate here?

docs/src/devdocs/dofhandler.md
```diff
@@ -19,6 +19,7 @@ Ferrite.SurfaceOrientationInfo
 The main entry point for dof distribution is [`__close!`](@ref).
 
 ```@docs
+Ferrite.getgrid
```
fredrikekre, line 22: Perhaps this should be `get_grid`, but can be changed later.

src/Dofs/ConstraintHandler.jl
```diff
@@ -315,14 +315,14 @@ function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, b
     return local_face_dofs, local_face_dofs_offset
 end
 
-# Dirichlet on nodeset
-function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpolation::Interpolation, field_dim::Int, offset::Int, bcvalue::BCValues, cellset::Set{Int}=Set{Int}(1:getncells(ch.dh.grid)))
-    if interpolation !== default_interpolation(typeof(ch.dh.grid.cells[first(cellset)]))
+function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcnodes::Set{Int}, interpolation::Interpolation, field_dim::Int, offset::Int, bcvalue::BCValues, cellset::Set{Int}=Set{Int}(1:getncells(getgrid(ch.dh))))
+    grid = getgrid(ch.dh)
+    if interpolation !== default_interpolation(getcelltype(grid, first(cellset)))
```
fredrikekre, line 320: `getcelltype(grid, i)` should probably be typeof(getcell(grid, i))`. Although, `getcelltype` can be optimized for homogeneous grids, but not sure that matters.

src/Dofs/DofHandler.jl
```diff
@@ -1,5 +1,17 @@
...
+    getgrid(dh::AbstractDofHandler)
+
+Access the some grid representation for the dof handler.
+    
+!!! note
+    This API function is currently not well-defined. It acts as the interface between 
+    distributed assembly and assembly on a single process, because most parts oft the
+    functionality can be handled by only acting on the locally owned cell set.
+"""
+getgrid(dh::AbstractDofHandler)
```
fredrikekre, line 13: Ah, this PR introduces this function. Perhaps rename then? See comment above.

src/Dofs/DofHandler.jl
```diff
@@ -1,5 +1,17 @@
 abstract type AbstractDofHandler end
 
+"""
+    getgrid(dh::AbstractDofHandler)
+
+Access the some grid representation for the dof handler.
```
fredrikekre, line 6:
```suggestion
Access some grid representation for the dof handler.
```

src/Dofs/DofHandler.jl
```diff
@@ -1,5 +1,17 @@
...
+    
+!!! note
+    This API function is currently not well-defined. It acts as the interface between 
+    distributed assembly and assembly on a single process, because most parts oft the
```
fredrikekre, line 10:
```suggestion
    distributed assembly and assembly on a single process, because most parts of the
```

src/exports.jl
```diff
@@ -119,6 +119,8 @@ export
     Field,
     evaluate_at_grid_nodes,
     apply_analytical!,
+    getgrid,
+    getdim,
```
fredrikekre, line 123: Do these have to be exported? In particular `getdim` might be ambiguous. Perhaps if we want to export it should be split into `get_refdim` and `get_spacedim`?

fredrikekre (approved):
LGTM but see the comment about exporting the new names (and various other things).

src/Dofs/ConstraintHandler.jl
termi-official: I added it because I thought it is a nice convenience which is clearer in the intent.

src/exports.jl
termi-official: Good point. Should I do it in this PR or a separate one?

termi-official: I will make a separate one because I feel like this needs further discussion.

KnutAM:
I'll hold #664 until this one is merged (I think that should be the least amount of work regarding conflict resolution)

PR #657, opened 2023-03-30
Add mapping from cell id to field handler id

fredrikekre:
This patch adds a mapping in the `MixedDofHandler` struct from cell id to `FieldHandler` id. This makes it possible to store data in the `FieldHandler` struct, and still fetch it given only the dof handler and the cell id. For example, the number of dofs per cell is constant within each field handler, but currently it is stored for every cell in the dof handler. In fact, this vector is replaced in this PR, so memory wise the mapping is for free, but strictly more useful.

The extra indirection makes `ndofs_per_cell` approximately twice as expensive. The following benchmark, on a 1000x1000 grid, goes from 280μs to 583μs but ¯\_(ツ)_/¯.

```
function cound_dofs(dh)
    c = 0
    for i in 1:ncells
        c += ndofs_per_cell(dh, i)
    end
    return c
end
```

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -28,6 +28,10 @@ Notice that a `FieldHandler` can hold several fields.
 mutable struct FieldHandler
     fields::Vector{Field}
     cellset::Set{Int}
+    ndofs_per_cell::Int # set in close(::DofHandler)
+    function FieldHandler(fields, cellset)
+        return new(fields, cellset, -1)
+    end
 end
 
 """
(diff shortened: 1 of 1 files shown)


kimauth:
Have you tested if the performance of `celldofs!` is impacted by this?

lijas:
I like the idea of storing dof-information in the field handler. For example it would be nice to store the indices of the face dofs in the fieldhandler (PR: https://github.com/Ferrite-FEM/Ferrite.jl/pull/360)

fredrikekre:
> Have you tested if the performance of celldofs! is impacted by this?

This benchmark
```
function count_dofs(dh)
    buf = zeros(Int, ndofs_per_cell(dh))
    for i in 1:getncells(dh.grid)
        celldofs!(buf, dh, i)
    end
    return buf
end
```
goes from 13.986 ms to 14.148 ms.

PR #660, opened 2023-03-31
Use `BitSet` in `apply_analytical!` and `reshape_to_nodes`

fredrikekre:
This patch uses `BitSet` in `apply_analytical!` and `reshape_to_nodes` for `MixedDofHandler`. The benefit here is twofold: computing the intersection is much faster (basically just bitwise `&`) and the subsequent looping over the cells are done in ascending cell order.

This closes the performance gap between `MixedDofHandler` and `DofHandler` in benchmarks from #629 of `apply_analytical!`, `reshape_to_nodes`, and `vtk_point_data`. For example, here is the benchmark results for `apply_analytical!`:
```
387.853 ms (72 allocations: 34.50 MiB)  # MixedDofHandler master
 55.262 ms (38 allocations: 553.45 KiB) # MixedDofHandler patch
 41.861 ms (14 allocations: 2.27 KiB)   # DofHandler master/patch
```

### DIFF
--- a/src/Dofs/MixedDofHandler.jl
+++ b/src/Dofs/MixedDofHandler.jl
@@ -667,7 +667,7 @@ function reshape_to_nodes(dh::MixedDofHandler, u::Vector{T}, fieldname::Symbol) 
         field_idx === nothing && continue
         offset = field_offset(fh, field_idx)
 
-        reshape_field_data!(data, dh, u, offset, field_dim, fh.cellset)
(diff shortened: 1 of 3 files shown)


KnutAM (approved):
I don't know how it affects performance in a mixed case with sparse cellsets though, but I also don't think this function is very performance critical so more important to not decrease single-fieldhandler performance. 

Found one extra performance boost option for single field...

PR #662, opened 2023-03-31
AbstractGrid

KnutAM:
*I don't expect this to be merged as-is, just trying out some changes to see if it makes sense. In case some of these changes/thoughts makes sense, we can move forward, and I'm not sure if splitting the files makes sense since we only have one grid type currently, but perhaps it makes sense if a distributed grid type is added @termi-official?*

## Defined for all `AbstractGrid`s (can be overloaded for performance etc.)
- `getdim(::AbstractGrid)`
- `getcells(::AbstractGrid, ::Union{Int,Vector{Int})`
- `getncells(::AbstractGrid)`
- `getcelltype(::AbstractGrid, [::Int])`
- `getnodes(::AbstractGrid, ::Union{Int,Vector{Int}})`
- `getnnodes(::AbstractGrid)`
- `nnodes_per_cell(::AbstractGrid, [::Int])`
- `get_node_coordinate(::AbstractGrid, ::Int)` **New**
- `get_coordinate_eltype(::AbstractGrid)`
- `get_coordinate_type(::AbstractGrid)` **New** (also in #655)
- `getcoordinates(::AbstractGrid, ::Union{Int,AbstractCell})` (and equivalent `getcoordinates!`) (Would suggest changing these to `get_cell_coordinates`)
- `cellnodes!(::Vector{Int}, ::AbstractGrid, ::Int)`
- `compute_vertex_values(::AbstractGrid, ::Function)`
- `Base.show`

## Must be defined by a new `CustomGrid<:AbstractGrid`
- `getcells(::CustomGrid)`
- `getnodes(::CustomGrid)`

If nodes are not of type `::Node`, `get_node_coordinate` must be defined for this node. 
Custom cells are a different story too.

## Additional functions defined just for `::Grid`
- All `get*set[s](::Grid,...)` and `add*set!(::Grid, ...)`
- `get*(::Grid, ::String)`
- `transform!`

## Additional notes
Added also `getnodeidxs(::AbstractCell, ::Int)` to generalize `getcoordinates!`

### DIFF
--- a/src/Ferrite.jl
+++ b/src/Ferrite.jl
@@ -81,6 +81,8 @@ include("FEValues/common_values.jl")
 include("FEValues/face_integrals.jl")
 
 # Grid
+include("Grid/AbstractGrid.jl")
+include("Grid/ExclusiveTopology.jl")
 include("Grid/grid.jl")
 include("Grid/grid_generators.jl")
 include("Grid/coloring.jl")
--- a/src/Grid/AbstractGrid.jl
+++ b/src/Grid/AbstractGrid.jl
@@ -171,6 +171,38 @@ function Base.show(io::IO, ::MIME"text/plain", grid::AbstractGrid)
     print(io, " cells and $(getnnodes(grid)) nodes")
 end
 
+"""
+    function compute_vertex_values(grid::AbstractGrid, f::Function)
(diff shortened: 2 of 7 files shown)


KnutAM:
Nvm, I think this just contributes to extra noise (mostly my own thoughts anyways)

PR #664, opened 2023-04-01
get_*_coordinates[!] - consistent naming

KnutAM:
### Currently
- `getcoordinates(::Node)` -> node coordinate
- `getcoordinates[!]([::Vector], ::Grid, ::Union{Int, CellIndex, FaceIndex})` -> cell coordinates
- `getcoordinates(::CellCache)` -> cell coordinates
- `Ferrite.cellcoords!(::Vector, ::Union{Grid,DofHandler}, ::Union{Int, AbstractCell}` -> cell coordinates (internal)
 
### Suggestion 
 Deprecate `getcoordinates[!]` and `Ferrite.cellcoords!`, replace with 
 - `get_node_coordinate(::Node)` and add the new method `get_node_coordinate(::Grid, ::Int)`
 - `get_cell_coordinates[!]([::Vector], ::Grid, ::Union{Int,CellIndexFaceIndex)` 
 - `get_cell_coordinates(::CellCache)`
 
I find `get_node_coordinate(grid, nodenr)` clearer than the current options discussed in #655 (`getcoordinates(getnodes(grid, nodenr))` or `grid.nodes[nodenr].x`)

`cellcoords!(::Vector, dh::DofHandler, cell)` has no direct equivalent function, and `get_cell_coordinates!(::Vector, dh.grid, cell)` should be used instead. In this PR, I think it makes sense *not* to introduce new functionality wrt. giving a dofhandler to access grid functions and leaving such a convenience layer to a separate discussion. 
 
### Future perspective
- `get_face_coordinates[!]([x], ::Grid, ::FaceIndex)` (see #702 )
- `get_edge_coordinates[!]([x], ::Grid, ::EdgeIndex)` 

This will require a separate discussion which depends on the use case: Should be all the nodes on the face/edge or just those defining the face/edge (i.e. endpoints)?


src/deprecations.jl
```diff
@@ -8,3 +8,6 @@ import Base: push!
 @deprecate faces(ip::Interpolation) facedof_indices(ip) false
 @deprecate edges(ip::Interpolation) edgedof_indices(ip) false
 @deprecate nfields(dh::AbstractDofHandler) length(getfieldnames(dh)) false
+@deprecate getcoordinates(node::Node) get_node_coordinate(node) false
+@deprecate getcoordinates(args...) get_cell_coordinates(args...) false
+@deprecate getcoordinates!(args...) get_cell_coordinates!(args...) false
```
termi-official, line 13: I think the false here has to be removed, so users can see the deprecation.

src/Dofs/DofHandler.jl
```diff
@@ -137,7 +137,7 @@ function celldofs(dh::AbstractDofHandler, i::Int)
     return celldofs!(zeros(Int, ndofs_per_cell(dh, i)), dh, i)
 end
 
-#TODO: perspectively remove in favor of `getcoordinates!(global_coords, grid, i)`?
+#TODO: perspectively remove in favor of `get_cell_coordinates!(global_coords, grid, i)`?
 function cellcoords!(global_coords::Vector{Vec{dim,T}}, dh::DofHandler, i::Union{Int, <:AbstractCell}) where {dim,T}
     cellcoords!(global_coords, dh.grid, i)
 end
```
termi-official, line 143: Should this also be called `get_cell_coordinates!`?

src/Grid/grid.jl
```diff
@@ -753,43 +755,43 @@ function addnodeset!(grid::AbstractGrid, name::String, f::Function)
...
 
-@inline getcoordinates!(x::Vector{Vec{dim,T}}, grid::AbstractGrid, cell::CellIndex) where {dim, T} = getcoordinates!(x, grid, cell.idx)
-@inline getcoordinates!(x::Vector{Vec{dim,T}}, grid::AbstractGrid, face::FaceIndex) where {dim, T} = getcoordinates!(x, grid, face.idx[1])
+@inline get_cell_coordinates!(x::Vector{Vec{dim,T}}, grid::AbstractGrid, cell::CellIndex) where {dim, T} = get_cell_coordinates!(x, grid, cell.idx)
+@inline get_cell_coordinates!(x::Vector{Vec{dim,T}}, grid::AbstractGrid, face::FaceIndex) where {dim, T} = get_cell_coordinates!(x, grid, face.idx[1])
 
-# TODO: Deprecate one of `cellcoords!` and `getcoordinates!`, as they do the same thing
-cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid{dim}, i::Int) where {dim,T} = getcoordinates!(global_coords, grid, i) 
+# TODO: Deprecate one of `cellcoords!` and `get_cell_coordinates!`, as they do the same thing
+cellcoords!(global_coords::Vector{Vec{dim,T}}, grid::AbstractGrid{dim}, i::Int) where {dim,T} = get_cell_coordinates!(global_coords, grid, i)
```
termi-official, line 779: Should this also be called `get_cell_coordinates!`?

termi-official, line 779: Rename and deprecate in this PR for consistency?

src/exports.jl
```diff
@@ -76,6 +76,9 @@ export
     getvertexset,
     getcoordinates,
     getcoordinates!,
```
termi-official, line 78: Remove here?

src/Grid/grid.jl
KnutAM: Good point, that makes sense while cleaning up the API!

src/Dofs/DofHandler.jl
KnutAM: I wasn't sure, but I think it makes sense.

But this does introduce a new method, another option is to simply remove it all together and require users to pass a grid if they want grid info.
And with your PR on `getgrid(dh)` that should be easy...

src/deprecations.jl
KnutAM: I didn't know this affected visibility, I thought it only caused export or not.
So for `cellcoords!` (now below), it should be false I think? 
(since that wasn't exported before I don't want to export it now)

src/Dofs/DofHandler.jl
termi-official: Removing this will be breaking. I do not think it is bad to have this convenience function available, because it is a common pattern in our assembly loops to call stuff on the dof handler/field handler and get the correct stuff. What do you think?

src/deprecations.jl
termi-official: The last argument is whether the old name should be exported or not.

src/Dofs/DofHandler.jl
KnutAM: Yes, I believe that is the intention. 
So IMO it is a design choice: Should one expect grid-api functions to also by giving a dofhandler instead of a grid. 
In that case, there is a bunch of other functions that should accept `dh` input as well. 
I think it would be good to be consistent here, so rethinking this I will deprecate `cellcoords!(x,dh,args...)` here, as I think that design discussion should take place separately.

fredrikekre (approved):
LGTM. Does it matter that `get_cell_coordinates(grid, cellid)` allocates and `get_cell_coordinates(cellcache)` doesn’t?
Also, if we merge this then I suggest we commit to a consistent `get_*` naming for exported getters, i.e. `ndofs(dh)` -> `get_n_dofs(dh)` etc. Thoughts?

lijas:
Would it not make sense to keep getcoordinates! for when writing "boundary"-independent code, so that it can dispatch on boundary type:
```
getcoordinates!(x, grid, faceindex)
getcoordinates!(x, grid, edgeindex)
getcoordinates!(x, grid, vertexindex)
```

KnutAM:
> Does it matter that get_cell_coordinates(grid, cellid) allocates and get_cell_coordinates(cellcache) doesn’t?

From a purist point of view, I think it is sub-optimal (the allocating one should be called `collect_cell_coordinates`). But from a new user perspective, I find such a distinction a bit too much. Furthermore, it could be considered an implementation detail since some `AbstractGrid` could store the cell coordinates in each cell. So I think it is ok that one allocates and one doesn't. 

> consistent get_* naming for exported getters

I think this would be nice (I would add that it should be for all *documented* getters), since some (like `get_dim`) should not be exported due to possible ambiguities with other packages

> Would it not make sense to keep getcoordinates! for when writing "boundary"-independent code, so that it can dispatch on boundary type:
> ```
> getcoordinates!(x, grid, faceindex)
> getcoordinates!(x, grid, edgeindex)
> getcoordinates!(x, grid, vertexindex)
> ```

Yes, I've also thought that would be nice, but I haven't found a good use case. Do you have an example from embedded elements/shells or similar where this would be useful?
In any case, since this functionality currently doesn't exist (it is always returning cell coordinates), we could introduce `get_coordinates[!]([x], grid, *index)` that does exactly what you propose as an outer layer to `get_*_coordinates[!]`, without being breaking by following the new naming convention!

PR #667, opened 2023-04-03
Merge `DofHandler` and `MixedDofHandler`

### LINKED ISSUE #624: Merging `DofHandler` and `MixedDofHandler`
### Background

Ferrite.jl has, for quite some time, maintained two different DoF handlers: the original `DofHandler` and the `MixedDofHandler`. The `DofHandler` is limited to i) single element type (e.g. only triangles, or only quads) and ii) to uniform dof distribution (e.g. all fields live on all elements). `MixedDofHandler` enables both these use cases, so functionality-wise it it is a superset of `DofHandler`. Naturally it would be good to only have a single DoF handler instead of maintaining two.

Performance wise the `DofHandler` is generally better, since it doesn't have to deal with the `Union` element type (when combining different elements). A lot of work, mainly by @kimauth has been done to reduce the performance gap, and I think that we are now at a point where we can merge them into a new DoF handler that can support all usecases with good performance.

### `DofHandler`

```julia
dh = DofHandler(grid)
add!(dh, :u, 2, u_interp)
```

### `NewDofHandler`

Initially the plan was to simply remove `DofHandler` and then rename `MixedDofHandler` to `DofHandler`. However, while we 

### Implementation plan
 - [ ] "Fork" `NewDofHandler` from `MixedDofHandler`
 - [ ] Update `NewDofHandler` to new common interface
 - [ ] Compare and test performance and functionality between `NewDofHandler` and `DofHandler`/`MixedDofHandler`
 - [ ] Delete `DofHandler` and `MixedDofHandler`, rename `NewDofHandler` to `DofHandler`.

kimauth:
After a series of previous PRs, the `MixedDofHandler` is now equally performant as the `DofHandler` in all performance critical parts of the code base (see #629).

This PR removes `DofHandler` and those code paths that specialised for it. `MixedDofHandler` is renamed to `DofHandler`. 

This should represent a non-breaking change for users using `DofHandler`. Users using `MixedDofHandler` need to rename all occurrences of `MixedDofHandler` to `DofHandler`, but their code should keep working otherwise.

closes #624


CHANGELOG.md
```diff
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+### Added
+- `DofHandler` now supports fields on subdomains and mixed grids. ([#667][github-667])
+### Removed
+- **BREAKING**: `MixedDofHandler` has been renamed to `DofHandler`. Update by replacing 
+`MixedDofHandler` to `DofHandler`. ([#667][github-667])
```
termi-official, line 13: Maybe we can make this transition more visible by adding an actual deprecation (@deprecate_binding or similar)?

fredrikekre:
Run `git mv src/Dofs/MixedDofHandler.jl src/Dofs/DofHandler.jl` too?

fredrikekre:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/b2b86dd4857d9ee40c6e75ef8f82a817e9db0685/ext/FerriteMetis.jl#L32

fredrikekre:
Thanks @kimauth for the herculean effort!

PR #670, opened 2023-04-03
Update topology construction

### LINKED ISSUE #641: Topology broken for wedges
Forgot to update topology. We also do not have test coverage on this. Problem is that `face_npoints` is not dispatched (and will require adaption to the new interface. Maybe something like `face_npoints(c::Cell,i::Int) = length(faces(c)[i])` is sufficient for all cells.

termi-official:
Closes #641

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -135,19 +135,15 @@ function Base.show(io::IO, ::MIME"text/plain", n::EntityNeighborhood)
 end
 
 """
-    face_npoints(::AbstractCell{dim,N,M)
-Specifies for each subtype of AbstractCell how many nodes form a face
+    nvertices_on_face(cell::AbstractCell, local_face_index::Int)
+Specifies for each subtype of AbstractCell how many nodes form a face.
 """
-face_npoints(::Cell{2,N,M}) where {N,M} = 2
-face_npoints(::Cell{3,4,1}) = 4 #not sure how to handle embedded cells e.g. Quadrilateral3D
+nvertices_on_face(cell::AbstractCell, local_face_index::Int) = length(faces(cell)[local_face_index])
 """
-    edge_npoints(::AbstractCell{dim,N,M)
-Specifies for each subtype of AbstractCell how many nodes form an edge
+    nvertices_on_edge(::AbstractCell, local_edge_index::Int)
+Specifies for each subtype of AbstractCell how many nodes form an edge.
 """
-edge_npoints(::Cell{3,4,1}) = 2 #not sure how to handle embedded cells e.g. Quadrilateral3D
-face_npoints(::Cell{3,N,6}) where N = 4
-face_npoints(::Cell{3,N,4}) where N = 3
-edge_npoints(::Cell{3,N,M}) where {N,M} = 2
+nvertices_on_edge(cell::AbstractCell, local_edge_index::Int) = length(edges(cell)[local_edge_index])
 
 getdim(::Cell{dim}) where dim = dim
 
@@ -156,8 +152,7 @@ abstract type AbstractTopology end
 """
     ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
 `ExclusiveTopology` saves topological (connectivity) data of the grid. The constructor works with an `AbstractCell`
-vector for all cells that dispatch `vertices`, `faces` and in 3D `edges` as well as the utility functions
-`face_npoints` and `edge_npoints`.
+vector for all cells that dispatch `vertices`, `faces` and in 3D `edges`.
 The struct saves the highest dimensional neighborhood, i.e. if something is connected by a face and an
  edge only the face neighborhood is saved. The lower dimensional neighborhood is recomputed, if needed.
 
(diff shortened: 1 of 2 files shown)


test/test_grid_dofhandler_vtk.jl
```diff
@@ -271,9 +270,9 @@ end
     hexgrid = generate_grid(Hexahedron,(2,2,1))
     topology = ExclusiveTopology(hexgrid)
     @test topology.edge_neighbor[1,11] == Ferrite.EntityNeighborhood(EdgeIndex(4,9))
-    @test getneighborhood(topology,hexgrid,EdgeIndex(1,11),true) == [EdgeIndex(4,9),EdgeIndex(2,12),EdgeIndex(3,10),EdgeIndex(1,11)]
+    @test Set(getneighborhood(topology,hexgrid,EdgeIndex(1,11),true)) == Set([EdgeIndex(4,9),EdgeIndex(2,12),EdgeIndex(3,10),EdgeIndex(1,11)])
```
termi-official, line 273: We should not rely on a specific ordering of the neighborhood in the returns.

termi-official:
Tandem with https://github.com/Ferrite-FEM/FerriteViz.jl/pull/74 .

src/Grid/grid.jl
```diff
@@ -206,29 +201,51 @@ function ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
...
     cell_neighbor_table = Vector{EntityNeighborhood{CellIndex}}(undef, length(cells)) 
 
-    for (cellid, cell) in enumerate(cells)    
-        #cell neighborhood
-        cell_neighbors = getindex.((vertex_cell_table,), cell_vertices_table[cellid]) # cell -> vertex -> cell
-        cell_neighbors = unique(reduce(vcat,cell_neighbors)) # non unique list initially 
-        filter!(x->x!=cellid, cell_neighbors) # get rid of self neighborhood
-        cell_neighbor_table[cellid] = EntityNeighborhood(CellIndex.(cell_neighbors)) 
+    for (cellid, cell) in enumerate(cells)
+        cell_neighbors = union([Set{Int}(vertex_cell_table[vertex]) for vertex ∈ vertices(cell) if vertex_cell_table[vertex] != cellid]...)
```
koehlerson, line 205: can we omit the splatting?

src/Grid/grid.jl
```diff
@@ -206,29 +201,51 @@ function ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
...
+        cell_neighbor_table[cellid] = EntityNeighborhood(CellIndex.(collect(cell_neighbors)))
```
koehlerson, line 206:
```suggestion
        cell_neighbor_table[cellid] = EntityNeighborhood(CellIndex.(collect(cell_neighbors)))
```

Why have it as a Set if we collect it anyways?

src/Grid/grid.jl
KristofferC: `reduce(union, ...)`?

termi-official: Already done this locally, but thanks for the Pointer Kristoffer! :)

src/Grid/grid.jl
```diff
@@ -188,15 +183,15 @@ struct ExclusiveTopology <: AbstractTopology
...
 
 function ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
-    cell_vertices_table = vertices.(cells) #needs generic interface for <: AbstractCell
-    vertex_cell_table = Dict{Int,Vector{Int}}() 
-    
-    for (cellid, cell_nodes) in enumerate(cell_vertices_table)
-       for node in cell_nodes
-            if haskey(vertex_cell_table, node)
-                push!(vertex_cell_table[node], cellid)
+    cell_vertex_table = vertices.(cells) #needs generic interface for <: AbstractCell
```
koehlerson, line 186: why not keep the plural? you map from cell to all vertices of the cell and not to a single vertex

koehlerson (commented):
Other than that lgtm

src/Grid/grid.jl
koehlerson: this gets even a bit better by omitting the intermediate vector allocation in the set construction

src/Grid/grid.jl
```diff
@@ -435,6 +437,7 @@ function ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
     face_skeleton_local = Vector{FaceIndex}()
     fs_length = length(face_skeleton_global)
     for (cellid,cell) in enumerate(cells)
+        getdim(cell) == 1 && continue
```
koehlerson, line 440: shouldnt the 1D case also have a "face" to be evaluated for e.g. DG formulations?

termi-official: Yes, but it is already stored in the vertex neighborhood there. I can mirror it, but integrating this weird corner case in the face construction loop is probably not worth it. What do you think?

koehlerson: can we then just extra dispatch `face_skeleton`?
```julia
faceskeleton(top::ExclusiveTopology, grid::AbstractGrid{1}) =  vertex_neighborhood..
```

koehlerson: hm but this doesn't work for embedded lines properly right?

termi-official: Maybe like this?

termi-official: But for mixed dimensional DG stuff you have different problems to solve :D

koehlerson (approved):

termi-official:
Ready for take off @fredrikekre

PR #672, opened 2023-04-04
Rework code for `add!(::ConstraintHandler, ::Dirichlet)`

fredrikekre:
This patch cleans up the code for adding `Dirichlet` constraints to the `ConstraintHandler` a bit after the DofHandler merge. In particular, it deprecates the method
```
add!(::ConstraintHandler, ::FieldHandler, ::Dirichlet)
```
in favor of just
```
add!(::ConstraintHandler, ::Dirichlet)
```
which was added in #427. There is no need to be able to specify the constrained set by both the set passed to `Dirichlet` and by the set given to the `FieldHandler`. Looking at the tests modified in this patch, it seems this was never the intention anyway. This patch also removes the copy of the set in the `Dirichlet` constructor in favor of explicit partitioning of the set over the `FieldHandler`s.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -40,7 +40,7 @@ struct Dirichlet # <: Constraint
     local_face_dofs_offset::Vector{Int}
 end
 function Dirichlet(field_name::Symbol, faces::Set, f::Function, components=nothing)
-    return Dirichlet(f, copy(faces), field_name, __to_components(components), Int[], Int[])
+    return Dirichlet(f, faces, field_name, __to_components(components), Int[], Int[])
 end
 
 # components=nothing is default and means that all components should be constrained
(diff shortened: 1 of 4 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -818,81 +816,62 @@ end
...
     end
-
-    for nodeid in nodeset
-        if nodeid ∈ nodes
-            all || return true 
-        else
-            all && return false
-        end
+    for x in dbcset
+        x in ns && push!(r, x)
```
KnutAM, line 872:
```suggestion
    for nodeid in dbcset
        nodeid in ns && push!(r, nodeid)
```
(x confused me thinking of coordinate:))

KnutAM (approved):
Nice!
Renaming `x -> nodeid` or similar would be nice, otherwise looks good from my perspective

PR #674, opened 2023-04-09
Multiple cellvalues

KnutAM:
Edit: Minimized oldcomments for readability, see also discussion in #680

<details> 
<summary> old comments </summary>
In most cases with multiple cell values, the geometric interpolation and quadrature are the same for all fields. In this draft PR, I'm investigating a few options for creating a single object containing all cellvalues for a given element, making it more convenient, less bug risky, and slightly faster to work with multiple cellvalues. (No changes to the old methods)

Before fine-tuning a finished proposal, I would appreciate input on any of the following questions
1. Do you like the interface inside `assemble_up!` in the updated [incompressible elasticity example](https://ferrite-fem.github.io/Ferrite.jl/previews/PR674/examples/incompressible_elasticity/#Commented-program)?
2. Do you see any issues with options 1 and 2 below that would prevent you from adopting these?
3. Do you see any further options for performance enhancements that I've missed? (See notes in Benchmarking below)

# Implementation options
## 1) `cv::CellMultiValues`
- Stores only one set of geometric values (`M`, `dMdξ`, and `geo_interp`) in the existing `CellValues`: All queries related to geometry remain as before by using the functions on `cv`
- Has a `NamedTuple` with `FunctionValues` (scalar/vector/mixed). In places where functions relating to that specific interpolation, pass, e.g., `cv[:u]`, to the function: e.g. `shape_value(cv[:u], q_point, base_func)`
- Values related to quadrature are accessed as before with `cv`, e.g. `getdetJdV(cv, q_point)`

Compared to before, however, there is only one `cv`. This guarantees that all values have the same quadrature rule, reference shape, etc. The incompressible elasticity example in the docs preview has been updated for this option. 

```julia
struct CellMultiValues{dim,T,RefShape,FVS<:NamedTuple} <: CellValues{dim,T,RefShape}
    geo_values::GeometryValues{dim,T,RefShape}
    fun_values::FVS
    detJdV::Vector{T}
    qr::QuadratureRule{dim,RefShape,T}
end
```

*One thought I had with this approach is that it should generalize better for different mappings; see [comments](https://github.com/Ferrite-FEM/Ferrite.jl/blob/412fe32b85897ef78365878609f9b2b13c1fe6f5/src/FEValues/CellMultiValues.jl#L98-L115) for `apply_mapping!` in `src/FEValues/CellMultiValues.jl`* 

## 2) `cv::CellValuesGroup`
Simply group multiple cellvalues in a struct (internally as a named tuple). 
The calculation of J-inverse is only done for the geometry in the first cellvalue, and that value is used to update all others. Hence this has exactly the same operations during `reinit!` as for `CellMultiValues`, but store redundant `detJdV`, `M`, and `dMdξ`. 
The usage would be equivalent to `CellMultiValues`, except that accessing via the key will return a regular `CellScalarValues` or `CellVectorValues` object, from which even `getnquadpoints` and `getdetJdV` can be queried. 

## 3) `reinit!(cvs::Union{Tuple,NTuple}, x)`
No change in function, just create a `reinit!` function that updates all cellvalues at the same time, reusing the geometry mapping calculation for performance. Currently, this is only implemented simplified with `reinit_multiple!(x, cv1, cv2, ...)` (fixed for 1,2,4 `cv`s)

# Benchmarking
I don't fully understand why, but option 2: `CellValuesGroup` seem to perform the best in most cases. My guess would be that it is related to memory alignment etc., since option 1 stores less information and option 2 requires setting `detJdV` more times. The performance difference is especially evident for a single scalar cell value in a `CellMultiValues`. (There is even a test with a `CellSingleValues` showing that splitting into geometry values and function values seems to make this difference)

However, the differences are often quite small (except for a single cellvalue). It might also be different results depending on whose system it runs on.

The benchmarking script is temporarily included in this pr in the top level folder: `multivals.jl`. 
`CellValues` corresponds to reinit-ing each cell value individually. 

```julia
Triangle in 2D
1 CellValues: :   19.559 ns (0 allocations: 0 bytes)
1 MultiValues :   33.199 ns (0 allocations: 0 bytes)
1 SingleValues:   33.602 ns (0 allocations: 0 bytes)
1 Tuple{CV}   :   19.759 ns (0 allocations: 0 bytes)
1 ValuesGroup :   19.539 ns (0 allocations: 0 bytes)

2 CellValues  :   47.725 ns (0 allocations: 0 bytes)
2 MultiValues :   46.263 ns (0 allocations: 0 bytes)
2 Tuple{CV}   :   46.714 ns (0 allocations: 0 bytes)
2 ValuesGroup :   47.071 ns (0 allocations: 0 bytes)

4 CellValues  :   96.130 ns (0 allocations: 0 bytes)
4 MultiValues :   75.000 ns (0 allocations: 0 bytes)
4 Tuple{CV}   :   73.895 ns (0 allocations: 0 bytes)
4 ValuesGroup :   73.436 ns (0 allocations: 0 bytes)

QuadraticTriangle in 2D
1 CellValues: :   28.945 ns (0 allocations: 0 bytes)
1 MultiValues :   45.354 ns (0 allocations: 0 bytes)
1 SingleValues:   49.798 ns (0 allocations: 0 bytes)
1 Tuple{CV}   :   39.617 ns (0 allocations: 0 bytes)
1 ValuesGroup :   38.345 ns (0 allocations: 0 bytes)

2 CellValues  :   99.255 ns (0 allocations: 0 bytes)
2 MultiValues :   95.975 ns (0 allocations: 0 bytes)
2 Tuple{CV}   :   96.504 ns (0 allocations: 0 bytes)
2 ValuesGroup :   97.134 ns (0 allocations: 0 bytes)

4 CellValues  :   187.070 ns (0 allocations: 0 bytes)
4 MultiValues :   141.458 ns (0 allocations: 0 bytes)
4 Tuple{CV}   :   153.205 ns (0 allocations: 0 bytes)
4 ValuesGroup :   145.599 ns (0 allocations: 0 bytes)

Hexahedron in 3D
1 CellValues: :   291.579 ns (0 allocations: 0 bytes)
1 MultiValues :   360.194 ns (0 allocations: 0 bytes)
1 SingleValues:   359.346 ns (0 allocations: 0 bytes)
1 Tuple{CV}   :   291.416 ns (0 allocations: 0 bytes)
1 ValuesGroup :   283.571 ns (0 allocations: 0 bytes)

2 CellValues  :   846.552 ns (0 allocations: 0 bytes)
2 MultiValues :   772.222 ns (0 allocations: 0 bytes)
2 Tuple{CV}   :   702.963 ns (0 allocations: 0 bytes)
2 ValuesGroup :   753.488 ns (0 allocations: 0 bytes)

4 CellValues  :   1.890 μs (0 allocations: 0 bytes)
4 MultiValues :   1.430 μs (0 allocations: 0 bytes)
4 Tuple{CV}   :   1.490 μs (0 allocations: 0 bytes)
4 ValuesGroup :   1.400 μs (0 allocations: 0 bytes)

Tetrahedron in 3D
1 CellValues: :   104.686 ns (0 allocations: 0 bytes)
1 MultiValues :   121.783 ns (0 allocations: 0 bytes)
1 SingleValues:   123.820 ns (0 allocations: 0 bytes)
1 Tuple{CV}   :   116.167 ns (0 allocations: 0 bytes)
1 ValuesGroup :   108.504 ns (0 allocations: 0 bytes)

2 CellValues  :   266.207 ns (0 allocations: 0 bytes)
2 MultiValues :   236.740 ns (0 allocations: 0 bytes)
2 Tuple{CV}   :   246.083 ns (0 allocations: 0 bytes)
2 ValuesGroup :   231.739 ns (0 allocations: 0 bytes)

4 CellValues  :   531.915 ns (0 allocations: 0 bytes)
4 MultiValues :   376.214 ns (0 allocations: 0 bytes)
4 Tuple{CV}   :   339.500 ns (0 allocations: 0 bytes)
4 ValuesGroup :   356.938 ns (0 allocations: 0 bytes)
```
</details>


termi-official:
Do you have examples when we have the same quadrature (and geometry) but different interpolations?

KnutAM:
> Do you have examples when we have the same quadrature (and geometry) but different interpolations?

Yes, the incompressible elasticity does that: [link](https://ferrite-fem.github.io/Ferrite.jl/previews/PR674/examples/incompressible_elasticity/#Commented-program)
It uses linear ip for pressure, `Lagrange{2,RefTetrahedron,1}()`, and quadratic, `Lagrange{2,RefTetrahedron,2}()` for displacements. (Or do you mean something else like one `Lagrange` and one other type of interpolation. I don't think that makes a difference though)
(Even with the same interpolation, we would without this still need two separate cellvalues, one for vector and one for scalar which would share the geometry and quadrature)

termi-official:
But in this case we would usually use different quadrature rules because the bilinear forms deneding on the pressure does not need the same quadrature order as the bilinear form depending only on the deformation, or am I missing something?

KnutAM:
How would you formulate the interaction between the fields if you don't evaluate them in the same physical points?

One example I know of is selectively reduced integration when the determinant of the deformation gradient is evaluated as the average value for the entire element. (and then you wouldn't have different quadrature rules anyways) 
Normally I believe all fields have the same quadrature rule. At least I think all our examples use the same quadrature rule.

src/FEValues/CellMultiValues.jl
```diff
@@ -0,0 +1,288 @@
...
+    @inbounds for j in 1:getnbasefunctions(funvals)
+        funvals.dNdx[j, q_point] = funvals.dNdξ[j, q_point] ⋅ Jinv
+    end
+    return nothing
+end
+
+#= TODO: This function could be generalized as follows, but does not make sense before we 
+   implement other mappings (i.e. Piola-mappings)
+"""
+    apply_mapping!(funvals::FunctionValues, q_point::Int, detJ, J, Jinv, geo_values)
```
fredrikekre, line 101: Did you have something in mind for how to dispatch on this? I thought the other day that we should change the parametrization of https://github.com/Ferrite-FEM/Ferrite.jl/blob/247d7c3efdd6325a78c7274180d4a439921dcc78/src/FEValues/cell_values.jl#L89 to instead be parametrized on the function and geometric interpolation
```julia
struct CellScalarValues{IPF, IPG}
```
to make this kind of dispatches possible.

fredrikekre:
Interesting that it is only after 4 (maybe 3?) values that it makes a difference.

src/FEValues/CellMultiValues.jl
KnutAM: Yes, that is the primary solution I've thought about.

KnutAM:
> Interesting that it is only after 4 (maybe 3?) values that it makes a difference.

There are some differences even for 2 values, but I agree that they are surprisingly small!

src/FEValues/CellMultiValues.jl
termi-official: Maybe we should gather somewhere the ideas regarding breaking changes of the FEValues to fix some open issues, so we can group the PRs and do not run into the issue of having several breaking changes over several releases.

KnutAM: That would indeed be good, and if those are getting ready would be nice to include in the 0.4 release unless it causes a too long hold-back (I'm eager to start adapting the the merged dofhandlers:)) 
(This PR is actually non-breaking, though)

KnutAM:
Alternatively, just pass the closed DofHandler to a function to setup the appropriate/default cellvalues. This could even return the appropriate value collection for each `SubDofHandler`

I've often reused the same cellvalue for multiple fields though. But an elegant solution would be if a `MultiCellValues` object would have more keys than values, where multiple keys point to the same cellvalues (if the function interpolations for different fields are equal).

KnutAM:
Superseded by #680

KnutAM:
Re-opened this as it is beneficial for generalization compared to #680, but it is slower with the split into `FunctionValues` and `GeometryValues`: 
```
Triangle in 2D
1 CellValues        :   20.762 ns (0 allocations: 0 bytes) # Fix in this PR (should be separate)
1 CellValues(master):   35.347 ns (1 allocation: 80 bytes) # Current master
1 MultiCellValues   :   19.659 ns (0 allocations: 0 bytes) # PR680 
1 MultiCellValues2  :   36.731 ns (0 allocations: 0 bytes) # PR674 (this PR)
1 SingleCellValues  :   37.538 ns (0 allocations: 0 bytes) # PR674 (this PR, SingleCellValues should replace CellValues in that case)
```
Note that performance degradation do not scale with the problem size the additional time is approximately constant. Feel free to try to optimize to get the same performance as for the standard cell values!

PR #679, opened 2023-04-18
Rework the `AbstractCell` interface

### LINKED ISSUE #677: Change parameterization of (Abstract)Cell
Currently we have https://github.com/Ferrite-FEM/Ferrite.jl/blob/aa261308181a5d856742385eafaa88e87481ad16/src/Grid/grid.jl#L25 and https://github.com/Ferrite-FEM/Ferrite.jl/blob/aa261308181a5d856742385eafaa88e87481ad16/src/Grid/grid.jl#L37-L39 which ... is a bit of a mess for several reasons, IMO, in particular:

1. The number of faces, `M`, shouldn't be a type parameter. It is not required for the storage (i.e. an `NTuple` or similar) and we never use it for dispatch. It was introduced in #125 because at the time we stored an `M`-tuple with on-boundary info. This struct field was removed in #130, but not the type parameter. In #279 the `AbstractCell` was introduced, and that propagated `M` also to the abstract type.
2. All concrete cells implemented in Ferrite are squeezed in under the same struct: https://github.com/Ferrite-FEM/Ferrite.jl/blob/aa261308181a5d856742385eafaa88e87481ad16/src/Grid/grid.jl#L85-L106. While it is a bit "cute" perhaps, it does lead to ambiguities.

Here is a suggested re-parametrization:
```julia
abstract type AbstractCell{dim,spacedim,refshape} end

struct Line{spacedim=1} <: AbstractCell{1,spacedim,RefCube}
    nodes::NTuple{2,Int}
end

struct Triangle{spacedim=2} <: AbstractCell{2,spacedim,RefTetrahedron}
    nodes::NTuple{3,Int}
end

...
```
`Line{spacedim}` can then replace `Line`, `Line2D` and `Line3D`, for example.

For the grid, all cells would have the same spacedim (i.e. the dim of the grid). In the grid:
```julia
struct Grid{dim=2,CT}
    cells::Vector{CT}
    # ...
end
```
when mixing cell types, for example `Triangle`s and `Quadrilateral`s, we would then go from
```julia
CT = Union{Cell{2,3,3}, Cell{2,4,4}}
```
to
```julia
CT = Union{Triangle{2}, Quadrilateral{2}}
```
which shouldn't matter to the compiler (but let's benchmark).

fredrikekre:
This patch reworks the `AbstractCell` interface (refer to #677 for motivation and discussion for alternatives to this patch). Specifically, this patch

 - introduces the reference dimension as a type parameter on `AbstractRefShape`,
 - introduces `RefInterval`, `RefTriangle`, `RefQuadrilateral`, and `RefHexahedron` as new subtypes of `AbstractRefShape{refdim}` (note that interpolations are not updated to this system yet since it can be done in a separate patch),
 - deprecates the old `Cell{refdim, nnodes, nfaces}` struct,
 - introduces new structs for all supported cell types, e.g. `struct Triangle <: AbstractCell{RefTriangle}`,
 - defines the unique vertex/edge/face identifiers for DoF-distribution in terms of the reference shape.

The new parametrization makes it possible to dispatch on the reference shape instead of union of concrete implementations, i.e. `::AbstractCell{RefTriangle}` instead of `::Union{Triangle, QuadraticTriangle}`, and similarly for the reference dimension.

One drawback is that it is no longer possible to use "anonymous celltypes" after this patch, i.e. using, e.g., `Cell{3, 20, 6}` and defining methods for that instead of explicitly naming it `SerendipityQuadraticHexahedron` or similar. Instead, you now have to define a struct and some methods. Arguably this is a good thing -- nobody understands the meaning of `Cell{3, 20, 6}`.

For normal usage this patch is not breaking (see e.g. no required changes for examples), unless one dispatches on the `Cell` struct directly.

Specific things that remain to be decided:

 - `struct Triangle <: AbstractCell{2, RefTriangle}` or `struct Triangle <: AbstractRefCell{2, RefTriangle} <: AbstractCell`? I don't know if it is too restrictive to "force" `{refdim, refshape}` on all subtypes of `AbstractCell`. On the other hand, if one implements a non-refshape based element one can define a dummy reference shape. For a real-world example, looking at e.g. https://github.com/kimauth/FerriteCohesiveZones.jl/blob/main/src/cohesive_cell.jl I think that would simply be implemented as ```julia struct CohesiveQuadrilateral <: AbstractCell{2, RefQuadrilateral} nodes::NTuple{4, Int} end ``` which I don't see a direct problem with, other than the fact that the cell will now have 4 edges (according to the fallback definitions). To solve this, one would have to implement a method for `faces(::CohesiveQuadrilateral)` that only return the upper and lower edge. I don't think this is too much to ask for custom elements like this. Alternatively, with #581 merged it is possible to distribute different number of dofs for different edges, so this can also be solved on the (default) interpolation level by distributing 0 dofs on the left and right edge.

 - `AbstractCell{refdim, refshape}` or just `AbstractCell{refshape}`? The dimension is implicit from the shape now, and you can dispatch on `AbstractCell{<:AbstractRefShape{refdim}} where refdim`. Personally I am in favor of including the dimension, even if redundant. It does not cost anything, and sometimes you want to dispatch on the dimension, sometimes on the refshape.

Fixes #677.


KnutAM:
> `AbstractCell{refdim, refshape}`

Is there a way to ensure that `refdim==getrefdim(refshape)` for this case, or is there a risk to get out of sync somehow?

src/Grid/grid.jl
```diff
@@ -73,37 +75,180 @@ the vertices that define an *oriented face*. This function induces the
...
+        (ns[1], ns[2]), (ns[2], ns[3]), (ns[3], ns[1]), # f1, f2, f3
+    )
+end
+
+# RefQuadrilateral (refdim = 2): vertices for vertexdofs, faces for facedofs (edgedofs) and BC
+function vertices(c::AbstractCell{RefQuadrilateral})
+    ns = get_node_ids(c)
+    return (ns[1], ns[2], ns[3], ns[4]) # v1, v2, v3, v4
+end
+function faces(c::AbstractCell{RefQuadrilateral})
```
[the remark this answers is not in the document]
termi-official: Relates to https://github.com/Ferrite-FEM/Ferrite.jl/issues/394#issuecomment-946096302 . We do not have a consistent definition of what `face` and `edge` is, which raises these "problems" over and over again. Currently, we have sometimes that the atomic boundary of a quad is an edge, sometimes it is a face and sometimes it is both.

termi-official:
Supersedes the prototype given in #519 ?

fredrikekre:
> Is there a way to ensure that `refdim == getrefdim(refshape)` for this case, or is there a risk to get out of sync somehow?

That can be encoded in the type system:
```julia
julia> abstract type AbstractRefShape{dim} end

julia> struct RefInterval <: AbstractRefShape{1} end

julia> struct RefTriangle <: AbstractRefShape{2} end

julia> abstract type AbstractCell{refdim, refshape <: AbstractRefShape{refdim}} end

julia> struct Nope <: AbstractCell{2, RefInterval} end
ERROR: TypeError: in AbstractCell, in refshape, expected refshape<:AbstractRefShape{2}, got Type{RefInterval}
Stacktrace:
 [1] top-level scope
   @ REPL[5]:1

julia> struct Yep <: AbstractCell{2, RefTriangle} end
```

> Supersedes the prototype given in https://github.com/Ferrite-FEM/Ferrite.jl/pull/519 ?

Yea, this roughly corresponds to the Grid changes in that PR.

src/Grid/grid.jl
```diff
@@ -73,37 +62,176 @@ the vertices that define an *oriented face*. This function induces the
...
+# TODO: Deprecate
+const Line2D = Line
+const Line3D = Line
+const Quadrilateral3D = Quadrilateral
+edges(c::Quadrilateral3D) = faces(c)
+
+# Serendipity interpolation based cells
+# TODO: Naming?
+struct SQuadraticQuadrilateral <: AbstractCell{2, RefQuadrilateral} nodes::NTuple{ 8, Int} end
+struct SQuadraticHexahedron    <: AbstractCell{3, RefHexahedron}    nodes::NTuple{20, Int} end
```
KnutAM, line 225: I think this is a good name, as long as we add some docs to clarify the naming convention.

fredrikekre: What about using just `Q` for quadratic too then? `QTriangle`, `QQuadrilateral`, `SQQuadrilateral` etc.?

KnutAM: I'm more skeptical about that, to me that's a bit hard to interpret. 
I prefer the current version being spelled out (it is not so often one has to spell it, so avoiding e.g. `SerendipityQuadraticQuadrilateral` is IMO mostly to avoid having too long types when printing e.g. error messages and stack-traces.

termi-official: Since we are discussing naming: Another problem for higher order elements which we might want to consider is that we do not really give information about the node placement (currently only equidistant, but could also be e.g. GLL placement).

PR #680, opened 2023-04-19
MultiCellValues

KnutAM:
`MultiCellValues` holds multiple `CellValue`s in a named tuple. 

* Main advantage: Only pass around one object for a given domain, and it is easy to get the correct `CellValues` for a given field. 
* All queries regarding geometry values or quadrature rules can be done directly on the `cvs::MultiCellValues`, e.g. `getdetJdV(cvs, q_point)`
* Queries regarding function values are done as `cvs[name]` (where `name::Symbol` typically is the fieldname), e.g. `shape_value(cvs[:u], q_point, i)`
* ~During `reinit!(cvs::MultiCellValues, x::AbstractVector)`, the geometric calculation is only done once~ (Removed to simplify code, slower with many fields, but not worth the extra code)
* Supports having the same `CellValues` for multiple fields (i.e. `cvs[:a]===cvs[:b]` is possible)

From my current testing, this should now be close to equally performant as a single cellvalue if there is only one.

*Standard constructor*
```julia
qr = QuadratureRule{RefTriangle}(2)
ip_vector = Lagrange{RefTriangle,2}()^2
ip_scalar = Lagrange{RefTriangle,1}()
cv_vector = CellValues(qr, ip_vector)
cv_scalar = CellValues(qr, ip_scalar)
cvs = MultiCellValues(;u=cv_vector, p=cv_scalar, T=cv_scalar)
```

*Convenience constructor*
```julia
cvs = MultiCellValues(dh::Union{DofHandler,SubDofHandler}; qr)
```
* Keys (fieldname) and matching function interpolations from `dh`
* Geometric interpolation from `dh.grid`
* Creates a `QuadratureRule{RefShape}(qr)` where `RefShape` matches the celltype in `dh` (If `qr::QuadratureRule`, that is used instead)

(`::DofHandler` is only supported if it has a single `SubDofHandler`.)

Example:
```
julia> grid = generate_grid(Hexahedron, (2,2,2));
julia> dh = DofHandler(grid); add!(dh, :u, Lagrange{RefHexahedron,1}()^3); add!(dh, :p, Lagrange{RefHexahedron,1}()); add!(dh, :T, Lagrange{RefHexahedron,1}()); close!(dh);
julia> cvs = MultiCellValues(dh; qr=2)
MultiCellValues with 2 unique values. Access names:
u: CellVectorValues{3, Float64, RefCube, 9} with 24 shape functions and 8 quadrature points
p: CellScalarValues{3, Float64, RefCube} with 8 shape functions and 8 quadrature points
T: CellScalarValues{3, Float64, RefCube} with 8 shape functions and 8 quadrature points
``` 
(Old `show(::CellValues)`, the new one should be improved...)


termi-official:
Stupid question, but why do we have to stop at having a single RefShape? I mean, it should be possible to also support different RefShapes, right?

KnutAM:
It is even stronger, it requires the same concrete subtype of `AbstractCell`. My idea with "`MultiValues`" is that all its values share `QuadratureRule` and geometric interpolation (such that we can reuse the geometric mapping for all function values). 

Hence, there should be one `MultiCellValues` per `FieldHandler` (similar to that you now would need a set of `CellValues` for each fieldhandler) (ref mixed-dof-examle from [448](https://github.com/Ferrite-FEM/Ferrite.jl/blob/f8406fffda50e821e54bfb00d5effb899977dd6e/docs/src/literate/porous_media.jl#L284-L287))

PR #687, opened 2023-04-26
document vtk_grid(::String, ::DofHandler)

KnutAM:
Just moves up the function and add to the docstring the option of providing the dofhandler. 

This option was already used in some examples (e.g. [Helmholtz](https://ferrite-fem.github.io/Ferrite.jl/stable/examples/helmholtz/)) and documented in the [manual](https://ferrite-fem.github.io/Ferrite.jl/stable/manual/export/#Exporting-with-DofHandler)

### DIFF
--- a/src/Export/VTK.jl
+++ b/src/Export/VTK.jl
@@ -17,9 +17,11 @@ nodes_to_vtkorder(cell::AbstractCell) = collect(cell.nodes)
 
 """
     vtk_grid(filename::AbstractString, grid::Grid)
+    vtk_grid(filename::AbstractString, dh::DofHandler)
 
-Create a unstructured VTK grid from a `Grid`. Return a `DatasetFile`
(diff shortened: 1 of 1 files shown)


src/Export/VTK.jl
```diff
@@ -17,9 +17,11 @@ nodes_to_vtkorder(cell::AbstractCell) = collect(cell.nodes)
 
 """
     vtk_grid(filename::AbstractString, grid::Grid)
+    vtk_grid(filename::AbstractString, dh::DofHandler)
```
termi-official, line 20: I think we are missing docs for the compress arg.

KnutAM: Yes, I wasn't sure if that should be documented/public

One option would be to just use `kwargs...` and direct to `WriteVTK.jl`'s documentation, i.e https://juliavtk.github.io/WriteVTK.jl/stable/grids/syntax/#Data-formatting-options for those?
(Its default is also `compress=true`)

PR #689, opened 2023-04-28
More reference cell documentation

fredrikekre:


docs/src/devdocs/reference_cells.md
```diff
@@ -4,6 +4,29 @@ The reference cells are used to i) define grid cells, ii) define shape functions
...
+
+##### Vertices
+Vertices (nodes) are numbered based on their coordinate in the reference system in *reverse*
+lexicographic ordering, i.e. the inner-most numbering loop is for the ``ξ₁`` direction. See
+e.g. [`RefQuadrilateral`](@ref Ferrite.RefQuadrilateral) for an example.
+
+##### Edges
+Edges are defined by the 2-tuple of vertices it connects, starting with the smallest vertex
+number. To enumerate the edges of the reference cell lexicographic ordering is used on the
+vertex tuples. See e.g. [`RefQuadrilateral`](@ref Ferrite.RefQuadrilateral) for an example.
```
fredrikekre, line 22: This convention causes a seemingly random order on edges and faces, perhaps it is better to use coordinate ordering for the midpoint of the edge face instead?

termi-official: Maybe lexicographic ordering in the reference coordinates? I think this should be almost what we already do.

termi-official: You can interpret your coordinate vector as a word and the coordinates (floats) as a generator for your alphabet.

termi-official: I misread what you wrote. I think we already do something like this on master for the interpolation, right? https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/interpolations.jl#L696-L712

fredrikekre: That would stay (but we would permute in `Ferrite.value`). This PR is for the ordering of the edges themselves, which dof order will match

termi-official: Yes, but the facedofs/edgedofs in the interpolations are lexicographically ordered w.r.t. face/edge center coordinates, right?

fredrikekre: Those functions just connect dof index with entity index, so wouldn't change: We would still want e.g. dof 1 in vertex 1 and first edge dof in edge 1. This PR is for potentially redefining which vertex is vertex 1, and which face is face 1, etc.

docs/src/devdocs/reference_cells.md
termi-official, line 1: I think we lost the vertex coordinates in the translation process, which was the only information about the reference coordinate ranges.

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,263 @@
...
+It is important for internal consistency and correctness that the the same convention is
+used everywhere. The convention adopted by Ferrite.jl is documented below for each reference
+cell.
+
+**Edges** are identified by the 2-tuple of vertices it connects, starting with the smallest
+vertex number, i.e. ``(v_i, v_j)`` where ``i < j``.
+
+**Faces** are identified by the n-tuple of vertices constructing the face in anti-clockwise
+order (viewing the face from the outside of the reference cell), starting with the smallest
+vertex number, i.e. ``(v_i, v_j, v_k, ...)`` where ``i < j, i < k, i < ...``.
```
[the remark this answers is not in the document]
fredrikekre: Actually not right? We only require that the first node is the smallest.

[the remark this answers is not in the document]
termi-official: IIUC then we always sort these fully for faces, see https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Dofs/DofHandler.jl#L729-L735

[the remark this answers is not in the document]
termi-official: What are $v_i,v_j,...$ then? I am also not confident that the faces for all reference elements are counter-clockwise ordered. At least we had in the past the scenario that not all were, but I cannot remember if we changed this. The faces are ordered in a way such that they give us a proper entity of lower dimension.

Also note that the edges are always not ordered as above either.

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,263 @@
...
+```
+
+**Edge identifiers**
+```math
+e_1 = (v_1, v_2), \quad
+e_2 = (v_2, v_3), \quad
+e_3 = (v_3, v_4), \quad
+e_4 = (v_4, v_1)
+```
+
```
KnutAM, line 107:
````suggestion

**Face identifier**
```math
f_1 = (v_1, v_2, v_3, v_4)
```

````

File: docs/src/devdocs/reference_cells.md [hunk 4]
```diff
@@ -1,17 +1,263 @@
...
+### Reference tetrahedron
+
+```@raw html
+<figure>
+<img class="refcell" src="../../assets/ref-tetrahedron.svg" />
+<figcaption><em>
+Numbering of the vertices, edges, faces, and the cell for the reference tetrahedron. Source:
```
KnutAM, line 114:
```suggestion
Numbering of the vertices, edges, faces, and the volume for the reference tetrahedron. Source:
```

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,263 @@
...
+f_4 = (v_1, v_4, v_3)
+```
+
+### Reference hexahedron
+
+```@raw html
+<figure>
+<img class="refcell" src="../../assets/ref-hexahedron.svg" />
+<figcaption><em>
+Numbering of the vertices, edges, faces, and the cell for the reference hexahedron. Source:
```
KnutAM, line 154:
```suggestion
Numbering of the vertices, edges, faces, and the volume for the reference hexahedron. Source:
```

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,263 @@
...
+f_6 = (v_5, v_6, v_7, v_8)
+```
+
+### Reference prism
+
+```@raw html
+<figure>
+<img class="refcell" src="../../assets/ref-prism.svg" />
+<figcaption><em>
+Numbering of the vertices, edges, faces, and the cell for the reference prism. Source:
```
KnutAM, line 206:
```suggestion
Numbering of the vertices, edges, faces, and the volume for the reference prism. Source:
```

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,263 @@
...
 Ferrite.AbstractRefShape
+Ferrite.RefSimplex
+Ferrite.RefHypercube
 Ferrite.RefLine
 Ferrite.RefTriangle
 Ferrite.RefQuadrilateral
 Ferrite.RefTetrahedron
 Ferrite.RefHexahedron
 Ferrite.RefPrism
 ```
```
KnutAM, line 258:
````suggestion
Ferrite.RefPyramid
```
````

And include above?

**Review by KnutAM — COMMENTED** [review]
Even if obvious, for embedded elements including the face numbering for 2d makes sense IMO. (Same as the edge is already included for a line)

I really like that we move refshapes to a separate file! 
Also the following should probably be moved. 

https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/Grid/grid.jl#L120-L132

https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/Grid/grid.jl#L174-L225

File: docs/src/devdocs/reference_cells.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (KnutAM) [reply docs/src/devdocs/reference_cells.md:21]: Does this ordering really matter that much? 

I guess the 2d edges are just made such that `face = first.(e1, e2, ...)`, but for the edge I think one only have to care about this when setting up vector interpolations, such that the positive direction is in the same direction as the edge, see e.g. [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/714d67c483d292f8244b53d990e0ac9ae0df2c54/src/interpolations.jl#L1717-L1740)). During cellvalues mapping, we simply need to [reverse ](https://github.com/Ferrite-FEM/Ferrite.jl/blob/714d67c483d292f8244b53d990e0ac9ae0df2c54/src/FEValues/FunctionValues.jl#L198) if `edges(cell)[edgenr][1] > edges(cell)[edgenr][2]`.

And in 3d (haven't implemented Hdiv for that yet where it matters for me), but didn't you say @fredrikekre that we should consider having something like

```julia
function face_direction(cell, facenr)
    face = faces(cell)[facenr]
    N = length(face)
    imin = argmin(face
    return face[mod1(imin-1, N)] > face[mod1(imin+1, N)] ? 1 : -1
end
```
(Do we already use this somewhere?)
I'm too tired now to think if this will work independent of local ordering, but I think it will? 

To me, the convention of starting with the lowest local vertex index is reasonable. Not sure if we have to require counter-clockwise, but sounds like a good thing to enforce in any case. I looked through the docs preview, and at least there all reference shapes have are ordered counter-clockwise. However, with moving all reference information to this new file, I would suggest that we test the ordering, 

I think something like this
```julia 
x = reference_coordinates(...)
xc = sum(x)/lenght(x) # refshape center
for fk in faces(refshape)
    xfc = sum(x[i] for i in fk)/length(fk) # face center
    n = cross(x[fk[i+1]] - x[fk[i]], x[fk[i+2]] - x[fk[i+1]])
    @test cdot(n, xfc - xc) > 0 # Normal is pointing out
end
```
should do it?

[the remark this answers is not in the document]
fredrikekre: > IIUC then we always sort these fully for faces, see https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Dofs/DofHandler.jl#L729-L735

But the convention would be used to populate that `SurfaceOrientationInfo` I think?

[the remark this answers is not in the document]
fredrikekre: > I am also not confident that the faces for all reference elements are counter-clockwise ordered. At least we had in the past the scenario that not all were, but I cannot remember if we changed this. The faces are ordered in a way such that they give us a proper entity of lower dimension.

I am pretty sure this is true and tested. cellnodes for 2d shapes are counter clockwise, and faces for 3d shapes should be so too, starting with the lowest index.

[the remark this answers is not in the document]
fredrikekre: > Also note that the edges are always not ordered as described above either 

Yea I think I wrote this when I had the idea to renumber just like defelement. Not sure what issues switching the edges would cause though.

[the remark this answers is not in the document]
termi-official: At least https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/Grid/grid.jl#L212  face 3 for prism is not starting from the lowest index. Regarding the orientation, does this https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/test/test_interpolations.jl#L164-L166 test the counter clockwise orientation?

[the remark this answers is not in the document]
KnutAM: > test the counter clockwise orientation?

Nice, yes, but for the interpolation numbering AFAIU, 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/interpolations.jl#L881C50-L881C57

the refshape numbering should match for linear interpolations,
https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/Grid/grid.jl#L200-L203

It only tests for the first 3 though, so for quad faces, it would be possible to have ordering like 
```julia
3  4
1  2
```
pass.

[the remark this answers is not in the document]
termi-official: > > test the counter clockwise orientation?
> 
> Nice, yes, but for the interpolation numbering AFAIU, 
> 
> https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/interpolations.jl#L881-L881
> 
> the refshape numbering should match for linear interpolations,
> 
By our convention via `geometric_interpolation` the interpolation numbering here *must* match the numbering in the reference cell, right?

> https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdd5b41f1485aef8274953a4d2dd361582d69e94/src/Grid/grid.jl#L200-L203
> 
> It only tests for the first 3 though, so for quad faces, it would be possible to have ordering like
> 
> ```julia
> 3  4
> 1  2
> ```
> 
> pass.

termi-official:
When checking the defelement PR I just noticed that we never merged this. Should we rebase and go merge? I think this piece is nice to have in the docs.

KristofferC:
Oh boy, this PR has some sweet svgs and ascii art.

fredrikekre:
Funny that you bring up this PR todays cause I thought about it today too and was gonna update it :D

KristofferC:
(gonna need some dark mode version of those figures though...)

docs/src/devdocs/reference_cells.md
```diff
@@ -1,17 +1,273 @@
...
 
+## Numbering and identification of entities
+
+The local numbering of vertices, edges, and faces, on the reference cells, specifies, for
+example, the cell node order, and the order in which interpolations distribute their DoFs.
+It is important for internal consistency and correctness that the the same convention is
+used everywhere. The convention adopted by Ferrite.jl is documented below for each reference
+cell.
+
+**Edges** are identified by the 2-tuple of vertices it connects, starting with the smallest
```
KristofferC: Don't think this is true, for the ref triangle we have `e_3 = (v_3, v_1)`?

fredrikekre: Yea IIRC I had the ambition to maybe move to using DefElement definitions while writing this.

fredrikekre:
Updated.

termi-official:
> (gonna need some dark mode version of those figures though...)

Wanna do it here or separately?

fredrikekre:
It was already included via the CSS rule but let's check the preview if it works still.

KristofferC:
The CSS seems to work well

fredrikekre:
Only for the default dark theme I noticed.

PR #690, opened 2023-04-28
Added citations for the project

pranavvatsa:
created a references.bib file to help with citations

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -14,6 +14,11 @@ using Documenter, Ferrite, FerriteGmsh, FerriteMeshParser
 
 const is_ci = haskey(ENV, "GITHUB_ACTIONS")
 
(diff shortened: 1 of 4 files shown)


docs/make.jl
```diff
@@ -14,6 +14,11 @@ using Documenter, Ferrite, FerriteGmsh, FerriteMeshParser
 
 const is_ci = haskey(ENV, "GITHUB_ACTIONS")
 
+using DocumenterCitations
+
+bib = CitationBibliography("references.bib", sorting = :nyt)
+# makedocs(bib)
```
KnutAM, line 20:
```suggestion
```

KnutAM (commented):
Thanks! I think this would be very nice!

After adding references from docs, I suggest making a separate page for references.

PR #691, opened 2023-04-29
Rename vtk_point_data

### LINKED ISSUE #278: Maybe remove the fallback `vtk_point_data`
Today I spent ~4 hours wondering why my exports looked like:

Screenshot 2020-03-09 at 14 35 40

(note the weird looking thing in the second row).

Turns out that I wrote

```jl
vtk_point_data(vtk, u, "u")
```

Instead of

```jl
vtk_point_data(vtk, dh, u, "u")
```

This meant that this function got used:

https://github.com/KristofferC/JuAFEM.jl/blob/8224282ab4d67cb523ef342e4a6ceb1716764ada/src/Export/VTK.jl#L58-L63

instead of the one that takes into account the order of the dofs. Maybe we should just call it `_vtk_point_data` or something?

KnutAM:
Suggestion to rename `vtk_point_data` instead of overloading the method from `WriteVTK.jl`, would close #278 

My name suggestion is `vtk_node_data` as the function reshapes the data to the nodes. But this could of course be confusing, thinking that the function requires the data to already be sorted by the nodes. 

Some other naming suggestions could be 
- `vtk_field_data` (from #278)
- `vtk_dof_data`
- `<x>_data_at_nodes` where `<x>` could be `save`, `write`, or `export`

Feel free to add other suggestions!
I will add deprecation once the name has been chosen.


fredrikekre:
Maybe it would be better to have e.g. `Ferrite.VTK.write` etc as a thin wrapper around WriteVTK.jl? It is a bit difficult to squeeze it under the same API, and slightly awkward that some of the `vtk_*` functions come from WriteVTK and some from Ferrite. Perhaps something like:
```
VTK.open(filename::String, dh::DofHandler) do vtk
   VTK.write_solution(vtk, dh, u)
   VTK.write_celldata(vtk, sig, "stress")
end
```
Would be nice with just `VTK.write` but that becomes ambiguous in some cases I think.

KristofferC:
I also think overloading the WriteVTK functions is non ideal. More extensible if we have our own function.

KnutAM:
I agree that not mixing overloading and own functions is better.

Even further, perhaps one could make it more flexible/hackable by not restricting to VTK directly, but rather have that as the implemented exporting backend, something like 
```julia
Ferrite.open_vtk(filename::String, dh::DofHandler) do vtk
   write_solution(vtk::Ferrite.VTKStream, u)
   write_celldata(vtk::Ferrite.VTKStream, sig, "stress")
   write_nodedata(vtk::Ferrite.VTKStream, dp, label)
end
```
where `VTKStream` could include dh in its struct.

fredrikekre:
Then the functions from WriteVTK wouldnt work, unless we add methods to them. But perhaps that isn't so useful to support.

KnutAM:
> Then the functions from WriteVTK wouldnt work, unless we add methods to them. But perhaps that isn't so useful to support.

This should be possible with #692 by passing the `vtk` field in `VTKStream`. 

Closing this PR as it is superseded by #692

Issue #686, opened 2023-04-25
Safety net for for vtk export


termi-official:
Currently `vtk_point_data` does not necessarily error if we do something stupid, and does not even warn us. We should at least add some warnings and possibly stop with some meaningful error message if possible.

KristofferC:
Dup of https://github.com/Ferrite-FEM/Ferrite.jl/issues/278 maybe?

termi-official:
Derp.

PR #692, opened 2023-05-01
Abstraction layer around vtk export

### LINKED ISSUE #278: Maybe remove the fallback `vtk_point_data`

KnutAM:
closes #278 (replaces #691, xref #686)

Following the discussion in #691, this PR introduces an abstraction layer for exporting from Ferrite to VTK. With this PR, no functions from `WriteVTK.jl` should be called directly by the user. (If still needed for special/custom reasons, it is possible by accessing the internal methods and or the `vtk` field in `VTKFile`)

Using `vtk_grid` with an `AbstractGrid` or `AbstractDofHandler` will error telling how to upgrade

Main motivations
1) Avoid re-exporting and extending methods from WriteVTK. Makes dispatch and function owner ship more clear and flexible
2) Provides infrastructure for backend-agnostic export by allowing users to define their own backends (see e.g. https://github.com/Ferrite-FEM/Ferrite.jl/pull/692#discussion_r1267038217 below)

The new workflow is
```julia
VTKFile(filename, grid_or_dh; kwargs...) do vtk
    write_solution(vtk, dh::DofHandler, u)
    write_projection(vtk, proj::L2Projector, vals)
    write_cell_data(vtk, celldata)
    write_node_data(vtk, nodedata)
    Ferrite.write_cellset(vtk, grid, [setnames])
    Ferrite.write_nodeset(vtk, grid, setname)
    Ferrite.write_constraints(vtk, ch::ConstraintHandler)
    Ferrite.write_cell_colors(vtk, grid, cell_colors)
end
```
or alternatively
```julia
vtk = VTKFile(filename, grid_or_dh; kwargs...)
# write ....
close(vtk) # Overload of Base.close
```

For writing paraview collections, the workflow is 
```julia
pvd = VTKFileCollection(filename, grid_or_dh; kwargs...)
for t in range(0, end_time, num_steps)
    # calculate e.g. the dof-vector u and cell data s
    addstep!(pvd, t, [grid_or_dh]) do vtk
        write_solution(vtk, dh, u)
        write_cell_data(vtk, s, "stress")
    end
end
close(pvd)
```
Alternatively, 
```julia
pvd = VTKFileCollection(filename, grid_or_dh; kwargs...)
for (n, t) in pairs(range(0, end_time, num_steps))
    # calculate e.g. the dof-vector u and cell data s
    vtk = VTKFile(string(filename, "_", n), grid_or_dh)
    write_solution(vtk, dh, u)
    write_cell_data(vtk, s, "stress")
    addstep!(pvd, vtk, t)
end
close(pvd)
```


koehlerson:
having something like `write_projected` sounds very good to me at least from a new user perspective it seems quite convenient to have this

src/Export/VTK.jl
```diff
@@ -1,3 +1,75 @@
+struct VTKStream{VTK<:WriteVTK.DatasetFile, DH<:Union{DofHandler,Grid}}
+    vtk::VTK
+    grid_or_dh::DH
+end
+"""
+    Base.close(vtks::VTKStream)
+
+Close the vtk stream and save the data to disk. 
+"""
+Base.close(vtks::VTKStream) = WriteVTK.vtk_save(vtks.vtk)
```
fredrikekre, line 10: `close_vtk`?

src/Export/VTK.jl
```diff
@@ -111,46 +171,167 @@ function component_names(::Type{S}) where S
...
-the cell is in the set and 0 otherwise.
+Saves the dirichlet boundary conditions to a vtkfile.
+Values will have a 1 where bcs are active and 0 otherwise
 """
-vtk_cellset(vtk::WriteVTK.DatasetFile, grid::AbstractGrid, cellset::String) =
-    vtk_cellset(vtk, grid, [cellset])
+function write_dirichlet(vtks, ch::ConstraintHandler)
+    
+    if getgrid(vtks) !== ch.dh.grid 
+        @warn("The grid saved in $(typeof(vtks)) and ConstraintHandler are not aliased, no checks are performed to ensure that they are equal")
```
fredrikekre, line 286: Why not error?

src/exports.jl
```diff
@@ -150,12 +150,6 @@ export
     end_assemble,
 
 # VTK export
-    vtk_grid,
```
fredrikekre, line 153: Should the new stuff not be exported?

src/Export/VTK.jl
```diff
@@ -75,29 +137,27 @@ function toparaview!(v, x::SecondOrderTensor{D}) where D
...
+end
+function _vtk_write_nodedata(vtk::WriteVTK.DatasetFile, nodedata::Vector{<:Real}, name::AbstractString)
+    return WriteVTK.vtk_point_data(vtk, nodedata, name)
+end
+function _vtk_write_nodedata(vtk::WriteVTK.DatasetFile, nodedata::Matrix{<:Real}, name::AbstractString; component_names=nothing)
+    return WriteVTK.vtk_point_data(vtk, nodedata, name; component_names=component_names)
 end
 
-function component_names(::Type{S}) where S
+function get_component_names(::Type{S}) where S
```
fredrikekre, line 160: Seems unnecessary?

src/Export/VTK.jl
```diff
@@ -111,46 +171,167 @@ function component_names(::Type{S}) where S
...
     z = zeros(getnnodes(grid))
     z[collect(getnodeset(grid, nodeset))] .= 1.0
-    vtk_point_data(vtk, z, nodeset)
+    WriteVTK.vtk_point_data(vtks.vtk, z, nodeset)
+    return vtks
 end
 
 """
-    vtk_cellset(vtk, grid::Grid)
+    write_cellset(vtks)
```
fredrikekre, line 257: This now becomes somewhat weird looking when you don't pass the grid. Perhaps it is fine to still do that? (And same with dofhandler).

src/Export/VTK.jl
KnutAM: I chose the function names to be independent of vtk to make the interface suitable for potential future, or user, extensions. One example could be a JLD2-interface (cc: @koehlerson). Could use `close_stream` if you prefer not overloading `close`?

fredrikekre: Was just thinking it should match `open_vtk`.

src/Export/VTK.jl
KnutAM: In case a user (for some reason I cannot think of) needs to use unaliased meshes. But let's error for now, and change to warning if a need arises.

src/Export/VTK.jl
KnutAM: Do you mean to allow both not passing and passing the grid? (for `write_celldata`, `write_nodedata`, `write_cellset`, `write_nodeset`, and `write_cell_colors`). 
And allow both not passing and passing the dofhandler with `write_solution`? 
Or alternatively, drop the internal dofhandler, and require that the grid or dofhandler is passed?

src/Export/VTK.jl
KnutAM: Fair point! 
Perhaps `open_vtk` should be replaced by `VTKStream` instead?

KnutAM: But is it "ok" to have a constructor in a do-block?

src/Export/VTK.jl
fredrikekre: I was thinking about the latter option. In general I think the grid/dofhandler will be available so you don't really need to pass some extra arguments just to send it to the vtk output function.

src/Export/VTK.jl
KnutAM: My thinking was that the user (since direct JLD2 support in Ferrite doesn't seem to be on the roadmap currently) can define e.g., 
```julia
struct JLDStream{T<:"JLD2File?"}
   jldfile::T
end
``` 
and then define e.g.
* `write_solution(jld::JLDStream, dh, u)`
* `write_nodedata(jld::JLDStream, grid, data, name)`
* `write_celldata(jld::JLDStream, grid, data, name)`
* ... 

Where less complicated transforms than for vtk are required for e.g. tensorial data since `Vector{Tensor}` can be saved directly.

This makes it possible to use the same code for export independent of the backend.

If we would make a JLD2 extension (or another saving backend) in the future, this could be included there.

src/Export/VTK.jl
KnutAM: Currently, `vtk_cell_data`, `vtk_point_data`, and `vtk_cell_data_colors` don't require a `grid` input. The two first relies on `WriteVTK`'s grid representation, and `vtk_cell_data_colors` first use the `Ncls` field of `DatasetFile`. 

By some not-so-nice deprecations, I can still make it backward compatible, but perhaps it is better to just provide an error on `vtk_grid(filename, ::AbstractGrid)` with an update guide instead? (Similar to for CellScalarValues and CellVectorValues)

test/test_l2_projection.jl
```diff
@@ -267,13 +267,14 @@ function test_export(;subset::Bool)
     end
 
     mktempdir() do tmp
-        fname = vtk_grid(joinpath(tmp, "projected"), grid) do vtk
-            vtk_point_data(vtk, p, p_scalar, "p_scalar")
-            vtk_point_data(vtk, p, p_vec, "p_vec")
-            vtk_point_data(vtk, p, p_tens, "p_tens")
-            vtk_point_data(vtk, p, p_stens, "p_stens")
+        fname = joinpath(tmp, "projected")
```
KnutAM, line 270: @fredrikekre: Note this change. Should it be kept that close(::VTKFile) returns a vector with a single element being the filename when the normal close does not?

src/exports.jl
```diff
@@ -149,14 +148,14 @@ export
...
 
-# VTK export
-    vtk_grid,
-    vtk_point_data,
-    vtk_cell_data,
-    vtk_nodeset,
-    vtk_cellset,
-    vtk_save,
-
+# export data (using e.g. Ferrite.VTKFile)
```
KnutAM, line 151: Chose not to export `VTKFile` since `WriteVTK.jl` use that name internally as well (supertype of e.g. `DatasetFile`), to make it clear that this one belongs to Ferrite.

KnutAM:
I also thought about abstracting the collection interface today. I think this would improve generalization. Syntax could be 
```julia
pvd = PVDFile("test", grid) # Only place specific file-type is referred to 
for t in range(0, 2, 4)
    # Solve the timestep to find u and σeff 
    Ferrite.addstep!(pvd, t) do io 
        write_solution(io, dh, u)
        write_celldata(io, grid, σeff, "Effective stress")
    end
end
```
See https://github.com/Ferrite-FEM/Ferrite.jl/commit/2398490bdf3058d21a0f82ed07447c2234d3d371 for implementation. Can also be done as a follow-up PR but is breaking unless we want to let the user still use functions from WriteVTK.

termi-official:
Two quick questions on the suggestion:
1. How would this interface deal with changing grids?
2. Why do we have to pass the grid for `write_celldata` again?

KnutAM:
> How would this interface deal with changing grids?

`addstep!(pvd, t, grid=pvd.grid)` is the full function, which allows changing grid

> Why do we have to pass the grid for write_celldata again?

For consistency; most functions require a grid/dofhandler/projector while `write_nodedata` and `write_celldata` don't. So my thought was that the general interface should be `write_x(<file>, <how_to_distribute_data>, <data_x>)`, but perhaps that is not such a good idea?

termi-official:
Yea, we are running here into the problem that the dof handler and the solution vector are decoupled. Which is fine as a design. Maybe we want to have some kind of abstraction layer here (as a separate issue).
`write_x(<file>, <data_abstraction_x>; <optional_args>)` maybe?

KnutAM:
> Maybe we want to have some kind of abstraction layer here (as a separate issue).
> write_x(<file>, <data_abstraction_x>; <optional_args>) maybe?

Do you mean something like `SolutionData(dh, u)`? I'm not sure if I follow here; aren't the combinations, such as `grid + celldata`, `dh + solutionvector`, and `projector + vals` that type of layer in an implicit fashion? What would be the advantage of an explicit additional layer?

termi-official:
I am still a bit split here. The advantage is that we have a coupled representation for the solution vector (and parts of it). I am also leaning towards that this code is something that a user designs, but on the other hand it feels like something that everyone has to do in one way or another. Thing e.g. about the scenarion where you have a heat problem, but your diffusion tensor is also defined as a piece-wise polynomial over your domain, then you already end up with at least 2 dof handlers and 2 vectors to manage.

docs/src/topics/export.md
```diff
@@ -10,98 +10,54 @@ u = rand(ndofs(dh)); σ = rand(getncells(grid))
...
+To write VTK-files, Ferrite comes with an export interface with a 
+[`WriteVTK.jl`](https://github.com/jipolanco/WriteVTK.jl) backend to simplify
 the exporting.
 
-First we need to create a file, based on the grid. This is done with the
-`vtk_grid` function:
-
-```@example export
-vtk = vtk_grid("my-solution", grid)
-# hide
```
KnutAM, line 22: Doesn't hide...

docs/src/topics/export.md
```diff
@@ -10,98 +10,50 @@ u = rand(ndofs(dh)); σ = rand(getncells(grid))
...
+a `VTKFileCollection` can be used, which will write one paraview datafile (`.pvd`)
+file and one `VTKFile` (`.vtu`) for each time step. 
+
+```@example pvdexport 
+pvd = VTKFileCollection("my_results", grid)
+for t in range(0, 1, 5)
+    # Do calculations to update u
+    addstep!(pvd, t) do vtk
+        write_solution(vtk, dh, u)
+    end
```
fredrikekre, line 55: Perhaps this could be
```suggestion
    VTKFile(pvd, t) do vtk
        write_solution(vtk, dh, u)
    end
```
?

docs/src/reference/export.md
```diff
@@ -24,10 +24,16 @@ evaluate_at_grid_nodes
...
 
 ## VTK Export
-
 ```@docs
-vtk_grid(filename::AbstractString, grid::Grid{dim,C,T}; compress::Bool) where {dim,C,T} 
-vtk_point_data
-vtk_cellset
-vtk_cell_data_colors
+VTKFile
+VTKFileCollection
```
termi-official, line 29: Why not `ParaviewCollection` collection?

docs/src/topics/export.md
KnutAM: The motivation for the `addstep!` function, or any other generic name, is that it allows replacing `pvd` with any other collection type, allowing you to only change the output file type in a single location.

docs/src/reference/export.md
KnutAM: @fredrikekre and I concluded that `VTKFileCollection` makes the intent/purpose more clear. 
Technically, it is a "paraview datafile", the "paraview collection" is a term that I think is only introduced by `WriteVTK.jl`?

termi-official: I think I am missing something. `.pvd` is one file format to group vtk files in a specific way (which due to paraview the most prominent one) and nothing specific to the WriteVTK.jl package. Another way to group vtk files is for example the `.pvtk` format or the `.xdmf` format. Or is this one not about exporting to paraview anymore?

KnutAM: No, pvd (Paraview datafile) is fine, but we just thought that the name "Paraview Collection" is not as descriptive or generally used (outside WriteVTK). But I could be wrong. Initially, I had PVDFile as another alternative.

termi-official: Just that I understand correctly: We then control which type of "grouping" we write by another input argument for the constructor call to `VTKFileCollection`? E.g. `VTKFileCollection("my_output"; type=ParaViewOutput)`?

KnutAM: Hm, I haven't thought about that! That sounds like a nice idea!

And for sure that would be fairly easy to extend to if more vtk collection types are desired. 
And probably a good idea to gather all of them (vtk-style) the same type, if possible.
But it only makes sense for time series though, so not for grouping multiple files for the same time step.

termi-official: Thinking about FerriteDistributed.jl - how would we handle a ParaviewCollection with pvtk files in the new abstraction layer?

KnutAM: Nice! Yes, that is a good way to check how well it generalizes. 
I think that we should relax the type demands on the `vtk` field in `VTKFile` to `WriteVTK.VTKFile` instead of `WriteVTK.DatasetFile`, to allow `WriteVTK.PVTKFile` as well. Then the `VTKFile` constructor would just need to be overloaded for the `AbstractNODGrid` or (dofhandler), and if needed write* functions too. 

This should make the collection work automatically. Not sure how the distributed files work though... Do they all get the same names and need to be combined at the end somehow?

Do you have an example with the old `paraview_collection` already?

fredrikekre: Can pvd files not refer to pvtk files?

lijas:
Do you think that multi-block files will work with this api? https://juliavtk.github.io/WriteVTK.jl/stable/metadata/multiblock/

Sometimes the mesh does not represent the actual geometry e.g in immersed method or if you have rigid bodies (the rigid-body-element contain no information about the geometry of the body). In these cases you have to "project" the results on some new mesh that represents the geometry, and then add that geometry to the vtk-output via a vtk multi-block.

KnutAM:
> Do you think that multi-block files will work with this API?

Without having used that feature myself, I would think that any block-structure meta-information has to be implemented as a separate type, similar to the collection type. As we aim to provide a "thin wrapper" around WriteVTK, we need to explicitly support any function ourselves. 

```julia
vtm = VTKMultiBlock(args...)
VTKFile(vtm, [filename], grid) do vtk
 write_*(vtk, ...)
end
```
should be straightforward. 
To generalize, I guess something like 
```
vtm = VTKMultiBlock(args...)
addpart!(vtm, grid) do io
 write_*(io, ...)
end
```
could work - but without experience I wouldn't know what is suitable for a general interface. 


Of course, `Ferrite.create_vtk_grid` and `Ferrite.project_to_nodes` can be used to make custom solutions. Probably a good point to document these/their use perhaps to show how users can hook in to make custom solutions.

KnutAM:
Realized customization required some minor changes. Something like this can now be done for custom solutions, and then we can add such cases into the API when needed without being breaking at that point. Do you think documenting something like this makes sense @lijas?

```julia
using Ferrite
import WriteVTK

g1 = generate_grid(Quadrilateral, (10,10), zero(Vec{2}), ones(Vec{2}))
g2 = generate_grid(Triangle, (10, 10), -ones(Vec{2}), zero(Vec{2}))

dh1 = close!(add!(DofHandler(g1), :u, Lagrange{RefQuadrilateral,1}()^2))
dh2 = close!(add!(DofHandler(g2), :u, Lagrange{RefTriangle,1}()^2))
a1 = apply_analytical!(zeros(ndofs(dh1)), dh1, :u, identity)
a2 = apply_analytical!(zeros(ndofs(dh2)), dh2, :u, x->2*x)

gd1 = Ferrite.create_vtk_griddata(g1)
gd2 = Ferrite.create_vtk_griddata(g2)

vtm = WriteVTK.vtk_multiblock("my_vtm_file")
# First block.
let vtk = WriteVTK.vtk_grid(vtm, gd1...)
    vtk["u"] = Ferrite.evaluate_at_grid_nodes(dh1, a1, :u)
end

# Second block.
let vtk = WriteVTK.vtk_grid(vtm, gd2...)
    vtk["u"] = Ferrite.evaluate_at_grid_nodes(dh2, a2, :u)
end

WriteVTK.vtk_save(vtm)
```
I thought `evaluate_at_grid_nodes` for `Vec` data would require the vtk=Val(true) from the internal method, but seems to work without... - any changes on the `WriteVTK` / Paraview side? @fredrikekre ?

lijas:
@KnutAM Nice. I don't necessarily think that it has to be documentet, I just wanted to check that it was still possible to use multi-blocks with this PR :). Good work with this PR

docs/src/reference/export.md
KnutAM: Good point @fredrikekre!
Modifying the pvtk example from WriteVTK seems to work, but perhaps some redundant pointers to different files. Opening `pvd1.pvd` below in Paraview also works fine, but even `pvd2.pvd` seems to contain all the data...
In any case, by using the extra options that would be available when we overload this ourselves in `FerriteDistributed`, I'm quite confident that we would be in an even better situation wrt. making sure it works together for those cases (one option could be to only save the collection on the main processor, and that collection should only refer to the `pvtu` files, which again points to the `vtu` files for each processor at the current time step). 
@termi-official: Do you think this looks to be promising from your point of view?
```julia
all_data = [
    # Process 1
    (
        points = rand(3, 5),  # 5 points on process 1
        cells = [             # 2 cells  on process 1
            MeshCell(VTKCellTypes.VTK_TRIANGLE, [1, 4, 2]),
            MeshCell(VTKCellTypes.VTK_QUAD, [2, 4, 3, 5]),
        ],
    ),

    # Process 2
    (
        points = rand(3, 4),  # 4 points on process 2
        cells = [             # 1 cell   on process 2
            MeshCell(VTKCellTypes.VTK_QUAD, [1, 2, 3, 4]),
        ]
    ),
]

saved_files = Vector{Vector{String}}(undef, 2)  # files saved by each "process"
pvds = [paraview_collection("pvd$i") for i in 1:2]

for (i, t) in enumerate((0.0, 0.2))
    for part = 1:2
        data = all_data[part]
        saved_files[part] = pvtk_grid(
                "simulation$i", data.points, data.cells;
                part = part, nparts = 2,
            ) do pvtk
            pvtk["Pressure"] = sum(data.points; dims = 1)
            pvds[part][t] = pvtk
        end
    end
end
vtk_save.(pvds)
```
gives 
```julia
2-element Vector{Vector{String}}:
 ["pvd1.pvd", "simulation1.pvtu", "simulation1\\simulation1_1.vtu", "simulation2.pvtu", "simulation2\\simulation2_1.vtu"]
 ["pvd2.pvd", "simulation1\\simulation1_2.vtu", "simulation2\\simulation2_2.vtu"]
 ```

**Comment by lijas** [comment]
1.
Should we consider adding an `AbstractVTKFile`? For example, if someone has a rigid body in their simulation, they probably need to export the rigid body in a separate VTK block. In that case, it could be nice to allow users to create a custom `VTkRigidBodyFile <: AbstractVTKFile`. I also require special output for IGA.jl and my immersed stuff. In such cases, it would be convenient to sub-type `AbstractVTKFile`.

```
vtm = WriteVTK.vtk_multiblock("my_vtm_file")

# First block.
vtk = VTKFile(vtm, dh)
write_solution(vtk, dh, a)

# Rigid body.
vtk = VTkRigidBodyFile(vtm, rgb_geometry)
write_solution(vtk, dh, a, rigid_body_cellid) #compute movement of the rigid body geometry

WriteVTK.vtk_save(vtm)
```

2.
addstep!(pvd, "", dh) do vtk` is a nice syntax for time dependent problems, but I also think it should be possible to it they way it is done in VTK, with like `collection_add_timestep(pvd, vtk, time)`.

**Comment by KnutAM** [comment]
Thanks for the comments Elias! 

1) I thought of doing that but I haven't been able to determine a suitable interface for such an abstract type. So far, I've tried writing the functions that fall back on more basic write routines (e.g. `write_cellset` doesn't require `::VTKFile`, and just calculates and then forwards to `write_celldata`), hoping that this should work quite well. As of now, I don't think there is anywhere that currently dispatches on `VTKFile` which instead could dispatch on a supertype. But perhaps it is possible to write some functions smarter to make use of this? Or does the supertype help you in some other way? I'm not at all against adding it, I just want to be able to motivate it!

2) I didn't include this way in order to ensure backend-agnostic code, but I see now that this might dictate users' codes too much... **So yes, that is probably a good idea to support**, should just be to define `addstep!(pvd, vtk, time)`. Would you also like to keep `pvd[time] = vtk` (I must admit it took me very long before I realized that `time::AbstractFloat` is allowed, which is why I think this is not so nice). 

For the IGA / immersed stuff, do you need to save more information in the, e.g., `VTKFileIGA`? Or is it possible to use `VTKFile` and dispatch the `write_*` on the grid representation in the `dh` or `proj` type?

**Comment by lijas** [comment]
1. True, It would be difficult to share code between the different custom vtk output files, and there would probably not be a common interface other the "open" and "save". 
2. Yes, that was my worry too. It might not always possible for the user to call `addstep!` with the `do` syntax, and they might want to add the timestep with e.g `addstep!(pvd, vtk, time)`  at a later stage. I have never used pvd[time] = vtk myself, so does not matter to me :) . 

I think I will need to save more information in `VTKFileIGA` or `VTKFileImmersed`... I will not be able to use the `VTKFile`

**Comment by lijas** [comment]
Another thought:
For the DG stuff, we can probably create some VTK_DG_File which decouples all elements for visualization of the DG elements/results. What do you think @termi-official ?

(This might only be temporary, since VTK are working on adding DG support)

**Comment by termi-official** [comment]
Why not make this an internal detail? We can dispatch on the continuity type and only if we are H1 export as it is for now. For all other cases the solution is discontinuous across the element boundary anyway.

**Comment by KnutAM** [comment]
> Why not make this an internal detail?

https://github.com/Ferrite-FEM/Ferrite.jl/issues/828

File: src/Export/VTK.jl [hunk 11]
```diff
@@ -1,3 +1,136 @@
...
+
+See [`addstep!`](@ref) for examples for how to use `VTKFileCollection`.
+```
+"""
+mutable struct VTKFileCollection{P<:WriteVTK.CollectionFile,G_DH,KW}
+    pvd::P
+    grid_or_dh::G_DH
+    name::String
+    step::Int
+    vtk_kwargs::KW
```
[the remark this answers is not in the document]
KnutAM: Good catch - I missed that they have one keyword; `append`, which also exists for `vtk_grid` :(

The only solution I can come up with right now is to move the `pvd` `append` keyword argument to a regular argument with the same default as `paraview_collection` (i.e. `false`) and document this. Do you have other ideas?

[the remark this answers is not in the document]
lijas: Sound good. Either way they can always create a seperate vtk and use `addstep!(pvd, vtk, t)`.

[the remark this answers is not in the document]
KnutAM: Another option I thought is to not support passing vtk kwargs to VTKFileCollection for now, and rather aim for that functionality in the future by passing a VTKExporter object, xref #828 ?
Feels like the way this PR forwards kwargs is a bit unconventional/unnecessarily abstract?

In the future, one could then instead do something like 
```julia
pvd = VTKFileCollection("results", VTKExporter(grid_or_dh; vtk_kwargs...); pvd_kwargs...)
```
where `VTKFileCollection("results", grid_or_dh; pvd_kwargs...)=VTKFileCollection("results", VTKExporter(grid_or_dh); pvd_kwargs...)` is the default fallback giving the behavior of this PR. 

And for now, as you say, users can just have the current functionality which would be 
`addstep!(pvd, vtk, t)` if you want to add vtk keyword arguments.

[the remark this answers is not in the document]
lijas: You could also have the pvd_kwargs when creating the VTKFileCollection, and have the vtk-args when in `addsteps!(pvd, t; vtk_kwargs)`

KnutAM:
Not sure why the mac CI fails, seems unrelated to changes but I don't have a mac so cannot test locally.

docs/src/reference/export.md
```diff
@@ -24,10 +24,16 @@ evaluate_at_grid_nodes
...
+addstep!
+write_solution
+write_projection
+write_celldata
+write_nodedata
+Ferrite.write_cellset
+Ferrite.write_nodeset
+Ferrite.write_dirichlet
+Ferrite.write_cell_colors
```
fredrikekre, line 38: There is `celldata` (no `_`) and `cell_colors`. Perhaps be consistent? I think I prefer with the `_` but not sure. `cellset` and `nodeset` matches `getcellset` etc so perhaps they should be without..

src/Export/VTK.jl
```diff
@@ -1,3 +1,135 @@
...
+function VTKFile(f::Function, args...; kwargs...)
+    vtk = VTKFile(args...; kwargs...)
+    try
+        f(vtk)
+    finally
+        close(vtk)
+    end
+end
+
+Base.close(vtk::VTKFile) = WriteVTK.vtk_save(vtk.vtk)
```
fredrikekre, line 49: Perhaps this should be added as a finalizer to the `VTKFile` struct? Or is it already a finalizer added for the upstream struct?

src/Export/VTK.jl
```diff
@@ -113,46 +235,150 @@ function component_names(::Type{S}) where S
...
-        vtk_cell_data(vtk, z, cellset)
+        write_celldata(vtk, z, cellset)
     end
     return vtk
 end
+write_cellset(vtk, grid::AbstractGrid, cellset::String) = write_cellset(vtk, grid, [cellset])
 
 """
-    vtk_cellset(vtk, grid::Grid, cellset::String)
+    write_dirichlet(vtk::VTKFile, ch::ConstraintHandler)
```
fredrikekre, line 330: Perhaps `write_constraints` and we can add a field where affine-constrained dofs are 1? (Can be in a separate PR, but perhaps `write_constriants` could be changed here anyway)

test/runtests.jl
```diff
@@ -38,9 +38,9 @@ include("test_interfacevalues.jl")
 include("test_quadrules.jl")
 include("test_assemble.jl")
 include("test_dofs.jl")
-include("test_constraints.jl")
+include("test_constraints.jl")
```
fredrikekre, line 41:
```suggestion
include("test_constraints.jl")
```

test/runtests.jl
```diff
@@ -38,9 +38,9 @@ include("test_interfacevalues.jl")
 include("test_quadrules.jl")
 include("test_assemble.jl")
 include("test_dofs.jl")
-include("test_constraints.jl")
+include("test_constraints.jl") 
 include("test_grid_dofhandler_vtk.jl")
-include("test_abstractgrid.jl")
+include("test_abstractgrid.jl")
```
fredrikekre, line 43:
```suggestion
include("test_abstractgrid.jl")
```

test/test_l2_projection.jl
fredrikekre: I don't think it is very useful to get the filenames

fredrikekre (approved):

src/Export/VTK.jl
KnutAM: #920

src/Export/VTK.jl
KnutAM: Do you mean https://docs.julialang.org/en/v1/base/base/#Base.finalizer?

I cannot get this to work for a `struct` (complains not mutable), and I'm also not sure exactly what it would do. I don't know exactly how such finalizers works, but I thought that the following sufficed? 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/692cf786be4e6d6463f67a5ec97c1687eb5ce3cd/src/Export/VTK.jl#L40-L47

(But no hits when searching for `finalizer` in the WriteVTK.jl repo, so probably not in the upstream struct)

fredrikekre: > struct (complains not mutable),

Yea, you would have to make it `mutable struct`.

> I'm also not sure exactly what it would do.

Call `vtk_save` (or our `close` wrapper)

> I don't know exactly how such finalizers works, but I thought that the following sufficed?

Yea if you use the `do`-block, but some examples calls `close` directly for example. This could be a separate PR anyway, let's think on it.

src/deprecations.jl
```diff
@@ -59,6 +59,17 @@ Base.@deprecate_binding Line3D Line
 Base.@deprecate_binding Quadrilateral3D Quadrilateral
 export Line2D, Line3D, Quadrilateral3D
 
+using WriteVTK: vtk_grid 
+export vtk_grid # To give better error
+
+function WriteVTK.vtk_grid(::String, ::Union{AbstractGrid,AbstractDofHandler}; kwargs...)
+    error(join(("The vtk interface has been updated in Ferrite v1.0.",
+                "See https://github.com/Ferrite-FEM/Ferrite.jl/pull/679.",
```
koehlerson, line 67: the deprecation message points to the wrong PR

Issue #682, opened 2023-04-24
(Face|Cell)(Scalar|Vector)Values -> (Face|Cell)Values


fredrikekre:
Perhaps we can combine scalar and vector values into a single struct. The `CellScalarValues` and `CellVectorValues` structs are almost identical:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/32d2a589fafe2a01bf5e992aa263997a65532c75/src/FEValues/cell_values.jl#L37-L49

https://github.com/Ferrite-FEM/Ferrite.jl/blob/32d2a589fafe2a01bf5e992aa263997a65532c75/src/FEValues/cell_values.jl#L89-L101

It should be possible to combine these into something like:

```julia
struct CellValues{N_t, dNdx_t, ...}
    N::Matrix{N_t}
    dNdx::Matrix{dNdx_t}
    dNdξ::Matrix{dNdξ_t}
    detJdV::Vector{T}
    M::Matrix{T}
    dMdξ::Matrix{dMdξ_t}
    qr::QR
    func_interp::IPF
    geo_interp::IPG
end

const CellScalarValues = CellValues{<: Number}
const CellVectorValues = CellValues{<: AbstractVector}
const CellMatrixValues = CellValues{<: AbstractMatrix}
```

We only use the distinction for e.g. https://github.com/Ferrite-FEM/Ferrite.jl/blob/32d2a589fafe2a01bf5e992aa263997a65532c75/src/FEValues/common_values.jl#L107 which you can then instead dispatch on `CellValues{<:AbstractVector}`.

Question is if it is worth it.

termi-official:
I kinda like the idea, as it allows to us address several open issues in a simpler way (e.g. complex-valued problems and general tensor-valued problems, ...) and we can reduce some code duplication. However, I am not super confident regarding the interaction with Tensors.jl here, as well as with constraints between the types which are necessary to avoid performance issues.

PR #694, opened 2023-05-08
Implement `(Scalar|Vector)Interpolation` and `VectorizedInterpolation{<:ScalarInterpolation}`

fredrikekre:
This patch adds two subtypes to `Interpolation`: `ScalarInterpolation` and `VectorInterpolation`, and make the existing interpolations (`Lagrange`, `Serendipity`, and `CrouzeixRaviart`) subtypes of `ScalarInterpolation`).

A `ScalarInterpolation` is characterized by `shape_value` returning a scalar and `shape_gradient` a vector, whereas a `VectorInterpolation` is characterized by `shape_value` returning a vector and `shape_gradient` second order tensor.

This patch also adds `VectorizedInterpolation{vdim}` to vectorize a scalar interpolation to `vdim` dimensions by simply duplicating the shape functions `vdim` times. With this we can (re)move the implicit vectorization in the `CellVectorValues` constructor, and the explicit vectorization in `add!(::DofHandler)` to just one (explicit) vectorization when constructing the interpolation.

For example, the following code:
```julia
ip = Lagrange{D,RefCube,1}()
cv = CellVectorValues(ip) # implicit vectorization to D dimensions
add!(dh, :u, 2, ip)       # explicit vectorization to D dimensions
```
can (should) now be written as:
```julia
ip = Lagrange{D,RefCube,1}() ^ 2 # explicit vectorization
cv = CellValues(ip)
add!(dh, :u, ip)
```


src/FEValues/cell_values.jl
```diff
@@ -123,19 +137,8 @@ function CellVectorValues(::Type{T}, quad_rule::QuadratureRule{dim,shape}, func_
...
-                N_comp[comp] = N_temp
-                N[basefunc_count, qp] = Vec{dim,T}((N_comp...,))
-
-                dN_comp = zeros(T, dim, dim)
-                dN_comp[comp, :] = dNdξ_temp
-                dNdξ[basefunc_count, qp] = Tensor{2,dim,T}((dN_comp...,))
-                basefunc_count += 1
-            end
+        for basefunc in 1:n_func_basefuncs
+            dNdξ[basefunc, qp], N[basefunc, qp] = gradient(ξ -> value(func_interpol, basefunc, ξ), ξ, :all)
```
fredrikekre, line 141: This "just works" now without this ugly stuff :)

src/Dofs/ConstraintHandler.jl
```diff
@@ -828,11 +828,15 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
         # Fetch information about the field on this FieldHandler
         field_idx = find_field(fh, dbc.field_name)
         interpolation = getfieldinterpolation(fh, field_idx)
-        field_dim = getfielddim(fh, field_idx)
+        # TODO: Internally here we use the devectorized version for now
```
termi-official, line 831: Is that really an issue?

src/interpolations.jl
```diff
@@ -40,6 +40,16 @@ julia> getnbasefunctions(ip)
...
 
+abstract type ScalarInterpolation{      refdim, refshape, order} <: Interpolation{refdim, refshape, order} end
+abstract type VectorInterpolation{vdim, refdim, refshape, order} <: Interpolation{refdim, refshape, order} end
+
+# Number of components for the interpolation.
+n_components(::ScalarInterpolation)                    = 1
+n_components(::VectorInterpolation{vdim}) where {vdim} = vdim
+# Number of components that are allowed to prescribe in e.g. Dirichlet BC
+n_dbc_components(ip::Interpolation) = n_components(ip)
+# n_dbc_components(::Union{RaviartThomas,Nedelec}) = 1
```
termi-official, line 51: Docs/devdocs?

src/Dofs/DofHandler.jl
```diff
@@ -201,20 +209,27 @@ function _check_cellset_intersections(dh::DofHandler, fh::FieldHandler)
...
+# TODO: Deprecate?
 function add!(dh::DofHandler, name::Symbol, dim::Int)
     celltype = getcelltype(dh.grid)
     isconcretetype(celltype) || error("If you have more than one celltype in Grid, you must use add!(dh::DofHandler, fh::FieldHandler)")
     add!(dh, name, dim, default_interpolation(celltype))
 end
 
+# TODO: Deprecate: If you are passing the interpolation you might as well vectorize it.
+function add!(dh::DofHandler, name::Symbol, dim::Int, ip::ScalarInterpolation)
+    add!(dh, name, dim == 1 ? ip : VectorizedInterpolation{dim}(ip))
```
termi-official, line 221: Should we add the warning now?

src/interpolations.jl
```diff
@@ -1147,3 +1176,46 @@ function value(ip::CrouzeixRaviart{2,1}, i::Int, ξ::Vec{2})
...
+function Base.literal_pow(::typeof(^), ip::ScalarInterpolation, ::Val{vdim}) where vdim
+    return VectorizedInterpolation{vdim}(ip)
+end
+
+# Helper to get number of copies for DoF distribution
+get_n_copies(::VectorizedInterpolation{vdim}) where vdim = vdim
+
+function getnbasefunctions(ipv::VectorizedInterpolation{vdim}) where vdim
+    return vdim * getnbasefunctions(ipv.ip)
+end
```
termi-official, line 1212: Do we have some internal invariants relying on `getnbasefunctions` in combination with the new dof api?

src/Dofs/apply_analytical.jl
```diff
@@ -56,7 +56,8 @@ function _apply_analytical!(
     ref_points = reference_coordinates(ip_fun)
     dummy_weights = zeros(length(ref_points))
     qr = QuadratureRule{dim, RefShape}(dummy_weights, ref_points)
-    cv = CellScalarValues(qr, ip_fun, ip_geo)
+    # Note: Passing ip_geo as the function interpolation here, it is just a dummy.
+    cv = CellScalarValues(qr, ip_geo, ip_geo)
```
termi-official, line 60: This should be constructed via some helper dispatch to extract the scalar-valued interpolation in case of vectorization, right? (e.g. when we use Serendipity for the solution and Lagrange for the geometry)

src/FEValues/cell_values.jl
```diff
@@ -96,23 +96,37 @@ struct CellVectorValues{dim,T<:Real,refshape<:AbstractRefShape,M} <: CellValues{
...
     CellVectorValues(Float64, quad_rule, func_interpol, geom_interpol)
 end
 
-function CellVectorValues(::Type{T}, quad_rule::QuadratureRule{dim,shape}, func_interpol::Interpolation,
-        geom_interpol::Interpolation=func_interpol) where {dim,T,shape<:AbstractRefShape}
+# TODO: Maybe deprecate this auto-vectorizing method?
+function CellVectorValues(::Type{T}, quad_rule::QuadratureRule, func_interpol::ScalarInterpolation,
+        geom_interpol::Interpolation=func_interpol) where {T}
+    return CellVectorValues(T, quad_rule, VectorizedInterpolation(func_interpol), geom_interpol)
+end
```
termi-official, line 119: We might want to add a warning/info log here, at least.

src/FEValues/cell_values.jl
termi-official: Finally some cleanup here :P We still might need to wrap this one into an external function for basis function correction.

src/PointEval/point_values.jl
```diff
@@ -1,47 +1,43 @@
 # Optimized version of PointScalarValues which avoids i) recomputation of dNdξ and
 # ii) recomputation of dNdx. Only allows function evaluation (no gradients) which is
 # what is used in get_point_values.
-struct PointScalarValuesInternal{dim,T<:Real,refshape<:AbstractRefShape} <: CellValues{dim,T,refshape}
-    N::Vector{T}
+struct PointValuesInternal{dim,T<:Real,refshape<:AbstractRefShape,N_t} <: CellValues{dim,T,refshape}
+    N::Vector{N_t}
```
termi-official, line 5: cc @koehlerson we need to make a PR in Ferrite viz to fix this. 

Fredrik: Maybe we can add an internal deprecation of this for now?

termi-official: Also some docs regarding the invariants between $N_t$ and $T$ might be nice. I think I am not really understanding the motivation here anyway.

src/interpolations.jl
```diff
@@ -72,34 +82,53 @@ struct InterpolationInfo
     ncelldofs::Int
     reference_dim::Int
     adjust_during_distribution::Bool
+    n_copies::Int
     function InterpolationInfo(interpolation::Interpolation{3})
+        n_copies = 1
+        if interpolation isa VectorizedInterpolation
+            n_copies = get_n_copies(interpolation)
+            interpolation = interpolation.ip
+        end
```
termi-official, line 91: Why not dispatch `get_n_copies(::Interpolations) = 1`?

termi-official (commented):
I left a first round of comments. I hope these are helpful.

docs/src/literate/stokes-flow.jl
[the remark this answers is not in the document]
fredrikekre: I will update them, deprecate the old style, and adjust tests when the implementation is accepted :)

src/Dofs/apply_analytical.jl
fredrikekre: Doesn't really matter, it is only used for `spatial_coordinate`, could be an [empty interpolation](https://dealii.org/developer/doxygen/deal.II/classFE__Nothing.html#a69940158d653b46222943f09c6b25a4c).

src/PointEval/point_values.jl
fredrikekre: This basically is https://github.com/Ferrite-FEM/Ferrite.jl/issues/682 for this implementation. Previously the vectorization was done on the outside, with awkward reshapes for vector valued functions.

src/interpolations.jl
fredrikekre: Sure, mostly just proof of concept at this point though, since I think quite a bit in the ConstraintHandler need to change for true vector interpolations :)

src/Dofs/apply_analytical.jl
termi-official: Ah, I see. So we do not actually need a full CellScalarValues here but something cheaper could be sufficient. Separate issue then.

fredrikekre: Could extract
```
struct Mapping
    M
    dMdξ
end
end
```
that can be used standalone.

src/interpolations.jl
termi-official: I am pretty sure that we do not need more changes for true vector interpolations.

src/PointEval/point_values.jl
termi-official: Why not constraint T differently then? T is now unused, right?

fredrikekre: It is a type param of the supertype (for now), unfortunately.

termi-official: Will address in https://github.com/Ferrite-FEM/Ferrite.jl/pull/651 then after this is merged. This with PR we should also have everything inplace for the merge (after I merge new master).

src/interpolations.jl
termi-official: Okay we just need to figure out the correct expression on the boundary (e.g. tangent or normal) to make the inner product with.

docs/src/literate/stokes-flow.jl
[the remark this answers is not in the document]
fredrikekre: Done.

src/deprecations.jl
```diff
@@ -57,3 +57,59 @@ Base.@deprecate_binding Line2D Line
...
+    )
+    ip = default_interpolation(celltype)
+    add!(dh, name, dim == 1 ? ip : VectorizedInterpolation{dim}(ip))
+end
+
+function add!(dh::DofHandler, name::Symbol, dim::Int, ip::ScalarInterpolation)
+    Base.depwarn(
+        "`add!(dh::DofHandler, name::Symbol, dim::Int, ip::ScalarInterpolation)` is " *
+        "deprecated. Instead, vectorize the interpolation to the appropriate dimension " *
+        "and add it (`vip = ip^dim; add!(dh, name, vip)`). See CHANGELOG for more details.",
```
termi-official, line 81: Maybe we can be more specific

```suggestion
        "and add it (`vip = ip^dim; add!(dh, name, vip)`). See CHANGELOG for version 0.4.0 for more details.",
```

termi-official: Same for the warning above and below.

termi-official (approved):
Left some comments. LGTM otherwise.

.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: '1'
+          version: '1.8'
```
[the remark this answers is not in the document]
fredrikekre: The meaning of `1` changed to 1.9 now, but `docs/Manifest.toml` is not updated to that version.

src/deprecations.jl
fredrikekre: We can add a link instead (when that section of the CHANGELOG is written).

PR #698, opened 2023-05-09
Deprecate `function_*(::VectorValues, ::Int, ::VectorDofs)`

fredrikekre:
`function_value`, `function_gradient`, `function_divergence`, and `function_curl` together with `CellVectorValues` supports specifying the local dof vector using `Vec` entries. This patch deprecates this. There is no point in casting `T` to `Vec`, just to then recast to `T` for the evaluation. Deprecations are added, but in practice I estimate that zero people use this functionality.

### DIFF
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -153,11 +153,6 @@ function function_value(::FieldTrait, fe_v::Values{dim}, q_point::Int, u::Abstra
     return val
 end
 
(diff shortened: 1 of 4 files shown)


termi-official:
Which ones are you referring to?

fredrikekre:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/2fe73a2c9ff3c65888178c897958e8c282548472/src/FEValues/common_values.jl#L201-L211

PR #699, opened 2023-05-09
Remove `project_to_nodes` option in L2 projection

fredrikekre:


fredrikekre:
The current default is wrong, indicated by having to set it to `false` for all the (practical) examples, and the default behavior of getting the data in the  grid nodes isn't particularly useful (and one can do that transformation separately instead, as is done in `vtk_point_data`).

PR #633, opened 2023-03-23
Making Lagrange2Tri345 interpolation values exact

AbdAlazezAhmed:
Fixes #582, turns out the issue was with `γ` and Float64 precision. using `γ` directly (multiplying `1` and `-( ξ_x +  ξ_y )` by `order` individually) seems to solve the issue.
Interpolation tests passed:
```Julia
Test Summary:  | Pass  Total   Time
interpolations | 3466   3466  47.5s
```

PR #701, opened 2023-05-10
Added interpolation Lagrange{2,RefCube,3}

JanM12:
introduced interpolation Lagrange{2,RefCube,3} to src/interpolations.jl

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -84,6 +84,20 @@ if !liveserver
     @timeit dto "deploydocs" deploydocs(
         repo = "github.com/Ferrite-FEM/Ferrite.jl.git",
         push_preview=true,
+        versions = [
+            "stable" => "v^",
+            "v#.#",
+            "v0.3.13",
+            "v0.3.12",
+            "v0.3.11",
+            "v0.3.10",
+            "v0.3.9",
+            "v0.3.8",
+            "v0.3.7",
+            "v0.3.6",
+            "v0.3.5",
+            "dev" => "dev"
+        ]
     )
 end
 


src/interpolations.jl
```diff
@@ -489,6 +489,56 @@ function value(ip::Lagrange{2,RefCube,2}, i::Int, ξ::Vec{2})
...
+            Vec{2, Float64}((-1/3,  1.0)),
+            Vec{2, Float64}((-1.0,  1/3)),
+            Vec{2, Float64}((-1.0, -1/3)),
+            Vec{2, Float64}((-1/3, -1/3)),
+            Vec{2, Float64}(( 1/3, -1/3)),
+            Vec{2, Float64}((-1/3,  1/3)),
+            Vec{2, Float64}(( 1/3,  1/3))]
+end
+
+function value(ip::Lagrange{2,RefCube,3}, i::Int, ξ::Vec{2})
```
koehlerson, line 520:
```suggestion
function value(ip::Lagrange{2,RefCube,3}, i::Int, ξ::Vec{2})
# see https://defelement.com/elements/examples/quadrilateral-Q-3.html
```

koehlerson (changes requested):
need to add the new interpolation to the test suite, see: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_interpolations.jl#L3-L36

fredrikekre:
Looks like the tolerance here https://github.com/Ferrite-FEM/Ferrite.jl/blob/ade9292688770f54caaccae6eb1af7e3d712a49e/test/test_interpolations.jl#L106needs to be adjusted (or possibly the long expressions here can be reordered to give a smaller error like what was done in #633, but maybe not possible).

termi-official:
Give me some more time to think about this adjustment, because I am not really sure about what could break downstream due to such spurious evaluations and what the related thresholds to prevent numerical oscillations are. Also I do not know how these affect the overall assembly precision. Generally speaking we should still be able to reorder the expressions to eliminate errors in the tested delta property (even though it is not trivial).

fredrikekre:
Perhaps we can BigFloats and truncate to Float64 after evaluation.

termi-official:
But this will be a big performance hit for the value call then.

fredrikekre:
Maybe, but it is only called in the constructor.

AbdAlazezAhmed:
How about using `Rational` ?

termi-official:
Thanks for this suggestions!

> Maybe, but it is only called in the constructor.

For order 3 (and maybe order 4) this should be fine, but when moving to higher order one quickyl runs into the issue that assembly degrades solver performance quite severly. Here usually matrix-free techniques are deployed which require reevaluation of the basis functions. I think `BigFloat` could hurt performance in this scenario quite a bit. To not delay this PW we can go either way and discuss this in a separate issue/discussion, investigating this in more detail (and maybe gathering some literature on this).


> How about using `Rational` ?

This would certainly fix the tests, but IIRC then general quadrature points are unfortunately Irrational. However, I am happy to get corrected if I am wrong on this.

AbdAlazezAhmed:
> This would certainly fix the tests, but IIRC then general quadrature points are unfortunately Irrational. However, I am happy to get corrected if I am wrong on this.

You're right. [AccurateArithmetic.jl](https://github.com/JuliaMath/AccurateArithmetic.jl) can be helpful?

termi-official:
Okay maybe we can bring this over the finish line as follows. We can increase the number of points in the partition of unity test here (points outside the domain should be rejected!)

https://github.com/Ferrite-FEM/Ferrite.jl/blob/1fadc1fc63bf6751e8f56ac1619c40d8a3c87465/test/test_interpolations.jl#L50-L55

and increase the tolerance here https://github.com/Ferrite-FEM/Ferrite.jl/blob/1fadc1fc63bf6751e8f56ac1619c40d8a3c87465/test/test_interpolations.jl#L107 to 1e-14 . This should be reasonably small for now.

Alternatively we can try to overdub the value calls in the tests with Kahan summations, but this might be a bit more work. It think the linked package fails with AD (just skipped over the implementation).

PR #703, opened 2023-05-10
Update VTK export to use CellValues for evalutation

### LINKED ISSUE #167: Export of super parametric elements

fredrikekre:
This patch deprecates `reshape_to_nodes` in favor of the new `evaluate_at_grid_nodes` function which more or less does the same thing. `evaluate_at_grid_nodes` uses the interpolation of the field to evaluate the approximated function in the cell coordinates (by using `reference_coordinates` of the cell as quadrature points). This is much simpler, and more correct, compared to figuring out which Dofs live on which node etc. In particular, this now support non-standard interpolations which may not even have dofs in the nodes.

Fixes #167.


src/Dofs/DofHandler.jl
```diff
@@ -766,45 +766,72 @@ getfieldinterpolation(fh::FieldHandler, field_idx::Int) = fh.field_interpolation
...
+end
 
+# Internal method that have the vtk option to allocate the output differently
+function _evaluate_at_grid_nodes(dh::DofHandler, u::Vector{T}, fieldname::Symbol, ::Val{vtk}=Val(false)) where {T, vtk}
+    # Make sure the field exists
+    fieldname ∈ getfieldnames(dh) || error("Field $fieldname not found.")
+    # Figure out the return type (scalar or vector)
+    field_idx = find_field(dh, fieldname)
+    ip = getfieldinterpolation(dh, field_idx)
+    if vtk
```
KnutAM, line 789: Would it be good to call this `pad_zero` or something like that?

fredrikekre: Maybe, it doesn't just enable padeing, also makes the output a matrix, for example.

KnutAM: Yes, probably not the best naming suggestion either; As it is internal I guess it can be changed later if there are cases for which it makes sense (I was just a bit on the side that it would be nice not to have references to the vtk-format at different places in the code-base when a part-goal of #692 is to centralize that to one place)

I suppose it is too inefficient to first create the "standard" output and then reformat it for vtk?

lijas (commented):
Nice. Finally we can use super parametric elements :laughing:

src/Dofs/DofHandler.jl
```diff
@@ -766,45 +766,72 @@ getfieldinterpolation(fh::FieldHandler, field_idx::Int) = fh.field_interpolation
...
     for fh in dh.fieldhandlers
-        # check if this fh contains this field, otherwise continue to the next
+        # Check if this fh contains this field, otherwise continue to the next
         field_idx = _find_field(fh, fieldname)
         field_idx === nothing && continue
-        offset = field_offset(fh, field_idx)
-
-        reshape_field_data!(data, dh, u, offset, field_dim, BitSet(fh.cellset))
+        # Set up CellValues with the local node coords as quadrature points
+        CT = getcelltype(dh.grid, first(fh.cellset))
```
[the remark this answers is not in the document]
fredrikekre: Agreed

PR #706, opened 2023-05-11
Improve boundschecks during assemble

KnutAM:
I think this should be safer, but I haven't benchmarked if there is a performance regression.


fredrikekre (approved):
I suppose requiring the same length is enough if we instead do
```
for (i, j) in zip(keys(ge), keys(dofs))
    addindex!(g, ge[i], dofs[j])
end
```
but in practice you would never utilize this extra degree of freedom I don't think.

PR #707, opened 2023-05-11
Deprecate start_assemble(f, K)

KnutAM:
Seems unnecessary to allow both options for starting to assemble, of which only one is documented.


KnutAM (commented):
derp

PR #710, opened 2023-05-15
Cross-elements coupling for `DiscontinuousLagrange` sparsity patterns.

AbdAlazezAhmed:
It iterates over all cells twice, once to get the length of I,J and another to fill them. There is probably a better way to estimate their length.
Also, codegen here looks a bit complicated, I'm still learning how to use it.
I think explicit tests for sparsity patterns are missing (they're used inside other tests, but they themselves don't have a standalone test)

Manual tests:
- [x] Test with a single `DiscontinuousLagrange`
- [x] Test with multiple `DiscontinuousLagrange`
- [x] Test with older tests to make sure nothing breaks
- [x] Test with  `DiscontinuousLagrange` and  `Lagrange`
- [x] Test `coupling`

Tasks:
- [x] Write tests
- [x] Code works fine and doesn't break anything. I hope
- [x] Update docs

Things to check:
- [x] `push!` vs size precalculation
- [x] L2 trait

* I think I should stop removing the extra whitespaces :)


fredrikekre:
Nice. Didn't look at the details, but the code is surprisingly short.

> It iterates over all cells twice, once to get the length of I,J and another to fill them.

Probably you can just `push!` then instead? Would be interesting to compare the two approaches. `push!` is generally pretty fast, but if you push into an already large vector it might have to be reallocated, which is expensive. That is why we currently make an estimate.

src/Dofs/sparsity_pattern.jl
```diff
@@ -233,3 +287,41 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
...
+            $(pre_f)
+            element_dof_start = 0
+            cnt = 0
+            for (fhi, fh) in pairs(dh.fieldhandlers)
+                
+                for fi in fh.field_interpolations
+                    # if(!(typeof(fi)<:DiscontinuousLagrange))
+                    #     element_dof_start += getnbasefunctions(fi)
+                    #     continue
+                    # end
```
AbdAlazezAhmed, line 306: Commenting this makes mixed interpolation symmetric, at least for full coupling for now,
```julia
julia> grid = generate_grid(Line, (3,));

julia> topology = ExclusiveTopology(grid);

julia> ip = DiscontinuousLagrange{1, RefCube, 2}();

julia> ipc = Lagrange{1, RefCube, 1}();

julia> dh = DofHandler(grid);

julia> add!(dh, :u, 1,ip);

julia> add!(dh, :v, 1,ipc);

julia> close!(dh);;

julia> K = create_sparsity_pattern(dh, topology = topology, cross_element_full_coupling = true)
13×13 SparseMatrixCSC{Float64, Int64} with 137 stored entries:
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
  ⋅    ⋅    ⋅    ⋅   0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
  ⋅    ⋅    ⋅    ⋅   0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
  ⋅    ⋅    ⋅    ⋅   0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
  ⋅    ⋅    ⋅    ⋅   0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -233,3 +287,41 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
...
+)
+    @eval begin
+        function $(func)(dh::AbstractDofHandler, topology::ExclusiveTopology, full_cross_element::Bool; max_buffer_length::Int = 0)
+            $(pre_f)
+            element_dof_start = 0
+            cnt = 0
+            for (fhi, fh) in pairs(dh.fieldhandlers)
+                
+                for fi in fh.field_interpolations
+                    if(!(typeof(fi)<:DiscontinuousLagrange) && !full_cross_element)
```
termi-official, line 303: We might want to think about this type of branching further. I have thought about adding a traits system to the interpolations to mark them e.g. as "L2-conforming", so users can add new interpolations (with corresponding dispatches), add the trait and things will just work out of the box. What do you think? cc @fredrikekre

KristofferC: Yes, something more extensible that `typeof(fi)<:DiscontinuousLagrange` would be to prefer (like traits as already mentioned). Something similar to https://docs.julialang.org/en/v1/manual/interfaces/:

```
IsDisontinuous(::Type) = false
...
IsDisontinuous(::Type{<:DiscontinuousLagrange}) = true

if IsDisontinuous(typeof(fi)) ...
```

test/test_dofs.jl
```diff
@@ -504,3 +504,72 @@ end
...
+                            end
+                        end
+                    end
+                end
+            end
+        end
+        check_I = repeat(1:dh.ndofs[], dh.ndofs[])
+        check_J = repeat(1:dh.ndofs[], inner = dh.ndofs[])
+        @test is_stored.(Ref(K), check_I, check_J) == reshape(K_check,length(check_I))
+    end
```
AbdAlazezAhmed, line 550: This function tests for field coupling. I'll work on generalizing it for component coupling.

src/Dofs/sparsity_pattern.jl
```diff
@@ -309,6 +309,10 @@ for (func,                              pre_f,
                                 cell_dofs = celldofs(dh,cell_idx)[element_dof_start + 1 : element_dof_start + getnbasefunctions(fi)]
                                 neighbour_dof_start = 0
                                 for fi2 in fh.field_interpolations
+                                    if(!IsDiscontinuous(typeof(fi2)<: VectorizedInterpolation ? typeof(fi2.ip) : typeof(fi2)))
```
termi-official, line 312: I would like to propose to dispatch this in the interpolation api via something like `IsDiscontinuous(ip::VectorizedInterpolation) = IsDiscontinuous(ip.ip)`

src/Dofs/sparsity_pattern.jl
```diff
@@ -233,3 +283,56 @@ function _condense_sparsity_pattern!(K::SparseMatrixCSC{T}, dofcoefficients::Vec
...
+                            end
+                        end
+                    end
+                    element_dof_start += getnbasefunctions(fi)
+                end
+            end
+            return $(return_values)
+        end
+    end
+end
```
termi-official, line 338:
```suggestion
end

```

src/Dofs/sparsity_pattern.jl
AbdAlazezAhmed: Done, thanks!
It's now dispatched for both variables and types, we can remove one if it can cause confusion.

AbdAlazezAhmed:
checking performance: ~5K allocations for simple test.... I think I did something horrible :D

fredrikekre:
That usually indicates a type instability,  [Cthulhu](https://github.com/JuliaDebug/Cthulhu.jl) might be a helpful tool to investigate with.

AbdAlazezAhmed:
Thanks @fredrikekre !
It seems to be the case, many `::Any` fallbacks when using `::Interpolation`.
```julia
315                     fi::Interpolation = cell_field::Field.interpolation::Interpolation
316                     if(!(IsDiscontinuous(fi::Interpolation)::Bool)::Bool)
317                         (element_dof_start::Any += getnbasefunctions(fi::Interpolation)::Any)::Any
318                         continue
319                     end
```
I'm not sure why/how to fix it but I'll look into it.

fredrikekre:
Ah, that is expected since we store interpolations in an abstractly typed vector: https://github.com/Ferrite-FEM/Ferrite.jl/blob/20145f35b541fd23ea99567911a9ab537971b48b/src/Dofs/DofHandler.jl#L40

To get around this issue in Dof distribution we have the `InterpolationInfo` struct which precomputes the necessary information before entering the hot loop. Perhaps you can use that in this case too? (If you just add the `isdiscontinuous` boolean to this struct perhaps)

AbdAlazezAhmed:
Thanks, that was very helpful!
Now the issue is with iterating over `topology.face_neighbor` as it is abstractly typed `EntityNeighborhood`, and if I understand correctly buffering them like `nbasefunctions` would result in the same allocations problem as we're doing the same number of iterations but somewhere else.
Profiler output for 200*200 grid:
```julia
        - for (func,                              pre_f,                                                                                      inner_f,                                return_values) in (
        -     (:cross_element_coupling_count,     :(nothing),                                                                                 :(nothing),                             :(cnt)),
183398448     (:cross_element_coupling,           :(I = Vector{Int}(undef, max_buffer_length); J = Vector{Int}(undef, max_buffer_length)),    :(I[cnt] = dofi; J[cnt] = dofj;),       :(I, J)),
        - )
        -     @eval begin
        -         function $(func)(dh::AbstractDofHandler, topology::ExclusiveTopology, sym::Bool, keep_constrained::Bool, couplings::Union{AbstractVector{<:AbstractMatrix{Bool}},Nothing} ; max_buffer_length::Int = 0)
183398448             $(pre_f)
        -             cnt = 0
        0             for (fhi, fh) in pairs(dh.fieldhandlers)
      128                 ip_infos = InterpolationInfo[]
      128                 nbasefunctions = Int[]
        0                 for cell_field in fh.fields
      480                     ip_info = InterpolationInfo(cell_field.interpolation)
     1088                     push!(ip_infos, ip_info)
        0                     push!(nbasefunctions, getnbasefunctions(cell_field.interpolation))
        0                 end
        -                 element_dof_start = 0
        0                 isnothing(couplings) || (coupling_fh = couplings[fhi])
        0                 for (cell_field_i, cell_field) in enumerate(fh.fields)
        0                     fii = ip_infos[cell_field_i]
        0                     if(!fii.is_discontinuous)
        0                         element_dof_start += nbasefunctions[cell_field_i]
        -                         continue
        -                     end
        0                     for cell_idx in eachindex(getcells(dh.grid))
        0                         cell_field ∈ dh.fieldhandlers[dh.cell_to_fieldhandler[cell_idx]].fields || continue
        0                         current_face_neighborhood = getdim(dh.grid.cells[cell_idx]) >1 ? topology.face_neighbor[cell_idx,:] : topology.vertex_neighbor[cell_idx,:]
        0                         shared_faces_idx = findall(!isempty,current_face_neighborhood)
        0                         for face_idx in shared_faces_idx
 40755200                             for neighbor_face in current_face_neighborhood[face_idx]
        0                                 cell_dofs = @view celldofs(dh,cell_idx)[element_dof_start + 1 : element_dof_start + nbasefunctions[cell_field_i]]
        -                                 neighbour_dof_start = 0
        0                                 for (neighbor_field_i, neighbor_field) in enumerate(fh.fields)
        0                                     fii2 = ip_infos[neighbor_field_i]
        0                                     neighbor_field ∈ dh.fieldhandlers[dh.cell_to_fieldhandler[neighbor_face[1]]].fields || continue
        0                                     if(!fii2.is_discontinuous)
        0                                         neighbour_dof_start += nbasefunctions[neighbor_field_i]
        -                                         continue
        -                                     end
        0                                     neighbour_dofs = @view celldofs(dh,neighbor_face[1])[neighbour_dof_start + 1 : neighbour_dof_start + nbasefunctions[neighbor_field_i]]
        -                                     # neighbour_unique_dofs = neighbour_dofs[.!(neighbour_dofs .∈ Ref(celldofs(dh,cell_idx)))]
        0                                     for j in eachindex(neighbour_dofs), i in eachindex(cell_dofs)
        0                                         isnothing(couplings) || coupling_fh[i+element_dof_start,j+neighbour_dof_start] || continue
        0                                         dofi = cell_dofs[i]
        0                                         dofj = neighbour_dofs[j]
        0                                         sym && (dofi > dofj && continue)
        0                                         !keep_constrained && (haskey(ch.dofmapping, dofi) || haskey(ch.dofmapping, dofj)) && continue
        0                                         cnt += 1
        0                                         $(inner_f)
        0                                     end
        0                                     neighbour_dof_start += nbasefunctions[neighbor_field_i]
        0                                 end
        0                             end
        0                         end
        0                     end
        0                     element_dof_start += nbasefunctions[cell_field_i]
        0                 end
        0             end
        0             return $(return_values)
        -         end
        -     end
        - end
```
I thought of using `topology.cell_neighbor` but this would make 1 and 4 neighbors and I need them to have shared faces.
```
+--------+--------+
|        |        |
|   3    |   4    |
|        |        |
+--------+--------+
|        |        |
|   1    |   2    |
|        |        |
+--------+--------+
```

koehlerson:
Then the topology should be parametrized in the type of the matrices to avoid it. I think that's not too hard to do. I can take care of that but I have to think about how to change the `EntityNeighborhood` thing in a nice way

AbdAlazezAhmed:
I tested some things * not a solution yet * that decreased number of allocations to the point at which it's almost independent of grid size.

- Abuse `@view` wherever there is copy allocation in loops
- `collect` `topology.face_neighbor` outside loops
- `face_neighbor::SparseMatrixCSC{EntityNeighborhood{FaceIndex},Int}`, which is wrong because it can have `EdgeIndex`
Some results:
```julia
julia> using Ferrite, BenchmarkTools
[ Info: Precompiling Ferrite [c061ca5d-56c9-439f-9c0e-210fe06d3992]

julia> coupling =  [
                   true    true    true    false
                   true    true    true    true
                   true    true    true    true
                   false    true    true    true
               ];

julia>     grid = generate_grid(Quadrilateral, (200, 200));

julia>     topology = ExclusiveTopology(grid);

julia>     dh = DofHandler(grid);

julia>     add!(dh, :u, DiscontinuousLagrange{RefQuadrilateral,1}()^2);

julia>     add!(dh, :p, DiscontinuousLagrange{RefQuadrilateral,1}());

julia>     add!(dh, :w, Lagrange{RefQuadrilateral,1}());

julia>     close!(dh);

julia> 

julia> @btime create_sparsity_pattern(dh; coupling=coupling, topology = topology)
  983.640 ms (84 allocations: 2.03 GiB)
520401×520401 SparseArrays.SparseMatrixCSC{Float64, Int64} with 31606001 stored entries:
⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦

julia>     dh = DofHandler(grid);

julia>     add!(dh, :u, Lagrange{RefQuadrilateral,1}()^2);

julia>     add!(dh, :p, Lagrange{RefQuadrilateral,1}());

julia>     add!(dh, :w, Lagrange{RefQuadrilateral,1}());

julia>     close!(dh);

julia> @btime create_sparsity_pattern(dh; coupling=coupling, topology = topology)
  129.338 ms (42 allocations: 428.83 MiB)
161604×161604 SparseArrays.SparseMatrixCSC{Float64, Int64} with 5056814 stored entries:
⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦
 
```
Turned out there were some allocations that didn't appear in profiler output.
```julia
for (func,                              pre_f,                                                                                      inner_f,                                return_values) in (
    (:cross_element_coupling_count,     :(nothing),                                                                                 :(nothing),                             :(cnt)),
    (:cross_element_coupling,           :(I = Vector{Int}(undef, max_buffer_length); J = Vector{Int}(undef, max_buffer_length)),    :(I[cnt] = dofi; J[cnt] = dofj;),       :(I, J)),
)
    @eval begin
        function $(func)(dh::AbstractDofHandler, topology::ExclusiveTopology, sym::Bool, keep_constrained::Bool, couplings::Union{AbstractVector{<:AbstractMatrix{Bool}},Nothing} ; max_buffer_length::Int = 0)
            $(pre_f)
            cnt = 0
            tp = collect(topology.face_neighbor)
            for (fhi, fh) in pairs(dh.fieldhandlers)
                ip_infos = InterpolationInfo[]
                nbasefunctions = Int[]
                for cell_field in fh.fields
                    ip_info = InterpolationInfo(cell_field.interpolation)
                    push!(ip_infos, ip_info)
                    push!(nbasefunctions, getnbasefunctions(cell_field.interpolation))
                end
                element_dof_start = 0
                isnothing(couplings) || (coupling_fh = couplings[fhi])
                for (cell_field_i, cell_field) in enumerate(fh.fields)
                    fii = ip_infos[cell_field_i]
                    if(!fii.is_discontinuous)
                        element_dof_start += nbasefunctions[cell_field_i]
                        continue
                    end
                    for cell_idx in eachindex(getcells(dh.grid))
                        cell_field ∈ dh.fieldhandlers[dh.cell_to_fieldhandler[cell_idx]].fields || continue
                        # current_face_neighborhood = getdim(dh.grid.cells[cell_idx]) >1 ? (@view topology.face_neighbor[cell_idx,:]) : (@view topology.vertex_neighbor[cell_idx,:])
                        cell_dofs = @view dh.cell_dofs[dh.cell_dofs_offset[cell_idx] : dh.cell_dofs_offset[cell_idx] + ndofs_per_cell(dh, cell_idx) - 1]
                        cell_field_dofs = @view cell_dofs[element_dof_start + 1 : element_dof_start + nbasefunctions[cell_field_i]]
                        for face_idx in @view tp[cell_idx,:]
                            isempty(face_idx) && continue
                            for neighbor_face in face_idx
                                neighbour_dof_start = 0
                                for (neighbor_field_i, neighbor_field) in enumerate(fh.fields)
                                    fii2 = ip_infos[neighbor_field_i]
                                    neighbor_field ∈ dh.fieldhandlers[dh.cell_to_fieldhandler[neighbor_face[1]]].fields || continue
                                    if(!fii2.is_discontinuous)
                                        neighbour_dof_start += nbasefunctions[neighbor_field_i]
                                        continue
                                    end
                                    # @info @allocated  neighbour_dofs = @view celldofs(dh,neighbor_face[1])[neighbour_dof_start + 1 : neighbour_dof_start + nbasefunctions[neighbor_field_i]]
                                    neighbor_dofs = @view dh.cell_dofs[dh.cell_dofs_offset[neighbor_face[1]] : dh.cell_dofs_offset[neighbor_face[1]] + ndofs_per_cell(dh, neighbor_face[1]) - 1]
                                    neighbor_field_dofs = @view neighbor_dofs[neighbour_dof_start + 1 : neighbour_dof_start + nbasefunctions[neighbor_field_i]]
                                    # neighbour_unique_dofs = neighbour_dofs[.!(neighbour_dofs .∈ Ref(celldofs(dh,cell_idx)))]
                                    for j in eachindex(neighbor_field_dofs), i in eachindex(cell_field_dofs)
                                        isnothing(couplings) || coupling_fh[i+element_dof_start,j+neighbour_dof_start] || continue
                                        dofi = cell_field_dofs[i]
                                        dofj = neighbor_field_dofs[j]
                                        sym && (dofi > dofj && continue)
                                        !keep_constrained && (haskey(ch.dofmapping, dofi) || haskey(ch.dofmapping, dofj)) && continue
                                        cnt += 1
                                        $(inner_f)
                                    end
                                    neighbour_dof_start += nbasefunctions[neighbor_field_i]
                                end
                            end
                        end
                    end
                    element_dof_start += nbasefunctions[cell_field_i]
                end
            end
            return $(return_values)
        end
    end
end
```
I think using `push!()` would be better than precalculating size at this point.
I'll clean the code/use better names once the performance issues are solved.
Thanks!

AbdAlazezAhmed:
Adding (`cell_face_neighbor::Vector{EntityNeighborhood{CellIndex}}`: cellid to all cells connected by face ) to topology and using it instead can be a solution ~ at least for now ~
Now the same example in the last comment takes about 80 allocations.

AbdAlazezAhmed:
Might revert last commit if the loss in speed outweighs the extra few allocations (probably it is).
I think the PR can move from draft now. Feel free to review it whenever you have time, I'm not in a hurry :D
Thanks all for your help!

src/interpolations.jl
termi-official, line 1: I think we can roll back this file for now.

`get_continuous_interpolation` is the piece for the other PR, right? Then make sure to cherry pick this into a separate PR.

termi-official (changes requested):
Design looks quite good already. However, we have to think about the user facing side again. I have left some comments with a proposal to add a way which does not add artificial overhead for the case when we do not need to couple L2 functions. Think e.g. about FSI. Not super confident yet tho.

We may want to discuss what exactly the feature is.

src/Dofs/sparsity_pattern.jl
```diff
@@ -161,7 +161,7 @@ function cross_element_coupling!(dh::AbstractDofHandler, topology::ExclusiveTopo
                 cell_field ∈ dh.fieldhandlers[dh.cell_to_fieldhandler[cell_idx]].fields || continue
                 cell_dofs = @view dh.cell_dofs[dh.cell_dofs_offset[cell_idx] : dh.cell_dofs_offset[cell_idx] + ndofs_per_cell(dh, cell_idx) - 1]
                 cell_field_dofs = @view cell_dofs[element_dof_start + 1 : element_dof_start + nbasefunctions[cell_field_i]]
-                for neighbor_cell in topology.cell_face_neighbor[cell_idx]
+                for neighbor_cell in (getdim(dh.grid.cells[cell_idx]) >1 ? topology.cell_face_neighbor[cell_idx] : topology.cell_neighbor[cell_idx])
```
termi-official, line 164: Also marking this line as a reminder to check if this is a proper fix for https://github.com/Ferrite-FEM/Ferrite.jl/issues/728

src/Dofs/sparsity_pattern.jl
```diff
@@ -147,7 +243,7 @@ function _create_sparsity_pattern(dh::AbstractDofHandler, ch#=::Union{Constraint
             end
         end
     end
-
+    has_discontinuous_ip && (cnt = cross_element_coupling!(dh,topology,sym, keep_constrained, couplings, cnt, I, J))
```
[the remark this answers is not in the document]
fredrikekre: I was also going to comment about this. Perhaps, if you request cross element couplings, a reasonable default is to couple discontinuous elements?

fredrikekre:
In https://github.com/Ferrite-FEM/Ferrite.jl/tree/fe/pattern-by-local-condense I played around with a slightly different approach to creating the sparsity pattern by essentially simulating the assembly procedure (while using local constraint condensation to automatically insert entries for affine constraints without having to build the matrix first). That approach might be useful here too later on when there are a nice way to iterate also over the interfaces. Not suggesting this should be done in this PR, just as a note.

src/interpolations.jl
AbdAlazezAhmed: Indeed, but `IsDiscontinuous` is still needed here

src/Dofs/sparsity_pattern.jl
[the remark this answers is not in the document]
AbdAlazezAhmed: I did some edits for this, now there are 2 coupling matrices. Elements coupling defaults to no coupling while same-cell coupling defaults to full coupling as it used to. Elements coupling also requires interpolation to be discontinuous to be coupled.
I'll look into the resulting matrices to make sure I haven't messed something. Thanks!

test/test_dofs.jl
```diff
@@ -504,3 +504,126 @@ end
...
```
AbdAlazezAhmed, line 589: I'm not sure but the test looks like it's just repeating the implementation in a less optimized way. I'm thinking of a way to automate it.

termi-official: Indeed. Unit tests should never mirror internals, but check for simple invariants.

test/test_dofs.jl
```diff
@@ -504,3 +504,154 @@ end
...
+                    for i_idx in i_dofs, j_idx in j_dofs
+                        i = dh.cell_dofs[i_idx + dh.cell_dofs_offset[cell_idx] - 1]
+                        j = dh.cell_dofs[j_idx + dh.cell_dofs_offset[neighbor_idx.idx] - 1]
+                        !is_discontinuous && continue
+                        @test is_stored(K, i, j) == (elements_coupling[index])
+                    end
+                end
+            end
+        end
+    end
```
AbdAlazezAhmed, line 589: I think this now acts like the other tests but somewhat automated * and hopefully doesn't mirror internals *

test/test_dofs.jl
```diff
@@ -504,3 +504,154 @@ end
...
+    u2dofs = udofs[2:2:end]
+    pdofs = dof_range(dh, :p)
+    wdofs = dof_range(dh, :w)
+    dofs4 = [   (u1dofs, u1dofs, :u, :u) (u1dofs, u2dofs, :u, :u) (u1dofs, pdofs, :u, :p) (u1dofs, wdofs, :u, :w)
+                (u2dofs, u1dofs, :u, :u) (u2dofs, u2dofs, :u, :u) (u2dofs, pdofs, :u, :p) (u2dofs, wdofs, :u, :w)
+                (pdofs, u1dofs, :p, :u) (pdofs, u2dofs, :p, :u) (pdofs, pdofs, :p, :p) (pdofs, wdofs, :p, :w)
+                (wdofs, u1dofs, :w, :u) (wdofs, u2dofs, :w, :u) (wdofs, pdofs, :w, :p) (wdofs, wdofs, :w, :w)]
+    dofs3 = [   (udofs, udofs, :u, :u)  (udofs, pdofs, :u, :p) (udofs, wdofs, :u, :w)
+                (pdofs, udofs, :p, :u)  (pdofs, pdofs, :p, :p) (pdofs, wdofs, :p, :w)
+                (wdofs, udofs, :w, :u)  (wdofs, pdofs, :w, :p) (wdofs, wdofs, :w, :w)]
```
AbdAlazezAhmed, line 609: I'm testing something that'll probably eliminate the need of these matrices.

test/test_dofs.jl
termi-official: Why not use `celldofs` instead of `dh.cell_dofs`?

AbdAlazezAhmed: Missed it, will change it with next push. 😄

src/Dofs/sparsity_pattern.jl
```diff
@@ -147,7 +209,7 @@ function _create_sparsity_pattern(dh::AbstractDofHandler, ch#=::Union{Constraint
             end
         end
     end
-
+    has_discontinuous_ip && !isnothing(elements_coupling) && any(elements_coupling, dims=[1,2])[] && (cnt = cross_element_coupling!(dh,topology,sym, keep_constrained, elements_couplings, cnt, I, J))
```
fredrikekre, line 212:
```suggestion
    # Add cross element couplings if there are discontinuous interpolations
    if has_discontinuous_ip && !isnothing(elements_coupling) && any(elements_coupling)
        cnt = cross_element_coupling!(dh,topology,sym, keep_constrained, elements_couplings, cnt, I, J)
    end
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -96,21 +100,78 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+        # Buffering interpolation information for type stability/better allocations
+        ip_infos = InterpolationInfo[]
+        nbasefunctions = Int[]
+        for ip in sdh.field_interpolations
+            ip_info = InterpolationInfo(ip)
+            push!(ip_infos, ip_info)
+            push!(nbasefunctions, getnbasefunctions(ip))
+        end
+        isnothing(couplings) || (coupling_sdh = couplings[sdh_idx])
+        for (cell_field_i, cell_field) in enumerate(sdh.field_names)
```
fredrikekre, line 122:
```suggestion
        for (cell_field_i, cell_field) in pairs(sdh.field_names)
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -96,21 +100,78 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+            fii = ip_infos[cell_field_i]
+            # Couple element for discontinuous interpolations only
+            fii.is_discontinuous || continue
+            for cell_idx in eachindex(getcells(dh.grid))
+                # Making sure current field is defined for current cell
+                cell_field ∈ dh.subdofhandlers[dh.cell_to_subdofhandler[cell_idx]].field_names || continue
```
fredrikekre, line 128: You can/should loop directly over the cellset associated with this subdofhandler:
```suggestion
            for cell_idx in BitSet(sdh.cellset)
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -96,21 +100,78 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+                dofrange1 = dof_range(sdh, cell_field)
+                cell_dofs = celldofsview(dh, cell_idx)
+                cell_field_dofs = @view cell_dofs[dofrange1]
+                # For 1D case, cells must share faces to be neighbors. Otherwise use cell_face_neighbor for higher dimensions
+                for neighbor_cell in (getdim(dh.grid.cells[cell_idx]) >1 ? topology.cell_face_neighbor[cell_idx] : topology.cell_neighbor[cell_idx])
+                    for (neighbor_field_i, neighbor_field) in enumerate(sdh.field_names)
```
fredrikekre, line 134:
```suggestion
                    for (neighbor_field_i, neighbor_field) in pairs(sdh.field_names)
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -96,21 +100,78 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+        for (cell_field_i, cell_field) in enumerate(sdh.field_names)
+            fii = ip_infos[cell_field_i]
+            # Couple element for discontinuous interpolations only
+            fii.is_discontinuous || continue
+            for cell_idx in eachindex(getcells(dh.grid))
+                # Making sure current field is defined for current cell
+                cell_field ∈ dh.subdofhandlers[dh.cell_to_subdofhandler[cell_idx]].field_names || continue
+                dofrange1 = dof_range(sdh, cell_field)
+                cell_dofs = celldofsview(dh, cell_idx)
+                cell_field_dofs = @view cell_dofs[dofrange1]
```
fredrikekre, line 131: You can probably use CellCaches here (instead of the view):
```
cc1 = CellCache(dh)
cc2 = CellCache(dh)

reinit!(cc1, cell_idx)
reinit!(cc2, neighbor_cell)
```
etc

src/Dofs/sparsity_pattern.jl
```diff
@@ -96,21 +100,78 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+                # For 1D case, cells must share faces to be neighbors. Otherwise use cell_face_neighbor for higher dimensions
+                for neighbor_cell in (getdim(dh.grid.cells[cell_idx]) >1 ? topology.cell_face_neighbor[cell_idx] : topology.cell_neighbor[cell_idx])
+                    for (neighbor_field_i, neighbor_field) in enumerate(sdh.field_names)
+                        fii2 = ip_infos[neighbor_field_i]
+                        sdh2 = dh.subdofhandlers[dh.cell_to_subdofhandler[neighbor_cell.idx]]
+                        neighbor_field ∈ sdh2.field_names && fii2.is_discontinuous || continue
```
fredrikekre, line 137: So just to be clear, coupling is added only if the neighbor interpolation is also discontinuous? I think maybe we even warn in DofHandler construction now if all interpolations for a field don't belong to the same "family".

src/exports.jl
```diff
@@ -111,6 +111,7 @@ export
     ndofs_per_cell,
     celldofs!,
     celldofs,
+    celldofsview,
```
fredrikekre, line 114: Don't think this needs to be exported, and I don't think you need it internally either, use CellCache (possibly with more fine grained update flags to only update the dofs).

src/interpolations.jl
```diff
@@ -363,6 +367,15 @@ boundarydof_indices(::Type{FaceIndex}) = Ferrite.facedof_indices
 boundarydof_indices(::Type{EdgeIndex}) = Ferrite.edgedof_indices
 boundarydof_indices(::Type{VertexIndex}) = Ferrite.vertexdof_indices
 
+"""
+    is_discontinuous(::Interpolation)
+    is_discontinuous(::Type{<:Interpolation})
+
+Checks whether the interpolation is discontinuous (i.e. `DiscontinuousLagrange`)
+"""
+is_discontinuous(::Interpolation) = false
```
fredrikekre, line 376: Can canonicalize the answers like so:
```suggestion
is_discontinuous(ip::Interpolation) = is_discontinuous(typeof(ip))
```

src/interpolations.jl
```diff
@@ -429,6 +442,9 @@ function shape_value(ip::DiscontinuousLagrange{shape, 0}, ::Vec{dim, T}, i::Int)
     return one(T)
 end
 
+is_discontinuous(::Type{<:DiscontinuousLagrange}) = true
+is_discontinuous(::DiscontinuousLagrange) = true
```
fredrikekre, line 446:
```suggestion
```

src/interpolations.jl
```diff
@@ -1369,3 +1385,6 @@ function shape_gradient_and_value(ipv::VectorizedInterpolation{vdim, shape}, ξ:
 end
 
 reference_coordinates(ip::VectorizedInterpolation) = reference_coordinates(ip.ip)
+
+is_discontinuous(ipv::VectorizedInterpolation) = is_discontinuous(ipv.ip)
```
fredrikekre, line 1389:
```suggestion
```

src/interpolations.jl
```diff
@@ -1369,3 +1385,6 @@ function shape_gradient_and_value(ipv::VectorizedInterpolation{vdim, shape}, ξ:
...
+is_discontinuous(::Type{<:VectorizedInterpolation{vdim, refshape, order, ip}}) where {vdim, refshape, order, ip<:DiscontinuousLagrange}= is_discontinuous(ip)
```
fredrikekre, line 1390:
```suggestion
is_discontinuous(::Type{<:VectorizedInterpolation{<:Any, <:Any, <:Any, ip}}) where {ip} = is_discontinuous(ip)
```

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,55 +1,59 @@
 """
-    create_sparsity_pattern(dh::DofHandler; coupling)
+    create_sparsity_pattern(dh::DofHandler; coupling, topology::Union{Nothing, AbstractTopology}, elements_coupling)
```
fredrikekre, line 2: Can you clarify how one opt-in to getting cross element couplings now. Do you need to pass `elements_coupling` *and* topology?

fredrikekre (commented):
Some comments.

src/Dofs/sparsity_pattern.jl
AbdAlazezAhmed: You mean I should implement this : https://github.com/Ferrite-FEM/Ferrite.jl/blob/ccd0e71570993ee91c3adaa4ab97569f21a476e1/src/Dofs/DofHandler.jl#L231 ? If so, check for same interpolation or check for same family?

fredrikekre: No you don't have to do that, not in this PR. I just meant that if we add that check then this should be fine, because then `:u` must always be discontinuous

fredrikekre: (I think the interpolation type (Lagrange, DiscontinuousLagrange, ...), and the order, should be checked, but refshape might differ).

src/Dofs/sparsity_pattern.jl
AbdAlazezAhmed: I think we can do it but not sure if should (would require more changes for coupling different fields on different subdofhandlers), the idea here is iterating over all cells that have a field.
For example, the test:
```julia
    dh = DofHandler(grid)
    sdh1 = SubDofHandler(dh, Set(1))
    add!(sdh1, :u, DiscontinuousLagrange{RefQuadrilateral,1}()^2)
    add!(sdh1, :y, DiscontinuousLagrange{RefQuadrilateral,1}())
    add!(sdh1, :p, Lagrange{RefQuadrilateral,1}())
    sdh2 = SubDofHandler(dh, Set(2))
    add!(sdh2, :u, DiscontinuousLagrange{RefQuadrilateral,1}()^2)
    close!(dh)
```
 Would fail because the second subdofhandler doesn't have `:y` field, thus it won't couple `:u` with `:y` for second cell dofs
 would result in:
 ```julia
 ⎡⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎤
⎢⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎥
⎢⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎥
⎢⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⎥
⎢⣿⣿⣿⣿⠀⠀⠀⠀⣿⣿⣿⣿⎥
⎣⣿⣿⣿⣿⠀⠀⠀⠀⣿⣿⣿⣿⎦
```
while the current implementation results in:
```julia
⎡⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎤
⎢⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎥
⎢⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⎥
⎢⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⎥
⎢⣿⣿⣿⣿⣿⣿⠀⠀⣿⣿⣿⣿⎥
⎣⣿⣿⣿⣿⣿⣿⠀⠀⣿⣿⣿⣿⎦
```
Notice some couplings are missing.
Please correct me if I'm wrong/they shouldn't be coupled.

termi-official:
Can you test and clarify the case when we use discontinuous interpolations with shared dofs? (e.g. Crouzeix-Raviart)

src/Dofs/sparsity_pattern.jl
fredrikekre: I am not sure I follow, sorry, can you post the full code for producing those two patterns so I can play around a bit? Is the problem that you only add the dofs in one "direction" when looping over the cells?

AbdAlazezAhmed: Yes, the problem is that the code added dofs in one direction, so if one cell has more fields than its neighbor they will be coupled in one direction only. This was mitigated by:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/bb16acbee3f77d283a5864788dc54cdf75c4b863/src/Dofs/sparsity_pattern.jl#L110-L122

https://github.com/Ferrite-FEM/Ferrite.jl/blob/bb16acbee3f77d283a5864788dc54cdf75c4b863/src/Dofs/sparsity_pattern.jl#L161-L164

In the last commits, so we add once in each direction.
For the full code IIRC:
```julia
grid = generate_grid(Quadrilateral, (2, 1))
topology = ExclusiveTopology(grid)

dh = DofHandler(grid)
sdh1 = SubDofHandler(dh, Set(1))
add!(sdh1, :u, DiscontinuousLagrange{RefQuadrilateral,1}()^2)
add!(sdh1, :y, DiscontinuousLagrange{RefQuadrilateral,1}())
add!(sdh1, :p, Lagrange{RefQuadrilateral,1}())
sdh2 = SubDofHandler(dh, Set(2))
add!(sdh2, :u, DiscontinuousLagrange{RefQuadrilateral,1}()^2)
close!(dh)

create_sparsity_pattern(dh; topology = topology, elements_coupling = trues(3,3))
```

fredrikekre: Okay, so right now we add in both directions for every cell then?

AbdAlazezAhmed: indeed

PR #711, opened 2023-05-17
Use the new reference shapes for `Interpolation`s and `QuadratureRule`

fredrikekre:
This patch replaces `RefCube{1, 2, 3}` with `RefLine`, `RefQuadrilateral`, and `RefHexahedron`; and `RefTetrahedron{2}` with `RefTriangle`.

Since the reference dimension is now encoded in the shape, this patch also introduces shorthands for constructing interpolations, i.e. `Lagrange{RefTriangle, order}()` as a shorthand for `Lagrange{2, RefTriangle, order}()`.

The same is done for `QuadratureRule`, for constructing cell quadrature rules. A face quadrature rule is still written as `QuadratureRule{D-1, shape}(...)`, where `shape` is the reference shape of the *cell*, i.e. face quadrature for the hexahedron is constructed as `QuadratureRule{2, RefHexahedron}()`. The new reference shapes make this non-ambiguous since we can now directly that a face quadrature is constructed.


fredrikekre:
I haven't updated the tests yet, want to see that the deprecation warnings capture everything.

PR #713, opened 2023-05-17
replaced str with str_scalar/str_vector for CellValues warning

JanM12:
I think the warning that appears when one tries to use `CellScalarValues` has a mistake in it. For my understanding the marked CellScalarValues needs to be replaced by a CellVectorValues to make sense.

grafik


termi-official (approved):

KnutAM:
Nicely caught @JanM12

PR #714, opened 2023-05-18
Hex27 VTK export

termi-official:
Upstream of https://github.com/Ferrite-FEM/FerriteGmsh.jl/pull/29

### DIFF
--- a/src/Export/VTK.jl
+++ b/src/Export/VTK.jl
@@ -9,11 +9,41 @@ cell_to_vtkcell(::Type{SerendipityQuadraticQuadrilateral}) = VTKCellTypes.VTK_QU
 
 cell_to_vtkcell(::Type{Hexahedron}) = VTKCellTypes.VTK_HEXAHEDRON
 cell_to_vtkcell(::Type{SerendipityQuadraticHexahedron}) = VTKCellTypes.VTK_QUADRATIC_HEXAHEDRON
+cell_to_vtkcell(::Type{QuadraticHexahedron}) = VTKCellTypes.VTK_TRIQUADRATIC_HEXAHEDRON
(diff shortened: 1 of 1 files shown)


src/Export/VTK.jl
```diff
@@ -9,11 +9,41 @@ cell_to_vtkcell(::Type{SerendipityQuadraticQuadrilateral}) = VTKCellTypes.VTK_QU
...
+    cell.nodes[17],
+    cell.nodes[18],
+    cell.nodes[19],
+    cell.nodes[20],
+    cell.nodes[25], # faces
+    cell.nodes[23],
+    cell.nodes[22],
+    cell.nodes[24],
+    cell.nodes[21],
+    cell.nodes[26],
```
KnutAM, line 44: Looks strange that it is different from the vtk order, considering that it says in the source code that Ferrite's cell is based on the VTK cell:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/18437f2d72ab33499773a2f865ebadd56a430da9/src/interpolations.jl#LL754C43-L754C131
(But as far as I can see, the PR is correct wrt. the reference coordinates, but in that case, probably the reference in the link should be removed as well). It might also change with #689 later though.

termi-official: I think we either changed the order at some point or we messed up in the first commit already :D

termi-official: Could find https://github.com/Ferrite-FEM/Ferrite.jl/pull/481/files#diff-3e76902e9a74000d5b26953f9acefca30977e4844e04df825a59ebe4ab77469e which seems to be the reason why we have to permutate here, but I think I do not understand what the exact motivation of these changes has been cc @fredrikekre

PR #715, opened 2023-05-19
Embedded elements showcase: Gray-Scott model on a sphere

termi-official:
This needs quite a bit of polishing, but it is at least a functional example.

reactive-surface

## TODO

* [ ] ~~Parameter passing for diffusivity~~
* [x] Documentation
* [ ] ~~Faster solver~~
* [x] Cleaner assembly code
* [x] ~~Update FerriteGmsh to allow parsing of embedded meshes~~ Use FerriteGmsh to generate the mesh on the fly


termi-official:
Little hack to get the mesh

```
# TODO replace with actual mesh generator
# try
# togrid("sphere.msh")
# catch
# end
# nodes = tonodes();
# elements, gmsh_elementidx = toelements(2);
```

termi-official:
If we want to include this in the next release, then this is as good as it might get, because we need the next release first to make the FerriteGmsh update (due to the breaking changes).

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,239 @@
...
+#
+# This tutorial gives a quick tutorial on how to assemble and solve time-dependent problems
+# on embedded surfaces.
+#
+# For this showcase we use the well known Gray-Scott model, which is a well-known reaction-diffusion
+# system to study pattern formation. The strong form is given by
+#
+# ```math
+#  \partial_t r_1 = \nabla \cdot (D_1 \nabla r_1) - r₁*r₂^2 + F *(1 - r₁) \quad \textbf{x} \in \Omega,
+#  \partial_t r_2 = \nabla \cdot (D_2 \nabla r_2) + r₁*r₂^2 - r₂*(F + k ) \quad \textbf{x} \in \Omega,
```
fredrikekre, line 23: I think you need `\begin{aligned}` or something around these, see https://ferrite-fem.github.io/Ferrite.jl/previews/PR715/tutorials/reactive_surface/#Introduction

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,239 @@
...
+# ```
+#
+# where $r_1$ and $r_2 are the reaction fields, $D_1$ and $D_2$ the diffusion tensors,
```
fredrikekre, line 26: See https://ferrite-fem.github.io/Ferrite.jl/previews/PR715/tutorials/reactive_surface/#Introduction
```suggestion
# where $r_1$ and $r_2$ are the reaction fields, $D_1$ and $D_2$ the diffusion tensors,
```

termi-official:
Ready again @fredrikekre

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
+# Putting this flag to false reproduces the figure shown in the example #src
+# We check for laminar flow development in the CI                       #src
+if @isdefined is_ci    #hide
+    IS_CI = is_ci      #hide
+else                   #hide
+    IS_CI = false      #hide
+end                    #hide
```
fredrikekre, line 7: Does this work? I think the flag is set in `Main` (from `make.jl`) but this is always evaluated in a separate module I think?

termi-official: Thanks for the pointer. Now this makes sense. And you are right, this is not working at all.

fredrikekre: I guess `isdefined(Main, :is_ci)` would work.
```suggestion
if isdefined(Main, :is_ci) #hide
    IS_CI = Main.is_ci     #hide
else                       #hide
    IS_CI = false          #hide
end                        #hide
```

termi-official: Or should we just do
```suggestion
const IS_CI = haskey(ENV, "GITHUB_ACTIONS")      #hide
```
I am fine with either.

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
+# Putting this flag to false reproduces the figure shown in the example #src
+# We check for laminar flow development in the CI                       #src
+if isdefined(Main, :is_ci) #hide
+    IS_CI = Main.is_ci     #hide
+else                       #hide
+    IS_CI = false          #hide
+end                        #hide
+# # [Reactive Surface](@id tutorial-reactive-surface)
+#
+# ![](reactive_surface.gif)
```
fredrikekre, line 10: This file is missing. Is it large? In that case we can put it on the gh-pages branch directly like we do with some other files and then download during the build.

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+# ```
+#
+# where $r_1$ and $r_2$ are the reaction fields, $D_1$ and $D_2$ the diffusion tensors,
+# $k$ is the conversion rate, $F$ is the feed rate and $\Omega$ the domain. Depending on the choice of
+# parameters a different pattern can be observed. Please also note that the domain does not have a
+# boundary. The corresponding weak form can be derived as usual.
+#
+# For simplicity we will solve the problem with the Lie-Troter-Godunov operator splitting technique with
+# the classical reaction-diffusion split. In this method we split our problem in two problems, i.e. a heat
+# problem and a pointwise reaction problem, and solve them alternatingly to advance in time.
```
fredrikekre, line 42: Since this is one of the "highlights" of this tutorial as advertised by the snipped in `tutorial/index.md` I think it would be nice to expand this a bit with some equations (e.g. the weak form of the heat equation and the strong form of the reaction problem.

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+    close!(dh);
+
+    M = create_sparsity_pattern(dh; coupling=[true false;false true])
+    D = create_sparsity_pattern(dh; coupling=[true false;false true])
+    assemble_matrices!(M, D, cellvalues, dh);
+
+    ## Since the heat problem is linear and has no time dependent parameters, we precompute the
+    ## decomposition of the system matrix to speed up the linear system solver.
+    A = M + Δt .* D
+    Alu = cholesky(A)
```
fredrikekre, line 202:
```suggestion
    cholA = cholesky(A)
```

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+    ## And prepare output for visualization.
+    pvd = VTKFileCollection("reactive-surface.pvd", dh);
+    addstep!(pvd, 0.0) do io
+        write_solution(io, dh, uₜ₋₁)
+    end
+
+    ## This is now the main solve loop.
+    for (iₜ, t) ∈ enumerate(Δt:Δt:T)
+        ## First we solve the heat problem
+        uₜ .= Alu \ (M * uₜ₋₁)
```
fredrikekre, line 218:
```suggestion
        uₜ .= cholA \ (M * uₜ₋₁)
```

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+
+        ## Then we solve the point-wise reaction problem with the solution of
+        ## the heat problem as initial guess.
+        rvₜ = reshape(uₜ, (2, length(grid.nodes)))
+        for i ∈ 1:length(grid.nodes)
```
fredrikekre, line 223: This somehow feels a bit iffy... but maybe we don't have a better way to do this now. Would
```suggestion
        rvₜ = reshape(uₜ, (2, ndofs(dh) ÷ 2))
        for i ∈ axes(rvₜ, 2)
```
be equivalent?

fredrikekre (commented):
Very nice. Since the main reason for this tutorial to exist is to show embedding I think it could be highlighted a bit more. However, since Ferrite is now pretty general there is really nothing extra you need to do to make embedding work :D Perhaps just add a remark somewhere that highlights that we use 2D elements embedded in 3D space?

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+    gmsh.finalize()
+    grid = Grid(elements, nodes);
+
+    ip = Lagrange{RefTriangle, 1}()
+    ip_geo = Lagrange{RefTriangle, 1}()
+    qr = QuadratureRule{RefTriangle}(2)
+    cellvalues = CellValues(qr, ip, ip_geo^3);
+
+    dh = DofHandler(grid);
+    add!(dh, :reactants, ip^2);
```
fredrikekre, line 192: Does it make sense to think of this as a vector field? Perhaps it should be two scalar fields?

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+    gmsh.model.occ.synchronize()
+
+    gmsh.model.mesh.renumberNodes()
+    gmsh.model.mesh.renumberElements()
+
+    ## Generate surface elements and refine several times
+    gmsh.model.mesh.generate(2)
+    for _ in 1:refinements
+        gmsh.model.mesh.refine()
+    end
```
fredrikekre, line 178: Perhaps extract this into a separate function and then the embedding stuff can be discussed before this new function? (i.e. we create a 3D sphere but only mesh the surface etc)

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
+# Putting this flag to false reproduces the figure shown in the example #src
+# We check for laminar flow development in the CI                       #src
+if isdefined(Main, :is_ci) #hide
+    IS_CI = Main.is_ci     #hide
+else                       #hide
+    IS_CI = false          #hide
+end                        #hide
```
fredrikekre, line 7:
```suggestion
end                        #hide
nothing                    #hide
```
To hide the `false` printout in https://ferrite-fem.github.io/Ferrite.jl/previews/PR715/tutorials/reactive_surface/

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,259 @@
...
+# # [Reactive Surface](@id tutorial-reactive-surface)
```
fredrikekre, line 8: I think we have switched to
```suggestion
# # [Reactive surface](@id tutorial-reactive-surface)
```
in most places.

docs/src/tutorials/index.md
```diff
@@ -145,18 +145,28 @@ for the time-integration.
 
 ---
 
+#### [Tutorial 10: Reactive Surface](@ref tutorial-reactive-surface)
```
fredrikekre, line 148:
```suggestion
#### [Tutorial 10: Reactive surface](@ref tutorial-reactive-surface)
```

termi-official:
> Very nice. Since the main reason for this tutorial to exist is to show embedding I think it could be highlighted a bit more. However, since Ferrite is now pretty general there is really nothing extra you need to do to make embedding work :D
I also struggled a bit with this one. Maybe I should add a note to the devdocs for users who might be interested in the internals?

> Perhaps just add a remark somewhere that highlights that we use 2D elements embedded in 3D space?
Will do.

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,260 @@
...
+        ## Get the quadrature weight
+        dΩ = getdetJdV(cellvalues, q_point)
+        ## Loop over test shape functions
+        for i in 1:n_basefuncs
+            ∇δuᵢ = shape_gradient(cellvalues, q_point, i)
+            ## Loop over trial shape functions
+            for j in 1:n_basefuncs
+                ∇δuⱼ = shape_gradient(cellvalues, q_point, j)
+                ## Add contribution to Ke
+                De[2*i-1, 2*j-1] += 2*0.00008 * (∇δuᵢ ⋅ ∇δuⱼ) * dΩ
```
KristofferC, line 97: Are these "material parameters"?

termi-official: Oh, right. That is a material parameter (diffusion coefficient). Let me fix it.

docs/src/literate-tutorials/reactive_surface.jl
termi-official: Here it is: https://private-user-images.githubusercontent.com/9196588/239651993-a2975c20-36c0-4a6b-979d-a4a0553223ec.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTcxNjA4MTgsIm5iZiI6MTcxNzE2MDUxOCwicGF0aCI6Ii85MTk2NTg4LzIzOTY1MTk5My1hMjk3NWMyMC0zNmMwLTRhNmItOTc5ZC1hNGEwNTUzMjIzZWMuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDUzMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA1MzFUMTMwMTU4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9ZGE5OWZmOTVmNTNkMzM4M2M3MzA3MWNiMGEwYmJmMjgzNjdiOTI4NDJlZWNjNjA4OWE5NzZlM2IxM2ExNTVmYiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.DZLlv5Kw_qM3O2XpvhW-dnci5B2e8pnkjWvnKp_PGx4 
Can you add it to the branch? I will then update the downloader.

docs/src/literate-tutorials/reactive_surface.jl
termi-official: Is this sufficient or do you have more in mind?

docs/src/literate-tutorials/reactive_surface.jl
termi-official: I expanded the explanation on this, is it more clear now?

docs/src/literate-tutorials/reactive_surface.jl
termi-official: This is a technical trick to have a fast query for the nodal dofs. See the updated docs for why we need this.

docs/src/literate-tutorials/reactive_surface.jl
termi-official: Good point, thanks. Should be addressed now.

docs/src/literate-tutorials/reactive_surface.jl
fredrikekre: Link doesn't seem to work. Is the the same as in the first post?

docs/src/literate-tutorials/reactive_surface.jl
fredrikekre: Otherwise it would just be the transpose though, but doesn't really matter.

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -0,0 +1,333 @@
...
+# ```math
+#   \partial_t \mathbf{r} = \mathcal{D}\mathbf{r} + R(\mathbf{r}) \quad \textbf{x} \in \Omega
+# ```
+# where $\mathcal{D}$ is the diffusion operator and $R$ is the reaction operator. Notice that the right
+# hand side is just the sum of two operators. Now with our operator splitting scheme we can advance a
+# solution $\mathbf{r}(t_1)$ to $\mathbf{r}(t_2)$ by first solving a heat problem
+# ```math
+#   \partial_t \mathbf{r}^{\mathrm{\mathrm{A}}} = \mathcal{D}\mathbf{r}^{\mathrm{A}} \quad \textbf{x} \in \Omega
+# ```
+# with $\mathbf{r^{\mathrm{A}}(t_1) = \mathbf{r}(t_1)$ on the time interval $t_1$ to $t_2$ and use
```
fredrikekre, line 58: This doesn't seem to render for some reason (https://ferrite-fem.github.io/Ferrite.jl/previews/PR715/tutorials/reactive_surface/#Solver-details)

docs/src/literate-tutorials/reactive_surface.jl
termi-official: Yes its the same as here https://github.com/notifications#issue-1717915883

PR #716, opened 2023-05-20
`QuadratureRule`: remove `dim` parameter and introduce `FaceQuadratureRule`

fredrikekre:
This patch removes the `dim` parameter from `QuadratureRule` constructors, since it is now implicit from the reference shape.

Unlike for interpolations (#711) the `dim` parameter for `QuadratureRule` was not completely redundant since it was used to construct quadrature rules for face integration. This patch therefore introduces `FaceQuadratureRule` as a replacement:
`FaceQuadratureRule{refshape}` instead of `QuadratureRule{dim - 1, refshape}`. The new syntax generalizes better to elements such as wedge or pyramid which doesn't have a unique `dim - 1` reference shape.

With the new syntax it is also possible to materialize the rule for each face already in the `FaceQuadratureRule` constructor, instead of doing it in the `FaceValues` constructor.

### DIFF
--- a/docs/src/literate/heat_equation.jl
+++ b/docs/src/literate/heat_equation.jl
@@ -53,11 +53,11 @@ grid = generate_grid(Quadrilateral, (20, 20));
 # this we need to specify an interpolation space for the shape functions.
 # We use Lagrange functions
 # based on the two-dimensional reference quadrilateral. We also define a quadrature rule based on
-# the same reference cube. We combine the interpolation and the quadrature rule
+# the same reference element. We combine the interpolation and the quadrature rule
 # to a `CellValues` object.
 dim = 2
 ip = Lagrange{RefQuadrilateral, 1}()
-qr = QuadratureRule{dim, RefQuadrilateral}(2)
+qr = QuadratureRule{RefQuadrilateral}(2)
 cellvalues = CellValues(qr, ip);
 
 # ### Degrees of freedom
--- a/docs/src/literate/heat_equation.jl
+++ b/docs/src/literate/heat_equation.jl
@@ -55,7 +55,6 @@ grid = generate_grid(Quadrilateral, (20, 20));
 # based on the two-dimensional reference quadrilateral. We also define a quadrature rule based on
 # the same reference element. We combine the interpolation and the quadrature rule
 # to a `CellValues` object.
-dim = 2
 ip = Lagrange{RefQuadrilateral, 1}()
 qr = QuadratureRule{RefQuadrilateral}(2)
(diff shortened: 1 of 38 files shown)


docs/src/literate/heat_equation.jl
```diff
@@ -53,11 +53,11 @@ grid = generate_grid(Quadrilateral, (20, 20));
 # this we need to specify an interpolation space for the shape functions.
 # We use Lagrange functions
 # based on the two-dimensional reference quadrilateral. We also define a quadrature rule based on
-# the same reference cube. We combine the interpolation and the quadrature rule
+# the same reference element. We combine the interpolation and the quadrature rule
 # to a `CellValues` object.
 dim = 2
```
termi-official, line 58:
```suggestion
```

docs/src/literate/helmholtz.jl
```diff
@@ -55,8 +55,8 @@ grid = generate_grid(Quadrilateral, (150, 150))
 
 dim = 2
```
termi-official, line 56:
```suggestion
```

docs/src/literate/transient_heat_equation.jl
```diff
@@ -66,7 +66,7 @@ grid = generate_grid(Quadrilateral, (100, 100));
 # Again, we define the structs that are responsible for the `shape_value` and `shape_gradient` evaluation.
 dim = 2
```
termi-official, line 67:
```suggestion
```

docs/src/reference/quadrature.md
```diff
@@ -7,7 +7,9 @@ DocTestSetup = :(using Ferrite)
 
 ```@docs
 QuadratureRule
-AbstractRefShape
```
termi-official, line 10: So ref shapes are now gone from docs? :P

test/test_apply_rhs.jl
```diff
@@ -2,7 +2,7 @@ function test_apply_rhs()
     grid = generate_grid(Quadrilateral, (20, 20))
     dim = 2
```
termi-official, line 3:
```suggestion
```

test/test_l2_projection.jl
termi-official, line 1: `test_projection` has an unused `dim` variable

termi-official (approved):
Some cleanup can be done. LGTM otherwise.

docs/src/reference/quadrature.md
fredrikekre: It was included elsewhere and caused a Documenter warning.

PR #721, opened 2023-05-22
Fold `Ferrite.value` into methods of `shape_value`

### LINKED ISSUE #609: Rename `Ferrite.value` and `Ferrite.derivative`
We now have
https://github.com/Ferrite-FEM/Ferrite.jl/blob/84e9aefd1cafc54d6aefd4587cecbf0e555d20f6/src/interpolations.jl#L65
and
https://github.com/Ferrite-FEM/Ferrite.jl/blob/84e9aefd1cafc54d6aefd4587cecbf0e555d20f6/src/interpolations.jl#L76
for computing the value and derivate of shape functions in the reference domain. These are not great names. We could (re)use `shape_value` and `shape_gradient` which is already used for `CellValues`, but in that case it is not in the reference domain, and would also "leak" these internal methods.

I don't have any good ideas,  hence the issue, but perhaps `reference_shape_value` and `reference_shape_gradient`.

fredrikekre:
This patch deprecates `Ferrite.value(ip::Interpolation, i::Int, ξ::Vec)` in favor of `shape_value(ip::Interpolation, ξ::Vec, i::Int)`. Note that the argument order is flipped to resemble the `(Cell|Face)Values` API more. Compare e.g. `shape_value(ip, ξ, i)` with `shape_value(cv, qp, i)`: the second argument becomes the "quadrature point" argument (`ξ` ~ `qp`), and the third argument is the shape function index, in both cases.

Since `shape_value` is exported this introduces new (public) functionality to users -- for some applications it is nice to be able to evaluate the shape functions in the reference domain without having to go through `CellValues`. For the same reason this patch also adds the method `shape_gradient(ip::Interpolation, ξ::Vec, i::Int)` which falls back to AD from Tensors.jl.

Fixes #609.

### DIFF
--- a/src/FEValues/cell_values.jl
+++ b/src/FEValues/cell_values.jl
@@ -69,10 +69,10 @@ function CellValues{IP, N_t, dNdx_t, dNdξ_t, T, dMdξ_t, QR, GIP}(qr::QR, ip::I
 
     for (qp, ξ) in pairs(getpoints(qr))
         for basefunc in 1:n_func_basefuncs
-            dNdξ[basefunc, qp], N[basefunc, qp] = gradient_and_value(ip, basefunc, ξ)
+            dNdξ[basefunc, qp], N[basefunc, qp] = shape_gradient_and_value(ip, ξ, basefunc)
         end
         for basefunc in 1:n_geom_basefuncs
-            dMdξ[basefunc, qp], M[basefunc, qp] = gradient_and_value(gip, basefunc, ξ)
+            dMdξ[basefunc, qp], M[basefunc, qp] = shape_gradient_and_value(gip, ξ, basefunc)
         end
     end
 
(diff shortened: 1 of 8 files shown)


lijas (approved):
I think this is a good change. For example in problems with contact or mortar methods I think it is common to eval the shape functions directly (and to not pre-compute the shape functions using e.g CellValues)

KnutAM:
Not sure if it makes sense, but at least worth considering IMO;
should we then also support the remaining 
- `shape_symmetric_gradient`
- `shape_divergence`
- `shape_curl`

with the same interface? The current implementation could just be on top using the tensor operations, but it might be possible to have some optimized implementations in the future.

fredrikekre:
Maybe, note that these return unmapped values, so not sure how useful it is?

PR #726, opened 2023-05-23
Make create_incidence_matrix work for AbstractGrid

lijas:
These changes would be useful in IGA.jl

### DIFF
--- a/src/Grid/coloring.jl
+++ b/src/Grid/coloring.jl
@@ -1,13 +1,11 @@
 # Incidence matrix for element connections in the grid
-function create_incidence_matrix(g::Grid, cellset=1:getncells(g))
+function create_incidence_matrix(g::AbstractGrid, cellset=1:getncells(g))
     cell_containing_node = Dict{Int, Set{Int}}()
     for cellid in cellset
         cell = getcells(g, cellid)
         for v in cell.nodes
-            if !haskey(cell_containing_node, v)
-                cell_containing_node[v] = Set{Int}()
-            end
-            push!(cell_containing_node[v], cellid)
+            _set = get!(cell_containing_node, v, Set{Int}())
+            push!(_set, cellid)
         end
     end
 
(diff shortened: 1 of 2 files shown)


src/Grid/coloring.jl
```diff
@@ -1,13 +1,11 @@
...
+function create_incidence_matrix(g::AbstractGrid, cellset=1:getncells(g))
     cell_containing_node = Dict{Int, Set{Int}}()
     for cellid in cellset
         cell = getcells(g, cellid)
         for v in cell.nodes
-            if !haskey(cell_containing_node, v)
-                cell_containing_node[v] = Set{Int}()
-            end
-            push!(cell_containing_node[v], cellid)
+            _set = get!(cell_containing_node, v, Set{Int}())
```
fredrikekre, line 7:
```suggestion
            _set = get!(Set{Int}, cell_containing_node, v)
```

fredrikekre (commented):
Don't you need `getnodes(cell)` on L5 etc?

koehlerson:
> No, my IGA grid also has a field called `nodes`

but then it should nonetheless use `getnodes` otherwise it's not really for `AbstractGrid` working

lijas:
This was just a quick fix for IGA.jl, but I can update it so it uses `getnodes` instead.

lijas:
Hmm, where to you want me to use `getnodes`?  I want the `nodeids` from the cell (`cell.nodes`) which is not part of the `AbstractGrid` interface?

termi-official:
I think the interface is now `get_node_ids`

koehlerson:
> I think the interface is now `get_node_ids`

https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L75

fredrikekre (approved):
Add a test using the SmallGrid in test/abstractgrid.jl?

PR #729, opened 2023-05-24
`ConstraintHandler` for discontinuous interpolations

AbdAlazezAhmed:
I haven't tested it with a problem, but it seems to work upon checking debugging mode.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -833,6 +833,13 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
         if interpolation isa VectorizedInterpolation
             interpolation = interpolation.ip
         end
+
+        if getorder(interpolation) == 0 
+            @warn("No dof prescribed for order 0 interpolations")
+            dbc_added = true
+            continue
+        end
+    
         # Set up components to prescribe (empty input means prescribe all components)
         components = isempty(dbc.components) ? collect(Int, 1:n_comp) : dbc.components
         if !all(c -> 0 < c <= n_comp, components)
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -833,13 +833,7 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
         if interpolation isa VectorizedInterpolation
             interpolation = interpolation.ip
         end
-
-        if getorder(interpolation) == 0 
-            @warn("No dof prescribed for order 0 interpolations")
-            dbc_added = true
-            continue
-        end
-    
+        getorder(interpolation) == 0 && error("No dof prescribed for order 0 interpolations")
         # Set up components to prescribe (empty input means prescribe all components)
         components = isempty(dbc.components) ? collect(Int, 1:n_comp) : dbc.components
         if !all(c -> 0 < c <= n_comp, components)
@@ -852,7 +846,7 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
             EntityType = FaceIndex
         end
         CT = getcelltype(ch.dh.grid, first(fh.cellset)) # Same celltype enforced in FieldHandler constructor
-        bcvalues = BCValues(interpolation , default_interpolation(CT), EntityType)
+        bcvalues = BCValues(interpolation, default_interpolation(CT), EntityType)
         # Recreate the Dirichlet(...) struct with the filtered set and call internal add!
         filtered_dbc = Dirichlet(dbc.field_name, filtered_set, dbc.f, components)
         _add!(
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -304,7 +304,7 @@ function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, b
     @assert issorted(components)
     local_face_dofs = Int[]
     local_face_dofs_offset = Int[1]
-    for (_, face) in enumerate(boundaryfunc(interpolation))
+    for (_, face) in enumerate(boundaryfunc(IsDiscontinuous(interpolation) ? get_continuous_interpolation(interpolation) : interpolation))
         for fdof in face, d in 1:field_dim
             if d in components
                 push!(local_face_dofs, (fdof-1)*field_dim + d + offset)
@@ -845,7 +845,7 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
             EntityType = FaceIndex
         end
         CT = getcelltype(ch.dh.grid, first(fh.cellset)) # Same celltype enforced in FieldHandler constructor
-        bcvalues = BCValues(interpolation, default_interpolation(CT), EntityType)
+        bcvalues = BCValues(IsDiscontinuous(interpolation) ? get_continuous_interpolation(interpolation) : interpolation , default_interpolation(CT), EntityType)
         # Recreate the Dirichlet(...) struct with the filtered set and call internal add!
         filtered_dbc = Dirichlet(dbc.field_name, filtered_set, dbc.f, components)
         _add!(
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -304,7 +304,7 @@ function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, b
     @assert issorted(components)
     local_face_dofs = Int[]
     local_face_dofs_offset = Int[1]
-    for (_, face) in enumerate(boundaryfunc(IsDiscontinuous(interpolation) ? get_continuous_interpolation(interpolation) : interpolation))
+    for (_, face) in enumerate(boundaryfunc(interpolation))
         for fdof in face, d in 1:field_dim
             if d in components
                 push!(local_face_dofs, (fdof-1)*field_dim + d + offset)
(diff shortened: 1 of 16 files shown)


AbdAlazezAhmed:
Setup:
```julia
julia> using Ferrite

julia> grid = generate_grid(Quadrilateral, (3, 3));

julia> ip = DiscontinuousLagrange{2, RefCube, 1}()
DiscontinuousLagrange{RefQuadrilateral, 1}()

julia> dh = DofHandler(grid);

julia> add!(dh, :u, 1,ip);
 (generic function with 10 methods)

julia> close!(dh);;

julia> ch = ConstraintHandler(dh);

julia> ∂Ω = ∪(
           getfaceset(grid, "left"),
           getfaceset(grid, "right"),
           getfaceset(grid, "top"),
           getfaceset(grid, "bottom"),

       );

julia> dbc = Dirichlet(:u, ∂Ω, (x, t) -> 0)
Dirichlet(var"#5#6"(), Set(FaceIndex[FaceIndex((3, 2)), FaceIndex((7, 3)), FaceIndex((8, 3)), FaceIndex((1, 4)), FaceIndex((3, 1)), FaceIndex((9, 3)), FaceIndex((2, 1)), FaceIndex((4, 4)), FaceIndex((7, 4)), FaceIndex((6, 2)), FaceIndex((9, 2)), FaceIndex((1, 1))]), :u, Int64[], Int64[], Int64[])

julia> add!(ch, dbc);

julia> close!(ch)
ConstraintHandler:
  BCs:
    Field: u, Components: [1]

```
Master branch:
```julia
julia> ch.prescribed_dofs
Int64[]
```
This PR:
```julia
julia> ch.prescribed_dofs
20-element Vector{Int64}:
  1
  2
  4
  5
  ⋮
 34
 35
 36
```

fredrikekre:
I wonder if we can define `facedof_indices` for the interpolations, instead? That is only used for boundary conditions I think.

What happens for the piecewise constant (order 0) interpolations? I suppose for those one need to impose the boundary condition by integrating the jump, just as is done for element-element couplings?

AbdAlazezAhmed:
Indeed, the `get_continuous_interpolation()` approach is faulty for piecewise constant interpolations. I'll check it, thanks!

termi-official:
> I wonder if we can define `facedof_indices` for the interpolations, instead? That is only used for boundary conditions I think.
> 

Note that this one is also designed for the distributed assembly stuff to sync process boundaries - I just started chery picking stuff from the PR to make the review easier.

> What happens for the piecewise constant (order 0) interpolations? I suppose for those one need to impose the boundary condition by integrating the jump, just as is done for element-element couplings?

If we have no basis function on the interior of the boundary of the element, then AFAIK we cannot use the strong Dirichlet enforcement anymore and fall back to penalty approaches.

fredrikekre:
> If we have no basis function on the interior of the boundary of the element, then AFAIK we cannot use the strong Dirichlet enforcement anymore and fall back to penalty approaches.

Right, but penalty approach (integrating the jump from the element to the prescribed value(?)) would work also for higher order elements, right? This PR is for supporting strong Dirichlet when possible then, which should simply not be supported for order 0.

AbdAlazezAhmed:
> This PR is for supporting strong Dirichlet when possible then, which should simply not be supported for order 0.

Indeed. It currently prescribes no dofs for order 0.

> Note that this one is also designed for the distributed assembly stuff to sync process boundaries

Maybe we can do something like:
```julia
function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, boundaryfunc::F=facedof_indices) where F
    @assert issorted(components)
    local_face_dofs = Int[]
    local_face_dofs_offset = Int[1]
    getorder(interpolation) == 0 && return [], [1 for _ in 1:nfaces(interpolation)+1]
    for (_, face) in enumerate(boundaryfunc(IsDiscontinuous(interpolation) ? get_continuous_interpolation(interpolation) : interpolation))
        for fdof in face, d in 1:field_dim
            if d in components
                push!(local_face_dofs, (fdof-1)*field_dim + d + offset)
            end
        end
        push!(local_face_dofs_offset, length(local_face_dofs) + 1)
    end
    return local_face_dofs, local_face_dofs_offset
end
```

termi-official:
> > This PR is for supporting strong Dirichlet when possible then, which should simply not be supported for order 0.
> 
> Indeed. It currently prescribes no dofs for order 0.
> 

Yea this case is quite tricky. Users may accidentally prescribe different values on different faces in the corners of a domain which leads to weird ambiguities. It also raises the question at which point of the boundary should a constraint be evaluated. We should just throw a warning in this case and do nothing - which should be reflected in the tests, because this feels like there are possibilities to introduce unexpected regressions.

> > Note that this one is also designed for the distributed assembly stuff to sync process boundaries
> 
> Maybe we can do something like:
> 
> ```julia
> function _local_face_dofs_for_bc(interpolation, field_dim, components, offset, boundaryfunc::F=facedof_indices) where F
>     @assert issorted(components)
>     local_face_dofs = Int[]
>     local_face_dofs_offset = Int[1]
>     getorder(interpolation) == 0 && return [], [1 for _ in 1:nfaces(interpolation)+1]
>     for (_, face) in enumerate(boundaryfunc(IsDiscontinuous(interpolation) ? get_continuous_interpolation(interpolation) : interpolation))
>         for fdof in face, d in 1:field_dim
>             if d in components
>                 push!(local_face_dofs, (fdof-1)*field_dim + d + offset)
>             end
>         end
>         push!(local_face_dofs_offset, length(local_face_dofs) + 1)
>     end
>     return local_face_dofs, local_face_dofs_offset
> end
> ```

I should also note that the `get_continuous_interpolation` trick will only work for the implemented discontinuous interpolation, which is not the usually used one. Most of the interpolations in DG do not have the nodes on the boundary, but for example at the Gauss-Legendre points (or other ones, depending on your needs). It is still nice (and wanted) to have strong Dirichlet enforcement for the existing interpolation! This e.g. allows us to compare approaches in a unified framework.

fredrikekre:
`dirichlet_*dof_indices` sounds like a good plan

src/Dofs/ConstraintHandler.jl
```diff
@@ -833,6 +833,13 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
         if interpolation isa VectorizedInterpolation
             interpolation = interpolation.ip
         end
+
+        if getorder(interpolation) == 0 
+            @warn("No dof prescribed for order 0 interpolations")
```
fredrikekre, line 838: Any reason not to error here?

src/Dofs/ConstraintHandler.jl
```diff
@@ -845,7 +852,7 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
             EntityType = FaceIndex
         end
         CT = getcelltype(ch.dh.grid, first(fh.cellset)) # Same celltype enforced in FieldHandler constructor
-        bcvalues = BCValues(interpolation, default_interpolation(CT), EntityType)
+        bcvalues = BCValues(interpolation , default_interpolation(CT), EntityType)
```
fredrikekre, line 855: 🤔

src/interpolations.jl
```diff
@@ -347,6 +391,11 @@ getnbasefunctions(::DiscontinuousLagrange{shape,0}) where {shape} = 1
 # This just moves all dofs into the interior of the element.
 celldof_interior_indices(ip::DiscontinuousLagrange) = ntuple(i->i, getnbasefunctions(ip))
 
+# Mirror the Lagrange element for now to avoid repeating.
+dirichlet_facedof_indices(ip::DiscontinuousLagrange) = dirichlet_facedof_indices(get_continuous_interpolation(ip))
```
fredrikekre, line 395: If you end up removing `get_continuous_interpolation`
```suggestion
dirichlet_facedof_indices(ip::DiscontinuousLagrange{shape, order}) where {shape, order} = dirichlet_facedof_indices(Lagrange{shape, order}())
```

src/Dofs/ConstraintHandler.jl
AbdAlazezAhmed: Not really.

fredrikekre (approved):
Looks good, perhaps add a simple test?

AbdAlazezAhmed:
> Looks good, perhaps add a simple test?

Sure, thanks!

AbdAlazezAhmed:
> Can you add tests for the periodic case as regression tests, too? Also, I think the devdocs are missing.

I'm not sure what regression tests can be written for the periodic case as I haven't changed its code, it throws the same mirroring error both before and after this PR.

src/interpolations.jl
```diff
@@ -249,6 +249,20 @@ match the vertex enumeration of the corresponding geometrical cell.
...
+    dirichlet_vertexdof_indices(ip::Interpolation)
+
+A tuple containing tuples of local dof indices for the respective vertex in local
+enumeration on a cell defined by [`vertices(::Cell)`](@ref). The vertex enumeration must
+match the vertex enumeration of the corresponding geometrical cell.
+Used internally in [`ConstraintHandler`](@ref) and defaults to [`vertexdof_indices(ip::Interpolation)`](@ref) for continuous interpolation.
+
+!!! note
+    The dofs appearing in the tuple must be continuous and increasing! The first dof must be
+    the 1, as vertex dofs are enumerated first.
```
AbdAlazezAhmed, line 262: Should this repeat `vertexdof_indices(ip::Interpolation)` or just refer to it?
I repeated it so it appears in the LSP/IDE tip

termi-official: I mean the resulting docstring should be a bit different, right? There is overlap because we use the same convention, which (as you already do) should be documented.

PR #730, opened 2023-05-26
`CellIterator` broadcast safety doc warning

### LINKED ISSUE #267: CellIterator not broadcast safe
I haven't looked into the details of that issue (Opened it as a reminder) and tbh I don't know much about the inner workings of broadcast and materialize. As the current id is mutated in the `CellIterator` the first example only picks up the last id. Any ideas if there's a possible fix for that?

### Wrong
```
julia> cellid.(CellIterator(dh))
22754-element Array{Int64,1}:
 22754
 22754
 22754
 22754
 22754
 22754
     ⋮
 22754
 22754
 22754
 22754
 22754
```

### Correct
```
julia> map(CellIterator(dh)) do cell
           cellid(cell)
       end
22754-element Array{Int64,1}:
     1
     2
     3
     4
     5
     6
     ⋮
 22750
 22751
 22752
 22753
 22754
```

AbdAlazezAhmed:
Adds warning to docstring

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -168,7 +168,6 @@ Base.IteratorSize(::Type{<:CellIterator}) = Base.HasLength()
 Base.IteratorEltype(::Type{<:CellIterator}) = Base.HasEltype()
 Base.eltype(::Type{<:CellIterator{CC}}) where CC = CC
 Base.length(ci::CellIterator) = length(ci.set)
-Base.collect(::CellIterator) = error("CellIterator is stateful and should not be used in broadcasting nor collected")
 
 
 function _check_same_celltype(grid::AbstractGrid, cellset)
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -128,6 +128,10 @@ for idx in cellset
     # ...
 end
 ```
+!!! warning
+
+    `CellIterator` is stateful and should not be used in broadcasting nor collected
+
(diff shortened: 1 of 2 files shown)


fredrikekre:
I don't think we should do this -- it is too easily circumvented (`[c for c in CellIterator(dh)]` for example), and is just not how `CellIterator` is supposed to be used. Perhaps we can just add a note in the docstring that the iterator is stateful?

AbdAlazezAhmed:
Ah, I get it, I thought it could have some use case. Maybe we can throw an error then?
```julia
julia> cellid.(CellIterator(dh))
ERROR: CellIterator is stateful and should not be used in broadcasting or collected
```

fredrikekre:
I think I prefer just the note in the documentation, it isn't feasible to catch every misuse like this anyway.

src/iterators.jl
```diff
@@ -128,6 +128,10 @@ for idx in cellset
     # ...
 end
 ```
+!!! warning
+
+    `CellIterator` is stateful and should not be used in broadcasting nor collected
+
```
fredrikekre, line 134:
```suggestion
!!! warning
    `CellIterator` is stateful and should not be used for things other than `for`-looping
    (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
```

Issue #620, opened 2023-03-17
Introduce citation system to docs


termi-official:
We should start introducing some citation system compatible with Documenter.jl to our docs (e.g. DocumenterCitations.jl).

PR #737, opened 2023-06-07
Structure the documentation according to the Diataxis framework.

fredrikekre:
See https://diataxis.fr/ and also https://youtu.be/t4vKPhjcMZg for a nice overview in form of a conference talk.

The framework suggest keeping the following topics clearly separated
1. Tutorials
2. How-to guides
3. Explanation
4. Reference

We almost follow this but not in a "committed" way:
 - current Examples map to Tutorials *and* in some cases to How-to guides
 - current Manual map to Explanation
 - current Reference map to Reference

I suggest we make these distinction more clear, and in particular that we commit to this style.

Regarding Tutorials vs How-to guides: currently our Examples do a bit of both. I suggest that we use Tutorials to introduce new concepts in a more clear path and avoid repeating things. Each new Tutorial should introduce a new concept (i.e. vector problem, non-linear problem, new type of boundary condition, ...). Then we can use How-to guides to build on top of these. As a concrete example, https://ferrite-fem.github.io/Ferrite.jl/stable/examples/threaded_assembly/ would be a How-to for threaded assembly, which just builds on the corresponding Tutorial program.

This patch currently just reworks the front page though, preview here: https://ferrite-fem.github.io/Ferrite.jl/previews/PR737/.


termi-official:
Leaving https://github.com/kylelobo/The-Documentation-Compendium for more in-depth information on docs writing.

KnutAM:
Nice!
Maybe we should also include a section of "research examples" or something like that, where authors that have written a paper using Ferrite can share a runnable example related to their research, showing how they used Ferrite?
We already have the topology optimization example that could be a candidate. It could also be good to keep a list of papers that have used Ferrite to highlight Ferrite's impact.

termi-official:
> Nice! Maybe we should also include a section of "research examples" or something like that, where authors that have written a paper using Ferrite can share a runnable example related to their research, showing how they used Ferrite? We already have the topology optimization example that could be a candidate. It could also be good to keep a list of papers that have used Ferrite to highlight Ferrite's impact.

xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/620

PR #739, opened 2023-06-08
Add stress computation to incompressible elasticity example

gagankaushikmanyam:
Added stress computation from incompressible_elasticity.jl, and kept the introduction the same as well as comments for individual functions.

### DIFF
--- a/docs/src/literate-tutorials/incompressible_elasticity.jl
+++ b/docs/src/literate-tutorials/incompressible_elasticity.jl
@@ -22,16 +22,15 @@
 # What follows is a program spliced with comments.
 #md # The full program, without comments, can be found in the next
 #md # [section](@ref incompressible_elasticity-plain-program).
-using Ferrite
+using Ferrite, Tensors
 using BlockArrays, SparseArrays, LinearAlgebra
-using Tensors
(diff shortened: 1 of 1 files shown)


docs/src/literate/incompressible_elasticity.jl
```diff
@@ -24,43 +24,47 @@
...
 end;
 
 # Next we define a function to set up our cell- and facevalues.
 function create_values(interpolation_u, interpolation_p)
-    ## quadrature rules
-    qr      = QuadratureRule{RefTriangle}(3)
-    face_qr = FaceQuadratureRule{RefTriangle}(3)
+    # quadrature rules
+    qr = QuadratureRule{2,RefTetrahedron}(3)
+    face_qr = QuadratureRule{1,RefTetrahedron}(3)
```
termi-official, line 46: Why is this reverted?

termi-official (commented):
Hi @gagankaushikmanyam and thanks for the contribution here!

Is it possible that you have developed this PR with an "old" Ferrite version (i.e. not the current master)? I see a lot of changes which change code from the new interface back to the old interface.

PR #743, opened 2023-06-18
`InterfaceValues` for DG interface integration

AbdAlazezAhmed:
Tasks:

- [x] `InterfaceValues`
- [x] syncing/transforming quadrature points
- [x] docs
- [x] tests

##  After merge

* [ ] open issue to discuss mixed grid support for `InterfaceValues`
* [ ] check subdomain support and possibly open issue
* [ ] open issue about the investigation of mortaring with the new infrastructure

PR #747, opened 2023-06-22
WIP: `InterfaceCache` & `InterfaceIterator`

AbdAlazezAhmed:
Used to be in #743 , now its own PR.

Tasks: 
- [ ] `InterfaceCache`
- [ ] `InterfaceIterator`
- [ ] docs
- [ ] tests


fredrikekre:
From the discussion in #743, it would be nice to be able to access all dofs of the interface

fredrikekre:
The concatenation of dofs for both cells, `[dofs_a; dofs_b]`.

src/iterators.jl
```diff
@@ -156,6 +210,47 @@ end
...
+    reinit!(cache.face_b, face_b)
+    return cache
+end
+
+getnodes(ic::InterfaceCache, use_cell_a::Bool = true) = getnodes(use_cell_a ? ic.face_a : ic.face_b)
+get_cell_coordinates(ic::InterfaceCache, use_cell_a::Bool = true) = get_cell_coordinates(use_cell_a ? ic.face_a : ic.face_b)
+cellid(ic::InterfaceCache) = (cellid(ic.face_a), cellid(ic.face_b))
+interfacedofs(ic::InterfaceCache) = vcat(celldofs(ic.face_a), celldofs(ic.face_b))
+interfacedofranges(ic::InterfaceCache) = (1 : length(celldofs(ic.face_a)), length(celldofs(ic.face_a)) + 1 : length(celldofs(ic.face_a))  + length(celldofs(ic.face_b)))
+
```
AbdAlazezAhmed, line 253: What do you think these should be like? 
+ reminder missing docs.

src/iterators.jl
```diff
@@ -261,9 +356,77 @@ end
...
+        grid = gridordh isa Grid ? gridordh : get_grid(gridordh)
+        i = 1;
+        grid_dim = getdim(grid)
+        neighborhood = grid_dim == 1 ? pairs(topology.vertex_vertex_neighbor) : pairs(topology.face_face_neighbor)
+        interface_skeleton = Array{NTuple{2, FaceIndex}}(undef, count(pair -> !isempty(pair[2]) && pair[2].neighbor_info[][1] > pair[1][1], neighborhood))
+        for (idx, face) in neighborhood
+            !isempty(face.neighbor_info) && face.neighbor_info[][1] > idx[1] || continue
+            interface_skeleton[i] = (face.neighbor_info[], FaceIndex(idx[1], idx[2])) #Assumes one neighbor only
+            i+=1
+        end
```
AbdAlazezAhmed, line 406: Ref #759

src/iterators.jl
```diff
@@ -261,9 +356,77 @@ end
...
+!!! warning
+    `InterfaceIterator` is stateful and should not be used for things other than `for`-looping
+    (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
+"""
+struct InterfaceIterator{Cache<:InterfaceCache}
+    cache::Cache
+    set::Set{NTuple{2, FaceIndex}}
+end
+
+function InterfaceIterator(gridordh::Union{Grid,AbstractDofHandler},
```
AbdAlazezAhmed, line 391: The set here is [Tuple([FaceIndex(cell_a, face_a), FaceIndex(cell_b, face_b)])...].
I think user can obtain it using faceset + topology.face_face_neighbor. and it's probably more consistent with how `FaceCache` uses `Set{FaceIndex}`. If you're ok with this approach I can document it.
Note that it does not check if the user passes "bad" faceindex (not neighbors?). the check happens when reiniting interfacevalues IIRC

test/test_grid_dofhandler_vtk.jl
```diff
@@ -452,19 +481,19 @@ end
...
-                jump_abs += abs(u_5_n + u_neighbor) * dΩ
-            end
+            normal_a = getnormal(fv_ele, q_point)
+            u_5_n = function_value(fv_ele, q_point, cellid(ic)[1] == 5 ? u_ele5 : u_neighbors) ⋅ normal_a
+            reinit!(fv_neighbor, ic.face_b.cc, ic.face_b.current_faceid[])
+            normal_b = getnormal(fv_neighbor, q_point)
+            u_neighbor = function_value(fv_neighbor, q_point, cellid(ic)[1] == 5 ? u_neighbors : u_ele5) ⋅ normal_b
+            jump_int += (u_5_n + u_neighbor) * dΩ
+            jump_abs += abs(u_5_n + u_neighbor) * dΩ
         end
```
AbdAlazezAhmed, line 497: This should work with #743

AbdAlazezAhmed (commented):
Some comments

src/iterators.jl
termi-official: How does this differ from faceskeleton?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+ - `getnodes(ic, use_cell_a = true)`: get the global node ids of cell A or cell B of the interface
+ - `get_cell_coordinates(ic, use_cell_a = true)`: get the coordinates of cell A or cell B of the interface
+ - `interfacedofs(ic)`: get the global dof ids of the interface cells
+ - `interfacedofranges(ic)`: get the interface dof ranges of the interface cells
+See also [`InterfaceIterator`](@ref).
+"""
+struct InterfaceCache{FC<:FaceCache}
+    face_a::FC
+    face_b::FC
+end
```
fredrikekre, line 183: I think we should cache the union of the dofs here, it will be expensive to do the `vcat` inside `interfacedofs(...)`:
```suggestion
struct InterfaceCache{FC<:FaceCache}
    face_a::FC
    face_b::FC
    dofs::Vector{Int}
end
```

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+Create a cache object with pre-allocated memory for the coordinates and faces of an
+interface. The cache is updated for a new cell by calling `reinit!(cache, face_a, face_b)` where
+`face_a::FaceIndex` and `face_b::FaceIndex` are the interface faces.
+**Struct fields of `InterfaceCache`**
+ - `ic.face_a :: FaceCache`: current cell node coordinates
+ - `ic.face_b :: FaceCache`: neighbor cell node coordinates
+**Methods with `InterfaceCache`**
+ - `reinit!(cache::InterfaceCache, face_a::FaceIndex, face_b::FaceIndex)`: reinitialize [`InterfaceCache`](@ref)
+ - `cellid(ic)`: get cell ids of the current interface
+ - `getnodes(ic, use_cell_a = true)`: get the global node ids of cell A or cell B of the interface
```
fredrikekre, line 174: Perhaps the way to use this would be through `getnodes(ic.a)` and `getnodes(ic.b)` instead?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+"""
+struct InterfaceCache{FC<:FaceCache}
+    face_a::FC
+    face_b::FC
+end
+
+function InterfaceCache(gridordh::Union{AbstractGrid, AbstractDofHandler})
+    face_a = FaceCache(gridordh)
+    face_b = FaceCache(gridordh)
+    return InterfaceCache(face_a, face_b)
```
fredrikekre, line 188:
```suggestion
    return InterfaceCache(face_a, face_b, Int[])
```

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+end
+
+function reinit!(cache::InterfaceCache, face_a::FaceIndex, face_b::FaceIndex)
+    reinit!(cache.face_a, face_a)
+    reinit!(cache.face_b, face_b)
```
fredrikekre, line 193:
```suggestion
    reinit!(cache.face_b, face_b)
    resize!(cache.dofs, length(face_a.dofs) + length(face_b.dofs))
    for (i, d) in pairs(face_a.dofs)
        cache.dofs[i] = d
    end
    for (i, d) in pairs(face_b.dofs)
        cache.dofs[i + length(face_a.dofs)] = d
    end
```

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+    return cache
+end
+
+getnodes(ic::InterfaceCache, use_cell_a::Bool = true) = getnodes(use_cell_a ? ic.face_a : ic.face_b)
```
fredrikekre, line 197: Maybe remove in favor of `getnodes(ic.a)` and `getnodes(ic.b)`?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+get_cell_coordinates(ic::InterfaceCache, use_cell_a::Bool = true) = get_cell_coordinates(use_cell_a ? ic.face_a : ic.face_b)
```
fredrikekre, line 198: Same thing here; `get_cell_coordinates(ic.a)` and `get_cell_coordinates(ic.b)`?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+cellid(ic::InterfaceCache) = (cellid(ic.face_a), cellid(ic.face_b))
```
fredrikekre, line 199: and here (`cellid(ic.a)` and `cellid(ic.b)`)?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+interfacedofs(ic::InterfaceCache) = vcat(celldofs(ic.face_a), celldofs(ic.face_b))
```
fredrikekre, line 200:
```suggestion
interfacedofs(ic::InterfaceCache) = ic.dofs
```

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+interfacedofranges(ic::InterfaceCache) = (1 : length(celldofs(ic.face_a)), length(celldofs(ic.face_a)) + 1 : length(celldofs(ic.face_a))  + length(celldofs(ic.face_b)))
```
fredrikekre, line 201: What is the indended use for this?

src/iterators.jl
```diff
@@ -156,6 +159,47 @@ end
...
+ - `cellid(ic)`: get cell ids of the current interface
+ - `getnodes(ic, use_cell_a = true)`: get the global node ids of cell A or cell B of the interface
+ - `get_cell_coordinates(ic, use_cell_a = true)`: get the coordinates of cell A or cell B of the interface
+ - `interfacedofs(ic)`: get the global dof ids of the interface cells
+ - `interfacedofranges(ic)`: get the interface dof ranges of the interface cells
+See also [`InterfaceIterator`](@ref).
+"""
+struct InterfaceCache{FC<:FaceCache}
+    face_a::FC
+    face_b::FC
```
fredrikekre, line 182: With the other suggestions, perhaps call these
```suggestion
    a::FC
    b::FC
```
?

src/iterators.jl
fredrikekre: Perhaps the set functionality doesn't need to be here and we loop over all of them? You can always skip elements inside the loop:
```julia
for ic in InterfaceIterator(...)
    if some_condition(ic)
        continue
    end
    # do stuff
end
```

src/iterators.jl
AbdAlazezAhmed: Can be used for 
```julia
function_(value|gradient|...)(iv::InterfaceValues, qp::Int, ue::Vector{Float64}, dof_range_a::UnitRange, dof_range_b::UnitRange)
```
?

fredrikekre: Hmm, but the dof ranges there would be dof ranges for the fields within each "interface dofrange", i.e. to select the u-dofs if you have both u and p.

src/iterators.jl
AbdAlazezAhmed: Btw, should set be changed to a vector/Array as its size is already known and the elements are unique? It may have better performance IIUC

docs/src/devdocs/elements.md
```diff
@@ -34,5 +34,6 @@ Ferrite.getvertexinstances
 Ferrite.filterfaces
 Ferrite.filteredges
 Ferrite.filtervertices
+Ferrite.InterfaceOrientationInfo
```
fredrikekre, line 37: Doesn't seem included in this PR
```suggestion
```

src/iterators.jl
```diff
@@ -261,9 +310,71 @@ end
...
+    neighborhood = grid_dim == 1 ? topology.vertex_vertex_neighbor : topology.face_face_neighbor
+    interface_skeleton = Array{NTuple{2, FaceIndex}}(undef, count(face -> !isempty(neighborhood[face[1], face[2]]), faceskeleton(topology, grid)))
+    for face_a in faceskeleton(topology, grid)
+        !isempty(neighborhood[face_a[1], face_a[2]]) || continue
+        neighbor = neighborhood[face_a[1], face_a[2]].neighbor_info[]
+        face_b = neighbor isa VertexIndex ? FaceIndex(neighbor[1], neighbor[2]) : neighbor
+        interface_skeleton[i] = (face_a, face_b) #Assumes one neighbor only
+        i+=1
+    end
+    set = Set(interface_skeleton)
```
fredrikekre, line 360: I wonder if this can be done lazily. There doesn't seem to be a need to materialize the set here, right? Just loop over all of them and skip the ones which doesn't have a neighbor

AbdAlazezAhmed: Like this?
https://github.com/Ferrite-FEM/Ferrite.jl/blob/7fe9140555b73137afbdd7188b808d1c468d98df/src/iterators.jl#L341-L371

PR #753, opened 2023-06-29
Topology vroom 1

### LINKED ISSUE #617: Topology construction is slow
Implementation of the topology construction is quite slow right now and can be optimized (type stability, allocations, ...).

Concrete ideas are
1. [ ] Preallocate sparsity pattern
2. [ ] Preallocate buffers with sufficient size to reduce allocations
3. [ ] Separate mixed dimensional code path from non-mixed dimensional code path

termi-official:
Modifies the topology construction algorithm and separates additional stuff like stencil and skeleton construction from the topology struct. I will add an interface for mixed-dimensional stuff in a subsequent PR, because I am not sure on the design. We can either add the mixed-dimensional stuff directly to the topology or to a different topology.

Resolves #617 mostly. We should be able to replace the sparse matrices for edges and faces with something more efficient, but I doubt that this will result in a significant runtime boost (however it will certainly be less memory-hungry).


termi-official:
Still not super fast

[c2104c0](https://github.com/Ferrite-FEM/Ferrite.jl/pull/753/commits/c2104c04c6522e80f3513e59bf03ecc163db3bb1)
```
julia> @btime ExclusiveTopology(generate_grid(Hexahedron, (80, 80, 80)));
  12.227 s (20234928 allocations: 2.90 GiB)
```

[7207632](https://github.com/Ferrite-FEM/Ferrite.jl/pull/753/commits/72076326638a8130b2108fbb02e07ea6e577c8b9)
```
julia> @btime ExclusiveTopology(generate_grid(Hexahedron, (80, 80, 80)));
  11.129 s (46327260 allocations: 4.34 GiB)
```

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+- `vertex_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}`: `vertex_neighbor[cellid,local_vertex_id]` -> neighboring vertex
+- `edge_neighbor::SparseMatrixCSC{EntityNeighborhood,Int}`: `edge_neighbor[cellid_local_vertex_id]` -> neighboring edge
+"""
+struct ExclusiveTopology <: AbstractTopology
+    # maps a global vertex id to all cells containing the vertex
+    vertex_to_cell::Vector{Set{Int}}
+    # index of the vector = cell id ->  all other connected cells
+    cell_neighbor::Vector{EntityNeighborhood{CellIndex}}
+    # face_neighbor[cellid,local_face_id] -> exclusive connected entities (not restricted to one entity)
+    face_face_neighbor::Matrix{EntityNeighborhood{FaceIndex}}
```
koehlerson, line 68: Why a dense matrix over a sparse one?

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+        println(io, "$(n.neighbor_info[1])")
+    else
+        println(io, "$(n.neighbor_info...)")
+    end
+end
+
+abstract type AbstractTopology end
+
+"""
+    ExclusiveTopology(cells::Vector{C}) where C <: AbstractCell
```
koehlerson, line 49: docstrings are outdated and more important there should be a note that mixed dimensional connectivity is now not working. Previously it was working to some degree in the sense of dispatched `edges,faces,vertices`

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+function getstencil(top::Dict{Int,EntityNeighborhood{VertexIndex}}, grid::AbstractGrid, vertex_idx::VertexIndex)
+    return top[toglobal(grid, vertex_idx)].neighbor_info
+end
+
+"""
+    compute_face_skeleton(topology, grid) -> Iterable{FaceIndex}
+Creates an iterateable face skeleton. The skeleton consists of `FaceIndex` that can be used to `reinit`
+`FaceValues`.
+"""
+function compute_face_skeleton(top::ExclusiveTopology, grid::Grid)
```
koehlerson, line 325: why is this outside of the struct?

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+"""
+    getstencil(top::StencilType, grid::AbstractGrid, vertex_idx::VertexIndex) -> Iterable{EntityNeighborhood}
+Get an iterateable over the stencil members for a given local entity.
+"""
+function getstencil(top::Dict{Int,EntityNeighborhood{VertexIndex}}, grid::AbstractGrid, vertex_idx::VertexIndex)
+    return top[toglobal(grid, vertex_idx)].neighbor_info
+end
+
+"""
+    compute_face_skeleton(topology, grid) -> Iterable{FaceIndex}
```
koehlerson, line 321: why not document the actual return?

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+    end
+    if include_self
+        return unique([top.edge_edge_neighbor[cellid, local_edgeidx].neighbor_info; self_reference_local; edgeidx])
+    else
+        return unique([top.edge_edge_neighbor[cellid, local_edgeidx].neighbor_info; self_reference_local])
+    end
+end
+
+"""
+    compute_vertex_star_stencils(top, grid) -> StencilType
```
koehlerson, line 287: The return docstring is for a user useless since `StencilType` is not a struct as far as I understand. I would rather resort to the actual return type or at least explain what indexing the return type does

src/Grid/topology.jl
```diff
@@ -0,0 +1,341 @@
...
+the given entity is included in the returned list as well.
+"""
+getneighborhood
+
+struct EntityNeighborhood{T<:Union{BoundaryIndex,CellIndex}}
+    neighbor_info::Vector{T}
+end
+
+EntityNeighborhood(info::T) where T <: BoundaryIndex = EntityNeighborhood([info])
+Base.zero(::Type{EntityNeighborhood{T}}) where T = EntityNeighborhood(T[])
```
koehlerson, line 22: I guess most of the stuff until 34 can be removed, since it was only needed for sparse matrices?

src/Grid/topology.jl
termi-official: It seems to be a bit faster and uses less memory, because we had almost all entries filled in the sparse matrix (which is above the rule-of-thumb threshold of nlog(n) entries).

src/Grid/topology.jl
termi-official: Good catch.

src/Grid/topology.jl
termi-official: Two reasons here. First, the skeleton is not always needed if we query basic topological information and second the algorithm is actually separate. I think this is also what @fredrikekre had in mind here.

src/Grid/topology.jl
koehlerson: > because we had almost all entries filled in the sparse matrix

ah I understand the misconception I had. I thought with increasing cells, there should be less non-connectivities encoded. However, non-connectivies are encoded by not having more entries in the EntityNeighborhood struct and certainly each cell does have a connectivity except for the boundary elements. So this makes absolute sense

src/Grid/topology.jl
koehlerson: How about a keyword argument in the constructor and have a union with nothing in the struct?

src/Grid/topology.jl
termi-official: I am still not super sure if we can eliminate this struct completely (and if this gives us more performance).

koehlerson: at least for vertex connections you cannot a priori how much other vertices are connected in the general case, can you?

src/exports.jl
```diff
@@ -67,6 +67,8 @@ export
     ExclusiveTopology,
     getneighborhood,
     faceskeleton,
+    compute_vertex_star_stencils,
```
koehlerson, line 70: maybe just `vertex_star_stencil` ? `compute` is kinda true in any case with computers

termi-official: Not really. For example, getters do not compute stuff.

termi-official: I am also not really understanding the point here, we have several functions even in Ferrite with this prefix https://github.com/search?q=repo%3AFerrite-FEM%2FFerrite.jl%20compute_&type=code

src/Grid/topology.jl
termi-official: I think this is true for any connectivity. I was more thinking about whether we really need EntityNeighborhood (which is basically a type wrapper around vector) or if we can use directly use a vector.

src/exports.jl
koehlerson: Those are mostly not Ferrite exposed functions. Its a better name without the compute imo.

koehlerson: To clarify: we would never write compute_function_value or compute_function_divergence since it would be a prefix which is to some extent almost true whenever you call something that is not 100% cached. So where do you draw the line and what is the advantage over documenting in the docs that it's actually computed on the fly

termi-official:
Still slow and unfortunately I introduces quite a bit of fluctuation in runtime.

```
julia> @btime ExclusiveTopology(generate_grid(Hexahedron, (80, 80, 80)));
  10.042 s (33015261 allocations: 3.75 GiB)

julia> @btime ExclusiveTopology(generate_grid(Hexahedron, (80, 80, 80)));
  8.195 s (33015261 allocations: 3.75 GiB)
  ```

termi-official:
Since this topology is specific to some AMR algorithms I will try to add a different topology struct which will hopefully be faster to construct in the future.

koehlerson (approved):

src/Grid/topology.jl
koehlerson: let's delete all the functions at least that were specified in order to make `SparseMatrixCSC` usage work

fredrikekre:
Clearer squashed commit message would be nice the next time :)

koehlerson:
Whoopsie, will do next time!

Issue #741, opened 2023-06-10
Naming convention for getters


KnutAM:
In #664 the question of consistent naming of (exported/documented) getters came up. In particular changing `get*` to `get_*`, changes a lot of functions. Changing some of these would be very breaking (but should also be ok by leaving as deprecated) for user code, but here is just a list of the different getters we have exported, to facilitate the discussion.  

# Already following the convention
```julia 
get_rhs_data,
get_point_values
get_grid # #655 
get_*dim # #655
get_cell_coordinates # #664 
get_cell_coordinates! # #664 
get_node_coordinate # #664 
```

# Exports not following the `get_*` convention
## `Interpolation`
```julia
getnbasefunctions
```

## `QuadratureRule`
```julia
getnquadpoints,
getweights,
getpoints,
```

## `FEValues`
```julia
shape_value,
shape_gradient,
shape_symmetric_gradient,
shape_divergence,
shape_curl,
function_value,
function_gradient,
function_symmetric_gradient,
function_divergence,
function_curl,
spatial_coordinate,
getnormal,
getdetJdV,
```

## `AbstractGrid`
```julia
getneighborhood,
faceskeleton,
getcells,
getncells,
getnodes,
getnnodes,
getcelltype,
getcellset,
getnodeset,
getfaceset,
getedgeset,
getvertexset,
getcoordinates,  # Deprecated in #664
getcoordinates!, # Deprecated in #664
getcellsets,
getnodesets,
getfacesets,
getedgesets,
getvertexsets,
nfaces,
```

## `DofHandler`
```julia
ndofs,
ndofs_per_cell,
celldofs!,
celldofs,
dof_range,
```

## `CellCache`
```julia
cellid
```

KnutAM:
My suggestion would be to be consistent in prepending `get`, but not enforcing the underscore if readable without (following the Base [convention](https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/)). The exception would be `FEValues`

With this, the following changes could be made
```julia
faceskeleton -> get_face_skeleton
ndofs -> getndofs # compare with e.g. getncells
ndofs_per_cell -> get_num_celldofs # Note sure what is best here
celldofs! -> get_cell_dofs!
celldofs -> get_cell_dofs
cellid -> get_cell_id
dof_range -> get_dof_range
```

[...]

PR #757, opened 2023-07-04
getcoordinates

KnutAM:
This PR partially reverts #664 as requested in #741. 
The current state just adds docstrings, I'll replace in the code after the interface is settled. 

As I understood from @fredrikekre, @kimauth wanted to keep `getcoordinates` for getting the cell coordinates, and that is what I've added to the docstring now. 

* `Ferrite.get_node_coordinate`: Gives the coordinate of a single node (replaces `getcoordinates` from before #664)
* `getcoordinates`: Gives the coordinate of a cell (specified in different ways, `::AbstractGrid, ::Union{Int, CellIndex, FaceIndex}` or `::CellCache`)
* `getcoordinates!`: Same as above, excluding `::CellCache`

However, this interface would not play well with @lijas [suggestion](https://github.com/Ferrite-FEM/Ferrite.jl/pull/664#issuecomment-1585525736) of having different output depending on the input. One option could be to remove the `::FaceIndex` option as suggested in #702, which only leaves the `getcoordinates(grid, ::Int)` option ambiguous wrt. what to expect (cell or node).

Please comment on which option you think is the best.


fredrikekre:
I think the behavior documented in this PR now is good. Also would be fine to remove the method for FaceIndex.

src/Grid/grid.jl
```diff
@@ -644,41 +643,45 @@ function addnodeset!(grid::AbstractGrid, name::String, f::Function)
...
+    getcoordinates(grid::AbstractGrid, idx::Union{Int,CellIndex})
+    getcoordinates(cache::CellCache)
+    
+Get a vector with the coordinates of the cell corresponding to `idx` or `cache`
 """
-@inline function get_cell_coordinates!(x::Vector{Vec{dim,T}}, grid::Ferrite.AbstractGrid, cellid::Int) where {dim,T} 
-    cell = getcells(grid, cellid)
-    get_cell_coordinates!(x, grid, cell)
+@inline function getcoordinates(grid::AbstractGrid, idx::Int)
+    CT = get_coordinate_type(grid)
```
KnutAM, line 652: Any reason not to use `get_coordinate_type` that was introduced in #655 here?

PR #758, opened 2023-07-06
Fix neighborhood queries for topology

termi-official:
@AbdAlazezAhmed found that neighborhood queries are still broken sometimes. This PR increases test coverage and fixes the reported issues.


koehlerson:
what's the difference or purpose of vertex_star_stencils after this? isn't the query identical if I request for all vertices?

termi-official:
No the queries are completely different.

```julia
julia> using Ferrite

julia> grid = generate_grid(Quadrilateral,(2,2))
Grid{2, Quadrilateral, Float64} with 4 Quadrilateral cells and 9 nodes

julia> topology = ExclusiveTopology(grid)
ExclusiveTopology

  Vertex neighbors: (4, 4)
  Face neighbors: (4, 4)
  Edge neighbors: (4, 0)

julia> stencil = vertex_star_stencils(topology, grid)
Dict{Int64, Ferrite.EntityNeighborhood{VertexIndex}} with 9 entries:
  5 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((4, 1)), VertexI…
  4 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((3, 1)), VertexI…
  6 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((4, 2)), VertexI…
  7 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((3, 4)), VertexI…
  2 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((2, 1)), VertexI…
  9 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((4, 3)), VertexI…
  8 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((4, 4)), VertexI…
  3 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((2, 2)), VertexI…
  1 => EntityNeighborhood{VertexIndex}(VertexIndex[VertexIndex((1, 1)), VertexI…

julia> getstencil(stencil, grid, VertexIndex(1,3))
12-element Vector{VertexIndex}:
 VertexIndex((4, 1))
 VertexIndex((4, 2))
 VertexIndex((4, 4))
 VertexIndex((2, 4))
 VertexIndex((2, 1))
 VertexIndex((2, 3))
 VertexIndex((3, 2))
 VertexIndex((3, 1))
 VertexIndex((3, 3))
 VertexIndex((1, 3))
 VertexIndex((1, 2))
 VertexIndex((1, 4))

julia> getneighborhood(topology, grid, VertexIndex(1,3))
3-element Vector{VertexIndex}:
 VertexIndex((4, 1))
 VertexIndex((2, 4))
 VertexIndex((3, 2))
```

koehlerson:
Can we add a little bit more docs to the functions?


```
x-----------o-----------x
|           |           |
|           |           |
|     3     |     4     |
|           |           |
|           |           |
o-----------⊗-----------o
|           |           |
|           |           |
|     1     |     2     |
|           |           |
|           |           |
x-----------o-----------x
```
something like this with markers that explain the neighborhood and the stencil?

Wouldn't it be more useful to have the stencil as a unique list? So non duplicated nodes as a global version:

```
julia> getstencil(stencil, grid, VertexIndex(1,3))
12-element Vector{VertexIndex}:
  2
  4
  6
  8
```

termi-official:
> Can we add a little bit more docs to the functions?
> 
> ```
> x-----------o-----------x
> |           |           |
> |           |           |
> |     3     |     4     |
> |           |           |
> |           |           |
> o-----------⊗-----------o
> |           |           |
> |           |           |
> |     1     |     2     |
> |           |           |
> |           |           |
> x-----------o-----------x
> ```
> 
> something like this with markers that explain the neighborhood and the stencil?
> 

Sure, but I think this is a different matter from fixing the queries. This should be a follow up PR to write a topic guide+how to's.

>
> Wouldn't it be more useful to have the stencil as a unique list? So non duplicated nodes as a global version:
> 
> ```
> julia> getstencil(stencil, grid, VertexIndex(1,3))
> 4-element Vector{VertexIndex}:
>   2
>   4
>   6
>   8
> ```

Correct me if I am wrong, but for this purpose we have the `toglobal` function, right?

```julia
julia> Ferrite.toglobal(grid, getstencil(stencil, grid, VertexIndex(1,3)))
5-element Vector{Int64}:
 5
 6
 8
 2
 4
```

 But I would say that the stencil return is fine as it is, because we do not know a priori whether we need volumetric information (i.e. which elements are attached to the vertices). The separate `toglobal` call gives users more flexibility IMO, because they don't need to recompute inverse information, if they run into the scenarion they need it.

koehlerson:
If you need volumetric information you can always query them by the `vertex_to_cell` I just don't see a usecase where you want to iterate over it with duplication. If there is a use case then it's fine

termi-official:
I still see 2 problems with this:

1. The "global vertex index" is technically not well defined
2. There is no accessor for the dofs, given a "global vertex index"

Again, in this PR I am trying to solve the topological queries on lines and not making the interface better. Should we discuss interface design improvements separately?

koehlerson (approved):

termi-official:
Can you confirm it also fixes your issue @AbdAlazezAhmed  ? If so I think this is good to merge.

PR #767, opened 2023-07-18
Interface elements

DRollin:
This PR implements cells, interpolations and cell values to be used for interface elements.
The main idea is that an interface element consists of two embedded elements.
This way the functionalities for embedded elements can be reused.
Since it does not affect the core functionalities and is more like a self-contained extension, I added the code in a separate file.

Some things still need some work in the future:
- The interface should be consistent with what is being done for DG.
- The way of referring to the two side needs to be discussed. Suggestions: `:here`&`:there`; `:side1`&`:side2`; `true`&`false`
- A way of computing the normals of the interface will be needed for some applications and is not implemented yet.


src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+    there = CellValues(T, qr, ip.ip.there^vdim, gip.ip.there^sdim)
+    return InterfaceCellValues(ip, here, there)
+end
+
+reinit!(cv::InterfaceCellValues, cc::CellCache) = reinit!(cv, cc.coords)
+
+function reinit!(cv::InterfaceCellValues, x::AbstractVector{Vec{sdim,T}}) where {sdim, T}
+    nsplit = getngeobasefunctions(cv.here)
+    reinit!(cv.here, x[1:nsplit])
+    reinit!(cv.there, x[nsplit+1:end])
```
fredrikekre, line 334:
```suggestion
    reinit!(cv.here, @view(x[1:nsplit]))
    reinit!(cv.there, @view(x[nsplit+1:end]))
```

src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+    if s == :nodes
+        return (c.here.nodes..., c.there.nodes...) # Order used to reinit InterfaceCellValues
+    else
+        return getfield(c,s)
+    end
+end
+
+
+####################################################################################
+# Interoplation
```
fredrikekre, line 99:
```suggestion
# Interpolation
```

src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+####################################################################################
+
+
+# The constructors of `InterpolationInfo` require a `VectorizedInterpolation{InterfaceCellInterpolation}`
+# and not an `InterfaceCellInterpolation{VectorizedInterpolation,VectorizedInterpolation}`.
+# To create a `VectorizedInterpolation{InterfaceCellInterpolation}`, `InterfaceCellInterpolation` needs to be a `ScalarInterpolation`.
+
+"""
+    InterfaceCellInterpolation(here::ScalarInterpolation, there::ScalarInterpolation) <: ScalarInterpolation
```
fredrikekre, line 108: Maybe remove `Cell` here since I don't think we will have a `InterfaceInterpolation` that is something else. But I guess it matches `InterfaceCell` and `InterfaceCellValues`, so perhaps it should be kept as is.
```suggestion
    InterfaceInterpolation(here::ScalarInterpolation, there::ScalarInterpolation) <: ScalarInterpolation
```

src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+
+"""
+    shape_value_average(cv::InterfaceCellValues, qp::Int, i::Int)
+
+Return the value of shape function `i` evaluated in quadrature point `qp`
+for computing the average value on an interface.
+"""
+function shape_value_average(cv::InterfaceCellValues, qp::Int, i::Int)
+    side, baseindex = get_side_and_baseindex(cv, i)
+    return side == :here ? 0.5*shape_value(cv.here, qp, baseindex) : 0.5*shape_value(cv.there, qp, baseindex)
```
fredrikekre, line 415: I think it would be better to divide by 2 since otherwise it will promote to `Float64` (since `0.5` is a `Float64`):
```suggestion
    return side == :here ? shape_value(cv.here, qp, baseindex) / 2 : shape_value(cv.there, qp, baseindex) / 2
```

src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+
+"""
+    shape_gradient_average(cv::InterfaceCellValues, qp::Int, i::Int)
+
+Return the gradient of shape function `i` evaluated in quadrature point `qp`
+for computing the average gradient on an interface.
+"""
+function shape_gradient_average(cv::InterfaceCellValues, qp::Int, i::Int)
+    side, baseindex = get_side_and_baseindex(cv, i)
+    return side == :here ? 0.5*shape_gradient(cv.here, qp, baseindex) : 0.5*shape_gradient(cv.there, qp, baseindex)
```
fredrikekre, line 426:
```suggestion
    return side == :here ? shape_gradient(cv.here, qp, baseindex) / 2 : shape_gradient(cv.there, qp, baseindex) / 2
```

src/interface_elements.jl
DRollin: Just for my understanding:
This would avoid allocations, right?
(Or what would be the benefit of your suggestion?)

src/interface_elements.jl
DRollin: I would prefer being consistent with the other types and use `InterfaceCell...`
At least to me that would be easier to use.

src/interface_elements.jl
```diff
@@ -0,0 +1,449 @@
...
+    return  collect( getproperty(c, side).nodes[baseindex] for (side, baseindex) in get_sides_and_base_indices(c) )
+end
+
+function Base.getproperty(c::InterfaceCell, s::Symbol)
+    if s == :nodes
+        return (c.here.nodes..., c.there.nodes...) # Order used to reinit InterfaceCellValues
+    else
+        return getfield(c,s)
+    end
+end
```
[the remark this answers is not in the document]
DRollin: I wanted to avoid that the correct node indices have to be figured out at every `reinit!`-call and was a bit lazy with the solution.
To be consistent with the usual ordering, I changed this:
Now the required indices for sorting the coordinates are determined when creating the `InterfaceCellValues` and stored in vectors.
I am not sure if there is a better way of doing it, but this way, this is at least only done once.

src/interface_elements.jl
```diff
@@ -0,0 +1,456 @@
...
+
+get_sides_and_base_indices(::Type{Quadrilateral}, ::Type{Quadrilateral}) = ((:here,1), (:here,2), (:here,3), (:here,4), (:there,1), (:there,2), (:there,3), (:there,4))
+get_sides_and_base_indices(::Type{QuadraticQuadrilateral}, ::Type{Quadrilateral}) = ((:here,1), (:here,2), (:here,3), (:here,4), (:there,1), (:there,2), (:there,3), (:there,4), (:here,5), (:here,6), (:here,7), (:here,8), (:here,9))
+get_sides_and_base_indices(::Type{Quadrilateral}, ::Type{QuadraticQuadrilateral}) = ((:here,1), (:here,2), (:here,3), (:here,4), (:there,1), (:there,2), (:there,3), (:there,4), (:there,5), (:there,6), (:there,7), (:there,8), (:there,9))
+get_sides_and_base_indices(::Type{QuadraticQuadrilateral}, ::Type{QuadraticQuadrilateral}) = ((:here,1), (:here,2), (:here,3), (:here,4), (:there,1), (:there,2), (:there,3), (:there,4), (:here,5), (:here,6), (:here,7), (:here,8), (:there,5), (:there,6), (:there,7), (:there,8), (:here,9), (:there,9))
+
+function get_node_ids(c::InterfaceCell)
+    return  collect( getproperty(c, side).nodes[baseindex] for (side, baseindex) in get_sides_and_base_indices(c) )
```
lijas, line 86: Perhaps this would be non-allocating: (?)
```
side_index = get_side_and_base_indeces(c)
n = length(side_index)
return ntuple( i -> getproperty(c, side_index[1]).nodes[side_index[2]), n)
```

lijas (commented):
Nice work!

I guess functions like `function_jump_value` and `function_jump_gradient` could/should also be added?

src/interface_elements.jl
DRollin: Good point! :+1:

DRollin: `function_value_jump` etc. are now implemented.
I also fixed a bug related to evaluating gradients.

src/interface_elements.jl
```diff
@@ -0,0 +1,546 @@
...
+    return InterfaceCellInterpolation(ip.here.ip, ip.there.ip)^sdim
+end
+function default_geometric_interpolation(ip::VectorizedInterpolation{<:Any, <:Any, <:Any, <:InterfaceCellInterpolation})
+    return ip
+end
+
+getorder(ip::InterfaceCellInterpolation) = getorder(ip.here) == getorder(ip.there) ? getorder(ip.here) : (getorder(ip.here), getorder(ip.there))
+getorder(ip::VectorizedInterpolation{<:Any,<:Any,<:Any,<:InterfaceCellInterpolation}) = getorder(ip.ip)
+
+Base.:(^)(ip::InterfaceCellInterpolation, vdim::Int) = VectorizedInterpolation{vdim}(ip)
```
kimauth, line 248: I don't think this definition is needed since `InterfaceCellInterpolation <: ScalarInterpolation`

kimauth:
I can't distribute dofs on 2D interface elements:
```julia
dim = 2
grid = generate_grid(Quadrilateral, (3,1))

# Ferrite default grid:
# 5 ____ 6 ____ 7 ____ 8
# |      |      |      |
# |      |      |      |
# 1 ____ 2 ____ 3 ____ 4

# make middle cell zero-thickness:
grid.nodes[2:3] = [Node(mean(n->n.x, grid.nodes[2:3])) for _ in 1:2]
grid.nodes[6:7] = [Node(mean(n->n.x, grid.nodes[6:7])) for _ in 1:2]

# 1D-line elements for construction of InterfaceCell
line1 = Line((2,6))
line2 = Line((3,7))

cohesivecell = InterfaceCell(line1, line2)

cohesive_grid = Grid([grid.cells[1], grid.cells[3], cohesivecell], grid.nodes)

ip_quads = Lagrange{RefQuadrilateral, 1}()#^dim
ip_base = Lagrange{RefLine, 1}()
ip_cohesive = InterfaceCellInterpolation(ip_base)#^dim

dh = DofHandler(cohesive_grid)
sdh_quads = SubDofHandler(dh, Set(1:2))
add!(sdh_quads, :u, ip_quads)
sdh_coh = SubDofHandler(dh, Set(3))
add!(sdh_coh, :u, ip_cohesive)
close!(dh)
```

```julia
julia> close!(dh)                                                                                                                                                                                                                                       [3/1802]
ERROR: MethodError: no method matching facedof_interior_indices(::Lagrange{RefLine, 1, Nothing})                                                                                                                                                                
                                                                                                                                                                                                                                                                
Closest candidates are:                                                                                                                                                                                                                                         
  facedof_interior_indices(::InterfaceCellInterpolation)
   @ Ferrite ~/.julia/packages/Ferrite/9NWtq/src/interface_elements.jl:215
  facedof_interior_indices(::CrouzeixRaviart)
   @ Ferrite ~/.julia/packages/Ferrite/9NWtq/src/interpolations.jl:1296
  facedof_interior_indices(::Lagrange{RefQuadrilateral, 2})
   @ Ferrite ~/.julia/packages/Ferrite/9NWtq/src/interpolations.jl:530
```

src/interface_elements.jl
```diff
@@ -0,0 +1,546 @@
...
+end
+
+"""
+    function_gradient_jump(cv::InterfaceCellValues, qp::Int, u::AbstractVector)
+
+Compute the jump of the function gradient in a quadrature point.
+"""
+function function_gradient_jump(cv::InterfaceCellValues, qp::Int, u::AbstractVector)
+    return function_gradient(cv, qp, u, false) - function_gradient(cv, qp, u, true)
+end
```
termi-official, line 546:
```suggestion
end

```

test/test_interface_elements.jl
```diff
@@ -0,0 +1,390 @@
...
+            ) # The nodes have been chosen as the numbers representing their expected ordering
+        Chere = typeof(here)
+        Cthere = typeof(there)
+
+        @test InterfaceCell{sdim, rdim, shape, rshape, Chere, Cthere}(here, there) isa InterfaceCell{sdim, rdim, shape, rshape, Chere, Cthere}
+        @test InterfaceCell(here, there) isa InterfaceCell{sdim, rdim, shape, rshape, Chere, Cthere}
+        cell = InterfaceCell(here, there)
+
+        @test Ferrite.nvertices(cell) == Ferrite.nvertices(here) + Ferrite.nvertices(there)
+        @test Ferrite.nfaces(cell) == 2
```
termi-official, line 26: Where does this assumption come from?

src/interface_elements.jl
```diff
@@ -0,0 +1,546 @@
...
+
+"""
+    get_interface_index(ip::InterfaceCellInterpolation, side::Symbol, i::Integer)
+
+Return the base function index for an `InterfaceCellInterpolation` given a `side` (`:here` or `:there`)
+and the local base function index `i` on that face.
+"""
+function get_interface_index(ip::InterfaceCellInterpolation, side::Symbol, i::Integer)
+    nvhere, nvthere = get_n_dofs_on_side(vertexdof_indices, ip, :here), get_n_dofs_on_side(vertexdof_indices, ip, :there)
+    nfhere, nfthere = get_n_dofs_on_side(facedof_interior_indices, ip, :here), get_n_dofs_on_side(facedof_interior_indices, ip, :there)
```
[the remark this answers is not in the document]
kimauth: I think we should consider defining empty defaults such as
```julia
Ferrite.facedof_interior_indices(::Lagrange{RefLine,1}) = ((), ())
```
instead of introducing branches. This would just mean that a line does not have interior facedofs. Imo that's easier to understand than branches.

(Introducing this fixes the problem for 2d dof distribution reported above).

src/exports.jl
```diff
@@ -170,4 +170,18 @@ export
...
     PointIterator,
     PointLocation,
-    PointValues
+    PointValues,
+
+# Interface elements
+    InterfaceCell,
+    InterfaceCellInterpolation,
+    InterfaceCellValues,
+    get_side_and_baseindex,
```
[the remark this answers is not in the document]
DRollin: Except for `get_side_and_baseindex` definitely.
That function is something I planned to use in the future, where I need to evaluate values on both sides separately.
This is probably a quite specific case, but I thought that I might not be the only one using it like this at some point.

test/test_interface_elements.jl
DRollin: The idea is that the two sides of an interface are considered the faces of the cell.
To me it makes sense to think of it this way, but theses definitions can definitely be discussed.

src/interface_elements.jl
```diff
@@ -0,0 +1,546 @@
...
+    if isempty(basedofs)
+        return (Tuple{}(),)
+    else
+        return broadcast.(get_interface_index, ((ip,),), ((side,),), basedofs)
+    end
+end
+
+vertexdof_indices(ip::InterfaceCellInterpolation) = get_interface_dof_indices(vertexdof_indices, ip)
+
+function facedof_indices(ip::InterfaceCellInterpolation)
```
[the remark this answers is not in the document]
DRollin: I think of the two sides of the interface as faces.
In 2D that means the interface is like a Quadrilateral, but without two faces and no interior due to the zero thickness of the element.

DRollin:
Thank you all for your feedback!
After some discussions, we decided to create a separate package for the interface elements: [FerriteInterfaceElements.jl](https://github.com/Ferrite-FEM/FerriteInterfaceElements.jl)
Since it is an add-on and not a core functionality, we thought this a better way to do it.
Hence, I will close this PR.

PR #769, opened 2023-07-20
Various changes to FaceCache/InterfaceCache/InterfaceIterator

fredrikekre:
In this patch:
 - add a struct field `.dofs` to `FaceCache` that just aliases the underlying `.dofs` field of the `FaceCache`s `CellCache`,
 - fix `InterfaceIterator(::DofHandler)` when not passing the topology,
 - change `iterate(::InterfaceIterator)` to add a check that there is only a single neighboring element.

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -197,7 +197,7 @@ end
 function reinit!(cache::InterfaceCache, face_a::FaceIndex, face_b::FaceIndex)
     reinit!(cache.a, face_a)
     reinit!(cache.b, face_b)
-    resize!(cache.dofs, length(cache.a.dofs) + length(cache.b.dofs))
+    resize!(cache.dofs, length(celldofs(cache.a)) + length(celldofs(cache.b)))
     for (i, d) in pairs(cache.a.dofs)
         cache.dofs[i] = d
     end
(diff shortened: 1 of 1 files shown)


src/iterators.jl
```diff
@@ -193,12 +197,12 @@ end
 function reinit!(cache::InterfaceCache, face_a::FaceIndex, face_b::FaceIndex)
     reinit!(cache.a, face_a)
     reinit!(cache.b, face_b)
-    resize!(cache.dofs, length(cache.a.cc.dofs) + length(cache.b.cc.dofs))
-    for (i, d) in pairs(cache.a.cc.dofs)
+    resize!(cache.dofs, length(cache.a.dofs) + length(cache.b.dofs))
```
KnutAM, line 200: Why not
```suggestion
    resize!(cache.dofs, length(celldofs(cache.a)) + length(celldofs(cache.b)))
```
?

AbdAlazezAhmed (approved):
LGTM. Thanks!

src/iterators.jl
fredrikekre: Perhaps this should be called `facedofs` (like `interfacedofs`, although that one has a slightly different meaning which warrants another name).

PR #766, opened 2023-07-14
MatrixFree GPU operators

termi-official:
This PR shows how to implement matrix-free operators on the GPU.

## TODOs

* [ ] Fix bug when block size > 1
* [ ] Performance comparison
* [ ] GPUCellValues/MFCellValues or similar
* [ ] Move adapt stuff into extension
* [ ] Documentation entries
* [ ] Document example
* [ ] Check if AMDGPU.jl works, too

PR #779, opened 2023-07-30
Implement FaceQuadratureRule for RefPrism and RefPyramid 

### LINKED ISSUE #772: Implement FaceQuadratureRule for RefPrism and RefPyramid
Currently, `FaceQuadratureRule{RefPrism}(2)` yields `ERROR: FaceQuadratureRule for RefPrism not implemented`

AbdAlazezAhmed:
Closes #772 
Changes the transformations used in `create_face_quad_rule` to use the ones in FerriteViz as they're beneficial for optimizing #743 and IIUC make more sense in general as they respect the normals directions.
Also, the transformation is separated into its own functions such that they can be reused as in #743.
Edit: test coverage should improve with interface test, because it covers face->cell transformations


src/Grid/grid.jl
```diff
@@ -107,6 +108,9 @@ function faces(c::AbstractCell{RefLine})
     ns = get_node_ids(c)
     return ((ns[1],), (ns[2],)) # f1, f2
 end
+function faces(::Type{RefLine})
```
KnutAM, line 111: 1) Should this "new" function have the same function name when it returns the local node indices? 
2) Is it possible to use this function to create the existing `faces(::AbstractCell)` functions? Seems bad to duplicate the numbering if it can be avoided without performance losses...

src/Quadrature/quadrature.jl
```diff
@@ -209,21 +209,28 @@ end
...
     # Shift interval from (-1,1) to (0,1)
-    for i in eachindex(qr.weights, qr.points)
-        qr.weights[i] /= 2
-        qr.points[i] = (qr.points[i] + Vec{1,T}((1,))) / 2
-    end
-    return create_face_quad_rule(RefTriangle, qr.weights, qr.points)
+    # for i in eachindex(qr.weights, qr.points)
+    #     qr.weights[i] /= 2
+    #     qr.points[i] = (qr.points[i] + Vec{1,T}((1,))) / 2
+    # end
```
KnutAM, line 215: Perhaps a short note for each if/how the intervals are shifted? (And remove the old code)

src/interpolations.jl
```diff
@@ -489,6 +503,25 @@ function reference_coordinates(::Lagrange{RefLine,1})
             Vec{1, Float64}(( 1.0,))]
 end
 
+"""
+Mapping from 1D line to point.
+"""
+function transfer_point_cell_to_face(point::Union{Vec{1, T}, Vector{T}}, cell::Type{RefLine}, face::Int) where T
```
KnutAM, line 509: 1) Perhaps use regular comments, `#` above each specialization to avoid getting all these as docstrings. The general descriptions on the top should suffice IMO.
2) Is it necessary to support `Vector` as input? Do we use that?

test/test_facevalues.jl
```diff
@@ -92,18 +93,18 @@ for (scalar_interpol, quad_rule) in (
...
-        for fname in fieldnames(typeof(fv))
-            v = getfield(fv, fname)
-            v isa Ferrite.ScalarWrapper && continue
-            vc = getfield(fvc, fname)
-            if hasmethod(pointer, Tuple{typeof(v)})
-                @test pointer(v) != pointer(vc)
-            end
-            @test v == vc
-        end
+    #     # test copy
```
KnutAM, line 96: Re-enable?

test/test_interpolations.jl
```diff
@@ -46,7 +46,18 @@
 
         # Note that not every element formulation exists for every order and dimension.
         applicable(Ferrite.getlowerorder, interpolation) && @test typeof(Ferrite.getlowerorder(interpolation)) <: Interpolation{ref_shape,func_order-1}
-
+    @testset "transform face points" begin
+        ref_coord = Ferrite.reference_coordinates(Lagrange{ref_shape, 1}())
+        for face in 1:nfaces(interpolation)
+            face_nodes = Ferrite.faces(ref_shape)[face]
+            for node in face_nodes
+                coord = ref_coord[node]
```
KnutAM, line 54: Perhaps a "harder" test would be a linear combination of coordinates on the face, to avoid just checking exactly the reference coordinates?

KnutAM (commented):
Cool! 
I just took a quick look through and shared some thoughts (but couldn't look longer as my laptop is running out of battery:)

src/Grid/grid.jl
AbdAlazezAhmed: Thanks for the review! I think the name makes sense for the input, however, `reference_faces` might be better  given the input is a reference shape. What do you think?

src/Quadrature/quadrature.jl
AbdAlazezAhmed: I need to stop forgetting commented stuff :"D

src/interpolations.jl
AbdAlazezAhmed: The `Vector` was used in #743 IIRC, I'll check if it can be removed (and probably it can, I think I changed its calls in #743 but didn't remove `Vector`)

test/test_facevalues.jl
AbdAlazezAhmed: The copy test fails because of NaNs IIUS
```julia
    detJdV = fill(T(NaN), max_n_qpoints, n_faces)
```
How about changing the default to zeros?

src/FEValues/face_integrals.jl
```diff
@@ -149,3 +94,34 @@ function weighted_normal(J::Tensor{2,3}, ::Type{RefTetrahedron}, face::Int)
...
+
+function weighted_normal(J::Tensor{2,3}, ::Type{RefPyramid}, face::Int)
+    @inbounds begin
+        face == 1 && return J[:,2] × J[:,1]
+        face == 2 && return J[:,1] × J[:,3]
+        face == 3 && return J[:,3] × J[:,2]
+        face == 4 && return J[:,2] × (J[:,3]-J[:,1])
+        face == 5 && return (J[:,3]-J[:,2]) × J[:,1]
+    end
+    throw(ArgumentError("unknown face number: $face"))
```
AbdAlazezAhmed, line 125: btw it looks like these throws are redundant because there are outer checks IIUC. Should I remove them or add tests for them anyways? (IDK if user would access face weighted normals directly through this function instead of using facevalues)

test/test_facevalues.jl
KnutAM: The reason for filling with NaN's is to catch cases when it wasn't reinit!:ed. (For example if the user forgot)

Above that test case, the fv is reinited, right?
So it should not be filled with NaN, and also not the copied version

src/FEValues/face_integrals.jl
KnutAM: If you can show that there is a significant performance benefit by removing the error (could be), then that would be fine, but should probably be called `_unsafe_weighted_normal` or something (I don't think we do this at other places in Ferrite at the moment, although `shape_value(::Interpolation, ...)` could be a candidate, xref #766 )

But I would leave that for a separate PR

test/test_facevalues.jl
AbdAlazezAhmed: The thing is IIUC for pyramids and prisms not all faces has the same number of quadrature points. which makes some entries in the matrix NaNs even after init for the faces with nquadpoints <max_n_qpoints

KnutAM: Aha, I'm not sure what to do here.

One option would be to fill only those entries with zero on construction to make them behave equally, but let's ask @fredrikekre on this one...

src/FEValues/face_integrals.jl
AbdAlazezAhmed: I don't see performance benefit (or maybe I should use something other than `@benchmark` to benchmark operations that take about 1ns :"D)

Maybe I can add some tests for these error paths then to improve test coverage.

KnutAM: Yes, it is hard to benchmark due to const propagation I can imagine. But let's leave that for a separate issue.

src/interpolations.jl
```diff
@@ -489,6 +496,13 @@ function reference_coordinates(::Lagrange{RefLine,1})
             Vec{1, Float64}(( 1.0,))]
 end
 
+# Mapping from to 0D node to 1D line vertex.
+function transfer_point_face_to_cell(point::Vec{N, T}, cell::Type{RefLine}, face::Int) where {N, T}
```
fredrikekre, line 500: I think these methods belong in face_integrals.jl where the craete_face_quad_rule is

fredrikekre (commented):
`transfer_point_face_to_cell` could perhaps be called `reference_face_to_face` or something?

src/interpolations.jl
AbdAlazezAhmed: Thanks for the review!
Will move them and push after CI finishes (making sure no older version break, had some issues with 1.6 before :"D)

test/test_facevalues.jl
KnutAM: A third option could be to just change the test to allow, add a function in the tests along the lines of 
```julia
check_equal_or_nan(a::Number, b::Number) = a==b || (isnan(a) && isnan(b))
check_equal_or_nan(a::Vector, b::Vector) = all(check_equal_or_nan, a, b)
```
This would (1) do the proper testing and (2) give out NaN if one by mistake (somehow) tries to access the uninitialized values.

src/interpolations.jl
AbdAlazezAhmed: I'll push a commit for better errors/coverage after it passes locally, hopefully this fully covers the PR

src/FEValues/face_integrals.jl
```diff
@@ -1,113 +1,134 @@
...
-function weighted_normal(J::Tensor{2,2}, ::Type{RefQuadrilateral}, face::Int)
+function weighted_normal(J::Tensor{2,2}, cell_T::Type{RefQuadrilateral}, face::Int)
     @inbounds begin
         face == 1 && return Vec{2}(( J[2,1], -J[1,1]))
         face == 2 && return Vec{2}(( J[2,2], -J[1,2]))
         face == 3 && return Vec{2}((-J[2,1],  J[1,1]))
         face == 4 && return Vec{2}((-J[2,2],  J[1,2]))
     end
-    throw(ArgumentError("unknown face number: $face"))
+    throw(ArgumentError("unknown face number: $face for a cell of type $(cell_T)"))
```
termi-official, line 90: Can we do this without interpolation somehow? This kind of stuff breaks GPU compilation.

src/FEValues/face_integrals.jl
```diff
@@ -1,113 +1,134 @@
+"""
+    reference_face_to_face(point::Vec{N, T}, cell_T::Type{<:AbstractRefShape}, face::Int) 
+
+Transform point from face's reference (N-1)D coordinates to ND coordinates on the cell's face.
+"""
+reference_face_to_face
```
termi-official, line 6: I like the idea! Can we come up with a better name for the routine maybe?

src/Grid/grid.jl
```diff
@@ -68,6 +69,16 @@ Note that the vertices are sufficient to define an edge uniquely.
 """
 edges(::AbstractCell)
 
+"""
+    reference_faces(::AbstractRefShape)
+
+Returns a tuple of n-tuples containing the ordered local node indices corresponding to
+the vertices that define an *oriented face*.
```
termi-official, line 76: Can you give a reference to oriented face in the manual? Users might not be familiar with the concept.

src/Grid/grid.jl
```diff
@@ -210,14 +215,17 @@ function edges(c::AbstractCell{RefPyramid})
...
-        (ns[3], ns[5], ns[4]),                                      
+        (1, 3, 4, 2), (1, 2, 5), 
+        (1, 5, 3), (2, 4, 5), 
+        (3, 5, 4),                                      
     )
 end
+function faces(c::AbstractCell{refshape}) where refshape
+    ns = get_node_ids(c)
+    return ntuple(i->getindex.(Ref(ns), reference_faces(refshape)[i]), nfaces(refshape))
+end
```
termi-official, line 229: Is there a performance hit to this operation? If so, how severe and can we optimize it?

test/test_facevalues.jl
```diff
@@ -95,7 +96,7 @@ for (scalar_interpol, quad_rule) in (
         # test copy
         fvc = copy(fv)
         @test typeof(fv) == typeof(fvc)
-        for fname in fieldnames(typeof(fv))
+        Ferrite.getrefshape(func_interpol) ∈ (RefPrism, RefPyramid)|| for fname in fieldnames(typeof(fv))
```
termi-official, line 99: What is the problem here?

test/test_facevalues.jl
```diff
@@ -38,8 +40,7 @@ for (scalar_interpol, quad_rule) in (
                 u_scal[i] = V ⋅ xs[i]
             end
             u_vector = reinterpret(Float64, u)
-
-            for i in 1:length(getnquadpoints(fv))
+            for i in 1:getnquadpoints(fv)
```
termi-official, line 43: Good catch!

src/FEValues/face_integrals.jl
AbdAlazezAhmed: Thanks!
Will think about a better one.

src/FEValues/face_integrals.jl
AbdAlazezAhmed: Does throwing break GPU stuff or having a variable in the error?
I noticed unsafe stuff in #766 if that's what you mean. Also, @KnutAM suggested something like `_unsafe_weighted_normal` if it has performance benefits but I couldn't measure the benefit :smile:

test/test_facevalues.jl
AbdAlazezAhmed: Thanks!

src/FEValues/face_integrals.jl
fredrikekre: Knowing what the input face was isn't auper useful, so perhaps it can be removed. Or you can keep it as is until  this actually runs on GPU.

test/test_facevalues.jl
AbdAlazezAhmed: It's that the copy test fails for matrices with NaN entries, and for cells with faces that have different number of dofs/quadpoints we get NaN entries for the extra entries for the faces with ndofs/nqp < max ndofs/nqp

src/Grid/grid.jl
AbdAlazezAhmed: I don't think there's much of a performance hit
The PR:
```julia
julia> @benchmark Ferrite.faces($cell)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  2.945 ns … 31.522 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.956 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.994 ns ±  0.791 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

         ▁  ▂ ▂  ▃ ▃  ▅ ▆  █ █  ▇ ▇  ▇ ▇  ▄ ▁                 
  ▂▁▃▁▁▅▁█▁▁█▁█▁▁█▁█▁▁█▁█▁▁█▁█▁▁█▁█▁▁█▁█▁▁█▁█▁▁▇▁▇▁▁▆▁▆▁▁▅▁▂ ▄
  2.94 ns        Histogram: frequency by time        2.97 ns <

 ```
 Master:
 ```julia
 julia> @benchmark Ferrite.faces($cell)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  2.949 ns … 30.682 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.024 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.187 ns ±  1.418 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▇                                                         
  ▅██▆▆▄▄▄▃▃▆▄▃▇▇▄▄▄▃▃▂▂▂▂▃▃▃▃▃▂▂▂▂▂▂▂▂▂▃▂▁▁▂▁▂▂▁▁▁▁▁▁▁▁▁▂▂▃ ▃
  2.95 ns        Histogram: frequency by time        3.69 ns <

 ```

File: test/test_facevalues.jl (hunk 12 above)
Comment (termi-official) [reply test/test_facevalues.jl:99]: We should be able to fix this by introducing some "matrix"-like struct with varying row/col size. I.e. something like
```julia
struct VaryingMatrix{TOV, TDV}
   offsets::TIV
   data::TDV
end
```
where the access with two indices first looks up the offset in the offset vector and the second index computes the actual element in the data vector via data[offset_index[i]+j], where the offsets start at 0. We should then start using this all over the codebase. Does this make sense? cc @fredrikekre separate PR maybe?

AbdAlazezAhmed: I was thinking of using a vector of matrices instead of a 3-D array `Array{T, 3}` but I'm not sure what the performance impact would be due to changing memory layout

termi-official: Yea, we should try to keep the memory continuous here, because there is potentially quite a big performance hit, depending on the system.

src/FEValues/face_integrals.jl
```diff
@@ -1,113 +1,134 @@
...
+    face_quad_rule = QuadratureRule{RefShape, T, getdim(AbstractCell{cell_T})}[]
+    for face in 1:nfaces(cell_T)
+        new_points = [reference_face_to_face(N != 0 ? p[i] : Vec(zero(T)), cell_T, face) for i in 1:n_points] # ξ = 1-t-s, η = s, ζ = 0
+        push!(face_quad_rule, QuadratureRule{RefShape, T}(w, new_points))    
+    end
+    return FaceQuadratureRule(face_quad_rule)
+end
+
+# For cells with mixed faces
+function create_face_quad_rule(cell_T::Type{RefShape}, quad_faces::Vector{Int}, w_quad::Vector{T}, p_quad::Vector{Vec{N, T}}, tri_faces::Vector{Int}, w_tri::Vector{T}, p_tri::Vector{Vec{N, T}}) where {N, T, RefShape <: AbstractRefShape}
```
fredrikekre, line 37: This is only applicable for prism and pyramid I think?
```suggestion
function create_face_quad_rule(cell_T::Type{RefShape}, quad_faces::Vector{Int}, w_quad::Vector{T}, p_quad::Vector{Vec{N, T}}, tri_faces::Vector{Int}, w_tri::Vector{T}, p_tri::Vector{Vec{N, T}}) where {N, T, RefShape <: Union{RefPrism, RefPyramic}}
```

docs/src/devdocs/elements.md
```diff
@@ -11,6 +11,7 @@ of faces `M`.
 ```@docs
 Ferrite.vertices(::Ferrite.AbstractCell)
 Ferrite.edges(::Ferrite.AbstractCell)
+Ferrite.reference_faces(::Ferrite.AbstractRefShape)
```
fredrikekre, line 14: I don't see where this is used in this PR?

AbdAlazezAhmed: I checked it. It's not used directly but it's used to get `nfaces` for reference shapes which is used in `create_face_quad_rule`. It'll be used in the other PR in a more direct way, but I think it's necessary here too

src/FEValues/face_integrals.jl
```diff
@@ -1,9 +1,9 @@
 """
-    reference_face_to_face(point::Vec{N, T}, cell_T::Type{<:AbstractRefShape}, face::Int) 
+    element_face_transformation(point::Vec{N, T}, cell_T::Type{<:AbstractRefShape}, face::Int) 
 
 Transform point from face's reference (N-1)D coordinates to ND coordinates on the cell's face.
 """
-reference_face_to_face
+element_face_transformation
```
termi-official, line 6: Sorry, I got this backwards. Reading the docstring I think it should be rather `face_to_element_transformation`.

AbdAlazezAhmed: Nah It's my bad I missed it :D
Will rename once back to branch

src/FEValues/face_integrals.jl
```diff
@@ -1,113 +1,134 @@
...
+    create_face_quad_rule(cell_T::Type{RefShape}, quad_faces::Vector{Int}, w_quad::Vector{T}, p_quad::Vector{Vec{N, T}}, tri_faces::Vector{Int}, w_tri::Vector{T}, p_tri::Vector{Vec{N, T}}) 
+
+Creates ["FaceQuadratureRule"](@ref) with the given cell type, weights and points. If the cell has faces of different shapes
+(i.e. quadrilaterals and triangles) then each shape's faces indices, weights and points are passed separately.
+"""
+function create_face_quad_rule(cell_T::Type{RefShape}, w::Vector{T}, p::Vector{Vec{N, T}}) where {N, T, RefShape <: AbstractRefShape}
+    n_points = length(w)
+    face_quad_rule = QuadratureRule{RefShape, T, getdim(AbstractCell{cell_T})}[]
+    for face in 1:nfaces(cell_T)
+        new_points = [face_to_element_transformation(N != 0 ? p[i] : Vec(zero(T)), cell_T, face) for i in 1:n_points] # ξ = 1-t-s, η = s, ζ = 0
```
fredrikekre, line 30:
```suggestion
function create_face_quad_rule(::Type{RefShape}, w::Vector{T}, p::Vector{Vec{N, T}}) where {N, T, RefShape <: AbstractRefShape}
    n_points = length(w)
    face_quad_rule = QuadratureRule{RefShape, T, getdim(AbstractCell{RefShape})}[]
    for face in 1:nfaces(RefShape)
        new_points = [face_to_element_transformation(N != 0 ? p[i] : Vec(zero(T)), RefShape, face) for i in 1:n_points] # ξ = 1-t-s, η = s, ζ = 0
```
I think the compiler prefer this pattern.

Also, the comment at the end of the line seems outdated?

src/FEValues/face_integrals.jl
```diff
@@ -1,113 +1,134 @@
...
+function create_face_quad_rule(cell_T::Type{RefShape}, quad_faces::Vector{Int}, w_quad::Vector{T}, p_quad::Vector{Vec{N, T}}, tri_faces::Vector{Int}, w_tri::Vector{T}, p_tri::Vector{Vec{N, T}}) where {N, T, RefShape <: Union{RefPrism, RefPyramid}}
+    n_points_quad = length(w_quad)
+    n_points_tri = length(w_tri)
+    face_quad_rule = Array{QuadratureRule{RefShape, T, getdim(AbstractCell{cell_T})}}(undef, nfaces(cell_T))
+    for face in quad_faces
+        new_points = [face_to_element_transformation(N != 0 ? p_quad[i] : Vec(zero(T)), cell_T, face) for i in 1:n_points_quad]
+        face_quad_rule[face] = QuadratureRule{RefShape, T}(w_quad, new_points)
+    end
+    for face in tri_faces
+        new_points = [face_to_element_transformation(N != 0 ? p_tri[i] : T[], cell_T, face) for i in 1:n_points_tri]
```
fredrikekre, line 46:
```suggestion
function create_face_quad_rule(::Type{RefShape}, quad_faces::Vector{Int}, w_quad::Vector{T}, p_quad::Vector{Vec{N, T}}, tri_faces::Vector{Int}, w_tri::Vector{T}, p_tri::Vector{Vec{N, T}}) where {N, T, RefShape <: Union{RefPrism, RefPyramid}}
    n_points_quad = length(w_quad)
    n_points_tri = length(w_tri)
    face_quad_rule = Array{QuadratureRule{RefShape, T, getdim(AbstractCell{RefShape})}}(undef, nfaces(RefShape))
    for face in quad_faces
        new_points = [face_to_element_transformation(N != 0 ? p_quad[i] : Vec(zero(T)), RefShape, face) for i in 1:n_points_quad]
        face_quad_rule[face] = QuadratureRule{RefShape, T}(w_quad, new_points)
    end
    for face in tri_faces
        new_points = [face_to_element_transformation(N != 0 ? p_tri[i] : T[], RefShape, face) for i in 1:n_points_tri]
```

test/test_interpolations.jl
```diff
@@ -46,7 +46,7 @@
 
         # Note that not every element formulation exists for every order and dimension.
         applicable(Ferrite.getlowerorder, interpolation) && @test typeof(Ferrite.getlowerorder(interpolation)) <: Interpolation{ref_shape,func_order-1}
-
+
```
fredrikekre, line 49:
```suggestion

```

src/FEValues/face_integrals.jl
AbdAlazezAhmed: You mean
```julia
# For cells with mixed faces
```
?
I think it still works because the method underneath is for prisms and pyramids

fredrikekre: I meant `ξ = 1-t-s, η = s, ζ = 0`

AbdAlazezAhmed: Ah, you're right, my bad.

lijas:
Nice work on this PR. Before this PR, the spatial coordinate evaluated on "cell-level" was not always the same as the coordinate evaluated on "face level" In other words, we just took the parametric coordinate on e.g. on a RefLine (say \xi=[0.123]) and transformed in to the Quadrilateral (say to local faceid 4 \xi = [-1, 0.123), however, this would not always correspond to the same spatial coordiante due to the permutation of the face-nodes. This PR seems to fix that.

Is it worth adding test for this (I can do it in that case)?

Somethink like:
```
using Ferrite

grid = generate_grid(Ferrite.Pyramid, (3,3,3))
dim = Ferrite.getdim(grid)

cellid = 3
cell = grid.cells[cellid]
ccoords = getcoordinates(grid, cellid)
Vec_t = Vec{3,Float64}
Vec_face_t = Vec{2,Float64}

lfaceid = 2
facenodes = Ferrite.faces(cell)[lfaceid]
fcoords = zeros(Vec_t, length(facenodes))
for (i,nodeid) in enumerate(facenodes)
    x = grid.nodes[nodeid].x
    fcoords[i] = x
end

refshape =  Ferrite.getrefshape(cell)
ipcell = Lagrange{refshape,1}()
ipface = Lagrange{RefTriangle,1}()

ξface = rand(Vec_face_t)/4
ξcell = Ferrite.face_to_element_transformation(ξface, refshape, lfaceid)

xface = zero(Vec_t)
for i in 1:getnbasefunctions(ipface)
    xface += Ferrite.shape_value(ipface, ξface, i) * fcoords[i]
end

xcell = zero(Vec_t)
for i in 1:getnbasefunctions(ipcell)
    xcell += shape_value(ipcell, ξcell, i) * ccoords[i]
end

@test xcell == xface
```

AbdAlazezAhmed:
Thanks @lijas ! I'll work on adding the test :smile:

PR #785, opened 2023-08-12
Change type of PointIterator in PointEvalHandler.jl

suurj:
Change the type of the coordinate vector to a more general type in PointIterator. The old type works only if the spatial dimension of the coordinates is 2.

### DIFF
--- a/src/PointEval/PointEvalHandler.jl
+++ b/src/PointEval/PointEvalHandler.jl
@@ -297,7 +297,7 @@ end
 """
 PointIterator
 
-struct PointIterator{PH<:PointEvalHandler, V <: Vector{<: Vec}}
+struct PointIterator{PH<:PointEvalHandler, V <: Vec}
     ph::PH
     coords::V
 end
--- a/src/PointEval/PointEvalHandler.jl
+++ b/src/PointEval/PointEvalHandler.jl
@@ -299,7 +299,7 @@ PointIterator
 
 struct PointIterator{PH<:PointEvalHandler, V <: Vec}
     ph::PH
-    coords::V
+    coords::Vector{V}
 end
 
 function PointIterator(ph::PointEvalHandler{G}) where {D,C,T,G<:Grid{D,C,T}}
--- a/src/PointEval/PointEvalHandler.jl
+++ b/src/PointEval/PointEvalHandler.jl
@@ -297,9 +297,9 @@ end
 """
 PointIterator
 
-struct PointIterator{PH<:PointEvalHandler}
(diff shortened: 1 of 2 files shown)


src/PointEval/PointEvalHandler.jl
```diff
@@ -297,9 +297,9 @@ end
 """
 PointIterator
 
-struct PointIterator{PH<:PointEvalHandler}
+struct PointIterator{PH<:PointEvalHandler, V <: Vector{<: Vec}}
```
fredrikekre, line 300:
```suggestion
struct PointIterator{PH<:PointEvalHandler, V <: Vec}
```

src/PointEval/PointEvalHandler.jl
```diff
@@ -297,9 +297,9 @@ end
...
     ph::PH
-    coords::Vector{Vec{2,Float64}}
+    coords::V
```
fredrikekre, line 302:
```suggestion
    coords::Vector{V}
```

fredrikekre (commented):
Thanks! Thats a silly bug left over from testing most likely. Would you be able to add a simple test case for this in the `test/test_pointevaluation.jl` file?

suurj:
> Thanks! Thats a silly bug left over from testing most likely. Would you be able to add a simple test case for this in the `test/test_pointevaluation.jl` file?

I changed the `PointIterator` test in `test/test_pointevaluation.jl` from two-dimensional to three-dimensional. It seems to work fine.

test/test_pointevaluation.jl
```diff
@@ -129,10 +129,10 @@ end
 function dofhandler2()
     # Computes the L2 projection of a quadratic field exactly
     # but not using L2Projector since we want the DofHandler dofs
-    mesh = generate_grid(Quadrilateral, (20, 20))
-    ip_f = Lagrange{RefQuadrilateral,2}()
-    ip_f_v = ip_f^2
-    qr = QuadratureRule{RefQuadrilateral}(3)
+    mesh = generate_grid(Hexahedron, (10, 10, 10))
```
fredrikekre, line 132: We should still test 2D as well. Perhaps you can make the dimension an input variable to the function (`dofhandler2`) and call it once for 2D and once for 3D?

suurj: Thanks for the suggestion. I added a boolean parameter into `dofhandler2` to test both cases.

fredrikekre:
Thanks for the fix, and welcome as a contributor!

PR #786, opened 2023-08-12
use replace! instead of aliased map!

KnutAM:
I learned on [Slack](https://julialang.slack.com/archives/C6A044SQH/p1691785439327919?thread_ts=1691784875.039309&cid=C6A044SQH) yesterday that `map!(f, x, x)` isn't guaranteed safe due to aliasing, and just happens to work at the moment. 

My quick benchmarks show no measurable difference in performance.


fredrikekre:
Seems ridiculous that you can't use `map!` here...

KnutAM:
Agreed! I think it at least should be warned about if not safe. 
However, I would have preferred the syntax `map!(f, x)` (I find it strange that `map!(f, values(dict::AbstractDict))` is supported but not `map!(f, x::Vector)`).

PR #780, opened 2023-07-31
p4est baseline

koehlerson:
to be discussed at FerriteCon 23, register now: https://ferrite-fem.github.io/FerriteCon/

## TODO
- [x] p4est types
    - [x] `OctantBWG` morton index encoded octant  in 2 and 3D
    - [x] `OctreeBWG` linear tree with homogeneous `OctantBWG` leaves
    - [x] `ForestBWG` forest of homogeneous `OctreeBWG`
- [x] Octant Operations
    - [x] construct children
    - [x] construct parents
    - [x] intraoctree operation
        - [x] face neighbor
        - [x] edge neighbor
        - [x] corner neighbor
    - [x] interoctree operation
        - [x] `transform_face`
        - [x] `transform_corner`
        - [x] `transform_edge`  #902 
 - [x] Octree operations
     - [x]  balance intertree
     - [x] balance intratree
     - [x] coarsen
     - [x] refine
- [x] materializing a `Ferrite.Grid` for unstructured meshes
    - [x] 2D 
    - [x] 3D still need edge neighborhood branch #902
- [x] hanging nodes
    - [x] use affine constraints 
    - [x] passing to constraint handler #900
- [x] Docs
    - [x] Docs to the p4est types and how they interplay
    - [x] Docs to the yet to be defined interface
    - [x] quick and dirty implementation of elasticity `docs/src/literate-tutorials/adaptivity.jl` with (almost?maybe?) working but slow ZZ error estimator
    - [x] manufactured solution example for heat equation at `docs/src/literate-tutorials/heat-adaptivity.jl`
    - [x] adjust quick and dirty `docs/src/literate-tutorials/adaptivity.jl` to a proper tutorial
- [x] More debug statements to track down issues
- [x] multiple neighbors attached to edge and vertex
- [x] rotation test in 3D

## Followup PRs

- DofHandler integration
- Transfer of solution vector between dof handlers
- make it work for arbitrary polynomial order
- refine_and_coarsen! function
- error estimation interface
    - faster local ZZ error estimator
    - Kelly error estimator
    - generic Dörfler marking interface https://www.ams.org/journals/mcom/2020-89-326/S0025-5718-2020-03553-4/S0025-5718-2020-03553-4.pdf
- iterators 
    -  cell
    - face
    - edge
    - vertex
- conformization

PR #789, opened 2023-08-14
Suggestion to change faces(::Quad) to edges(::Quad)

lijas:
The main reasons for doing this is :

- I think the code base becomes nicer, see e.g. the dofhandler, topology, InterpolationInfo. Basically we can remove most `if dim== 2`  etc. We can also remove the "hacks" we have for embedded elements (shells, beams). In other words we will have better integration of embedded elements.

Another indication that this might be the right change is that many other FE-related packages uses this definition, e.g Finics, Gmsh, [defelements](https://defelement.com/ciarlet.html) (also deal.ii i think)

The drawback is:
 - Changes to the api. FaceIndex will not be "dimension independent" anymore. For 2d-problems, one would need to use e.g. `getedgeset()` instead of `getfaceset()` (actually, this is not strictly necessary because the we can keep backwards compatibility by defining: 
   `getfacset(grid::Grid{2}, name::String) = grid.edgesets[name]` .... however, it might be a bit strange that `getfaceset` returns an edgeset (for 2d-problems)).
   One possible way of keeping a "dimension independent" way of accessing boundary-sets (facesets/edgeset) is to introduce "facets", for example `getfacetset, addfacetset`, with the following meaning:
```
getfacetsets(grid::AbstractGrid{1}) = grid.vertexsets
getfacetsets(grid::AbstractGrid{2}) = grid.edgesets
getfacetsets(grid::AbstractGrid{3}) = grid.facesets
```

To be clear, this is only a suggestion. Please come with feedback :)


termi-official:
Thanks for the suggestion and working on this. A few comments on this.

[Deal.ii uses faces for lines in 2D](https://github.com/NiMlr/pyfunctionbases#where-evaluation-of-polynomials-can-fail). MFEM for example does not have a consistent definition for this. I also asked a bit around and the response whether people would call the boundary of a 2D element an edge or face is actually pretty split. 

Saying an edge is a topological 1D entity further raises the question whether the cell for a line element is also an edge, and e.g.  how do we handle this edge case in the code.

Regarding the removal of the internal hacks I thought about having an dimension-independent layer in the Ferrite core, which is independent of the actual definition of face/edge/... .

KnutAM:
Would be really nice if this could simplify the internals! 

IMO any such change must be done without removing the possibility of straight-forward writing dimension-independent code. Would this also turn `FaceValues` into `FacetValues`?

lijas:
@termi-official 
I can imagaine that there are many ways to implement it and there is no right ansewr!.

> Saying an edge is a topological 1D entity further raises the question whether the cell for a line element is also an edge,

I think this is true, a line element would also be an edge, but what would be the problem with this? What edge-cases (pun intended? :P ) needs to be handled.

@KnutAM 
I agree, it should still be possible  to write dimension independent code. And I think introducing the definition of facet would make this possible.

Regarding FaceValues, I think it would be nice to introduce FacetValues which returns FaceValues for 3d-surfaces and EdgeValues for Lines.

koehlerson:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/780 I have there a lot of `dim==3` code branches because I often have to look at the full surrounding of a cell. Could this be beneficial in such a case?

src/Grid/topology.jl
```diff
@@ -70,9 +70,14 @@ mutable struct ExclusiveTopology <: AbstractTopology
     edge_edge_neighbor::Matrix{EntityNeighborhood{EdgeIndex}}
     # lazy constructed face topology
     face_skeleton::Union{Vector{FaceIndex}, Nothing}
+    edge_skeleton::Union{Vector{EdgeIndex}, Nothing}
```
koehlerson, line 73: This seems only valid for 2D and should rather be `facet_skeleton` ?

lijas:
@KnutAM Those are my concerns too.

Regarding FacetValues... I think you are right, we might need to introduce a new type FacetValues, because EdgeValues typically does not store any normals, which is still something you might want for a 2d problem even though the boundary is technically an edge.

src/Grid/topology.jl
lijas: Hmm, I dont think so. My idea is that you generate the edge_skeleton, face_skeleton etc the same for both a 2d and 3d grid (because topologicaly they are definied in the same way now), and then if you want to loop over the dim-1 entity, you would dispatch on the dim-parameter:

```
function facetskeleton(top::ExclusiveTopology, grid::Grid{2})
    return top.edge_skeleton
end
function facetskeleton(top::ExclusiveTopology, grid::Grid{3})
    return top.face_skeleton
end
```

koehlerson: I think I'm wondering in general about the dispatch. If I'm not mistaken, then you define a `facet` as a dim-1 entity and you can store directly the `facetskeleton` without having the extra field in the struct

```
function facetskeleton(top::ExclusiveTopology, grid)
    return top.facetskeleton
end
```

Or is there some advantage of having both fields?

lijas: But what if you have a hex mesh? Then you will have both edge_skeleton and face_skeleton, and you will have a field called facet_skeleton which is identical to face_skeleton? 

I realise now though that the dispatch approach might not work for shells in 3d... the dispatch for facetskeleton(..., grid{3}) would return the face_skeleton, even though you probably would want egde_skeleton...

koehlerson: Ah okay! I think as of now, there is no construction of edge skeleton in 3D cases planned. Not sure if they are useful in some cases, but if we want to include them, then this makes sense

fredrikekre (approved):
Fix the things we discussed.

termi-official:
Thanks for the work on solving the issue here!

> @termi-official and @KnutAM, can you make a review of this PR? I know you have thought about this problem, so I would like your input here. I think that I have convinced Fredrik and Kim that this is a good solution, but I imagine you have a different approach in mind . 

Sure I will take a look. I do not think that I have a different solution per-se, but I share similar concerns as Kim about the entry barrier to Ferrite with these changes (and I mean with all possible solutions, not specific to this PR).

> 
> To reiterate the main idea of this PR. We keep the names vertices, edges and faces (e.g in `VertexIndex`, `EdgeIndex` and `FaceIndex`), but we make the definitions of these entities consistent for all cells (and interpolations):
> 
I already discussed this with Fredrik and want to bring this also up here: Now FaceIndex and EdgeIndex is a bit ambigous. Maybe we should introduce FacetIndex and RidgeIndex, because this is actually what these indices refer to on master.

>     * Vertex refers to points (nothing changes here)
I would rather reserve vertex for corners and nodes for points in general. I think this is a bit more consistent with our intent. We really need to clarify this in the docs, because I think this is already on master super confusing for users. 

> 
>     * edges always refers to lines between two vertices
> 
I think we all can agree on this definition.

>     * faces are surfaces with normals
>       (These are not formal definitions, but I think you get the point).
> 
I would drop the normal part here, because defining what the normal (field) is can be troublesome. E.g. what is the normal of a face in 4D or 2D?

> This change simplifies the code-base quite a lot at some places, especially in the dof-distribution, and solves most(all?) of the consistency issues we have
This would definitely solve the dof distribution consistency issue for the current set of interpolations, but not all.

> 
> The major drawback is now that users can no longer call e.g. `getfaceset(grid, "right")` in 2d problems anymore (`sdim=2`), since the boundary of the domain is now an edgeset. To account for this, we also introduce the concept of `facet`s, e.g. `getfacetset`, which changes meaning depening on the reference dim of the element:
> 
> ```
>  facets(c::AbstractCell{<:AbstractRefShape{1}}) = vertices(c)
>  facets(c::AbstractCell{<:AbstractRefShape{2}}) = edges(c)
>  facets(c::AbstractCell{<:AbstractRefShape{3}}) = faces(c)
> ```
> 
> Similar dispatches are used through out the code, e.g for `dirichlet_facetdof_indices`, `facetdof_indices` ...

For me personally this is also fine. But we should reflect this somewhere in the docs very carefully and visible for beginners. No idea how tho.

src/iterators.jl
```diff
@@ -325,6 +325,46 @@ end
...
+
+@inline _getcache(fi::VertexIterator) = fi.fc
+@inline _getset(fi::VertexIterator) = fi.set
+
+FaceIterator(gridordh::Union{Grid,AbstractDofHandler}, set::Set{EdgeIndex}, flags::UpdateFlags=UpdateFlags()) = 
+    EdgeIterator(gridordh, set, flags)
+
+FaceIterator(gridordh::Union{Grid,AbstractDofHandler}, set::Set{VertexIndex}, flags::UpdateFlags=UpdateFlags()) = 
+    VertexIterator(gridordh, set, flags)
+
```
KnutAM, line 367: Cannot a `FacetIterator` replace the current `FaceIterator`?

KnutAM (commented):
Thanks for starting this huge effort @lijas!
Some initial comments/thoughts

~My comments are based on my understanding that co-dimension is relative the spatial dimension (as opposed to the reference (element) dimension), so with my understanding, for `RefLine` embedded in 2d, the `edge` is the `facet` and the `point` is the `ridge`, but in a non-embedded case (1d), the `edge` is the `cell` and the `point` is the `facet`. Does this match your view?~
Edit: Updated comments after I understood that we use the codim relative the topological/reference shape dimension. 

I would have expected that for the grid we work with co-dimensional entities (basically as we were before, except we had the wrong names), i.e. we should just rename
```julia
FaceIndex -> FacetIndex
EdgeIndex -> RidgeIndex
VertexIndex -> PeakIndex
```
instead of changing depending on the `sdim` in the grid. 

As @termi-official mentions, I think we should have `FaceValues` -> `FacetValues` and `FaceQuadratureRule` to `FacetQuadratureRule`. AFAIU, from the user perspective, one would for most cases only need to care about these names (cell, facet, ridge, peak). 

Changing to using the dimensional entities (volume, face, edge, point) for interpolations, reference shapes, and dof distribution looks like it works great!

fredrikekre:
> My comments are based on my understanding that co-dimension is relative the spatial dimension (as opposed to the reference (element) dimension), so with my understanding, for RefLine embedded in 2d, the edge is the facet and the point is the ridge, but in a non-embedded case (1d), the edge is the cell and the point is the facet. Does this match your view?

Elias and I discussed this based on https://defelement.com/ciarlet.html#Cell+subentities where they explicitly define the names based on codimension of the reference element (topological dimension) and not the spatial dimension (which they refer to as geometric dimension).

So if we have 3 spatial dimension with hex and (embedded) quads we have 8 facets of the hex which are faces and we have 4 facets of the quad which are edges.

src/Grid/grid.jl
```diff
@@ -466,6 +523,59 @@ Returns all vertex sets of the grid.
...
+getfacetset(grid::AbstractGrid{sdim}, setname::String) where sdim = getfacetsets(grid)[setname]
+
+"""
+    getfacetsets(grid::AbstractGrid)
+
+Returns all facet sets of the grid.
+"""
+getfacetsets(grid::AbstractGrid{1}) = grid.vertexsets
+getfacetsets(grid::AbstractGrid{2}) = grid.edgesets
+getfacetsets(grid::AbstractGrid{3}) = grid.facesets
```
KnutAM, line 541: Since facet refers to `rdim-1` this isn't always true then, right?
So it would be better to actually just store the `peaksets`, `ridgesets`, and `facetsets` directly?

KnutAM:
> Vertex refers to points (nothing changes here)

Shouldn't this be renamed to "point" following the table, and "vertex" refers to the zero-dimensional reference shape?
image

koehlerson:
> > Vertex refers to points (nothing changes here)
> 
> Shouldn't this be renamed to "point" following the table, and "vertex" refers to the zero-dimensional reference shape? image

If we do `vertex == nodes` according to the table, then we need a new name for the current `vertex` which are all the corners of an element. It's already mixed up in the code base from time to time

KnutAM:
> If we do vertex == nodes, then we need a new name for the current vertex which are all the corners of an element. The naming is already mixed up in the code base from time to time. So point would be good substitute for nodes

I don't think `vertex == nodes` become the result? For example, what is a node in the middle of an edge, to me a node is a different concept (but could be wrong), relating to the interpolation (either geometric or algebraic). 

AFAIU `point` becomes the corners of an element, `vertex` the reference shape for 0-dimensional entities, and `peak == point` for 3d refshapes.

koehlerson:
I just meant that we should keep the distinction between corners which are vertices currently and all 0 dimension entities of an element, which should be currently `nodes` in the code base. 

> AFAIU point becomes the corners of an element, vertex the reference shape for 0-dimensional entities, and peak == point for 3d refshapes.

https://defelement.com/elements/examples/quadrilateral-lagrange-gll-1.html

Here for instance, the refer still to the corners as vertices, which is fine for me. I was just unsure if the table also gathers the inner 0 dimension entities as vertices.

KnutAM:
Interesting, so either I have further misunderstandings, or they are inconsistent 🤔 (But language-wise I'd switch vertex and point in the table)

fredrikekre:
> I guess the implication for the example in https://github.com/Ferrite-FEM/Ferrite.jl/issues/899 then would be that the "face" of the beam would not be accessible for adding boundary conditions to (using facets), but the endpoints would be available. And when thinking of this now, that makes sense because in the non-embedded case a load on the "face" of the beam would actually be a body load!

Yea, and thats why it make sense to use `CellValues` also for embedded elements and not `FacetValue` etc.

KnutAM:
> Yea, and thats why it make sense to use CellValues also for embedded elements and not FacetValue etc.

Yes, that makes sense, but what do you mean with etc.? (I still would use `FacetValues` to replace the current `FaceValues`, but I agree that the `CellValues` should be kept)

fredrikekre:
I mean that if it was relative the spatial dimension, then when integrating for example an embedded quad in 3d you would use a FacetValue and when integrating an embedded beam in 3d you would use RidgeValues.

KnutAM:
Aha, yes - I agree that codim relative rdim/topological dimension makes more sense!

Just to put up the a summary of fenics' definitions based on their manual, I think the only difference from defelement is the `vertex` vs `point`, (currently I'm leaning towards the fenics definition as the more logical one). 
(Feel free to edit this post with more names / links if I've missed something)

Co-dimensionally defined quantities
* [Cell](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Cell.html): "A Cell is a MeshEntity of topological codimension 0."
* [Facet](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Facet.html#dolfin.cpp.mesh.Facet): "A Facet is a MeshEntity of topological codimension 1."
* Ridge: *Not defined*
* Peak: *Not defined*

Topologically dimensionally defined quantities
* [Vertex](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Vertex.html): "A Vertex is a MeshEntity of topological dimension 0."
* [Edge](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Edge.html): "An Edge is a MeshEntity of topological dimension 1."
* [Face](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Face.html): "A Face is a MeshEntity of topological dimension 2."
* Volume: *Not defined*

Other relevant names
* [Point](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Point.html): "A Point represents a point in $\mathbb{R}^3$ with coordinates x,y,z, or alternatively, a vector in $\mathbb{R}^3$, supporting standard operations like the norm, distances, scalar and vector products etc."
* [FacetCell](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/FacetCell.html): "This class represents a cell in a mesh incident to a facet on the boundary. It is useful in cases where one needs to iterate over a boundary mesh and access the corresponding cells in the original mesh."

koehlerson:
> Aha, yes - I agree that codim relative rdim/topological dimension makes more sense!
> 
> Just to put up the a summary of fenics' definitions based on their manual, I think the only difference from defelement is the `vertex` vs `point`, (currently I'm leaning towards the fenics definition as the more logical one). (Feel free to edit this post with more names / links if I've missed something)
> 
> Co-dimensionally defined quantities
> 
>     * [Cell](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Cell.html): "A Cell is a MeshEntity of topological codimension 0."
> 
>     * [Facet](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Facet.html#dolfin.cpp.mesh.Facet): "A Facet is a MeshEntity of topological codimension 1."
> 
>     * Ridge: _Not defined_
> 
>     * Peak: _Not defined_
> 
> 
> Topologically dimensionally defined quantities
> 
>     * [Vertex](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Vertex.html): "A Vertex is a MeshEntity of topological dimension 0."
> 
>     * [Edge](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Edge.html): "An Edge is a MeshEntity of topological dimension 1."
> 
>     * [Face](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Face.html): "A Face is a MeshEntity of topological dimension 2."
> 
>     * Volume: _Not defined_
> 
> 
> Other relevant names
> 
>     * [Point](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/Point.html): "A Point represents a point in R3 with coordinates x,y,z, or alternatively, a vector in R3, supporting standard operations like the norm, distances, scalar and vector products etc."
> 
>     * [FacetCell](https://fenicsproject.org/olddocs/dolfin/1.3.0/python/programmers-reference/cpp/mesh/FacetCell.html): "This class represents a cell in a mesh incident to a facet on the boundary. It is useful in cases where one needs to iterate over a boundary mesh and access the corresponding cells in the original mesh."

In this definition of a vertex you set vertex == nodes or am I missing something? This is only valid for linear stuff

src/Grid/grid.jl
```diff
@@ -466,6 +523,59 @@ Returns all vertex sets of the grid.
...
+    @warn("addfaceset! for 1d-problems have been deprecated. Use getfacetset instead")
+     addvertexset!(grid, name, f; all)
+end
+function addfaceset!(grid::AbstractGrid{1}, name::String, set::Union{Set{VertexIndex},Vector{VertexIndex}}) 
+    addvertexset!(grid, name, set)
+end
+
+@inline facets(c::AbstractCell{<:AbstractRefShape{1}}) = vertices(c)
+@inline facets(c::AbstractCell{<:AbstractRefShape{2}}) = edges(c)
+@inline facets(c::AbstractCell{<:AbstractRefShape{3}}) = faces(c)
```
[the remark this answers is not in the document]
termi-official: I unresolved this, because I think we should add this to the documentation to have a single source of truth for the definitions. @lijas

koehlerson:
I'm just saying if we define something that includes **all** 0 dimension entities, there should be something for the corners, since they are very relevant for topological (geometry) constructions which was previously in the code base `vertices`. I think it's rather confusing going from vertices = corner to vertices = all 0 dimensional entities, but I don't mind how we name it in the future. I guess `points` refers to all 0 dimensional things, while vertices are just the corners, but I'm not sure and unfortunately I don't have access to the Ciarlet book. The website is not really clear on this subject for me

KnutAM:
> I guess points refers to all 0 dimensional things, while vertices are just the corners

This is my (current) understanding too.

koehlerson:
I'm confused by the book. For triangles it seems that corners == vertices, but for quads all points == vertices (they are defined with a single index)

image
vs. 
image

KnutAM:
Yes, I don't get their numbering scheme either, e.g. why 3 indices are used for 3rd order cases (I would have assumed `a_{ij}` where `i` is edge number and `j` is the internal node number for edge `i`.)

image
Fig 2.2.4

But I think we can agree that
* `Vertex`: End of line/interval segments (i.e. corners) (excludes internal points on edges/faces and in volumes)
* `Node`: Support points for geometric interpolation? (Includes internal points on edges/faces and in volumes)

?

termi-official:
Maybe it is easier for everyone if we just rename the thing to "corner"? I think this is the most direct and unambiguous term.

KnutAM:
> Maybe it is easier for everyone if we just rename the thing to "corner"? I think this is the most direct and unambiguous term.

Overall, from what I can read, vertices and corners are synonyms, but yet I wouldn't call the end-points of a line for corners, but rather vertices?

fredrikekre:
I suggest we use vertex then, and we can include a glossar in the docs with clarifications if needed.

lijas:
Knut pointed out that the dispatches

```
getfacetset(grid::AbstractGrid{1}) = grid.vertexsets
getfacetset(grid::AbstractGrid{2}) = grid.edgesets
getfacetset(grid::AbstractGrid{3}) = grid.facesets
```

are technically wrong, since facets is defined based on codimension of `rdim`, and the grid is parametrised with `sdim`. So my suggestion is to use the name `getboundaryset` instead, where "boundary" would be based on the co-dimension of `sdim` . What do you think?

I also think we can get away with not renaming `FaceValues` to `FacetValues`. My reasoning is:

- You would never want to integrate on the faces in 2d. So there is not really any ambiguity what FaceValues refer to here.
- In 3d, you would not use "FacetValues" to integrate along edges of e.g shell elements, since they do not have a defined normal direction (which facetvalues requires/stores). Instead, you would use something like "EdgeValues".

In this way, we would not need to introduce the concept of "facet" to the user, it would only be an internal thing.

KnutAM:
> getboundaryset instead, where "boundary" would be based on the co-dimension of sdim

In the following case, I would expect `boundaryset` to include the edges (4,3) and (3,6) as well as the vertex (3) if I want to integrate along the top boundary. These are all facets with co-dimension 1 wrt. rdim, so I think `facetset` makes more sense than introducing another dimension case. 

```julia
# Ferrite grid: Embedded line element is nr 2
# 4 ____ 3  3 3 ____ 6
# |      |  | |      |
# | (1)  |  | |  (3) |
# 1 ____ 2  2 2 ____ 5
```

I would therefore argue that we should change the fields in `Grid`:

* `facesets`: Change to to `facetsets`.
* `edgesets`: I have never used them. 
  *Could* become `ridgesets`, but since fenics don't seem to have it we could perhaps remove it? The only place I can find it used in the docs are for the shell example [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/d0187fd741cc2bcd48884650360b0f7ac50bc7ab/docs/src/literate-tutorials/linear_shell.jl#L44-L57), and that usage would be replaced by `facetset` after this PR.
* `vertexsets`: I think keeping this makes sense. (e.g. apply Dirichlet at one location)
  I don't think `peaksets` are useful because they can only be used in 3d, and then coincides with `vertexsets`

> You would never want to integrate on the faces in 2d. So there is not really any ambiguity what FaceValues refer to here.
> In 3d, you would not use "FacetValues" to integrate along edges of e.g shell elements, since they do not have a defined normal direction (which facetvalues requires/stores). Instead, you would use something like "EdgeValues".

I think that while doing this change now, we should be fully consistent with the notation and avoid "exceptions", such as saying that `FaceValues` refers to `facets` which happens to coincide with `faces` in 3d, but are `edges` with defined normals in `2d`, and `vertices` with defined normals in `1d`. 

> shell elements, since they do not have a defined normal direction

In this case, I would expect that `FacetValues` stores the outwards pointing vector perpendicular to the edge tangent and the face normal, isn't this well-defined? An `EdgeValues` object would, on the other hand, only store the edge direction as the normal to an edge of a polyhedron is not defined.

I see that in this setting, it would also be relevant to know the edge tangent direction though, but this would be similar wrt. an embedded `CellValues`, which currently don't store the normal information. 

I took a quick look at the code, and a change that could make sense is to include a direction-field in `GeometryMapping` (to be done in a separate PR, I don't think the current PR removes features that such a change would reenable). The direction field would change depending on the usage as something like
* Standard, non-embedded `CellValues`: `nothing`
* Standard, non-embdded `FacetValues`: Normal vector
* 2D `CellValues` embedded in 3D: Normal vector
* 1D `CellValues` embedded in 2D or 3D: Direction vector
* 2D `FacetValues` embedded in 3D (e.g. shell): Normal and direction vector
* 1D `FacetValues` embedded in 2D or 3D: Normal vector

fredrikekre:
> In the following case, I would expect boundaryset to include the edges (4,3) and (3,6) as well as the vertex (3) if I want to integrate along the top boundary.

But surely you couldn't treat these three entities the same in the integration? You would in this case do something for  the edges and something else for the vertex, right?

I think we should just not allow `facetsets` of different types. In this example you would instead have two facetsets for the top boundary, which would correspond how you treat them in the code anyway.

KnutAM:
> But surely you couldn't treat these three entities the same in the integration? ... 
I think we should just not allow facetsets of different types

Currently, these would both belong to the boundary faceset but different subdofhandlers. I don't see the problem for keeping this functionality when we rename `FaceIndex` to `FacetIndex`?

I would just split these by intersecting with the subdofhandlers' sets and dispatch (prob with a trait) on the `FacetValue`. I think both ways (as you and I wrote) should be possible.

PR #790, opened 2023-08-14
Arbitrary order Lagrange interpolations for hypercubes (with arbitrary basis) and triangle (equidistant basis only)

### LINKED ISSUE #626: Generic arbitrary order basis functions
Currently we have a fast path with some optimized fixed order interpolations. However, it might be beneficial to also provide several generic interpolations with more flexibility (e.g. change between equidistant and GLL points for the basis functions) for some users. The existing Lagrange triangles of order 3,4,5 may be a good starting point.

AbdAlazezAhmed:
For #626 
Some comments:
- using Vandermonde inverse resulted in worse floating point errors during Dirac-delta tests
- Tested order 10 with heat equation example, seems to work (haven't checked convergence tho but seems ok from paraview)

Note: I also edited the existing Tri345 to be arbitrary but it doesn't have the arbitrary basis/points thing


termi-official:
We definitely should be able to pass 1D basis functions at least for hypercubes, with GLL as default for continuous Lagrange and GL as default for discontinuous Lagrange (for the high order infrastructure), because we can rule out Runge's phenomenon.

src/interpolations.jl
```diff
@@ -1330,33 +1362,58 @@ function getPermLagrangeQ(order)
...
-function facedof_indices(ip::ArbitraryOrderLagrange{RefQuadrilateral,order}) where order
-    interior = facedof_interior_indices(ip)
-    face1 = SVector{order + 1}((i == 1 ? 1 : i == 2 ? 2 : interior[1][i-2] for i in 1:order+1))    
-    face2 = SVector{order + 1}((i == 1 ? 2 : i == 2 ? 3 : interior[2][i-2] for i in 1:order+1))    
-    face3 = SVector{order + 1}((i == 1 ? 3 : i == 2 ? 4 : interior[3][i-2] for i in 1:order+1))    
-    face4 = SVector{order + 1}((i == 1 ? 4 : i == 2 ? 1 : interior[4][i-2] for i in 1:order+1))    
-    return (face1, face2, face3, face4)
+for (name,  facefuncs) in (
+    (:ArbitraryOrderLagrange,               :facedof_indices),
+    (:ArbitraryOrderDiscontinuousLagrange,  :dirichlet_facedof_indices)
```
AbdAlazezAhmed, line 1387: I just realized that using GL points should result in `dirichlet_facedof_indices` being empty IIUC. This makes this function need to either check for faces on boundary or have predetermined options for the points (equidistant, GL, Radau) instead of passing the points in the constructor.
I think having predetermined options would be easier and more consistent (as we guarantee the ordering), what do you think?

termi-official: I would suggest that we just throw an error for now when trying to extract the Dirichlet dofs for all L2 elements which are not based on "symmetric" nodes, when the basis functions of the interior nodes are non-zero on the boundary (and for modal basis functions). If we want to work further on strong enforcement of the Dirichlet condition we have to think a bit how to do it correctly.

src/interpolations.jl
termi-official, line 1: @fredrikekre should we separate the fixed order implementation which we have from the arbitrary order infrastructure?

src/interpolations.jl
```diff
@@ -683,75 +683,61 @@ function shape_value(ip::Lagrange{RefTriangle, 2}, ξ::Vec{2}, i::Int)
...
-    order == 5 && return ((1,2,4,5,6,7), (2,3,8,9,10,11), (3,1,12,13,14,15))
+function getPermLagrangeTri(order)
+    verts = SVector{3}(order + 1, (order + 1) * (order + 2) ÷ 2, 1)
+    edge1 = SVector{order-1}((sum(i:order+1) for i in order:-1:2))
+    edge2 = SVector{order-1}((sum(i:order+1)+1 for i in 3:(order+1)))
+    edge3 = SVector{order-1}((i for i in 2:order))
+    # TODO: fix this bottleneck, nested generators don't work so one has to use ...
+    interior = NTuple{(order - 2) * (order - 1) ÷ 2}((i for j in (order+1):-1:4 for i in sum(j:order+1)+2 : sum(j:order+1)+j-2))
+    return SVector((verts..., edge1..., edge2..., edge3..., interior...))
+end
```
termi-official, line 703: I think it is worth to investigate a code generator here in the future (e.g. to just generate the elements from the Ciarlet definition). So, no need to invest too much time here if it is not absurdely slow. :)

src/interpolations.jl
```diff
@@ -1264,6 +1250,429 @@ function shape_value(ip::BubbleEnrichedLagrange{RefTriangle, 1}, ξ::Vec{2}, i::
...
+    return result
+end
+
+
+for (name,  default_coords) in (
+    (:ArbitraryOrderLagrange,               :both),
+    (:ArbitraryOrderDiscontinuousLagrange,  :neither)
+    )
+    @eval begin
+        function $(name){RefQuadrilateral, order}(points::Vector{Float64} = GaussQuadrature.legendre(order+1, GaussQuadrature.$(default_coords))[1]) where order
```
termi-official, line 1371: Can we add doc strings for these, so users know about the functionality?

Also, how can users discover this feature? Maybe we should add some hint in the Lagrange doc string and a tip one of the examples? Not sure if it is worth to copy paste the full heat example here. In the future we can think about adding a Lagrangian formulation of some advection problem.

src/interpolations.jl
```diff
@@ -1264,6 +1250,429 @@ function shape_value(ip::BubbleEnrichedLagrange{RefTriangle, 1}, ξ::Vec{2}, i::
...
+    ncellintdofs = (order - 1)^2
+    totaldofs = getnbasefunctions(ip)
+    return SVector{ncellintdofs}((totaldofs-ncellintdofs+i for i in 1:ncellintdofs))
+end
+
+##########################
+# Lagrange RefHexahedron #
+##########################
+# Based on vtkTriQuadraticHexahedron (see https://kitware.github.io/vtk-examples/site/Cxx/GeometricObjects/IsoparametricCellsDemo/)
+# Permutation to switch numbering to Ferrite ordering
```
termi-official, line 1429: Can you elaborate?

termi-official (commented):
Comments below. I think we need to add some doc strings and think about discoverability. Not sure if we can study the convergence order at this point.

src/interpolations.jl
KnutAM: Passing `Val(order)` might be a remedy

src/interpolations.jl
AbdAlazezAhmed: The first line refers to the numbering convention used (copied from fixed order one)
The second one refers to the permutation that maps from the generated (-1:1)^n coordinates to the ones following the convention (such as {-1,...,1} ->{-1,1,...} in 1D)

src/interpolations.jl
```diff
@@ -32,6 +29,17 @@ The following interpolations are implemented:
 * `Serendipity{RefQuadrilateral,2}`
 * `Serendipity{RefHexahedron,2}`
 
+The following interpolations are implemented with arbitrary order:
+
+* `Lagrange{RefTriangle,order}`
```
termi-official, line 34: I think we should wrap this also into the arbitrary order Lagrange struct.

AbdAlazezAhmed: I think the name is kinda too long 😄. Maybe we can make it a fallback for Lagrange constructor so we don't have to use this long name?

termi-official:
Just as a reminder, this should come after https://github.com/Ferrite-FEM/Ferrite.jl/pull/829 right?

AbdAlazezAhmed:
> Just as a reminder, this should come after https://github.com/Ferrite-FEM/Ferrite.jl/pull/829 right?

Yeah I think so as #829 would eliminate the need for permutations IIUC.

termi-official:
I think this one here might be helpful for future reference on simplices: Warburton, T. An explicit construction of interpolation nodes on the simplex. J Eng Math 56, 247–262 (2006). https://doi.org/10.1007/s10665-006-9086-6

PR #792, opened 2023-09-07
Improvements to assembler

lijas:
The first commit only improves the error message when the user is trying to add a value to a location that has not been allocated in the sparse-matrix. This error is pretty uncommon to hit in a "standard fe-code", but I have gotten the error i bit recently because I have messed up when creating costum sparsity patterns.

The second commit is a suggestion for making the assembler a bit more multi-threaded friendly. Currently we have to create an assembler for each thread `assemblers = [start_assemble(K, f) for i in 1:nthreads]`. This PR makes it possible to instead create the `permutation` and `sorteddof` vectors as arguments to the assemble-function instead. `permutation` and `sorteddof` can be created for each thread at the same time as the other cache-variables (celldofs, coords etc).

```
for color in colors
        # Each color is safe to assemble threaded
        Threads.@threads :static for i in 1:length(color)
               (; coords, celldofs, permutation, sorteddofs) = cache[threadid()]
               .... 
               assemble!(assembler, celldofs, Ke, fe, permutation, sorteddofs)
        end
end
```


KnutAM:
The error message improvement is nice!

For the multithreading support, what is the advantage of storing both `permutation` and `sorteddof` over just storing the assembler in `cache`?

lijas:
It depends a bit on your code. In my code I have a init-phase where I create the cache for the cell (coords, dofs etc), and at that point I dont really have access to K and f so I can not create an assembler (but I can create permutation and sorteddof). So currently I have to create new assemblers (for each thread) each time a assemble the system. This is should ofcourse not be an issue performace wise, but is not super nice imo. 

Fredrik said that did not like that this PR leaked internals, and perhaps that is a good argument not to merge this.

termi-official:
Is there any argument against a custom assembler here instead of leaking internals? Not a good idea for a name tho, because `ThreadedAssembler` is not really precise and misleading.

KnutAM:
> that is a good argument not to merge this

But as mentioned, I think the better error message in the first commit will be a nice enhancement. So would be great to perhaps split into two separate PRs/issues to get that one in IMO!

termi-official (approved):

test/test_assemble.jl
```diff
@@ -139,11 +139,17 @@ end
     as = start_assemble(Symmetric(K))
     errr = ErrorException("some row indices were not found")
```
KnutAM, line 140:
```suggestion
    errr(i,j) = ErrorException(Ferrite._missing_sparsity_pattern_message(i,j))
```
And use below?

Alternatively, remove this line as not used.

PR #793, opened 2023-09-15
DocumenterCitations.jl baseline integration

termi-official:
As mentioned in https://github.com/Ferrite-FEM/Ferrite.jl/pull/737#issuecomment-1580977060 this prodives a baseline integration for citations in Documenter.jl . In subsequent PRs we have to feed it with our papers. Closes https://github.com/Ferrite-FEM/Ferrite.jl/issues/620 .


docs/make.jl
```diff
@@ -21,6 +21,11 @@ include("generate.jl")
 include("changelog.jl")
 create_documenter_changelog()
 
+bibtex_plugin = CitationBibliography(
+    joinpath(@__DIR__, "src", "assets", "references.bib"),
+    style=:authoryear
```
fredrikekre, line 26: I prefer the default `:numeric` style I think.

termi-official: The reasoning from my side was that people might be able to deduce some information from the name already when reading. Changed.

fredrikekre:
https://ferrite-fem.github.io/Ferrite.jl/previews/PR793/bibliography/ Can you also add the recommended CSS file to make it look a bit nicer?

PR #795, opened 2023-09-18
Add more references and some urls/dois

fredrikekre:
- Add references to workstream paper and dof distribution paper
 - Add urls and dois to existing references

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -582,18 +582,15 @@ element B in the example.
 
 In addition, we also have to preverse the ordering at each dof location.
 
-For more details we refer to [1] as we follow the methodology described therein.
+For more details we refer to Scroggs et al. [Scroggs2022](@cite) as we follow the methodology
+described therein.
 
-[1] Scroggs, M. W., Dokken, J. S., Richardson, C. N., & Wells, G. N. (2022).
-    Construction of arbitrary order finite element degree-of-freedom maps on
-    polygonal and polyhedral cell meshes. ACM Transactions on Mathematical
-    Software (TOMS), 48(2), 1-23.
-
-    !!!TODO Citation via DocumenterCitations.jl.
-
-    !!!TODO Investigate if we can somehow pass the interpolation into this function in a typestable way.
+# References
+ - [Scroggs2022](@cite) Scroggs et al. ACM Trans. Math. Softw. 48 (2022).
 """
 @inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int,Int}, orientation::PathOrientationInfo, adjust_during_distribution::Bool)
+    # TODO Investigate if we can somehow pass the interpolation into this function in a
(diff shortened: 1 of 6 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -582,18 +582,15 @@ element B in the example.
...
-    polygonal and polyhedral cell meshes. ACM Transactions on Mathematical
-    Software (TOMS), 48(2), 1-23.
+For more details we refer to Scroggs et al. [Scroggs2022](@cite) as we follow the methodology
+described therein.
 
-    !!!TODO Citation via DocumenterCitations.jl.
-
-    !!!TODO Investigate if we can somehow pass the interpolation into this function in a typestable way.
+# References
+ - [Scroggs2022](@cite) Scroggs et al. ACM Trans. Math. Softw. 48 (2022).
```
termi-official, line 589: For future reference: Is there something stopping us from generating this one automatically, similar to the reference generator in the examples?

src/Dofs/DofHandler.jl
```diff
@@ -582,18 +582,15 @@ element B in the example.
...
-[1] Scroggs, M. W., Dokken, J. S., Richardson, C. N., & Wells, G. N. (2022).
-    Construction of arbitrary order finite element degree-of-freedom maps on
-    polygonal and polyhedral cell meshes. ACM Transactions on Mathematical
-    Software (TOMS), 48(2), 1-23.
+For more details we refer to Scroggs et al. [Scroggs2022](@cite) as we follow the methodology
+described therein.
 
-    !!!TODO Citation via DocumenterCitations.jl.
-
-    !!!TODO Investigate if we can somehow pass the interpolation into this function in a typestable way.
```
termi-official, line 594: Debris or intended? If the latter, then we don't want to give this one another try to simplify the dof adjustment system?

termi-official (approved):
LGTM otherwise.

src/Dofs/DofHandler.jl
fredrikekre: Which reference generator?

src/Dofs/DofHandler.jl
fredrikekre: I just moved it down as a code comment instead if keeping it in the docstring.

src/Dofs/DofHandler.jl
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/docs/src/literate-gallery/topology_optimization.jl#L530-L534 where, for example, we put something else than `Pages`. Or even automate adding the references in docstrings with another plugin.

fredrikekre: Ah, okay. No, because this is in a docstring and there can't really be anything dynamic there since you want it to look nice also in the REPL help mode and such. See e.g. https://github.com/JuliaDocs/DocumenterCitations.jl/issues/21 for some related discussion.

Issue #742, opened 2023-06-10
Reverse CI integration with downstream Ferrite packages


termi-official:
We should add some basic integration testing for packages downstream of Ferrite core to catch internal problems earlier (currently FerriteViz.jl and FerriteDistributed.jl).

References 

https://discourse.julialang.org/t/github-actions-workflow-for-reverse-dependency-integration-testing/53536

https://github.com/JuliaDiff/ChainRulesCore.jl/blob/caf8692ca1bf5fabeb4afc86340eef9456da469c/.github/workflows/IntegrationTest.yml#L65

KnutAM:
> currently FerriteViz.jl and FerriteDistributed.jl)

and FerriteGmsh.jl and FerriteMeshParser.jl

> We should add some basic integration testing for packages downstream of Ferrite core

Yes, that would be nice for all such packages that are in the Ferrite-FEM org! 
I'm not sure how this would work in practice. Currently, for example, we haven't changed to 0.4 in master, but included a lot of breaking changes. Perhaps this would be fine just setting the version to 0.4.0-dev - not sure what is the "correct" approach.
However, what happens when a sub-package fixes one breaking change, allows Ferrite 0.4.0-dev, and then another new breaking change is introduced. Then the workflow will fail for all new PR's until the downstream package is fixed. Is this the desired result? Or perhaps I'm misunderstanding how this would work?

An alternative approach is that downstream packages can run timed (e.g. weekly) CI-runs against Ferrite#master to be "notified" about breaking changes in Ferrite.jl, see e.g. https://github.com/KnutAM/FerriteAssembly.jl/blob/main/.github/workflows/FerriteMasterCI.yml

termi-official:
> > currently FerriteViz.jl and FerriteDistributed.jl)
> 
> and FerriteGmsh.jl and FerriteMeshParser.jl
> 

What do you think is missing for basic test coverage in via what we have in docs? (assuming the porous media example will be merged soon)

> > We should add some basic integration testing for packages downstream of Ferrite core
> 
[...]

PR #797, opened 2023-09-20
Reverse CI

termi-official:
Resolves https://github.com/Ferrite-FEM/Ferrite.jl/issues/742 .


termi-official:
Preceded by https://github.com/Ferrite-FEM/Ferrite.jl/pull/1000 . I will work on it again afterwards.

PR #764, opened 2023-07-12
Split values into geometric mapping and function values

KnutAM:
This PR splits `Values` into geometric values and function values, the main advantages are 
* Reuse code for `reinit!` for `FaceValues` and `CellValues`
  - This solves [#616](https://github.com/Ferrite-FEM/Ferrite.jl/issues/616#issuecomment-1556138552)
  - When different mappings are added in the future (cf #798) re-use becomes more important since we get more `reinit!` methods. 
  - For making the `MultiCellValues` type (cf. #680) this allows using only a single geometry cache
* The new parameterization is free to allow e.g. GPU-arrays

For a more complete picture, please see #798.

For normal use, this is not breaking wrt. master (i.e. no direct field access).
However, `CellValues{<:VectorInterpolation}` cannot be dispatched on anymore. This can be added if required, but in that case perhaps (re-)introducing `CellVectorValues` and `CellScalarValues` as aliases is cleaner. 

<details><summary> Description of internal changes </summary>

# Description of internal changes
```julia
struct CellValues{FV, GV, QR, detT<:AbstractVector} <: AbstractCellValues
    fun_values::FV # FunctionValues
    geo_values::GV # GeometryValues
    qr::QR         # QuadratureRule
    detJdV::detT   # AbstractVector{<:Number}
end

struct FaceValues{FV, GV, QR, detT, nT, V_FV<:AbstractVector{FV}, V_GV<:AbstractVector{GV}} <: AbstractFaceValues
    fun_values::V_FV # AbstractVector{FunctionValues}
    geo_values::V_GV # AbstractVector{GeometryValues}
    qr::QR           # FaceQuadratureRule
    detJdV::detT     # AbstractVector{<:Number}
    normals::nT      # AbstractVector{<:Vec}
    current_face::ScalarWrapper{Int}
end
```
where

`QuadratureRule/FaceQuadratureRule`: No changes, contains the location and 
weights for the quadrature on the reference element.

`GeometryValues` contain information about the geometric interpolation on 
the reference element. These are used to calculate the required quantities 
for (1) mapping mutable parts of `FunctionValues` from the reference element to the 
actual element and (2) calculating properties for the current cell, such as the `detJdV`
for each quadrature point and the face `normals`. 

```julia
struct GeometryValues{IP, M_t, dMdξ_t, d2Mdξ2_t}
    ip::IP             # ::Interpolation                Geometric interpolation 
    M::M_t             # ::AbstractVector{<:Number}     Values of geometric shape functions
[...]

Issue #193, opened 2018-04-15
My wishlist


louisponet:
After learning how to use/using JuAFEM for about 2 weeks, I must say that it is a great tool, very understandable and hackable design, and it seems general enough to be expanded upon greatly!

I've used JuAFEM basically as a discretization tool for calculating a continuum model's total energy and gradients/hessians, as discussed on  [juafem-forwarddiff-nlsolve-finite-elements-for-nonlinear-problems](https://discourse.julialang.org/t/juafem-forwarddiff-nlsolve-finite-elements-for-nonlinear-problems/10085) It performs very well, vastly outperforming other implementations that colleagues of mine have done in c++ or Mathematica (duh). 

I'll now give my wishlist for features as requested in the thread:

- Easier ways to access the fields inside a `DofHandler`, and create starting conditions e.g.:
```julia
function startingconditions!(dofvec, dh)
    mid = div(ndofs_per_cell(dh), 2)
    for cell in CellIterator(dh)
        globaldofs = celldofs(cell)
        for i=1:3:mid
            it = div(i, 3) + 1
            dofvec[globaldofs[mid+i]] = -0.53
            dofvec[globaldofs[mid+i+1]] = 0.53tanh(cell.coords[it][1]/5)
            dofvec[globaldofs[mid+i+2]] = 0.53
        end
    end
end
```
In light of the other design, something like `fielddofs(:Field, cell)` similar to `celldofs(cell)`, seems a good way. Also `fielddofs(:Field, dh::DofHandler)` would be very nice to calculate certain things from the finalized result. 

- A way to thread over the `CellIterator` loop. 
This seems a bit tricky since assembling residus or stiffness tensors is not trivial multithreaded. It really depends on the grid what needs to be done. The way I do it now, which is very much not scalable is pretty brutal (I have two fields, :u, :P, and thus a dofvector uP):
```julia
mutable struct ModelCaches{T, CV, CI, CF}
    ∇fs::Vector{Vector{T}}
    cvus::Vector{CV}
    cvPs::Vector{CV}
    cellits::CI
    configs::CF
end

function ModelCaches(∇f::Vector{T}, cvu, cvP, dh) where T
    cellits = [CellIterator(dh) for t=1:nth]
    ellen   = length(celldofs(cellits[1]))
    ∇fs     = [zeros(∇f) for t=1:nth]
    cvus    = [copy(cvu) for t=1:nth]
    cvPs    = [copy(cvP) for t=1:nth]
[...]

PR #798, opened 2023-09-23
WIP: New vector interpolations and mappings

### LINKED ISSUE #569: Support for H(curl) and H(div) elements

KnutAM:
Implements Hcurl (Nedelec) and Hdiv (RaviartThomas and BrezziDouglasMarini) interpolations and the associated mappings.

To ensure that the infrastructure works well, I would aim to have the following here

* [x] Tests for H(div) and H(curl) on reference element 
* [x] Continuity testing for mapped H(div) and H(curl)
* [x] Test homogeneous Dirichlet BC for all interpolations
* [x] Tutorial with H(div) elements: [preview](https://ferrite-fem.github.io/Ferrite.jl/previews/PR798/tutorials/heat_equation_hdiv) (Superseded by #1388)
* [x] Tutorial with H(curl) elements [preview](https://ferrite-fem.github.io/Ferrite.jl/previews/PR798/tutorials/maxwell_good_bad_ugly/)

# Finalization plan
The diff is too large for effective review, so this PR should not be merged, but I will split it into several smaller PRs.

1. Bulk part (main implementation and testing) #1145
2. ProjectedDirichlet #1151
3. H(div) projection how-to #1161
4. H(curl) how-to: Not started


docs/src/devdocs/mapping.md
```diff
@@ -0,0 +1,103 @@
...
+\begin{align*}
+    \boldsymbol{x}(\boldsymbol{\xi}) =& \sum_{i}^N \hat{\boldsymbol{x}}_i \hat{N}_i^g(\boldsymbol{\xi}) \\
+    \boldsymbol{J} :=& \frac{\mathrm{d}\boldsymbol{x}}{\mathrm{d}\boldsymbol{\xi}} = \sum_{i}^N \hat{\boldsymbol{x}}_i \otimes \frac{\mathrm{d} \hat{N}_i^g}{\mathrm{d}\boldsymbol{\xi}}\\
+    \boldsymbol{\mathcal{H}} :=&
+    \frac{\mathrm{d} \boldsymbol{J}}{\mathrm{d} \boldsymbol{\xi}} = \sum_{\alpha=1}^N \hat{\boldsymbol{x}}_\alpha \otimes \frac{\mathrm{d}^2 \hat{N}^g_\alpha}{\mathrm{d} \boldsymbol{\xi}^2}
+\end{align*}
+```
+where the defined $\boldsymbol{J}$ is the jacobian of the mapping, and in some cases we will also need the corresponding hessian, $\boldsymbol{\mathcal{H}}$ (3rd order tensor).
+
+We require that the mapping from reference coordinates to real coordinates is bijective, meaning that we can express $\boldsymbol{x} = \boldsymbol{x}(\boldsymbol{\xi}(\boldsymbol{x}))$, such that
```
KnutAM, line 19:
```suggestion
We require that the mapping from reference coordinates to real coordinates is [diffeomorphic](https://en.wikipedia.org/wiki/Diffeomorphism), meaning that we can express $\boldsymbol{x} = \boldsymbol{x}(\boldsymbol{\xi}(\boldsymbol{x}))$, such that
```
Thanks @termi-official

docs/src/literate-tutorials/maxwell.jl
```diff
@@ -0,0 +1,72 @@
+# Maxwell's equations 
+# For simplicity, we start with the very basic example from 
+# https://www.math.colostate.edu/~bangerth/videos.676.33.html
+# Specifically,
```
termi-official, line 4: Maybe some context might help, because there is quite a big jump from Maxwell to this specific form
```suggestion
# Specifically, if we investigate an electromagentic problem where we are interested in the steady state of either the electrical or magentical field, in the case when neither electrical nor magnetical sources are present, then we can state the weak form as follows
```

termi-official: Maybe we can also leave the hint for users that we can simplify further to obtain a Poisson or Laplace problem instead for this very specific case and that we only look at this case here to showcase how to use Nedelec elements to assemble a Hcurl conforming problem.

docs/src/assets/fe_mapping.svg
termi-official, line 1: I think this is quite valuable already!

Maybe we should also think about how we can visualize the relation our mappings also in this SVG, if we are already at introducing it to the docs? Maybe some arrow from the first to the second box with some text, or is this already too dense? Happy to discuss!

KnutAM: Thanks for the comments!
1) For the example (as discussed on slack), this was just some test cases. Started on the eigenvalue problem, but doesn't seem to work yet - have to look into that! (In case someone is interested in playing around, feel free to add more example files on this branch, I might have to pause this one for some time now)
2) For an illustration, that might be a good idea! If too crowded, I guess we could have one figure per mapping to show how the normal/tangential parts are preserved. Let's discuss at FerriteCon if we have time!
3) For examples, I think it would be nice to have one example for each (also as suggested on slack): One mixed formulation (with H-div) and one maxwell problem (with H-curl). I wouldn't be confident in merging until we have good working examples

(But it is becoming a massive PR with introducing both the split from #764, new interpolations, and tutorials... (even if for line numbers the SVG is the main sinner)

termi-official:
Does this now close everything in #193 ?

KnutAM:
> Does this now close everything in https://github.com/Ferrite-FEM/Ferrite.jl/issues/193 ?

"A way to add fancier constraints, such as bounds on fields etc" is still missing I guess. But perhaps for future reference it might be nice to note what is missing in that issue to avoid having to scan through all of it?

termi-official:
> > Does this now close everything in #193 ?
> 
> "A way to add fancier constraints, such as bounds on fields etc" is still missing I guess. But perhaps for future reference it might be nice to note what is missing in that issue to avoid having to scan through all of it?

Isn't this covered in the topology optimization example?

KnutAM:
> Isn't this covered in the topology optimization example?

I don't think so, since the bounds there are on the density factor which is not a field in the DofHandler?

docs/src/devdocs/FEValues.md
```diff
@@ -26,3 +26,17 @@ Ferrite.FunctionValues
   the jacobian, $\partial M/\partial x$, is calculated. 
 * In the second part of `reinit!`, The `MappingValues`, containing sufficient information for the current quadrature point, is given to `apply_mapping!`. 
   This allows the shape values and gradients stored in `FunctionValues`, to be mapped to the current cell by calling `apply_mapping!`.
+
+## Custom FEValues
+Custom FEValues, `fe_v`, should normally implement the `reinit!` method.
+Additionally, for normal functionality the `getnquadpoints` should be implemented.
+Note that asking for the `n`th quadrature point must be inside array bounds if 
+`1<=n<:getnquadpoints(fe_v)`
```
termi-official, line 34:
```suggestion
`1<=n<=getnquadpoints(fe_v)`
```

docs/src/devdocs/FEValues.md
```diff
@@ -26,3 +26,17 @@ Ferrite.FunctionValues
...
+(`checkquadpoint` can, alternatively, be dispatched to check that `n` is inbounds.)
+
+Supporting `function_value`, `function_gradient`, `function_symmetric_gradient`, `function_divergence`, and `function_curl`,
+requires implementing `getnbasefunctions`, `shape_value`, and `shape_gradient`.
```
termi-official, line 38: Maybe we should reference the new API (batch eval) here too.

docs/src/devdocs/FEValues.md
```diff
@@ -26,3 +26,17 @@ Ferrite.FunctionValues
...
   This allows the shape values and gradients stored in `FunctionValues`, to be mapped to the current cell by calling `apply_mapping!`.
+
+## Custom FEValues
+Custom FEValues, `fe_v`, should normally implement the `reinit!` method.
+Additionally, for normal functionality the `getnquadpoints` should be implemented.
+Note that asking for the `n`th quadrature point must be inside array bounds if 
+`1<=n<:getnquadpoints(fe_v)`
+(`checkquadpoint` can, alternatively, be dispatched to check that `n` is inbounds.)
+
+Supporting `function_value`, `function_gradient`, `function_symmetric_gradient`, `function_divergence`, and `function_curl`,
```
termi-official, line 37: I think this is not correct. We should rephrase that we have more requirements here (like for example the fields need to have specific names in master). Or we add more functions to query these, which I think is better. Maybe we can think about a trait system for cell values, too?

docs/src/devdocs/FEValues.md
```diff
@@ -26,3 +26,17 @@ Ferrite.FunctionValues
...
+requires implementing `getnbasefunctions`, `shape_value`, and `shape_gradient`. 
+Note that asking for the `i`th shape value or gradient must be inside array bounds if `1<=i<:getnbasefunctions(fe_v)`
+
+Supporting `spatial_coordinate` requires implementing `getngeobasefunctions` and `geometric_value`.
```
termi-official, line 41: Same as above.

docs/src/devdocs/FEValues.md
KnutAM: Yes, it should be documented, but I think that should go under devdocs for interpolations.

docs/src/devdocs/FEValues.md
KnutAM: Oops - thanks!

docs/src/devdocs/FEValues.md
KnutAM: Maybe I misunderstand, but there should be no requirements on specific field names left, see the example for `SimpleCellValues` that fulfill this specification. Or if you meant something else, could you elaborate?

docs/src/devdocs/FEValues.md
KnutAM: Same answer as above, but not shown in the `SimpleCellValues` example.

docs/src/devdocs/FEValues.md
termi-official: Maybe I missed something, but if I rename `dNdx` to `fun_by_x`in `SimpleCellValues`, then function* won't work anymore right?

KnutAM: Yes it will, the `function_value` just calls `shape_value`. (In this branch, the built-in cell values and face values don't have this field)

On master, there were a few (IMO) antipatterns in `common_values.jl` dispatching on `AbstractValue` inputs and expecting certain fields to exist, but these are now removed.

docs/src/literate-tutorials/heat_equation_hdiv.jl
```diff
@@ -0,0 +1,234 @@
+# # [Heat equation - Mixed H(div) conforming formulation)](@id tutorial-heat-equation-hdiv)
+# As an alternative to the standard formulation for solving the heat equation used in
+# the [heat equation tutorial](@ref tutorial-heat-equation), we can used a mixed formulation
+# where both the temperature, $u(\mathbf{x})$, and the heat flux, $\boldsymbol{q}(\boldsymbol{x})$,
+# are primary variables. From a theoretical standpoint, there are many details on e.g. which combinations
+# of interpolations that are stable. See e.g. [Gatica2014](@cite) and [Boffi2013](@cite) for further reading.
+# This tutorial is based on the theory in
+# [Fenics' mixed poisson example](https://fenicsproject.org/olddocs/dolfin/1.4.0/python/demo/documented/mixed-poisson/python/documentation.html).
```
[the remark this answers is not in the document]
KnutAM: Yes, I was struggling with that today. Not sure if that's just a mistake though, since it is still linked to from their GitHub webpage? 

https://github.com/FEniCS/docs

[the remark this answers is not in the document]
fredrikekre: Probably forgot to pay the bill. Perhaps we should buy it and redirect it to Ferrite :)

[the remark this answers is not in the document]
KristofferC: Seems to be back?

docs/src/literate-tutorials/maxwell.jl
```diff
@@ -0,0 +1,91 @@
...
+    Ferrite.zero_out_rows!(B, ch.dofmapping)
+    Ferrite.zero_out_columns!(B, ch.prescribed_dofs)
+    return A, B, dh
+end
+
+ip = Nedelec{2, RefTriangle, 1}()
+
+A, B, dh = setup_and_assemble(ip)
+
+vals, vecs, info = geneigsolve((A, B), 1, EigSorter(x -> abs(x - 5.0)); maxiter = 1000);
```
KnutAM, line 90:
```suggestion
# vals, vecs, info = geneigsolve((A, B), 1, EigSorter(x -> abs(x - 5.0)); maxiter = 1000);
```

src/interpolations.jl
```diff
@@ -1774,3 +1783,299 @@ function mapping_type end
...
+```
+"""
+RefTriangle
+=#
+# https://defelement.com/elements/examples/triangle-raviart-thomas-lagrange-2.html
+# Signs changed when needed to make positive direction outwards
+function reference_shape_value(ip::RaviartThomas{2, RefTriangle, 2}, ξ::Vec{2}, i::Int)
+    x, y = ξ
+    # Face 1 (keep ordering, flip sign)
+    i == 1 && return Vec(4x * (2x - 1), 2y * (4x - 1)) / 2
```
KnutAM, line 1847: The division by 2 compared to defelement gives an identity for the line integral of 1/num_edge_dofs, such that setting the actual dof-values to 1 results in a unit flux. @termi-official: Do you see any issues by scaling the shape functions like this?

test/test_interpolations.jl
```diff
@@ -252,6 +249,251 @@ using Ferrite: reference_shape_value, reference_shape_gradient
...
+
+    # Required properties of shape value Nⱼ of an edge-elements (Hcurl) on an edge with direction v, length L, and dofs ∈ 𝔇
+    # 1) Unit property: ∫(Nⱼ ⋅ v f(s) dS) = 1/length(𝔇) ∀ ∈ 𝔇
+    #    Must hold for
+    #    length(𝔇) ≥ 1: f(s) = 1
+    #    length(𝔇) = 2: f(s) = 1 - s or f(s) = s for 1st and 2nd dof, respectively.
+    #    Additionally, should be zero for
+    #    length(𝔇) = 2: f(s) = s or f(s) = 1 - s for 1st and 2nd dof, respectively.
+    #    s is the path parameter ∈[0,1] along the positive direction of the path.
+    # 2) Zero along other edges: Nⱼ ⋅ v = 0 if j∉𝔇
```
KnutAM, line 277: Associated test definitions for [comment above](https://github.com/Ferrite-FEM/Ferrite.jl/pull/798/files#r1854906581) for Hcurl

test/test_interpolations.jl
```diff
@@ -252,6 +249,251 @@ using Ferrite: reference_shape_value, reference_shape_gradient
...
+
+    # Required properties of shape value Nⱼ of an edge-elements (Hdiv) on an edge with normal n, length L, and dofs ∈ 𝔇
+    # 1) Unit property: ∫(Nⱼ ⋅ n f(s) dS) = 1/length(𝔇) ∀ j ∈ 𝔇
+    #    Must hold for
+    #    length(𝔇) ≥ 1: f(s) = 1
+    #    length(𝔇) = 2: f(s) = 1 - s or f(s) = s for 1st and 2nd dof, respectively.
+    #    Additionally, should be zero for
+    #    length(𝔇) = 2: f(s) = s or f(s) = 1 - s for 1st and 2nd dof, respectively.
+    #    s is the path parameter ∈[0,1] along the positive direction of the path.
+    # 2) Zero normal component on other edges: Nⱼ ⋅ n = 0 if j∉𝔇
```
KnutAM, line 322: Associated test definitions for [comment above](https://github.com/Ferrite-FEM/Ferrite.jl/pull/798/files#r1854906581) for Hdiv

src/interpolations.jl
KnutAM: The advantage is that this way, we don't have to do strange modifications to the values when applying boundary conditions, see the tests at the end of test_interpolations.jl

src/FEValues/EdgeValues.jl
termi-official, line 1: Can we put this feature in a separate PR?

KnutAM: Yes, it will. 
Sorry that didn't update much on the plan with this PR, but it will need to be split into several sub-PRs. However, it is easier to work with the full one, and it can also serve as a reference when reviewing the sub-PRs later.

docs/src/devdocs/interpolations.md
```diff
@@ -49,3 +49,24 @@ dofs defined on a specific entity. Hence, not overloading of the dof functions w
 element with zero dofs. Also, it should always be double checked that everything is consistent as
 specified in the docstring of the corresponding function, as inconsistent implementations can
 lead to bugs which are really difficult to track down.
+
+## Vector interpolation properties
+### Hdiv interpolations
```
termi-official, line 54:
```suggestion
### H(div) interpolations
```

docs/src/devdocs/interpolations.md
termi-official, line 1: It might make sense to add some comments about the De-Rahm complex here for readers not familiar with the topic.

docs/src/literate-tutorials/maxwell_good_bad_ugly.jl
```diff
@@ -9,9 +9,9 @@ solution for a Maxwell problem, when vectorized `Lagrange` interpolations conver
 
 ![Results for different discretizations](maxwell.png)
 
-**Figure 1**: The results of this tutorial, showing how the analytical solution is not found when discretizing
+**Figure 1**: The results of this tutorial, showing how the exact solution is not found when discretizing
```
termi-official, line 12: Stylistic detail, but maybe we can get rid of the negation here. Maybe something like:
```suggestion
**Figure 1**: The results of this tutorial, showing how the approximation converges to the wrong solution when discretizing
```

docs/src/literate-tutorials/maxwell_good_bad_ugly.jl
```diff
@@ -219,28 +213,28 @@ function create_data(tr::Triangulation, fieldname::Symbol, a; f = identity)
     return data
 end
 
-# ## Analytical implementation
+# ## Exact implementation
```
termi-official, line 216: Stylistic again, but `exact implementation` does sound off.

KnutAM: Yes, but it would be good to highlight that this refers to a chosen solution, not an analytical solution to an existing problem. But I'm not used to working with such "reverse" problems, so perhaps that is still conventionally called "analytical"?

KnutAM: And btw, I'll open a separate PR with the tutorials in a while. 
Working on separating out the functionality changes. Just keeping this branch including all just to avoid missing anything...

termi-official: You mean `manufactured solution`?

KnutAM: I think it's different, because it is one solution that fulfills the pde, exactly so no source terms are added. The influence comes from the boundary terms... 

In https://doi.org/10.1016/j.compstruc.2019.106175 they mention also "Method of exact solutions" but couldn't find any ref or explanation to this...

termi-official: Ohh, I see. I found an explanation for this term here https://onlinelibrary.wiley.com/doi/pdf/10.1002/fld.660 in 3.1. So essentially we are looking for analytical solutions for corner cases of the PDE.

KnutAM: Thanks for that paper that defines it. The statement, 
```
In the method of exact solutions, specialized cases are identied where exact solutions
exist to a given set of governing equations. For the Euler and Navier–Stokes sets of 
equations, there are only a limited number of exact solutions. Furthermore, these 
exact solutions may not exercise all of the terms in the governing equations. For
two infnite parallel plates, one moving relative to the other (Couette flow), the 
velocity profile is linear, hence the diffusion term, a second derivative of velocity, 
is identically zero and therefore would not be fully exercised.
```
I guess somewhat in that direction, but here we only (as I understood the reference page) choose some field fulfilling the pde without a source term, but the boundary conditions are found by evaluating the resulting reaction loads/boundary constraints. 

But since it is an exact fulfillment of the PDE, but not of the problem, I think exact makes sense. However, the header is still not good, probably better to have "Implementation of the exact solution" (or just "Exact solution").

termi-official:
What was the reason that we never merged two the tutorials?

KnutAM:
> What was the reason that we never merged two the tutorials?

my time to create PR

PR #799, opened 2023-09-27
Linear elasticity example

kimauth:
Here is a first suggestion for a linear elasticity example. The explanations could probably use a bit more work and I believe the logo.geo file for generating the mesh should be stored somewhere else? It's a start anyways :)


docs/src/literate-tutorials/logo.geo
termi-official, line 1: Assets folder maybe? Or we can directly use FerriteGmsh to generate it inplace.

docs/src/literate-tutorials/heat_equation.jl
```diff
@@ -173,9 +173,7 @@ end
 #
 # !!! note "Notation"
 #     Comparing again with [Introduction to FEM](@ref), `f` and `u` correspond to
-#     $\underline{\hat{f}}$ and $\underline{\hat{u}}$, since they represent the discretized
-#     versions. However, through the code we use `f` and `u` instead to reflect the strong
-#     connection between the weak form and the Ferrite implementation.
+#     $\underline{\hat{f}}$ and $\underline{\hat{u}}$, since they represent the discretized versions. However, through the code we use `f` and `u` instead to reflect the strong connection between the weak form and the Ferrite implementation.
```
termi-official, line 176: Automatic formating?

termi-official:
I guess this should be the second tutorial and it is the step from a scalar-valued linear PDE to a vector-valued linear PDE? So the "new concents" would be a first contact with Tensors.jl and how to treat vector-valued problems in general?

docs/src/literate-tutorials/heat_equation.jl
kimauth: Upps, I'll revert that.

docs/src/literate-tutorials/logo.geo
kimauth: Moved it to the assets folder now, should (hopefully) work the same way as it does in the homogenization example.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,243 @@
...
+#
+# ## Introduction
+#
+# The classical first finite element problem to solve in solid mechanics is a linear balance
+# of momentum problem. We will use this to introduce a vector valued field, as well as the
+# [`Tensors.jl`](https://github.com/Ferrite-FEM/Tensors.jl) toolbox.
+#
+# The strong form of the balance of momentum is given by
+# ```math
+#  -\boldsymbol{\sigma} \cdot \boldsymbol{\nabla} = \boldsymbol{b}  \quad \textbf{x} \in \Omega,
```
termi-official, line 20:
```suggestion
#  -\boldsymbol{\nabla} \cdot \boldsymbol{\sigma}(\boldsymbol{u}(\textbf{x})) = \boldsymbol{b}(\textbf{x})  \quad \textbf{x} \in \Omega,
```
maybe to highlight the dependencies?

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,243 @@
...
+# # [Linear elasticity](@id tutorial-linear-elasticity)
 #
-# TBW
+# ![](linear_elasticity.png)
+#
+# *Figure 1*: Linear elastically deformed Ferrite logo.
+#
+#md # !!! tip
+#md #     This example is also available as a Jupyter notebook:
+#md #     [`heat_equation.ipynb`](@__NBVIEWER_ROOT_URL__/examples/heat_equation.ipynb).
```
koehlerson, line 9:
```suggestion
#md #     [`linear_elasticity.ipynb`](@__NBVIEWER_ROOT_URL__/examples/linear_elasticityipynb).
```

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,243 @@
...
+#   \boldsymbol{\nabla} \otimes \boldsymbol{u}
+#   +
+#   \boldsymbol{u} \otimes \boldsymbol{\nabla}
+# \right)
+# ```
+#
+# The resulting weak form is given given as follows: Find ``\boldsymbol{u} \in \mathbb{U}`` such that
+# ```math
+# \int_\Omega 
+#   \boldsymbol{\sigma} : \left(\delta \boldsymbol{u} \otimes \boldsymbol{\nabla} \right)
```
fredrikekre, line 42: Testfunction to the left, please :)

fredrikekre (commented):
I think it would be good to add a note about plane strain/stress in this example too.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,243 @@
...
+            ∇Nᵢ = shape_gradient(cellvalues, q_point, i)# shape_symmetric_gradient(cellvalues, q_point, i)
+            fe[i] += σ ⊡ ∇Nᵢ * dΩ # add internal force to residual
+            for j in 1:n_basefuncs
+                ∇ˢʸᵐNⱼ = shape_symmetric_gradient(cellvalues, q_point, j)
+                ke[i, j] += (∂σ∂ε ⊡ ∇ˢʸᵐNⱼ) ⊡ ∇Nᵢ * dΩ
+            end
+        end
+    end
+end
+#md nothing # hide
```
termi-official, line 184: So you implement the solver via one Newton step if I understand this correctly? Maybe we should comment in this case.

docs/src/literate-tutorials/linear_elasticity.jl
kimauth: I'm not following either, why do you want the test function to the left?

docs/src/literate-tutorials/linear_elasticity.jl
kimauth: I usually do that, so this is sort of an artifact. Removing it now for making this truely linear.

docs/src/literate-tutorials/linear_elasticity.jl
kimauth: I left out the dependencies on x in order to make it more readable and easier to recognize. Do you think it's important for understanding? Perhaps an alternative is to highlight that in the variable explanation, i.e. something like "where \sigma(x) is the stress tensor" ?

kimauth: (The "x in Omega" at the end of the equation should go away though if I don't have x in the equation itself.)

termi-official: I think if someone without continuum mechanics background reads this, then he will have a hard time to follow due to the recursion. I usually prefer putting the dependence on the solution variable explictily in the equations, so it is clearer which terms are linear/nonlinear functions in the solution variables and which are independent. We can also add a bit more explanation below.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,232 @@
...
+# Note that the elastic stiffness tensor is constant. Thus is needs to be computed and once
+# and can then be used for all integration points.
+function assemble_cell!(ke, cellvalues, ∂σ∂ε)
+    fill!(ke, 0.0)
+
+    n_basefuncs = getnbasefunctions(cellvalues)
+    for q_point in 1:getnquadpoints(cellvalues)
+        dΩ = getdetJdV(cellvalues, q_point)
+        for i in 1:n_basefuncs
+            ∇Nᵢ = shape_gradient(cellvalues, q_point, i)# shape_symmetric_gradient(cellvalues, q_point, i)
```
termi-official, line 168: This should also be the symmetric gradient, right?

kimauth: I'm a bit undecisive what's the better way to write this. It's only valid to use symmetric gradient because of the symmetry of dsigma_depsilon. But of course that's symmetric in this case and it's also pretty common to use the symmetric version there.

termi-official: I would use the formally correct version in the implementation and elaborate in a note why the other choice is in this specific case also okay. What do you think?

kimauth:
Added a Neumann boundary condition instead of the Dirichlet boundary condition. Still need to add some explanation for that + a new picture.

docs/src/literate-tutorials/linear_elasticity.jl
KristofferC: It's the way.

KnutAM:
Updated the example to master and tried to make it more complete, as promised during the Swedish Mechanics days...
Shape functions are now on the left, CC @fredrikekre and @KristofferC :)

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,313 @@
...
+# Like for the Heat Equation example, we will use a unit square - but here we'll load the grid of the Ferrite logo!
+# This is done by downloading [`logo.geo`](logo.geo) and loading it using [`FerriteGmsh.jl`](https://github.com/Ferrite-FEM/FerriteGmsh.jl),
+using Downloads: download
+logo_mesh = "logo.geo"
+asset_url = "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/"
+isfile(logo_mesh) || download(string(asset_url, logo_mesh), logo_mesh)
+
+FerriteGmsh.Gmsh.initialize() # hide
+FerriteGmsh.Gmsh.gmsh.option.set_number("General.Verbosity", 2) #hide
+grid = togrid(logo_mesh);
```
termi-official, line 89:
```suggestion
grid = togrid(logo_mesh);
FerriteGmsh.Gmsh.finalize();
```

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -586,7 +586,7 @@ close(pvd);
 
 
 using Test                                                                      #hide
-if IS_CI                                                                        #hide
+if false # Disable to avoid sporadic doc build failures "if IS_CI"              #hide
```
termi-official, line 589: We should never remove a test, just because it fails, if the test is not inherently faulty...

termi-official (changes requested):

docs/src/literate-tutorials/ns_vs_diffeq.jl
KnutAM: Thanks for figuring this out - I did not see the connection via the global gmsh state!
(I see that my code comment misleads but in this PR I only removed the test temporarily be able to build the docs in CI)
Out of curiosity, why do we only test this when running CI?

KnutAM:
```suggestion
if IS_CI                                                                        #hide
```

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: We don't do this for any other tutorials as far as I can see, so presenting it this way is sufficiently clear, makes it more recognizable, and easier to read IMO.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: I've added a details block below with performance tips noting these possibilities.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+    ## Create a temporary array for the facet's local contributions to the external force vector
+    fe_ext = zeros(getnbasefunctions(facetvalues))
+    for face in FacetIterator(dh, facetset)
+        ## Update the facetvalues to the correct facet number
+        reinit!(facetvalues, face)
+        ## Reset the temporary array for the next facet
+        fill!(fe_ext, 0.0)
+        for qp in 1:getnquadpoints(facetvalues)
+            ## Calculate the global coordinate of the quadrature point.
+            x = spatial_coordinate(facetvalues, qp, getcoordinates(face))
```
KristofferC, line 180: Does it make sense to lift out the `getcoordinates`  call outside the `qp` loop? Not really for performance but to show that this is independent of the qp.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+apply!(K, f_ext, ch)
+u = K \ f_ext;
+
+# ### Exporting to VTK
+# To visualize the result we export the grid and our field, `u`,
+# to a VTK-file, which can be viewed in e.g. [ParaView](https://www.paraview.org/).
+# For fun we'll color the logo, and thus create cell data with numbers according
+# to the logo colors.
+color_data = zeros(Int, getncells(grid))
+colors = Dict(
```
KristofferC, line 319: The Julia logo colors are cute but I wonder if not showing how to output the stress or strain would be more instructive for your typical FEM beginner.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+    end
+    return f_ext
+end
+#md nothing #hide
+
+# ### Material behavior
+# Next, we need to define the material behavior.
+# In this tutorial, we use plane strain linear isotropic elasticity, with Hooke's law as
+# ```math
+# \boldsymbol{\sigma} = 2G \boldsymbol{\varepsilon}^\mathrm{dev} + 3K \boldsymbol{\varepsilon}^\mathrm{vol}
```
termi-official, line 200: There is a disconnect between this form and the form introduced in the beginning of the tutorial, which is not obvious for beginners (from my experience).

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+# where $G$ is the shear modulus and $K$ the bulk modulus. The volumetric, $\boldsymbol{\varepsilon}^\mathrm{vol}$,
+# and deviatoric, $\boldsymbol{\varepsilon}^\mathrm{dev}$ strains, are defined as
+# ```math
+# \begin{align*}
+# \boldsymbol{\varepsilon}^\mathrm{vol} &= \frac{\mathrm{tr}(\boldsymbol{\varepsilon})}{3}\boldsymbol{I}, \quad
+# \boldsymbol{\varepsilon}^\mathrm{dev} &= \boldsymbol{\varepsilon} - \boldsymbol{\varepsilon}^\mathrm{vol}
+# \end{align*}
+# ```
+#
+# Starting from Young's modulus, $E$, and Poisson's ratio, $\nu$, the shear and bulk modulus are
```
termi-official, line 211: You introduced E above as the elasticity tensor.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+# ```math
+# G = \frac{E}{2(1 + \nu)}, \quad K = \frac{E}{3(1 - 2\nu)}
+# ```
+# Finally, the stiffness tensor can be calculated as the derivative of stress wrt. strain.
+# Due to the linearity, we can calculate this at any point, and choose zero strain.
+Emod = 200e3 # Young's modulus [MPa]
+ν = 0.3      # Poisson's ratio [-]
```
termi-official, line 218: Do you agree that we might want to teach users to wrap these into a struct? Or is this too much for this tutorial?

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+            end
+        end
+    end
+    return ke
+end
+#md nothing #hide
+#
+#md # !!! details "Performance tips"
+#md #     1. In this tutorial, `∂σ∂ε`, is minor symmetric, and we could have used
+#md #        the symmetric part of test function gradient above.
```
termi-official, line 268: Why is this a performance tip?

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+#md #     2. The product `∇Nᵢ ⊡ ∂σ∂ε` can be done outside the "j"-loop, but is kept inside here
+#md #        to highlight the similarity to the finite element form.
+#md #
```
termi-official, line 271: 3. We can eliminate the fourth order tensor fully here if necessary (in the linear elastic model).

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+# where $\boldsymbol{\sigma}$ is the (Cauchy) stress tensor and $\boldsymbol{b}$ the body force.
+# The domain, $\Omega$, has the boundary $\Gamma$, consisting of a Dirichlet part, $\Gamma_\mathrm{D}$, and
+# a Neumann part, $\Gamma_\mathrm{N}$, with outward pointing normal vector $\boldsymbol{n}$.
+# $\boldsymbol{u}_\mathrm{D}$ denotes prescribed displacements on $\Gamma_\mathrm{D}$,
+# while $\boldsymbol{t}_\mathrm{N}$ the known tractions on $\Gamma_\mathrm{N}$.
+#
+# In this tutorial, we use linear elasticity, such that
+# ```math
+# \boldsymbol{\sigma} = \boldsymbol{\mathsf{E}} : \boldsymbol{\varepsilon}, \quad
+# \boldsymbol{\varepsilon} = \left(\boldsymbol{u} \otimes \boldsymbol{\nabla}\right)^\mathrm{sym}
```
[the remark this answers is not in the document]
KnutAM: Using `grad` instead

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+# The classical first finite element problem to solve in solid mechanics is a linear balance
+# of momentum problem. We will use this to introduce a vector valued field, the displacements
+# $\boldsymbol{u}(\boldsymbol{x})$. In addition, some features of the
+# [`Tensors.jl`](https://github.com/Ferrite-FEM/Tensors.jl) toolbox are demonstrated.
+#
+# ### Strong form
+# The strong form of the balance of momentum for quasi-static loading is given by
+# ```math
+# \begin{alignat*}{2}
+#   \boldsymbol{\sigma} \cdot \boldsymbol{\nabla} + \boldsymbol{b} &= 0 \quad &&\boldsymbol{x} \in \Omega \\
```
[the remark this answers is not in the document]
KnutAM: Ref: https://github.com/Ferrite-FEM/Tensors.jl/issues/210

Took the easy route of using `div` and `grad` :)

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Because I thought you'd save some computational time, but seems negligble and only detectable in 3d...
```julia
E4 = rand(SymmetricTensor{4,3});
xs = rand(SymmetricTensor{2,3});
x = rand(Tensor{2,3});
@btime dcontract($E4, $xs); # 3.200 ns (0 allocations: 0 bytes)
@btime dcontract($E4, $x);  # 3.500 ns (0 allocations: 0 bytes)
```

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: $E$ and $\mathsf{E}$, but maybe we should use $\matsf{C}$ instead.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: In this case I don't see a reason, since we only pass the stiffness tensor to the element routine.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Do you mean by calculating the volumetric and deviatoric terms separately? At least when I tried once, that was slower IIRC.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,354 @@
...
+# The generated grid lacks the facetsets for the boundaries, so we add them by using Ferrite's
+# [`addfacetset!`](@ref). It allows us to add facetsets to the grid based on coordinates.
+# Note that approximate comparison to 0.0 doesn't work well, so we use a tolerance instead.
+addfacetset!(grid, "top",    x -> x[2] ≈ 1.0) # faces for which x[2] ≈ 1.0 for all nodes
+addfacetset!(grid, "left",   x -> abs(x[1]) < 1e-6)
+addfacetset!(grid, "bottom", x -> abs(x[2]) < 1e-6);
+
+# ### Trial and test functions
+# In this tutorial, we use linear Lagrange functions, and apply the same functions as test and trial,
+# $\delta\boldsymbol{N}_i = \boldsymbol{N}_i$, following Galerkin's method.
```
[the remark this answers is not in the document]
KnutAM: Good catch (I only recalled from my basic course and didn't check that it was the proper name)! I decided to not include at all, to not add more theory than required for now.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Tried to clarify.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Hid this now then, and added stress calculation and export as both cell data and L2Projection.

docs/src/literate-tutorials/linear_elasticity.jl
KristofferC: I would argue the performance tips should all have been tested to see that they actually make a difference. Otherwise they are just noisy guesses.

termi-official: Can we add the benchmarks then? And also the hint that users shall not compute them separately (as this is also counterintuitive).

KnutAM: I've removed performance tips completely now. I don't think this is required for the introductory tutorial. Originally, I added the one about symmetric part in response to https://github.com/Ferrite-FEM/Ferrite.jl/pull/799#discussion_r1378740246, and then the hoisting of computation felt that it should also be mentioned.

Thinking more about it, this would rather fit in a separate performance tip how-to / advice section in the docs.

termi-official: xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/1023 for later

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,408 @@
...
+# +
+# \int_\Omega
+#   \delta \boldsymbol{u} \cdot \boldsymbol{b}
+# \, \mathrm{d}\Omega
+# \quad \forall \, \delta \boldsymbol{u} \in \mathbb{T},
+# ```
+# where $\mathbb{U}$ and $\mathbb{T}$ denote suitable trial and test function spaces.
+# $\delta \boldsymbol{u}$ is a vector valued test function and
+# $\boldsymbol{t} = \boldsymbol{\sigma}\cdot\boldsymbol{n}$ is the traction vector on
+# the boundary. In this tutorial, we will neglect body foces and the weak form reduces to
```
termi-official, line 64:
```suggestion
# the boundary. In this tutorial, we will neglect body forces (i.e. $\boldsymbol{b} = \boldsymbol{0}$) and the weak form reduces to
```

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,408 @@
...
+# ### Finite element form
+# Finally, the finite element form is obtained by introducing the finite element shape functions.
+# Since the displacement field, $\boldsymbol{u}$, is vector valued, we use vector valued shape functions
+# $\delta\boldsymbol{N}_i$ and $\boldsymbol{N}_i$ to approximate the test and trial functions:
+# ```math
+# \boldsymbol{u} \approx \sum_{i=1}^N \boldsymbol{N}_i (\boldsymbol{x}) \, \hat{u}_i
+# \qquad
+# \delta \boldsymbol{u} \approx \sum_{i=1}^N \delta\boldsymbol{N}_i (\boldsymbol{x}) \, \delta \hat{u}_i
+# ```
+# Here $N$ is the number of nodal variables, with $\hat{u}_i$ and $\delta\hat{u}_i$ representing the $i$-th nodal values.
```
termi-official, line 84:
```suggestion
# Here $N$ is the number of nodal variables, with $\hat{u}_i$ and $\delta\hat{u}_i$ representing the $i$-th nodal value.
```

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,408 @@
...
+# ```
+# where $\boldsymbol{\sigma}$ is the (Cauchy) stress tensor and $\boldsymbol{b}$ the body force.
+# The domain, $\Omega$, has the boundary $\Gamma$, consisting of a Dirichlet part, $\Gamma_\mathrm{D}$, and
+# a Neumann part, $\Gamma_\mathrm{N}$, with outward pointing normal vector $\boldsymbol{n}$.
+# $\boldsymbol{u}_\mathrm{D}$ denotes prescribed displacements on $\Gamma_\mathrm{D}$,
+# while $\boldsymbol{t}_\mathrm{N}$ the known tractions on $\Gamma_\mathrm{N}$.
+#
+# In this tutorial, we use linear elasticity, such that
+# ```math
+# \boldsymbol{\sigma} = \boldsymbol{\mathsf{C}} : \boldsymbol{\varepsilon}, \quad
```
termi-official, line 36: Maybe
```suggestion
# \boldsymbol{\sigma} = \boldsymbol{\mathbb{C}} : \boldsymbol{\varepsilon}, \quad
```
because at least on my machines the font is not really distinguishable from boldsymbol

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,408 @@
...
+#md nothing #hide
+
+# ### Material behavior
+# Next, we need to define the material behavior, specifically the elasticity tensor $\mathsf{C}$.
+# In this tutorial, we use plane strain linear isotropic elasticity, with Hooke's law as
+# ```math
+# \boldsymbol{\sigma} = 2G \boldsymbol{\varepsilon}^\mathrm{dev} + 3K \boldsymbol{\varepsilon}^\mathrm{vol}
+# ```
+# where $G$ is the shear modulus and $K$ the bulk modulus. The linearity of this expression implies that
+# can be represented by the general expression, $\boldsymbol{\sigma} = \mathsf{C}:\boldsymbol{\varepsilon}$,
```
termi-official, line 208: Sorry, I am not sure about the statement here. Maybe we can rephrase it a bit to be more precise here to not confuse user.

```suggestion
# where $G$ is the shear modulus and $K$ the bulk modulus. Note that the elasticity tensor can be expressed by the general expression, $\boldsymbol{\sigma} = \mathsf{C}:\boldsymbol{\varepsilon}$. We will use this form in the remaining tutorial.
```

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -1,3 +1,408 @@
...
+# G = \frac{E}{2(1 + \nu)}, \quad K = \frac{E}{3(1 - 2\nu)}
+# ```
+Emod = 200e3 # Young's modulus [MPa]
+ν = 0.3      # Poisson's ratio [-]
+
+Gmod = Emod / (2(1 + ν))  # Shear modulus
+Kmod = Emod / (3(1 - 2ν)) # Bulk modulus
+
+# Finally, we demonstrate `Tensors.jl`'s automatic differentiation capabilities when
+# calculating the stiffness tensor
```
termi-official, line 233: I think it is helpful to stick to a nomenclature, once introduced, and avoid synonyms, to not throw off readers without consolidated knowledge about linear elasticity (which might jump on the train of thought: Why is this now called differently? Is it a different object?).
```suggestion
# calculating the elasticity tensor
```

termi-official (changes requested):
I think we are getting closer. Also, can you run a on the file spell-checker locally?

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Changing to
```suggestion
# where $G$ is the shear modulus and $K$ the bulk modulus. This expression can be written as
# $\boldsymbol{\sigma} = \mathbb{C}:\boldsymbol{\varepsilon}$, with
```
hopefully this is easier.

docs/src/literate-tutorials/linear_elasticity.jl
KnutAM: Good catch! Now it says "elastic stiffness tensor" everywhere to be clear.

KnutAM:
> Also, can you run a on the file spell-checker locally?

typos gives no errors

termi-official (approved):

PR #802, opened 2023-09-29
More flexibility for internal types

### LINKED ISSUE #195: Support arbitrary precision

termi-official:
This PR fixes two weakly related things. 

First, it allows to actually pass the matrix type into the sparsity pattern constructor. Currently only SparseMatrixCSC with element type of Float64 was allowed. Now we can also use different element types. Other matrix types are not yet implemented and up to future PRs (also related to FerriteDistributed.jl).

The second thing is that at some code points we fixed the data types to Float64. This constraint has also been removed, allowing e.g. the usage of Float32. This change also opens the possibility to implement an example of a complex-valued problem (e.g. Schrödinger problem).

This PR precedes #766 , because it can be useful to use lower precision on the GPU, e.g. to increase throughput and because not every GPU has dedicated Float64 units.

Closes #195 .


KnutAM:
Nice! 
Just leaving the note here, not sure if already on the list to change, that many of the `shape_value(ip, \xi)` implementations hard-code Float64, but should be easy to change (might also not matter as it in many cases would convert to a container type of lower precision), e.g. https://github.com/Ferrite-FEM/Ferrite.jl/blob/383c0c5084ebf36f6be11eb81c7a141c9bd3fbbc/src/interpolations.jl#L535

termi-official:
Yea, this is why I have left it as a draft. I am not sure how to resolve this one properly, because there are several possibilities. First, we could make the value type part of e.g. the Lagrange struct parametrization. On the other hand, we also might be able to have extra dispatches on e.g. `shape_value`. Not entirely sure. But the body should be definitely in the correct precision!

KnutAM:
I think in most cases it should be as simple as 
```julia

function shape_value(ip::Lagrange{RefQuadrilateral, 1}, ξ::Vec{2}, i::Int)
    ξ_x = ξ[1]
    ξ_y = ξ[2]
    i == 1 && return (1 - ξ_x) * (1 - ξ_y) * 0.25
    i == 2 && return (1 + ξ_x) * (1 - ξ_y) * 0.25
    i == 3 && return (1 + ξ_x) * (1 + ξ_y) * 0.25
    i == 4 && return (1 - ξ_x) * (1 + ξ_y) * 0.25
    throw(ArgumentError("no shape function $i for interpolation $ip"))
end
```

to 

```julia
function shape_value(ip::Lagrange{RefQuadrilateral, 1}, ξ::Vec{2,T}, i::Int) where T
    ξ_x = ξ[1]
    ξ_y = ξ[2]
    i == 1 && return (oneunit(T) - ξ_x) * (oneunit(T) - ξ_y) / 4
    i == 2 && return (oneunit(T) + ξ_x) * (oneunit(T) - ξ_y) / 4
    i == 3 && return (oneunit(T) + ξ_x) * (oneunit(T) + ξ_y) / 4
    i == 4 && return (oneunit(T) - ξ_x) * (oneunit(T) + ξ_y) / 4
    throw(ArgumentError("no shape function $i for interpolation $ip"))
end
```
(I suppose this is what you meant with the body, but left it here anyways)

If supplied a quadrature rule with the correct precision/type of the xi-coordinates, I suppose it should work? But there are probably more entry points in some cases, such as the reference coordinates when using BCValues for Dirichlet boundary conditions.

termi-official:
Okay, this took longer than it should have taken. :D Interestingly 1 seems to work fine.

termi-official:
Following the suggestion by Knut in https://github.com/Ferrite-FEM/Ferrite.jl/pull/766#discussion_r1269085009 I also included a change from ArgumentError to BoundsError for interpolations (which might fix the issue in the linked PR).

src/interpolations.jl
termi-official, line 1: I am split about the design here. On the one hand it seems convenient to have the evaluation type as a type parameter. On the other hand we can also rearrange things to just deduce the evaluation type from the function calls. The problematic part is the `reference_coordinates` call, but this can be fixed by letting `reference_coordinates` accept some vector type to write into, i.e. `ref_coord_vector = Vector{Vec{2,Float32}}; reference_coordinates(ip, ref_coord_vector)`. 

I think I prefer this design, because I plan to provide eval function for interpolation stuff anyway for better GPU support and performance.

KnutAM: I also like the latter, having an extra type parameter on the interpolation which should imply the reference coordinate datatype doesn't make sense to me (I would have thought that referred to the shape function type instead then). 

And there are only a few places where `reference_coordinates` are used that I could find:
* [`apply_analytical!`](https://github.com/Ferrite-FEM/Ferrite.jl/blob/e9ae940a3b6965d78bfd1ae03827e7584bfddac4/src/Dofs/apply_analytical.jl#L56): Here we could simply take the coordinate's eltype from the grid
* [`evaluate_at_grid_nodes(proj, ...)`](https://github.com/Ferrite-FEM/Ferrite.jl/blob/e9ae940a3b6965d78bfd1ae03827e7584bfddac4/src/L2_projection.jl#L256) and [`evaluate_at_grid_nodes(dh, ...)`](https://github.com/Ferrite-FEM/Ferrite.jl/blob/e9ae940a3b6965d78bfd1ae03827e7584bfddac4/src/Dofs/DofHandler.jl#L898): Same here, can take info from the grid.
* [`BCValues`](https://github.com/Ferrite-FEM/Ferrite.jl/blob/e9ae940a3b6965d78bfd1ae03827e7584bfddac4/src/FEValues/face_values.jl#L207): A type parameter is already given as input

Why not have the signature for each interpolation as `reference_coordinates(T, ip)` and define `reference_coordinates(ip::Interpolation) = reference_coordinates(Float64, ip)` as the default? (Instead of providing the full vector as an input)?

Or would you like the ability to have `reference_coordinates(...)::StaticVector` or GPU array? In that case, perhaps we could define the reference coordinates as a tuple instead of an array, which could just be collected into the correct `AbstractArray` container? Not sure if this has any negative performance impact, but I would assume the compiler would figure this out, especially if we add an `@inline` to the default method.

KnutAM:
> Interestingly 1 seems to work fine.

Using `Int` would work as long as there are no units, but this would throw if someone tries to supply coordinate with units - but this will break a lot of other assumptions in the code base too, so perhaps not worth the extra hassle :)

src/interpolations.jl
termi-official: Not sure if we need these functions  on he GPU but I think it is nice to at least have the possibility in case a user runs into the scenario where he needs it.

src/interpolations.jl
```diff
@@ -451,7 +451,7 @@ function reference_coordinates(ip::DiscontinuousLagrange{RefTetrahedron,0,T}) wh
 end
 
 function shape_value(ip::DiscontinuousLagrange{shape, 0}, ::Vec{dim, T}, i::Int) where {dim, shape <: AbstractRefShape{dim}, T}
-    i > 1 && throw(ArgumentError("no shape function $i for interpolation $ip"))
+    i > 1 && throw(BoundsError(ip, i))
```
KnutAM, line 451: Just noting it here, since I looked here how you did for using it somewhere else. 
I'm not sure if the error message is informative enough.
```
ERROR: BoundsError: attempt to access Lagrange{RefQuadrilateral, 1, Nothing} at index [1]
```

termi-official: If there is a better way to do it in a type stable way I am happy to implement it. I chose `BoundsError` over `ArgumentsError` because the latter is incompatible with evaluating individual shape functions on the GPU due to the apparent type instability.

fredrikekre: BoundsError feels wrong here though. Is ArgumentError with a const string (no interpolation) also problematic?

termi-official: Why? The evaluation here acts like a lazy array access for the last index.

I remember that is also was an issue with constant strings, but I cannot quickly reproduce the issue in 1.9.3. Maybe this one has been fixed.

termi-official:
> > Interestingly 1 seems to work fine.
> 
> Using `Int` would work as long as there are no units, but this would throw if someone tries to supply coordinate with units - but this will break a lot of other assumptions in the code base too, so perhaps not worth the extra hassle :)

I think this is still interesting to explore. :)

test/test_interpolations.jl
```diff
@@ -98,18 +99,24 @@
...
-    @testset "dirac delta property of dof $dof" for dof in 1:n_basefuncs
-        for k in 1:n_basefuncs
-            N_dof = shape_value(interpolation, coords[dof], k)
-            if k == dof
-                @test N_dof ≈ 1.0
-            else
-                factor = interpolation isa Lagrange{RefQuadrilateral, 3} ? 200 : 4
-                @test N_dof ≈ 0.0 atol = factor * eps(Float64)
+        # Check for evaluation type stability of interpolation
+        @testset "type stability dof $dof" for dof in 1:n_basefuncs
```
KnutAM, line 103: Doesn't check for type-instabilities or?
```suggestion
        @testset "type correctness dof $dof" for dof in 1:n_basefuncs
```

test/test_interpolations.jl
```diff
@@ -153,19 +160,26 @@
...
+        v_interpolation_1 = interpolation^2
+        v_interpolation_2 = (d = 2; interpolation^d)
+        @test getnbasefunctions(v_interpolation_1) == getnbasefunctions(v_interpolation_2) ==
+            getnbasefunctions(interpolation) * 2
+        # pretty printing
+        @test repr("text/plain", v_interpolation_1) == repr(v_interpolation_1.ip) * "^2"
+        
+        # Check for evaluation type stability of vectorized interpolation
+        v_interpolation_3 = interpolation^ref_dim
+        @testset "vectorized case stability of dof $dof" for dof in 1:n_basefuncs
```
KnutAM, line 173:
```suggestion
        @testset "vectorized case type correctness of dof $dof" for dof in 1:n_basefuncs
```

KnutAM (commented):
Nice work! Quite a lot of detailed changes!

I've tried to go through carefully but especially in the ConstraintHandler, there seem to be several possibilities for `Float64` to sneak in. 
Not sure if it is necessary to prevent this as long as (a) we have type-stable code and (2) the output types are the correct types. I guess further improvements can be done in the GPU PR if issues arise with internal `Float64` then?

In addition to the JET testing, it would be nice to have a few `@btime` comparisons with master for usages of the `ConstraintHandler` with `Float64`.
For example
* `close!`
* `update!`
* `apply!`
* Creation of periodic boundary conditions
Where the 3 first are done for cases including affine constraints to avoid any regressions here. This could also, in addition, be tested with JET, if possible.

src/Dofs/ConstraintHandler.jl
```diff
@@ -90,11 +90,11 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     closed::ScalarWrapper{Bool}
 end
 
-function ConstraintHandler(dh::AbstractDofHandler)
+function ConstraintHandler(dh::AbstractDofHandler, T=Float64)
```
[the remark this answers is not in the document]
termi-official: Good point. Let us assume this coupling for now. If users will run into the scenario where we need more granularity of the types, then we can adjust on more specific use cases.

`T` is introduced to refer to everything related to the actual computations. I have not tried to decouple the individual concepts in this PR. We can merely have the discussion going about more involved types ( e.g. unitful or complex). However, I want to achieve for now that we can at least compute with different precision in a stable way.

src/Dofs/ConstraintHandler.jl
```diff
@@ -407,7 +407,7 @@ function update!(ch::ConstraintHandler, time::Real=0.0)
 end
 
 # for vertices, faces and edges
-function _update!(inhomogeneities::Vector{Float64}, f::Function, boundary_entities::Set{<:BoundaryIndex}, field::Symbol, local_face_dofs::Vector{Int}, local_face_dofs_offset::Vector{Int},
+function _update!(inhomogeneities::Vector{T}, f::Function, boundary_entities::Set{<:BoundaryIndex}, field::Symbol, local_face_dofs::Vector{Int}, local_face_dofs_offset::Vector{Int},
```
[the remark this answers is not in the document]
termi-official: I introduced this as a sanity check. So I assume this does not work and will just promote faulty arrays then?

src/Dofs/ConstraintHandler.jl
```diff
@@ -468,7 +468,7 @@ end
 
 # Saves the dirichlet boundary conditions to a vtkfile.
 # Values will have a 1 where bcs are active and 0 otherwise
-function WriteVTK.vtk_point_data(vtkfile, ch::ConstraintHandler)
+function WriteVTK.vtk_point_data(vtkfile, ch::ConstraintHandler{<: AbstractDofHandler,<: AbstractFloat})
```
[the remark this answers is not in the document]
termi-official: AFAIK VTK can just handle real-valued data natively. So other data types will require more special exports. Users will likely hit some kind of error way before coming this far, if they work with something different than float. However, I have not propagated <:AbstractFloat everywhere to see how far they come with their specific problem. I am not even sure what exactly to put into the docstrings to communicate this "restriction" properly.

src/Dofs/ConstraintHandler.jl
```diff
@@ -913,13 +913,13 @@ coordinate axes.
...
-struct PeriodicDirichlet
+struct PeriodicDirichlet{T}
     field_name::Symbol
     components::Vector{Int} # components of the field
     face_pairs::Vector{Pair{String,String}} # legacy that will populate face_map on add!
     face_map::Vector{PeriodicFacePair}
     func::Union{Function,Nothing}
-    rotation_matrix::Union{Matrix{Float64},Nothing}
+    rotation_matrix::Union{Matrix{T},Nothing}
 end
```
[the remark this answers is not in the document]
termi-official: I tried to not touch the ConstraintHandler (construction) logic, because I really don't want to break any assumption here -- I don't have an overview of what all the assumptions taken are. Hence I would leave refining/cleaning up everything related to the ConstraintHandler to future PRs.

src/Dofs/ConstraintHandler.jl
```diff
@@ -930,14 +930,15 @@ PeriodicDirichlet(fn::Symbol, fp::Union{Vector{<:Pair},Vector{PeriodicFacePair}}
 # add!(::ConstraintHandler, ...) instead
 function PeriodicDirichlet(fn::Symbol, fp::Vector{<:Pair}, f::Union{Function,Nothing}, c=nothing)
     face_map = PeriodicFacePair[] # This will be populated in add!(::ConstraintHandler, ...) instead
-    return PeriodicDirichlet(fn, __to_components(c), fp, face_map, f, nothing)
+    return PeriodicDirichlet{Nothing}(fn, __to_components(c), fp, face_map, f, nothing)
```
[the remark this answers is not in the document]
termi-official: No, why should it?

```julia
julia> periodic = PeriodicDirichlet(
           :u,
           ["left" => "right", "bottom" => "top"],
           [1, 2]
       )
ERROR: UndefVarError: `T` not defined
Stacktrace:
 [1] PeriodicDirichlet(field_name::Symbol, components::Vector{Int64}, face_pairs::Vector{Pair{String, String}}, face_map::Vector{PeriodicFacePair}, func::Nothing, rotation_matrix::Nothing)
   @ Ferrite ~/Repos/Ferrite.jl/src/Dofs/ConstraintHandler.jl:917
 [2] PeriodicDirichlet
   @ ~/Repos/Ferrite.jl/src/Dofs/ConstraintHandler.jl:933 [inlined]
 [3] PeriodicDirichlet(fn::Symbol, fp::Vector{Pair{String, String}}, c::Vector{Int64})
   @ Ferrite ~/Repos/Ferrite.jl/src/Dofs/ConstraintHandler.jl:926
 [4] top-level scope
   @ REPL[24]:1
```

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: Makes sense - I just find that there are a bit too long input args here for making the dispatch easily readable, but that's a separate issue. Can be kept as is then IMO!

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: In this case, I think it would be better to put this somewhere else for writing to VTK. I won't include it in #692, but can be done as a follow-up on that and this PR to give better error messages in case they are bad. 

In this particular case, I don't think having the dofs as complex-valued is a problem because it should only write a zero or one, depending on whether the dof is constrained or not.

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: Ah, sorry - yes. (That's what I don't like about the parameterization, now there are two different "nothing"s)

But if `rotation_matrix::Matrix{Float64}` (as below) it works, right?

src/Dofs/ConstraintHandler.jl
```diff
@@ -1309,7 +1310,7 @@ end
 
 Same as [`collect_periodic_faces`](@ref) but adds all matches to the existing `face_map`.
 """
-function collect_periodic_faces!(face_map::Vector{PeriodicFacePair}, grid::Grid, mset::Union{Set{FaceIndex},String}, iset::Union{Set{FaceIndex},String}, transform::Union{Function,Nothing}=nothing; tol::Float64=1e-12)
+function collect_periodic_faces!(face_map::Vector{PeriodicFacePair}, grid::Grid{dim, <: AbstractCell, T}, mset::Union{Set{FaceIndex},String}, iset::Union{Set{FaceIndex},String}, transform::Union{Function,Nothing}=nothing; tol::T=T(1e-12)) where {dim, T}
```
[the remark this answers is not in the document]
termi-official: Good point. But I think we should not mix different precisions here, because it gives a false sense of security.

termi-official:
> I've tried to go through carefully but especially in the ConstraintHandler, there seem to be several possibilities for `Float64` to sneak in. Not sure if it is necessary to prevent this as long as (a) we have type-stable code and (2) the output types are the correct types. I guess further improvements can be done in the GPU PR if issues arise with internal `Float64` then?
>

I think I would like to fix such issues in this PR.
 
> In addition to the JET testing, it would be nice to have a few `@btime` comparisons with master for usages of the `ConstraintHandler` with `Float64`. For example
> 
>     * `close!`
> 
>     * `update!`
> 
>     * `apply!`
> 
>     * Creation of periodic boundary conditions
>       Where the 3 first are done for cases including affine constraints to avoid any regressions here. This could also, in addition, be tested with JET, if possible.

```
using BenchmarkTools, Ferrite

dh = DofHandler(grid)
add!(dh, :u, ip)
close!(dh);

@btime ConstraintHandler(dh)

∂Ω = union(
    getfaceset(grid, "left"),
    getfaceset(grid, "right"),
    getfaceset(grid, "top"),
    getfaceset(grid, "bottom"),
);
@btime add!(ch, $dbc) setup=(ch = ConstraintHandler($dh)) evals=1;

function setup_ch(dh, dbc)
    ch = ConstraintHandler(dh)
    add!(ch, dbc)
    return ch
end
@btime close!(ch) setup=(ch = setup_ch($dh, $dbc)) evals=1;

function setup_ch(dh, dbc)
    ch = ConstraintHandler(dh)
    add!(ch, dbc)
    close!(ch)
    return ch
end
@btime update!(ch) setup=(ch = setup_ch($dh, $dbc)) evals=1;
```

shows only a difference in the ctor.

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
termi-official: Same issue below.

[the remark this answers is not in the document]
termi-official: Because we still have the union with function and nothing.

termi-official:
So I guess #818 precedes this. :D

src/Dofs/ConstraintHandler.jl
```diff
@@ -90,11 +90,13 @@ struct ConstraintHandler{DH<:AbstractDofHandler,T}
     closed::ScalarWrapper{Bool}
 end
 
-function ConstraintHandler(dh::AbstractDofHandler)
+ConstraintHandler(dh::AbstractDofHandler) = ConstraintHandler(Float64, dh)
+
+function ConstraintHandler(T, dh::AbstractDofHandler)
```
fredrikekre, line 95:
```suggestion
function ConstraintHandler(::Type{T}, dh::AbstractDofHandler) where T <: Number
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -69,9 +69,9 @@ struct AffineConstraint{T}
 end
 
 """
-    ConstraintHandler
+    ConstraintHandler([T=Float64], dh::AbstractDofHandler)
 
-Collection of constraints.
+A collection of constraints with type `T` on a dof handler `dh`.
```
fredrikekre, line 74:
```suggestion
A collection of constraints associated with the dof handler `dh`. `T` is the numeric type for stored values.
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -1123,15 +1126,15 @@ function _add!(ch::ConstraintHandler, pdbc::PeriodicDirichlet, interpolation::In
     # Any remaining mappings are added as homogeneous AffineConstraints
     for (k, v) in dof_map
         if dof_map_t === Int
-            ac = AffineConstraint(k, [v => 1.0], inhomogeneity_map === nothing ? 0.0 : inhomogeneity_map[k])
+            ac = AffineConstraint(k, [v => T(1.0)], inhomogeneity_map === nothing ? T(0.0) : inhomogeneity_map[k])
```
fredrikekre, line 1129:
```suggestion
            ac = AffineConstraint(k, [v => one(T)], inhomogeneity_map === nothing ? zero(T) : inhomogeneity_map[k])
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -1123,15 +1126,15 @@ function _add!(ch::ConstraintHandler, pdbc::PeriodicDirichlet, interpolation::In
...
             add!(ch, ac)
         else
             @assert inhomogeneity_map === nothing
             @assert rotation_matrix !== nothing
             for (i, ki) in pairs(k)
                 # u_mirror = R ⋅ u_image
                 vs = Pair{Int,eltype(T)}[v[j] => rotation_matrix[i, j] for j in 1:length(v)]
-                ac = AffineConstraint(ki, vs, 0.0)
+                ac = AffineConstraint(ki, vs, T(0.0))
```
fredrikekre, line 1137:
```suggestion
                ac = AffineConstraint(ki, vs, zero(T))
```

src/assembler.jl
```diff
@@ -4,10 +4,12 @@ struct Assembler{T}
     V::Vector{T}
 end
 
-function Assembler(N)
+Assembler(N) = Assembler(Float64, N)
+
+function Assembler(T, N)
```
fredrikekre, line 9:
```suggestion
function Assembler(::Type{T}, N) where T
```

src/Dofs/ConstraintHandler.jl
fredrikekre: See https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing, even though it shouldn't matter here.

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
termi-official: @fredrikekre after thinking more about this I think it makes sense to have two different structs for the constraint.  But I would leave this as future work, as this PR here is "only" concerned with paving the way for Float32 support (and possibly more types)

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: Resolved by https://github.com/Ferrite-FEM/Ferrite.jl/pull/802/commits/fbdfa2339b9aa0857e4963d7bcf21aa753ded8fc

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: Resolved / reverted in https://github.com/Ferrite-FEM/Ferrite.jl/pull/802/commits/1a097538b39ef3e32c9d175c8cfc493c5b38cd24

PR #803, opened 2023-09-29
remove allocations in reinit!(cv, xe)

kimauth:
The allocations were caused by the (potentially) unused type parameter `vdim`.

Since I was very confused about the meaning of `rdim`, `sdim` and `vdim` when I started fixing this, I also added a note about what they mean.

### DIFF
--- a/src/FEValues/cell_values.jl
+++ b/src/FEValues/cell_values.jl
@@ -97,6 +97,13 @@ function CellValues(::Type{T}, qr::QuadratureRule, ip::Interpolation, sgip::Scal
     return CellValues(T, qr, ip, VectorizedInterpolation(sgip))
 end
 
+################################################################# 
+# Note on dimensions:                                           #
+# sdim = spatial dimension (dimension of the grid)              # 
+# rdim = reference dimension (dimension in isoparametric space) #
+# vdim = vector dimension (dimension of the field)              #
+################################################################# 
+
 # Entrypoint for `ScalarInterpolation`s (rdim == sdim)
 function CellValues(::Type{T}, qr::QR, ip::IP, gip::VGIP) where {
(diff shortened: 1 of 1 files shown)


src/FEValues/cell_values.jl
```diff
@@ -97,6 +97,13 @@ function CellValues(::Type{T}, qr::QuadratureRule, ip::Interpolation, sgip::Scal
     return CellValues(T, qr, ip, VectorizedInterpolation(sgip))
 end
 
+################################################################# 
+# Note on dimensions:                                           #
+# sdim = spatial dimension (dimension of the grid)              #
```
termi-official, line 102:
```suggestion
# sdim = spatial dimension (dimension of the grid nodes)         # 
```

termi-official: We must have deleted them accidentally at some point. Sorry for the confusion. We also might want to add this to the devdocs or user docs to increase visibility.

termi-official (approved):
LGTM.

If you have an idea about how to catch such errors automatically, then I am happy to join the discussion. Also, can you open an issue or a discussion about the terminology issue you ran into, so we do not forget to fix it?

KnutAM:
One could add a test that it is allocation-free, 
e.g. `@test 0 == @allocated reinit(cv, x)` (just ensure that it compiled first)

kimauth:
I guess we could test that certain things are allocation free for instance, but not sure if it's worth the hassle. Nobody seems to have noticed this the past couple of months and now it's fixed.

kimauth:
Terminology is fine from my perspective now, I was just confused about `rdim`, `vdim` and `sdim`. That should be easy enough to understand for the next person touching this code now I hope. 😅

PR #806, opened 2023-10-02
Convenience constructors CellValues and FaceValues

KnutAM:
(Taken out and extended from #680)

With the new distinction between vector and scalar interpolations since #694, 
it is possible to make a convenience constructor of cell and face values, given a 
`SubDofHandler` (or `DofHandler` with a single `SubDofHandler`). 

Happy to hear feedback on this idea!

**Proposed syntax**
```julia
cv = CellValues(dh, fieldname; qr)
fv = FaceValues(dh, fieldname; qr)
```
where `qr::Union{Int,[Face]QuadratureRule}`. For standard quadrature rules, the syntax now allows e.g.,
```julia
cv = CellValues(dh, :u; qr=2)
fv = FaceValues(dh, :u, qr=1)
```
for single subdofhandler problems, and just replace `dh::DofHandler` with `sdh::SubDofHandler` for cases with multiple subdofhandlers. 

**Motivation**
The geometric interpolation matches the grid, without having to extract information by using e.g. `Ferrite.default_interpolation` on the user-side
Less verbose by not requiring the quadrature rule constructor when the standard quadrature points should be used.

### DIFF
--- a/src/convenience_constructors.jl
+++ b/src/convenience_constructors.jl
@@ -1,0 +1,52 @@
+# Convenience constructors often require multiple types to be defined.
+# Where necessary, these can be put in this file.
+
+_create_qr_rule(qr::QuadratureRule, args...) = qr 
+_create_qr_rule(order::Int, ::Interpolation{RefShape}) where {RefShape} = QuadratureRule{RefShape}(order)
+
+"""
+    CellValues(dh::DofHandler, fieldname::Symbol; qr::Union{Int,QuadratureRule})
+    CellValues(sdh::SubDofHandler, fieldname::Symbol; qr::Union{Int,QuadratureRule})
+
+Create a `CellValues` object by using the interpolation of `fieldname` in `dh` 
+or `sdh`, as well as the default geometric interpolation based on the cell type.
+A `DofHandler` input is only allowed with a single `SubDofHandler` in `dh`.
+The quadrature rule is created automatically as `QuadratureRule{dim,RefShape}(qr)`,
+where `dim` and `RefShape` are taken from `dh` or `sdh`, if `qr::Int` is given.
+"""
+function CellValues(dh::DofHandler, fieldname::Symbol; kwargs...)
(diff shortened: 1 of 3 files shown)


termi-official:
I absolutely like this! I very often run into the scenario where I have to manage deducing the correct geometric interpolation and it can easily lead to errors - especially for embedded elements! While I also think that Ferrite should be as low level as possible, I also think that this is now the correct level for the user-facing API.

I am just indecisive if we should provide the dispatch with the quadrature order (instead of just using the quadrature rule). Maybe it is also the design of the dispatches which makes this weird. Maybe the quadrature order dispatch should be 
```
CellValues(dh, fieldname, qr_order)
CellValues(dh, fieldname, qr_order, qr_symbol)
```
But then I am also not sure about qr_symbol, because it changes between refshape types.

termi-official:
Also, if we already think about convenience functions, what about dispatches for interpolations and quadrature rules taking in the grid or grid+cellset as input argument to deduce the ref shape type?

KnutAM:
> I absolutely like this!

Happy to hear!

> I am just indecisive if we should provide the dispatch with the quadrature order

I see the point that this is the only thing that is not fully inferrable from the input (i.e. the symbol). But since in practice all examples use the standard `QuadratureRule{RefShape}(order)` call, I thought this was a reasonable default, and if using a non-standard type (which, as you say depends on the refshape), I think it makes sense to pass the full rule. 

I think using keyword argument is better than positional argument for this case, as I think `CellValues(dh, :u; qr=1)` is more readable than `CellValues(dh, :u, 1)`

> Also, if we already think about convenience functions, what about dispatches for interpolations and quadrature rules taking in the grid or grid+cellset as input argument to deduce the ref shape type?

Interesting to explore possibilities for that too! I left a separate issue, #841 with my suggestion!

src/convenience_constructors.jl
termi-official, line 1: Why not into the FEValues folder?

KnutAM: BCValues in ConstraintHandler...

termi-official (approved):
LGTM.

PR #807, opened 2023-10-02
define 1-arg getcelltype with a subdofhandler

KnutAM:
Change the "shortcut" from `getcelltype(grid, subdofhandler)` to just `getcelltype(subdofhandler)`.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -46,12 +46,11 @@ function SubDofHandler(dh::DH, cellset) where {DH <: AbstractDofHandler}
     return sdh
 end
 
-# Shortcut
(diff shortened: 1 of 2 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -46,12 +46,11 @@ function SubDofHandler(dh::DH, cellset) where {DH <: AbstractDofHandler}
     return sdh
 end
 
-# Shortcut
-@inline getcelltype(grid::AbstractGrid, sdh::SubDofHandler) = getcelltype(grid, first(sdh.cellset))
+getcelltype(sdh::SubDofHandler) = getcelltype(get_grid(sdh.dh), first(sdh.cellset))
```
KnutAM, line 49:
```suggestion
@inline getcelltype(sdh::SubDofHandler) = getcelltype(get_grid(sdh.dh), first(sdh.cellset))
```

PR #809, opened 2023-10-03
getcellset(::SubDofHandler)

KnutAM:
Since we already have a function for getting the cellset from the grid, 
I find it logical to also provide `getcellset(::SubDofHandler)`, such that users
don't have to use direct field-access to obtain the `SubDofHandler`'s cellset.

### DIFF
--- a/ext/FerriteMetis.jl
+++ b/ext/FerriteMetis.jl
@@ -52,7 +52,7 @@ function Ferrite.compute_renumber_permutation(
         else
             count(couplings[sdhi][i, j] for i in 1:n, j in 1:n if i != j)
         end
-        buffer_length += entries_per_cell * length(sdh.cellset)
+        buffer_length += entries_per_cell * length(getcellset(sdh))
     end
     I = Vector{idx_t}(undef, buffer_length)
     J = Vector{idx_t}(undef, buffer_length)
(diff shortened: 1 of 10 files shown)


kimauth:
Any particular reason why you don't want to access the field itself? Since we don't have a "subdofhandler interface" I don't really see why this is problematic.
In case it's just because the field is technically internal, we could alternatively document that `SubDofHandler` has a `cellset` field, I don't think this is likely to change (similar to how the fields of `CellCache` are documented).

KnutAM:
Documenting the field is also an option.

I find that an access function for use outside `Ferrite.jl` is more user-friendly, e.g. when writing an assembly loop for one subdofhandler. And since we already have that function, I find such a dispatch logical. But for sure this is just a matter of taste.

termi-official:
This is a thing I would really like to discuss at FerriteCon, because not having an interface here causes some trouble downstream.

termi-official:
The name suggests that a set is returned. Do we want to enforce this? I played a bit around with using ranges instead of sets in https://github.com/Ferrite-FEM/Ferrite.jl/pull/625 which might be an efficient alternative for the case that subdofhandlers model physical subdomains.

KnutAM:
@fredrikekre suggested on slack that we could potentially support both 

* `getcells(::SubDofHandler) -> Vector{Int}`
* `getcellset(::SubDofHandler) -> OrderedSet{Int}`

Will be new feature, so to be done after 1.0

PR #812, opened 2023-10-07
Use 2 threads for docs build

KnutAM:
Just to check the impact of this


fredrikekre:
Nothing is multithreaded, right?

KnutAM:
Only the multithreading how-to :)

But was hoping that the precompilation would run faster, but about 10 % diff doesn't sound worth it to me...

Issue #811, opened 2023-10-07
Too tight tolerance for pyramid interpolation tests?


KnutAM:
Is this too tight tolerance for the test, or isn't the floating point error sufficiently low?
Leaving it here as a future todo, failed on Windows on this run: 
https://github.com/Ferrite-FEM/Ferrite.jl/actions/runs/6441434254/job/17491226328

```
Lagrange{RefPyramid, 2}(): Test Failed at D:\a\Ferrite.jl\Ferrite.jl\test\test_interpolations.jl:56
  Expression: sum([shape_value(interpolation, x, i) for i = 1:n_basefuncs]) ≈ 1.0
   Evaluated: 1.000000060768798 ≈ 1.0
```

lijas:
I am guessing this happens when z is close to one, where there is a singularity for pyramids.

Looking at the code that tests this ([ `x = rand(Tensor{1, ref_dim}) `.)](https://github.com/Ferrite-FEM/Ferrite.jl/blob/e9ae940a3b6965d78bfd1ae03827e7584bfddac4/test/test_interpolations.jl#L52C33-L52C33), it seems like the generated point can be outside the parametric bounds for triangles, pyramids etc. Maybe we only generate points within the parametric bounds? I dont know if that will fix this issue however.

KnutAM:
Nice! 
That sounds very reasonable to require!

I guess we could also add a printout of `x` so that we know that value in case it fails again?

lijas:
Sounds good. Is there a way to generate a point within a domain? I guess one could just generate a random point and check if it is inside the domain, but then the distrubtion of points will not be uniform in the domain I think.

KnutAM:
I think all reference shapes are convex, so a random linear combination with coefficients in [0,1] should work. Something like 
```julia
x_ref = reference_coordinates(ip)
i = rand(1:length(x_ref))
x = rand(length(x_ref)) .* (x_ref .- (x_ref[i],)) / (length(x_ref)-1) + x_ref[i]`
```

termi-official:
Btw, next time we hit tolerance issues we also might try out an automated expression optimization system like Herbie https://herbie.uwplse.org/doc/2.0/release-notes.html . I am not aware of a Julia equivalent.

PR #818, opened 2023-10-12
Uniform point sampling for reference geometry and interpolation test fix

termi-official:
Should address #811 .

### DIFF
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -50,25 +50,24 @@
         coords = Ferrite.reference_coordinates(interpolation)
         @test length(coords) == n_basefuncs
         f(x) = [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
+
+        #TODO prefer this test style after 1.6 is removed from CI
         # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
         x = sample_random_point(ref_shape)
-        @testset "Random point test" begin
+        random_point_testset = @testset "Random point test" begin
             # Check gradient evaluation
             @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
                 reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
             # Check partition of unity at random point.
-            ansatz_sum = sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs])
-            if ansatz_sum ≉ 1.0
-                # Show coordinate in case failure (see issue #811)
-                # Remove after 1.6 is removed from CI (see above)
-                println("Partition of unity test might fail at $x for $interpolation !")
-                @test ansatz_sum ≈ 1.0
-            end
+            @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
             # Check if the important functions are consistent
             @test_throws ArgumentError shape_value(interpolation, x, n_basefuncs+1)
             # Idempotency test
             @test shape_value(interpolation, x, n_basefuncs) == shape_value(interpolation, x, n_basefuncs)
         end
+        # Remove after 1.6 is removed from CI (see above)
+        # Show coordinate in case failure (see issue #811)
+        !isempty(random_point_testset.results) && println("^^^^^Random point test failed at $x for $interpolation !^^^^^")
 
         # Test whether we have for each entity corresponding dof indices (possibly empty)
         @test length(Ferrite.vertexdof_indices(interpolation)) == Ferrite.nvertices(interpolation)
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -51,7 +51,7 @@
         coords = Ferrite.reference_coordinates(interpolation)
         f = (x) -> [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
         # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
-        @testset "Random point test"
+        @testset "Random point test" begin
             x = sample_random_point(ref_shape)
             # Check gradient evaluation
             @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -62,7 +62,7 @@
                 # Show coordinate in case failure (see issue #811)
                 # Remove after 1.6 is removed from CI (see above)
                 println("Partition of unity test might fail at $x for $interpolation !")
-                @test ansatz_sum ≈ 1.0 atol=10*eps(Float64)
+                @test ansatz_sum ≈ 1.0
             end
             # Check if the important functions are consistent
             @test length(coords) == n_basefuncs
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -58,11 +58,11 @@
                 reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
             # Check partition of unity at random point.
             ansatz_sum = sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs])
-            if ansatz_sum !≈ 1.0
+            if ansatz_sum ≉ 1.0
                 # Show coordinate in case failure (see issue #811)
                 # Remove after 1.6 is removed from CI (see above)
-                println("Partition of unity test fails at $x for $interpolation !")
-                @test ansatz_sum ≈ 1.0
+                println("Partition of unity test might fail at $x for $interpolation !")
+                @test ansatz_sum ≈ 1.0 atol=10*eps(Float64)
             end
             # Check if the important functions are consistent
             @test length(coords) == n_basefuncs
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -49,7 +49,7 @@
 
         n_basefuncs = getnbasefunctions(interpolation)
         coords = Ferrite.reference_coordinates(interpolation)
-        f = (x) -> [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
+        f(x) = [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
         # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
         @testset "Random point test" begin
             x = sample_random_point(ref_shape)
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -37,7 +37,6 @@
                       #
                       CrouzeixRaviart{RefTriangle, 1}(),
     )
-
         # Test of utility functions
         ref_dim = Ferrite.getdim(interpolation)
         ref_shape = Ferrite.getrefshape(interpolation)
@@ -49,10 +48,11 @@
 
         n_basefuncs = getnbasefunctions(interpolation)
         coords = Ferrite.reference_coordinates(interpolation)
+        @test length(coords) == n_basefuncs
         f(x) = [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
         # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
+        x = sample_random_point(ref_shape)
         @testset "Random point test" begin
-            x = sample_random_point(ref_shape)
             # Check gradient evaluation
             @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
                 reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
@@ -65,7 +65,6 @@
                 @test ansatz_sum ≈ 1.0
             end
             # Check if the important functions are consistent
-            @test length(coords) == n_basefuncs
             @test_throws ArgumentError shape_value(interpolation, x, n_basefuncs+1)
             # Idempotency test
             @test shape_value(interpolation, x, n_basefuncs) == shape_value(interpolation, x, n_basefuncs)
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -50,7 +50,9 @@
         n_basefuncs = getnbasefunctions(interpolation)
         coords = Ferrite.reference_coordinates(interpolation)
         f = (x) -> [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
-        @testset let x = sample_random_point(ref_shape)
+        # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
+        @testset "Random point test"
+            x = sample_random_point(ref_shape)
             # Check partition of unity at random point.
             @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
                 reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -53,10 +53,17 @@
         # @testset let x = sample_random_point(ref_shape) # not compatible with Julia 1.6
         @testset "Random point test"
             x = sample_random_point(ref_shape)
-            # Check partition of unity at random point.
+            # Check gradient evaluation
             @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
                 reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
-            @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
+            # Check partition of unity at random point.
+            ansatz_sum = sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs])
+            if ansatz_sum !≈ 1.0
+                # Show coordinate in case failure (see issue #811)
+                # Remove after 1.6 is removed from CI (see above)
+                println("Partition of unity test fails at $x for $interpolation !")
+                @test ansatz_sum ≈ 1.0
+            end
             # Check if the important functions are consistent
             @test length(coords) == n_basefuncs
             @test_throws ArgumentError shape_value(interpolation, x, n_basefuncs+1)
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -47,19 +47,20 @@
         # Note that not every element formulation exists for every order and dimension.
         applicable(Ferrite.getlowerorder, interpolation) && @test typeof(Ferrite.getlowerorder(interpolation)) <: Interpolation{ref_shape,func_order-1}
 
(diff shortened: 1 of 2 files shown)


test/test_utils.jl
```diff
@@ -251,6 +251,28 @@ coords_on_faces(x, ::Serendipity{RefHexahedron, 2}) =
...
+    ξ₃ = rand(Float64)
+    return Vec{3}((ξ₁, ξ₂, ξ₃))
+end
+# TODO what to do here? The samplig is not uniform...
+function sample_random_point(::Type{RefPyramid})
+    ξ₃ = rand(Float64)
+    # If we fix a z coordinate we get a Quad with extends (1-ξ₃)
+    (ξ₁, ξ₂) = (1.0 - ξ₃) .* rand(Vec{2})
+    return Vec{3}((ξ₁, ξ₂, ξ₃))
+end
```
termi-official, line 274: I can't prove or disprove this for now. :/

termi-official:
Turns out the context capture for the test does not work on 1.6.

test/test_utils.jl
```diff
@@ -251,6 +251,28 @@ coords_on_faces(x, ::Serendipity{RefHexahedron, 2}) =
...
 check_equal_or_nan(a::Union{Tensor, Array}, b::Union{Tensor, Array}) = all(check_equal_or_nan.(a, b))
 
+# Hypercube is simply ⨂ᵈⁱᵐ Line :)
+sample_random_point(::Type{Ferrite.RefHypercube{ref_dim}}) where {ref_dim} = Vec{ref_dim}(2.0 .* rand(Vec{ref_dim}) .- 1.0)
+# Dirichlet type sampling
+function sample_random_point(::Type{Ferrite.RefSimplex{ref_dim}}) where {ref_dim} 
+    ξ = rand(Vec{ref_dim})
+    ξₜ = -log.(ξ)
+    return Vec{ref_dim}(ξₜ ./ sum(ξₜ))
+end
```
KnutAM, line 261: This one looks a bit strange to me. 
How can it get close to the origin?
My quick test shows that it is just on the far edge:
```julia
julia> function sample_random_point()
           ξ = rand(2)
           ξₜ = -log.(ξ)
           return (ξₜ ./ sum(ξₜ))
       end
sample_random_point (generic function with 1 method)

julia> x = zeros(1000); y = zeros(1000);

julia> for i in 1:length(x)
           x[i], y[i] = sample_random_point()
       end

julia> using UnicodePlots

julia> scatterplot(x, y)
     ┌────────────────────────────────────────┐
   1 │⠙⠦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠉⠲⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠙⠦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠲⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⠦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠲⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⠦⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢤⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠲⣄⠀⠀⠀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⠦⣀⠀⠀⠀⠀⠀│
     │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⢤⡀⠀⠀│
   0 │⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⠲⣄│
     └────────────────────────────────────────┘
     ⠀0⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀1⠀
```

test/test_utils.jl
```diff
@@ -251,6 +251,28 @@ coords_on_faces(x, ::Serendipity{RefHexahedron, 2}) =
...
+# Wedge = Triangle ⊗ Line
+function sample_random_point(::Type{RefPrism})
+    (ξ₁, ξ₂) = sample_random_point(RefTriangle)
+    ξ₃ = rand(Float64)
+    return Vec{3}((ξ₁, ξ₂, ξ₃))
+end
+# TODO what to do here? The samplig is not uniform...
```
KnutAM, line 268: I don't think uniform sampling is so important (as long as it is not super-skewed), the main point is that these points should (1) be inside the reference shape and (2) in practice cover all areas.

test/test_interpolations.jl
```diff
@@ -47,19 +47,20 @@
...
-    @test shape_value(interpolation, x, n_basefuncs) == shape_value(interpolation, x, n_basefuncs)
-    @test_throws ArgumentError shape_value(interpolation, x, n_basefuncs+1)
+        n_basefuncs = getnbasefunctions(interpolation)
+        coords = Ferrite.reference_coordinates(interpolation)
+        f = (x) -> [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
+        @testset let x = sample_random_point(ref_shape)
+            # Check partition of unity at random point.
+            @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
+                reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
+            @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
```
KnutAM, line 57: Since this was the problem in #811, perhaps it is good to add a temporary show here in case the test fails in the future, to make it possible to debug the problem then?
```suggestion
            @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
            if sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) !≈ 1.0
                @show x # Show coordinate in case failure, see issue #811 
            end
```

test/test_utils.jl
KnutAM: This looks reasonable, the fact that the point density will be higher towards the top seems fine. However, considering that we know that the solution is discontinuous at the very top, perhaps avoid the very top as suggested above?

test/test_utils.jl
```diff
@@ -251,6 +251,28 @@ coords_on_faces(x, ::Serendipity{RefHexahedron, 2}) =
...
+function sample_random_point(::Type{RefPyramid})
+    ξ₃ = rand(Float64)
```
KnutAM, line 270:
```suggestion
    ξ₃ = (1-1e-3)*rand(Float64) # Derivative is discontinuous at the top
```

test/test_utils.jl
termi-official: Oopsie is fixed in latest commit. Sampling was in the wrong dimension. Forgot to revert after getting lost on the broadcasting issue.

test/test_utils.jl
```diff
@@ -251,6 +251,28 @@ coords_on_faces(x, ::Serendipity{RefHexahedron, 2}) =
 check_equal_or_nan(a::Any, b::Any) = a==b || (isnan(a) && isnan(b))
 check_equal_or_nan(a::Union{Tensor, Array}, b::Union{Tensor, Array}) = all(check_equal_or_nan.(a, b))
 
+# Hypercube is simply ⨂ᵈⁱᵐ Line :)
+sample_random_point(::Type{Ferrite.RefHypercube{ref_dim}}) where {ref_dim} = Vec{ref_dim}(2.0 .* rand(Vec{ref_dim}) .- 1.0)
+# Dirichlet type sampling
+function sample_random_point(::Type{Ferrite.RefSimplex{ref_dim}}) where {ref_dim}
```
KnutAM, line 257: Looks beautiful :) Posting here as future reference
```julia
julia> using UnicodePlots

julia> function sample_random_point()
           ξ = rand(3)
           ξₜ = -log.(ξ)
           return ξₜ[1:2] / sum(ξₜ)
       end
sample_random_point (generic function with 1 method)

julia> x = zeros(1000); y = zeros(1000);

julia> for i in 1:length(x)
           x[i], y[i] = sample_random_point()
       end

julia> scatterplot(x, y)
```

image

KnutAM (approved):
Nice. 
Should fix the `ansatz_sum ≈ 1.0 atol=10*eps(Float64)` line, otherwise good to go!

test/test_interpolations.jl
```diff
@@ -47,19 +47,29 @@
...
-    @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
-
-    # Check if the important functions are consistent
-    coords = Ferrite.reference_coordinates(interpolation)
-    @test length(coords) == n_basefuncs
-    @test shape_value(interpolation, x, n_basefuncs) == shape_value(interpolation, x, n_basefuncs)
-    @test_throws ArgumentError shape_value(interpolation, x, n_basefuncs+1)
+        n_basefuncs = getnbasefunctions(interpolation)
+        coords = Ferrite.reference_coordinates(interpolation)
+        f = (x) -> [shape_value(interpolation, Tensor{1, ref_dim}(x), i) for i in 1:n_basefuncs]
```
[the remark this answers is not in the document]
termi-official: Isn't this literally the same? I understand the recommendation above e.g. as not doing `ForwardDiff.jacobian(x->f(x), Array(x))')` instead of `ForwardDiff.jacobian(f, Array(x))')`, or what am I missing here?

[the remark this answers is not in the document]
KnutAM: No, not literally, and it is in the test so not crucial, but this makes `f` not being a `const`, which is bad for type-stability etc. The recommendation I guess is just because (1) it means compiling another specialization if used in multiple places for the same `f` and (2) because it looks ugly :)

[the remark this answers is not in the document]
termi-official: Interesting. Can you add this to https://docs.julialang.org/en/v1/manual/style-guide/#Do-not-write-x-f(x) ? I think this is quite a crucial piece of information missing in the docs then.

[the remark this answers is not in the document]
KnutAM: I'm not sure how crucial, because I think it only matters when you are working in the global scope (as we do here during testing), e.g. 
```julia
julia> using BenchmarkTools

julia> f(x) = sin(x)
f (generic function with 1 method)

julia> g = x -> sin(x)
#3 (generic function with 1 method)

julia> function foo(x)
           foofun(y) = sin(y)
           return foofun(x)
       end
foo (generic function with 1 method)

julia> function bar(x)
           barfun = y -> sin(y)
           return barfun(x)
       end
bar (generic function with 1 method)

julia> @btime f(1.0);
  0.800 ns (0 allocations: 0 bytes)

julia> @btime g(1.0); #Solved by @btime $g(1.0)
  19.659 ns (1 allocation: 16 bytes)

julia> @btime foo(1.0);
  0.800 ns (0 allocations: 0 bytes)

julia> @btime bar(1.0);
  0.800 ns (0 allocations: 0 bytes)
```

test/test_utils.jl
termi-official: Any feelings regarding putting this into Ferrite instead of docs only? I think I have a legit use case for this already in the point finder (within its current design at least).

KnutAM: As procrastination I looked at the initial guessing a bit, and perhaps a reproducible way is better? For testing I think this is nice, but for actual computations I'm not currently in favor of introducing randomness in the code. 

Here is a code-snippet with some trial changes, where the guess is calculated based on the closest point on a line between the two closest cell coordinates. 
The main internally breaking change would be that `reference_coordinates` should return a `Tuple` instead of `Vector` to avoid allocations...
```julia
# Find the two closest cell coordinates to the target coordinate
function find_two_closest(x_cell, x_target)
    i1 = 0
    i2 = 0
    d1 = typemax(eltype(x_target))
    d2 = d1
    for (i, x) in pairs(x_cell)
        d = norm(x-x_target)
        if d < d1
            i2 = i1
            d2 = d1
            i1 = i
            d1 = d
        end
    end
    return i1, i2, d2/(d1+d2)
end

# Define this function as a simple test for testing `QuadraticQuadrilateral`
function reference_coords()
    return (Vec{2, Float64}((-1.0, -1.0)),
            Vec{2, Float64}(( 1.0, -1.0)),
            Vec{2, Float64}(( 1.0,  1.0)),
            Vec{2, Float64}((-1.0,  1.0)),
            Vec{2, Float64}(( 0.0, -1.0)),
            Vec{2, Float64}(( 1.0,  0.0)),
            Vec{2, Float64}(( 0.0,  1.0)),
            Vec{2, Float64}((-1.0,  0.0)),
            Vec{2, Float64}(( 0.0,  0.0)))
end

# See https://discourse.julialang.org/t/finding-the-value-of-a-field-at-a-spatial-location-in-juafem/38975/2
# TODO: should we make iteration params optional keyword arguments?
function find_local_coordinate(interpolation, cell_coordinates::Vector{V}, global_coordinate::V) where {dim, T, V <: Vec{dim, T}}
    n_basefuncs = getnbasefunctions(interpolation)
    @assert length(cell_coordinates) == n_basefuncs
    i1, i2, w = find_two_closest(cell_coordinates, global_coordinate)
    ξ1 = reference_coords()[i1]
    ξ2 = reference_coords()[i2]
    local_guess = ξ1*w + ξ2*(1-w)
    #local_guess = zero(V)
    max_iters = 10
    tol_norm = 1e-10
    converged = false
    local niter
    for iter in 1:max_iters
        global_guess = zero(V)
        J = zero(Tensor{2, dim, T})
        for j in 1:n_basefuncs
            dNdξ, N = shape_gradient_and_value(interpolation, local_guess, j)
            global_guess += N * cell_coordinates[j]
            J += cell_coordinates[j] ⊗ dNdξ
        end
        residual = global_guess - global_coordinate
        if norm(residual) <= tol_norm
            converged = true
            niter = iter
            break
        end
        local_guess -= inv(J) ⋅ residual
    end
    return converged, local_guess, niter
end
```

termi-official: For this case we can introduce a function `reference_coordinates!` which just writes into a buffer. :)

test/test_interpolations.jl
```diff
@@ -47,19 +47,29 @@
...
+            @test vec(ForwardDiff.jacobian(f, Array(x))') ≈
+                reinterpret(Float64, [shape_gradient(interpolation, x, i) for i in 1:n_basefuncs])
+            # Check partition of unity at random point.
+            ansatz_sum = sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs])
+            if ansatz_sum ≉ 1.0
+                # Show coordinate in case failure (see issue #811)
+                # Remove after 1.6 is removed from CI (see above)
+                println("Partition of unity test might fail at $x for $interpolation !")
+                @test ansatz_sum ≈ 1.0
+            end
```
KnutAM, line 66: Not necessary, but learned this on Slack today, perhaps a nicer way to print if it fails?
(Haven't checked julia version compat for this)
```suggestion
            ansatz_test = @test sum([shape_value(interpolation, x, i) for i in 1:n_basefuncs]) ≈ 1.0
            if ansatz_test isa Test.Fail
                # Show coordinate in case failure (see issue #811)
                # Remove after 1.6 is removed from CI (see above)
                println("Partition of unity test failed at $x for $interpolation !")
            end
```

KnutAM (approved):
LGTM!
(Nightly failure is unrelated)

test/test_interpolations.jl
termi-official: Maybe better like this? So we have also context if another test in this battery fails.

KnutAM: Yes! But looking forward to skipping 1.6 from CI:)

termi-official:
@fredrikekre this can be merged if you are okay with it. Should reduce some of the annoying CI failures experienced recently. And if it fails in the test, we can now reconstruct which exact point fails to eval.

Issue #763, opened 2023-07-12
Bug in PointValues


KnutAM:
Shouldn't `dMdξ` be updated here as well with the new quadrature point location...

https://github.com/Ferrite-FEM/Ferrite.jl/blob/fd0b99930394cafb4b5662cfb5f17a1694b01fec/src/PointEval/point_values.jl#L59-L66

PR #820, opened 2023-10-13
Test coverage for PointEval with nonlinear geometries

termi-official:
Point evaluation on master just does not work on nonlinear geometries. Here is an attempt to make it work. Fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/763 but might not completely fix the point eval.

## TODO

* [x] docs
* [x] ~~flexibility for inner solver~~ not sure how. ~~Should be a separate PR.~~ How is now included during review.
* [x] expose inner solver parameters
* [x] fix point eval for mixed-dimensional and embedded grids

### DIFF
--- a/src/FEValues/GeometryMapping.jl
+++ b/src/FEValues/GeometryMapping.jl
@@ -121,11 +121,11 @@ function otimes_returntype(#=typeof(x)=#::Type{<:Vec{dim,Tx}}, #=typeof(d2Mdξ2)
     return Tensor{3,dim,promote_type(Tx,TM)}
 end
 
-@inline function calculate_mapping(::GeometryMapping{0}, q_point, x)
+@inline function calculate_mapping(::GeometryMapping{0}, q_point::Int, x::AbstractVector{<:Vec})
     return MappingValues(nothing, nothing)
 end
 
-@inline function calculate_mapping(geo_mapping::GeometryMapping{1}, q_point, x)
+@inline function calculate_mapping(geo_mapping::GeometryMapping{1}, q_point::Int, x::AbstractVector{<:Vec})
     fecv_J = zero(otimes_returntype(eltype(x), eltype(geo_mapping.dMdξ)))
     @inbounds for j in 1:getngeobasefunctions(geo_mapping)
         #fecv_J += x[j] ⊗ geo_mapping.dMdξ[j, q_point]
@@ -148,6 +148,21 @@ end
 
 calculate_detJ(J::Tensor{2}) = det(J)
 calculate_detJ(J::SMatrix) = embedding_det(J)
+
+function calculate_jacobian_and_spatial_coordinate(gip::ScalarInterpolation, ξ::Vec{rdim,Tξ}, x::AbstractVector{<:Vec{sdim, Tx}}) where {Tξ, Tx, rdim, sdim}
+    n_basefuncs = getnbasefunctions(gip)
+    @boundscheck checkbounds(x, Base.OneTo(n_basefuncs))
+
+    fecv_J = zero(otimes_returntype(Vec{sdim,Tx}, Vec{rdim,Tξ}))
+    sx = zero(Vec{sdim, Tx})
+    @inbounds for j in 1:n_basefuncs
+        dMdξ, M = reference_shape_gradient_and_value(gip, ξ, j)
+        sx += M * x[j]
+        fecv_J += otimes_helper(x[j], dMdξ)
+    end
+    return fecv_J, sx
+end
+
 
 # Embedded
 
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -329,7 +329,10 @@ Compute the spatial coordinate in a quadrature point. `x` contains the nodal
 coordinates of the cell.
 
 The coordinate is computed, using the geometric interpolation, as
-``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{x}) \\mathbf{\\hat{x}}_i``
+``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{\\xi}) \\mathbf{\\hat{x}}_i``.
+
+where ``\\xi``is the coordinate of the given quadrature point `q_point` of the associated
+quadrature rule.
 """
 function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVector{<:Vec})
     n_base_funcs = getngeobasefunctions(fe_v)
@@ -342,6 +345,24 @@ function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVecto
     return vec
 end
 
+"""
+    spatial_coordinate(ip::ScalarInterpolation, ξ::Vec, x::AbstractVector{<:Vec{sdim, T}})
+
+Compute the spatial coordinate in a given quadrature point. `x` contains the nodal coordinates of the cell.
+
+The coordinate is computed, using the geometric interpolation, as
+``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{\\xi}) \\mathbf{\\hat{x}}_i``
+"""
+function spatial_coordinate(interpolation::ScalarInterpolation, ξ::Vec, x::AbstractVector{<:Vec})
+    n_basefuncs = getnbasefunctions(interpolation)
+    @boundscheck checkbounds(x, Base.OneTo(n_basefuncs))
+    vec = zero(eltype(x))
+    @inbounds for j in 1:n_basefuncs
+        M = reference_shape_value(interpolation, ξ, j)
+        vec += M * x[j]
+    end
+    return vec
+end
 
 # Utility functions used by GeometryMapping, FunctionValues
 _copy_or_nothing(x) = copy(x)
--- a/src/PointEvalHandler.jl
+++ b/src/PointEvalHandler.jl
@@ -1,13 +1,14 @@
+Base.@kwdef struct NewtonLineSearchPointFinder{T}
+    max_iters::Int = 10
+    max_line_searches::Int = 5
+    residual_tolerance::T = 1e-10
+end
+
 """
     PointEvalHandler(grid::Grid, points::AbstractVector{Vec{dim,T}}; kwargs...) where {dim, T}
 
 The `PointEvalHandler` can be used for function evaluation in *arbitrary points* in the
 domain -- not just in quadrature points or nodes.
-
-The `PointEvalHandler` takes the following keyword arguments:
- - `search_nneighbors`: How many nodes should be found in the nearest neighbor search for each
-   point. Usually there is no need to change this setting. Default value: `3`.
- - `warn`: Show a warning if a point is not found. Default value: `true`.
 
 The constructor takes a grid and a vector of coordinates for the points. The
 `PointEvalHandler` computes i) the corresponding cell, and ii) the (local) coordinate
@@ -28,10 +29,10 @@ There are two ways to use the `PointEvalHandler` to evaluate functions:
 """
 PointEvalHandler
 
-struct PointEvalHandler{G,dim,T<:Real}
+struct PointEvalHandler{G,T<:Real}
     grid::G
     cells::Vector{Union{Nothing, Int}}
-    local_coords::Vector{Union{Nothing, Vec{dim,T}}}
+    local_coords::Vector{Union{Nothing, Vec{1,T},Vec{2,T},Vec{3,T}}}
 end
 
 function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
@@ -45,20 +46,27 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
     end
 end
 
-function PointEvalHandler(grid::AbstractGrid, points::AbstractVector{Vec{dim,T}}; search_nneighbors=3, warn=true) where {dim, T}
+# Internals:
+# `PointEvalHandler` takes the following keyword arguments:
+#  - `search_nneighbors`: How many nodes should be found in the nearest neighbor search for each
+#    point. Usually there is no need to change this setting. Default value: `3`.
+#  - `warn::Bool`: Show a warning if a point is not found. Default value: `true`.
+#  - `newton_max_iters::Int`: Maximum number of inner Newton iterations. Default value: `10`.
+#  - `newton_residual_tolerance`: Tolerance for the residual norm to indicate convergence in the
+#    inner Newton solver. Default value: `1e-10`.
+function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim,T}}; search_nneighbors=3, warn::Bool=true, strategy = NewtonLineSearchPointFinder()) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
-    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn)
+    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn, strategy)
     return PointEvalHandler(grid, cells, local_coords)
 end
 
-function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn) where {dim, T<:Real, C}
-
+function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn, strategy::NewtonLineSearchPointFinder) where {dim, T<:Real, C}
     # set up tree structure for finding nearest nodes to points
     kdtree = KDTree(reinterpret(Vec{dim,T}, getnodes(grid)))
     nearest_nodes, _ = knn(kdtree, points, search_nneighbors, true)
 
     cells = Vector{Union{Nothing, Int}}(nothing, length(points))
-    local_coords = Vector{Union{Nothing, Vec{dim, T}}}(nothing, length(points))
+    local_coords = Vector{Union{Nothing, Vec{1, T},Vec{2, T},Vec{3, T}}}(nothing, length(points))
 
     for point_idx in 1:length(points)
         cell_found = false
@@ -70,7 +78,7 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
                 possible_cells === nothing && continue # if node is not part of the subdofhandler, try the next node
                 for cell in possible_cells
                     cell_coords = getcoordinates(grid, cell)
-                    is_in_cell, local_coord = point_in_cell(geom_interpol, cell_coords, points[point_idx])
+                    is_in_cell, local_coord = find_local_coordinate(geom_interpol, cell_coords, points[point_idx], strategy; warn)
                     if is_in_cell
                         cell_found = true
                         cells[point_idx] = cell
@@ -89,54 +97,85 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
     return cells, local_coords
 end
 
-# check if point is inside a cell based on physical coordinate
-function point_in_cell(geom_interpol::Interpolation{shape}, cell_coordinates, global_coordinate) where {shape}
-    converged, x_local = find_local_coordinate(geom_interpol, cell_coordinates, global_coordinate)
-    if converged
-        return _check_isoparametric_boundaries(shape, x_local), x_local
-    else
-        return false, x_local
-    end
+# check if point is inside a cell based on isoparametric coordinate
+function check_isoparametric_boundaries(::Type{RefHypercube{dim}}, x_local::Vec{dim, T}, tol) where {dim, T}
+    # All in the range [-1, 1]^dim
+    return all(x -> abs(x) - 1 ≤ tol, x_local)
 end
 
 # check if point is inside a cell based on isoparametric coordinate
-function _check_isoparametric_boundaries(::Type{RefHypercube{dim}}, x_local::Vec{dim, T}) where {dim, T}
-    tol = sqrt(eps(T))
-    # All in the range [-1, 1]
-    return all(x -> abs(x) - 1 < tol, x_local)
-end
-
-# check if point is inside a cell based on isoparametric coordinate
-function _check_isoparametric_boundaries(::Type{RefSimplex{dim}}, x_local::Vec{dim, T}) where {dim, T}
-    tol = sqrt(eps(T))
+function check_isoparametric_boundaries(::Type{RefSimplex{dim}}, x_local::Vec{dim, T}, tol) where {dim, T}
     # Positive and below the plane 1 - ξx - ξy - ξz
     return all(x -> x > -tol, x_local) && sum(x_local) - 1 < tol
 end
 
+cellcenter(::Type{<:RefHypercube{dim}}, _::Type{T}) where {dim, T} = zero(Vec{dim, T})
+cellcenter(::Type{<:RefSimplex{dim}}, _::Type{T}) where {dim, T} = Vec{dim, T}((ntuple(d->1/3, dim)))
+
+_solve_helper(A::Tensor{2,dim}, b::Vec{dim}) where {dim} = inv(A) ⋅ b
+_solve_helper(A::SMatrix{idim, odim}, b::Vec{idim,T}) where {odim, idim, T} = Vec{odim,T}(pinv(A) * b)
+
 # See https://discourse.julialang.org/t/finding-the-value-of-a-field-at-a-spatial-location-in-juafem/38975/2
-# TODO: should we make iteration params optional keyword arguments?
-function find_local_coordinate(interpolation, cell_coordinates::Vector{<:Vec{dim}}, global_coordinate::Vec{dim}; tol_norm = 1e-10) where dim
+function find_local_coordinate(interpolation::Interpolation{refshape}, cell_coordinates::Vector{<:Vec{sdim}}, global_coordinate::Vec{sdim}, strategy::NewtonLineSearchPointFinder; warn::Bool = false) where {sdim, refshape}
+    boundary_tolerance = √(strategy.residual_tolerance)
+
     T = promote_type(eltype(cell_coordinates[1]), eltype(global_coordinate))
     n_basefuncs = getnbasefunctions(interpolation)
     @assert length(cell_coordinates) == n_basefuncs
-    local_guess = zero(Vec{dim, T})
-    max_iters = 10
+    local_guess = cellcenter(refshape, T)
     converged = false
-    for _ in 1:max_iters
-        global_guess = zero(Vec{dim, T})
-        J = zero(Tensor{2, dim, T})
-        # TODO batched eval after 764 is merged.
-        for j in 1:n_basefuncs
-            dNdξ, N = reference_shape_gradient_and_value(interpolation, local_guess, j)
-            global_guess += N * cell_coordinates[j]
-            J += cell_coordinates[j] ⊗ dNdξ
-        end
+    for iter in 1:strategy.max_iters
+        # Setup J(ξ) and x(ξ)
+        J, global_guess = calculate_jacobian_and_spatial_coordinate(interpolation, local_guess, cell_coordinates)
+        # Check if converged
         residual = global_guess - global_coordinate
-        if norm(residual) <= tol_norm
-            converged = true
+        best_residual_norm = norm(residual) # for line search below
+        # Early convergence check
+        if best_residual_norm ≤ strategy.residual_tolerance
+            converged = check_isoparametric_boundaries(refshape, local_guess, boundary_tolerance)
+            if converged
+                @debug println("Local point finder converged in $iter iterations with residual $best_residual_norm to $local_guess")
+            else
+                @debug println("Local point finder converged in $iter iterations with residual $best_residual_norm to a point outside the element: $local_guess")
+            end
             break
         end
-        local_guess -= inv(J) ⋅ residual
+        if calculate_detJ(J) ≤ 0.0
+            warn && @warn "det(J) negative! Aborting! $(calculate_detJ(J))"
+            break
+        end
+        Δξ = _solve_helper(J, residual) # J \ b throws an error. TODO clean up when https://github.com/Ferrite-FEM/Tensors.jl/pull/188 is merged.
+        # Do line search if the new guess is outside the element
+        best_index = 1
+        new_local_guess = local_guess - Δξ
+        global_guess = spatial_coordinate(interpolation, new_local_guess, cell_coordinates)
+        best_residual_norm = norm(global_guess - global_coordinate)
+        if !check_isoparametric_boundaries(refshape, new_local_guess, boundary_tolerance)
+            # Search for the residual minimizer, which is still inside the element
+            for next_index ∈ 2:strategy.max_line_searches
+                new_local_guess = local_guess - Δξ/2^(next_index-1)
+                global_guess = spatial_coordinate(interpolation, new_local_guess, cell_coordinates)
+                residual_norm = norm(global_guess - global_coordinate)
+                if residual_norm < best_residual_norm && check_isoparametric_boundaries(refshape, new_local_guess, boundary_tolerance)
+                    best_residual_norm = residual_norm
+                    best_index = next_index
+                end
+            end
+        end
+        local_guess -= Δξ / 2^(best_index-1)
+        # Late convergence check
+        if best_residual_norm ≤ strategy.residual_tolerance
+            converged = check_isoparametric_boundaries(refshape, local_guess, boundary_tolerance)
+            if converged
+                @debug println("Local point finder converged in $iter iterations with residual $best_residual_norm to $local_guess")
+            else
+                @debug println("Local point finder converged in $iter iterations with residual $best_residual_norm to a point outside the element: $local_guess")
+            end
+            break
+        end
+        if iter == strategy.max_iters
+            @debug println("Failed to converge in $(strategy.max_iters) iterations")
+        end
     end
     return converged, local_guess
 end
@@ -180,8 +219,8 @@ function evaluate_at_points(ph::PointEvalHandler, proj::L2Projector, dof_vals::A
     evaluate_at_points(ph, proj.dh, dof_vals)
 end
 
-function evaluate_at_points(ph::PointEvalHandler{<:Any, dim, T1}, dh::AbstractDofHandler, dof_vals::AbstractVector{T2},
-                           fname::Symbol=find_single_field(dh)) where {dim, T1, T2}
+function evaluate_at_points(ph::PointEvalHandler{<:Any, T1}, dh::AbstractDofHandler, dof_vals::AbstractVector{T2},
+                           fname::Symbol=find_single_field(dh)) where {T1, T2}
     npoints = length(ph.cells)
     # Figure out the value type by creating a dummy PointValues
     ip = getfieldinterpolation(dh, find_field(dh, fname))
@@ -204,7 +243,7 @@ end
 
 # values in dof-order. They must be obtained from the same DofHandler that was used for constructing the PointEvalHandler
 function evaluate_at_points!(out_vals::Vector{T2},
-    ph::PointEvalHandler{<:Any, <:Any, T_ph},
+    ph::PointEvalHandler{<:Any, T_ph},
     dh::DofHandler,
     dof_vals::Vector{T},
     fname::Symbol,
--- a/test/test_pointevaluation.jl
+++ b/test/test_pointevaluation.jl
@@ -1,20 +1,29 @@
-function scalar_field()
+using Ferrite, Test
+
+function test_pe_scalar_field()
     # isoparametric approximation
-    mesh = generate_grid(QuadraticQuadrilateral, (20, 20))
-    f(x) = x[1]^2
+    mesh = generate_grid(QuadraticQuadrilateral, (3, 3))
+    perturbate_standard_grid!(mesh, 1/10)
+
+    f(x) = x[1]+x[2]
 
     ip_f = Lagrange{RefQuadrilateral,2}() # function interpolation
     ip_g = Lagrange{RefQuadrilateral,2}() # geometry interpolation
 
+    # points where we want to retrieve field values
+    points = Vec{2,Float64}[]
+
     # compute values in quadrature points
-    qr = QuadratureRule{RefQuadrilateral}(3) # exactly approximate quadratic field
+    qr = QuadratureRule{RefQuadrilateral}(3) # exactly integrate field
     cv = CellValues(qr, ip_f, ip_g)
     qp_vals = [Vector{Float64}(undef, getnquadpoints(cv)) for _ in 1:getncells(mesh)]
     for cellid in eachindex(mesh.cells)
         xe = getcoordinates(mesh, cellid)
         reinit!(cv, xe)
         for qp in 1:getnquadpoints(cv)
-            qp_vals[cellid][qp] = f(spatial_coordinate(cv, qp, xe))
+            x = spatial_coordinate(cv, qp, xe)
+            qp_vals[cellid][qp] = f(x)
+            push!(points, x)
         end
     end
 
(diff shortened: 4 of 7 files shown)


test/test_utils.jl
```diff
@@ -259,3 +259,15 @@ getfacerefshape(::Hexahedron, ::Int) = RefQuadrilateral
...
+    function perturbate(x::Vec{dim}) where dim
+        for d in 1:dim
+            if x[d] ≈ 1.0 || x[d] ≈ -1.0
+                return x
+            end
+        end
+        return x + Vec{dim}(0.5*strength .* (2 .* rand(Vec{dim}) .- 1.0))
+    end
+    transform_coordinates!(grid, perturbate)
+end
```
KnutAM, line 273: Just for reference, I did something similar for #798, but it was hard to make sure that negative determinants didn't occur, so decided to go with a smooth and non-random, but still nonlinear grid:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/3aad963b0726571052c10053e91415bdd207aa0a/test/test_interpolations.jl#L197-L198

termi-official: Since the used perturbations in the tests are rather small negative detJ isn't an issue. I also thought about analytical deformation, but then I was not sure about which function should be applied. Also I think that the mentioned function should be extended to "rotate" some random elements just to check that we do not have hidden assumptions on the element orientation.

KnutAM: Do you mean to modify "rotate" the node indices in some random cells? That would be very nice!

termi-official: Yes, cyclic shifts of the node ids, which should keep the grid intact but can trigger issues. This can be done without importing generated meshes via external packages.

src/PointEval/PointEvalHandler.jl
```diff
@@ -103,26 +105,26 @@ end
...
 end
 
 # See https://discourse.julialang.org/t/finding-the-value-of-a-field-at-a-spatial-location-in-juafem/38975/2
 # TODO: should we make iteration params optional keyword arguments?
-function find_local_coordinate(interpolation, cell_coordinates::Vector{V}, global_coordinate::V) where {dim, T, V <: Vec{dim, T}}
+function find_local_coordinate(interpolation::IP, cell_coordinates::Vector{V}, global_coordinate::V) where {dim, T, V <: Vec{dim, T}, ref_shape, IP <: Interpolation{ref_shape}}
     n_basefuncs = getnbasefunctions(interpolation)
     @assert length(cell_coordinates) == n_basefuncs
-    local_guess = zero(V)
+    local_guess = sample_random_point(ref_shape)
```
KnutAM, line 123: I would vote for not having this random if it can be avoided, see https://github.com/Ferrite-FEM/Ferrite.jl/pull/818#discussion_r1359338948

termi-official: First, let me give you two arguments for using randomization here over a single fixed guess.

1. There is no easy heuristic for a single initial guess which guarantees convergence from the nearest, because there is no trivial relation between basin of attractions for roots of the nonlinear problem. For the heuristic given in your comment you can easily construct examples (which we likely also hit in practice) such that the newton will be in the wrong basin and cannot converge, which is one issue I try to fix here.
2. For a grid search we can guarantee convergence with a *fine enough* grid. However, the grid might be quite large for nonlinear problems.

Now the presented randomization strategy is a tradeoff, which will also converge with enough samples (because we will eventually hit the basin of attraction), however in 3D we will likely use less points than a grid search, provided we have a proper random number generator.

However, if you have a better strategy on how we can guarantee to start in the basin of attraction (at best with 1 sample) I will take it. :)


On the other side we should make this algorithm swappable anyways. I thought about NonlinearSolve.jl as a dependency for the nonlinear problem (via extension?) and an external function to pass here to provide the outer loop (i.e. initial guess + eventual looping). What do you think?

koehlerson: Since it is a nonlinear multimodal optimization problem the approach should be tackle it as such. This is why I agree with Dennis' opinion that either a full multistart strategy with a grid of initial guesses or a random point multistart with less points should be used, since both is often use in such cases. Other than that, I would agree with any conventional method for this type of problem (trust region or the like)

KnutAM: > For the heuristic given in your comment you can easily construct examples (which we likely also hit in practice) such that the newton will be in the wrong basin and cannot converge, which is one issue I try to fix here.

I guess I'm underestimating the complexity of this problem then - thanks for the clarification! If you happen to have an example around, I would be interested in visualizing it with a Q&D code I have that shows the residual field over a `QuadraticQuadrilateral`. When trying random quad distortions and points, I've so far only ended up with nice problems. 

So sure, having guaranteed convergence would be better then! An option that is potentially quite efficient would be to have a reasonable heuristic first guess, followed by a multistart algorithm based on random guesses in case that one doesn't work.

termi-official: I think you are right. The second derivative for the minimization problem $min_\xi||x(\xi)-x_g||^2$ should be just J, so for an intact grid we should end up with a convex setup due to $det(J)>0$, as you have written on slack?

src/PointEval/PointEvalHandler.jl
```diff
@@ -8,6 +8,9 @@ The `PointEvalHandler` takes the following keyword arguments:
  - `search_nneighbors`: How many nodes should be found in the nearest neighbor search for each
    point. Usually there is no need to change this setting. Default value: `3`.
  - `warn`: Show a warning if a point is not found. Default value: `true`.
+ - `newton_max_iters`: Maximum number of inner Newton iterations. Default value: `10`.
+ - `newton_residual_tolerance`: Tolerance for the residual norm to indicate convergence in the
```
fredrikekre, line 12: These kwarg names don't match the code

src/Grid/utils.jl
```diff
@@ -340,3 +340,28 @@ for (func,                   entity_f,            entity_t,     filter_f,
     end
 end
 end
+
+# Hypercube is simply ⨂ᵈⁱᵐ Line :)
+sample_random_point(::Type{Ferrite.RefHypercube{ref_dim}}) where {ref_dim} = Vec{ref_dim}(2.0 .* rand(Vec{ref_dim}) .- 1.0)
+
+# Dirichlet type sampling
+function sample_random_point(::Type{Ferrite.RefSimplex{ref_dim}}) where {ref_dim} 
+    ξ = rand(ref_dim+1)
```
fredrikekre, line 349: To avoid the array allocation
```suggestion
    ξ = ntuple(_ -> rand(), ref_dim+1)
```

src/Grid/utils.jl
```diff
@@ -340,3 +340,28 @@ for (func,                   entity_f,            entity_t,     filter_f,
     end
 end
 end
+
+# Hypercube is simply ⨂ᵈⁱᵐ Line :)
+sample_random_point(::Type{Ferrite.RefHypercube{ref_dim}}) where {ref_dim} = Vec{ref_dim}(2.0 .* rand(Vec{ref_dim}) .- 1.0)
```
fredrikekre, line 345: To avoid the array allocation
```suggestion
sample_random_point(::Type{Ferrite.RefHypercube{ref_dim}}) where {ref_dim} = Vec{ref_dim}(i -> 2 * rand() - 1)
```

src/Grid/utils.jl
fredrikekre, line 349: Add a comment about this `+1` etc

src/PointEval/PointEvalHandler.jl
```diff
@@ -90,53 +92,88 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
-function point_in_cell(geom_interpol::Interpolation{shape}, cell_coordinates, global_coordinate) where {shape}
-    converged, x_local = find_local_coordinate(geom_interpol, cell_coordinates, global_coordinate)
-    if converged
-        return _check_isoparametric_boundaries(shape, x_local), x_local
-    else
-        return false, x_local
-    end
+# TODO linear case can be handled easier
+function point_in_cell(geom_interpol::Interpolation{shape}, cell_coordinates, global_coordinate, warn,  linesearch_max_substeps, max_iters, tol_norm) where {shape}
+    return find_local_coordinate(geom_interpol, cell_coordinates, global_coordinate, warn, linesearch_max_substeps, max_iters, tol_norm)
```
fredrikekre, line 97: Couldn't this function be removed? It just passes everything to `find_lcoal_coordinate`?

termi-official:
I moved the computation of J and x without the caches (i.e. GeometryMapping) into GeometryMapping.jl and common_values.jl and changed the signature for the PointEvalHandler to support also mixed-dimensional grids (which are e.g. obtained by mixing shells, beams and solid elements in one grid). These functions are up to further optimization in subsequent PRs (i.e. the stuff which I do in the GPU support PR).

I also fixed an oopsie in the line searcher (which we need to identify the cell boundary for nonlinear cells).

Docs have been added as suggested.

src/FEValues/GeometryMapping.jl
```diff
@@ -121,15 +121,32 @@ end
...
 end =# # PR798
 
-@inline function calculate_mapping(::GeometryMapping{0}, q_point, x)
+@inline function calculate_mapping(gip::VectorizedInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec})
+    return calculate_mapping(gip.ip, ξ, x)
+end
+
+@inline function calculate_mapping(gip::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec})
+    n_basefuncs = getnbasefunctions(gip)
+    @assert length(cell_coordinates) == n_basefuncs
```
KnutAM, line 130: Driveby: 
```suggestion
    @boundscheck checkbounds(cell_coordinates, Base.OneTo(n_basefuncs))
```
To be safe for `AbstractVector` and to allow escape by `@inbounds` from outside?
Could also be reasonable to check for the same length to catch user bugs?

Also good to note that the `calculate_mapping` function that I wrote above is unsafe regarding inbounds. 
It would be required to either remove the unchecked `@inbounds` and add a `@propagate_inbounds`, or add something like an `unsafe_calculate_mapping` which assumes inbounds are ok to use. But the performance was tricky and easily disturbed...

src/Grid/utils.jl
```diff
@@ -340,3 +340,38 @@ for (func,                   entity_f,            entity_t,     filter_f,
...
+    return Vec{3}((ξ₁, ξ₂, ξ₃))
+end
+
+# TODO what to do here? The samplig is not uniform...
+function sample_random_point(::Type{RefPyramid})
+    ξ₃ = (1-1e-3)*rand(Float64) # Derivative is discontinuous at the top
+    # If we fix a z coordinate we get a Quad with extends (1-ξ₃)
+    (ξ₁, ξ₂) = (1.0 - ξ₃) .* Vec{2}(ntuple(_ -> rand(), 2))
+    return Vec{3}((ξ₁, ξ₂, ξ₃))
+end
```
KnutAM, line 377: Are these used for anything but testing?

src/FEValues/GeometryMapping.jl
termi-official: Should I add the note here or rather open an issue?

src/Grid/utils.jl
termi-official: I use them downstream for testing, too, and thought I should just expose these here rather than copy paste them into the downstream packages.

If we don't want to have such functions in core we can start discussing about something like FerriteUtils.jl?

KnutAM: I see the need - but I'm just worried that it is a slippery slope adding convenience functions not used will bloat the core. A test dep utility package also adds to the maintenance load, but at least reduces the core package. Not sure what the best option is for these cases...

src/PointEvalHandler.jl
```diff
@@ -89,54 +91,66 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
             converged = true
             break
         end
-        local_guess -= inv(J) ⋅ residual
+        if calculate_detJ(J) ≤ 0.0
+            warn && @warn "det(J) negative! Aborting! $(calculate_detJ(J))" 
+            break
+        end
+        Δξ = _solve_helper(J, residual) # J \ b throws an error. TODO clean up when https://github.com/Ferrite-FEM/Tensors.jl/pull/188 is merged.
+        # Do line search if outside/on boundary
```
KnutAM, line 140:
```suggestion
        # Do line search if outside boundary
```
To be consistent with the change to `≥` in `check_isoparameteric_boundaries`?

KnutAM (commented):
Nice work - really cool with the ever-increasing support for embedded elements!

* I would suggest leaving the inclusion of `sample_random_point` into `src` to a separate pr for now. 

* I like the renaming of `x` to `cell_coordinates`, but on the other hand such a renaming is unrelated? If renamed it should probably also be renamed in the `reinit!` methods? Should this renaming take place in a separate pr then?

I'm unsure regarding iterations for embedded elements. I would have thought we should find the value of $\xi$ which gives zero distance to the sought coordinate, $\boldsymbol{x}_0$, projected onto the element. And then potentially say didn't find if the distance to the element surface is too large. 
```math
\begin{align}
\boldsymbol{x}(\boldsymbol{\xi}) &= \boldsymbol{x}_i N_i(\boldsymbol{\xi})\\
\boldsymbol{r}(\boldsymbol{\xi}) &= [\boldsymbol{x}(\boldsymbol{\xi}) - \boldsymbol{x}_0] \cdot \frac{\partial\boldsymbol{x}}{\partial \boldsymbol{\xi}}, \quad \text{Potentially normalizing each column in }\frac{\partial\boldsymbol{x}}{\partial \boldsymbol{\xi}} \ \\
\frac{\partial\boldsymbol{r}}{\partial \boldsymbol{\xi}} &= [\boldsymbol{x}(\boldsymbol{\xi}) - \boldsymbol{x}_0] \cdot \frac{\partial^2\boldsymbol{x}}{\partial \boldsymbol{\xi}^2} + \left[\frac{\partial\boldsymbol{x}}{\partial \boldsymbol{\xi}}\right]^T \cdot \left[ \frac{\partial\boldsymbol{x}}{\partial \boldsymbol{\xi}} \right] 
\end{align}
```
I suppose the `pinv` updates are then doing this in some quasi-newton fashion? (But last time I checked, this is quite slow, so probably the above algorithm is faster even with the higher order derivative calculations?)

But since we are using autodiff anyway to compute the gradients of hessians of the base functions, perhaps it would be equally efficient to just define the residual function and just iterate on this? I.e. 
```
function residual(ξ::Vec{dim}, x0::Vec{dim}, ip, cell_coordinates::AbstractVector{<:Vec{dim}) where dim
    return spatial_coordinate(ip, ξ, cell_coordinates) - x0
end
function residual(ξ::Vec{dim}, x0::Vec{dim}, ip, cell_coordinates::AbstractVector{<:Vec{dim}) where dim
    dxdξ, x = gradient(ξ_ -> spatial_coordinate(ip, ξ_, cell_coordinates), :all)
    # do scaling if desired or needed?
    return (x - x0) \cdot dxdξ
end
```
(And even if not used this modified residual, I suppose the first definition of residual can be used to reduce the need for the `calculate_mapping_and_spatial_coordinate` functions?

termi-official:
Thanks for taking the time @KnutAM and @fredrikekre !

>     * I would suggest leaving the inclusion of `sample_random_point` into `src` to a separate pr for now.
> 
>     * I like the renaming of `x` to `cell_coordinates`, but on the other hand such a renaming is unrelated? If renamed it should probably also be renamed in the `reinit!` methods? Should this renaming take place in a separate pr then?
> 

I am not a big fan of making many small 1-5 line PRs (i.e. add docs, move function with same name from one file to another, ..) due to the associated overhead and dependencies, but I can make many tiny PRs if it is preferred.

> 
> I'm unsure regarding iterations for embedded elements. I would have thought we should find the value of ξ which gives zero distance to the sought coordinate, x0, projected onto the element. And then potentially say didn't find if the distance to the element surface is too large.
> 
> I suppose the `pinv` updates are then doing this in some quasi-newton fashion? (But last time I checked, this is quite slow, so probably the above algorithm is faster even with the higher order derivative calculations?)

My initial idea was also to have better modularity for the inner solver, but I could not find a good way how to do it. The rationale behind the current Newton solver for embedded (nonlinear) stuff is that it does not converge to a root if the coordinate is not in the element. You can interpret the solution of the linear system in a least square sense, as pinv returns by default the Moore-Penrose pseudoinverse. The problem with the projection is that you need additional infrastructure to handle the projections properly. Both approaches are valid tho.

> 
> But since we are using autodiff anyway to compute the gradients of hessians of the base functions, perhaps it would be equally efficient to just define the residual function and just iterate on this? I.e.
> 
> ```
> function residual(ξ::Vec{dim}, x0::Vec{dim}, ip, cell_coordinates::AbstractVector{<:Vec{dim}) where dim
>     return spatial_coordinate(ip, ξ, cell_coordinates) - x0
> end
> function residual(ξ::Vec{dim}, x0::Vec{dim}, ip, cell_coordinates::AbstractVector{<:Vec{dim}) where dim
>     dxdξ, x = gradient(ξ_ -> spatial_coordinate(ip, ξ_, cell_coordinates), :all)
>     # do scaling if desired or needed?
>     return (x - x0) \cdot dxdξ
> end
> ```
> 
> (And even if not used this modified residual, I suppose the first definition of residual can be used to remove the need for the `calculate_mapping_and_spatial_coordinate` functions?

I am not sure if I follow the idea here. So you are proposing to use a gradient descend over a Newton with line search as default? I think this will end up with the same issues as master for points near (or even on) element boundaries.

src/PointEvalHandler.jl
```diff
@@ -89,54 +91,66 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
+    return all(x -> abs(x) - 1 ≤ tol, x_local)
 end
 
 # check if point is inside a cell based on isoparametric coordinate
-function _check_isoparametric_boundaries(::Type{RefSimplex{dim}}, x_local::Vec{dim, T}) where {dim, T}
-    tol = sqrt(eps(T))
+function check_isoparametric_boundaries(::Type{RefSimplex{dim}}, x_local::Vec{dim, T}, tol::T) where {dim, T}
     # Positive and below the plane 1 - ξx - ξy - ξz
-    return all(x -> x > -tol, x_local) && sum(x_local) - 1 < tol
+    return all(x -> x ≥ -tol, x_local) && sum(x_local) - 1 < tol
```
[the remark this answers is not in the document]
termi-official: This also allows points directly on the boundary.

src/FEValues/common_values.jl
```diff
@@ -272,7 +272,10 @@ Compute the spatial coordinate in a quadrature point. `x` contains the nodal
 coordinates of the cell.
 
 The coordinate is computed, using the geometric interpolation, as
-``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{x}) \\mathbf{\\hat{x}}_i``
+``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{\\xi}) \\mathbf{\\hat{x}}_i``.
+
+where ``\\xi``is the coordinate of the given quadrature point `q_point` of the associated
+quadrature rule.
 """
 function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVector{<:Vec})
```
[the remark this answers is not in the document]
termi-official: Here or separate PR?

src/PointEvalHandler.jl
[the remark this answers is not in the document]
KnutAM: Yes, the interaction with `tol` was why I was wondering why there was a need to change in the first place. In that case, perhaps it is better to leave all as strict inequality as they were before.

src/FEValues/common_values.jl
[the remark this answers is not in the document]
termi-official: Bump.

src/FEValues/GeometryMapping.jl
```diff
@@ -121,15 +121,32 @@ function otimes_returntype(#=typeof(x)=#::Type{<:Vec{dim,Tx}}, #=typeof(d2Mdξ2)
...
+
+    fecv_J = zero(otimes_returntype(Vec{sdim,T}, Vec{rdim,T}))
+    @inbounds for j in 1:n_basefuncs
+        dMdξ = shape_gradient(gip, ξ, j)
+        #fecv_J += cell_coordinates[j] ⊗ dMdξ[j]
+        fecv_J += otimes_helper(cell_coordinates[j], dMdξ[j])
+    end
+    return MappingValues(fecv_J, nothing)
+end
+
```
KnutAM, line 140: Separate PR? 
Doesn't seem used in this PR?

src/FEValues/GeometryMapping.jl
```diff
@@ -84,13 +84,13 @@ function GeometryMapping{2}(::Type{T}, ip::ScalarInterpolation, qr::QuadratureRu
...
-function precompute_values!(gm::GeometryMapping{0}, qr_points::Vector{<:Vec})
+function precompute_values!(gm::GeometryMapping{0}, qr_points::AbstractVector{<:Vec})
     reference_shape_values!(gm.M, gm.ip, qr_points)
 end
-function precompute_values!(gm::GeometryMapping{1}, qr_points::Vector{<:Vec})
+function precompute_values!(gm::GeometryMapping{1}, qr_points::AbstractVector{<:Vec})
     reference_shape_gradients_and_values!(gm.dMdξ, gm.M, gm.ip, qr_points)
 end
-function precompute_values!(gm::GeometryMapping{2}, qr_points::Vector{<:Vec})
+function precompute_values!(gm::GeometryMapping{2}, qr_points::AbstractVector{<:Vec})
```
KnutAM, line 93: Is this related to / required for this PR? 
(The change makes sense, but seems like it should be in a PR on relaxing container types)

src/FEValues/GeometryMapping.jl
```diff
@@ -149,6 +166,25 @@ end
 calculate_detJ(J::Tensor{2}) = det(J)
 calculate_detJ(J::SMatrix) = embedding_det(J)
 
+function calculate_mapping_and_spatial_coordinate(gip::VectorizedInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec})
+    return calculate_mapping_and_spatial_coordinate(gip.ip, ξ, cell_coordinates)
+end
```
KnutAM, line 171: Doesn't seem used, remove and only allow ScalarInterpolation?

src/FEValues/common_values.jl
```diff
@@ -342,6 +345,27 @@ function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVecto
     return vec
 end
 
+"""
+    spatial_coordinate(ip::VectorizedInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec{sdim, T}})
```
KnutAM, line 349:
```suggestion
    spatial_coordinate(ip::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec{sdim, T}})
```

src/FEValues/common_values.jl
```diff
@@ -342,6 +345,27 @@ function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVecto
...
+
+Compute the spatial coordinate in a given quadrature point. `cell_coordinates` contains the nodal coordinates of the cell.
+
+The coordinate is computed, using the geometric interpolation, as
+``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{\\xi}) \\mathbf{\\hat{x}}_i``
+"""
+spatial_coordinate(ip::VectorizedInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec}) = spatial_coordinate(ip.ip, ξ, cell_coordinates)
+
```
KnutAM, line 357: Not used, remove and only allow scalar interpolation?

src/FEValues/common_values.jl
```diff
@@ -342,6 +345,27 @@ function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVecto
...
+function spatial_coordinate(interpolation::ScalarInterpolation, ξ::Vec{<:Any}, cell_coordinates::AbstractVector{<:Vec{sdim, Tx}}) where {Tx, sdim}
```
KnutAM, line 358:
```suggestion
function spatial_coordinate(interpolation::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec})
```

src/FEValues/common_values.jl
```diff
@@ -342,6 +345,27 @@ function spatial_coordinate(fe_v::AbstractValues, q_point::Int, x::AbstractVecto
...
+    n_basefuncs = getnbasefunctions(interpolation)
+    @boundscheck checkbounds(cell_coordinates, Base.OneTo(n_basefuncs))
+
+    x = zero(Vec{sdim, Tx})
```
KnutAM, line 362:
```suggestion
    x = zero(eltype(cell_coordinates))
```

src/PointEvalHandler.jl
```diff
@@ -45,20 +40,27 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
...
+function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim,T}}; search_nneighbors=3, warn::Bool=true, newton_max_iters::Int = 10, newton_residual_tolerance::T = 1e-10) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
-    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn)
+    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn, #=linesearch_max_substeps=# 5, newton_max_iters, newton_residual_tolerance)
     return PointEvalHandler(grid, cells, local_coords)
 end
 
-function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn) where {dim, T<:Real, C}
-
+function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn, linesearch_max_substeps, max_iters, tol_norm) where {dim, T<:Real, C}
```
KnutAM, line 57: keyword arguments?

src/PointEvalHandler.jl
```diff
@@ -89,54 +91,70 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
+            for next_index ∈ 2:linesearch_max_substeps
+                new_local_guess = local_guess - Δξ/2^(next_index-1)
+                global_guess = spatial_coordinate(interpolation, new_local_guess, cell_coordinates)
+                residual_norm = norm(global_guess - global_coordinate)
+                if residual_norm < best_residual_norm && check_isoparametric_boundaries(ref_shape, new_local_guess, sqrt(tol_norm))
+                    best_residual_norm = residual_norm
+                    best_index = next_index
+                end
+            end
+        end
```
KnutAM, line 151: So long since we discussed this that I don't recall, but do you have an example which triggers this block? Would be nice to comment on when this could happen.

src/PointEvalHandler.jl
```diff
@@ -89,54 +91,70 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
-        # TODO batched eval after 764 is merged.
-        for j in 1:n_basefuncs
-            dNdξ, N = reference_shape_gradient_and_value(interpolation, local_guess, j)
-            global_guess += N * cell_coordinates[j]
-            J += cell_coordinates[j] ⊗ dNdξ
-        end
+    for iter in 1:max_iters
+        # Setup J(ξ) and x(ξ)
+        mapping, global_guess = calculate_mapping_and_spatial_coordinate(interpolation, local_guess, cell_coordinates)
+        J = getjacobian(mapping)
```
KnutAM, line 122:
```suggestion
        J, global_guess = calculate_jacobian_and_spatial_coordinate(interpolation, local_guess, cell_coordinates)
```

src/FEValues/GeometryMapping.jl
```diff
@@ -149,6 +166,25 @@ end
...
+
+function calculate_mapping_and_spatial_coordinate(gip::ScalarInterpolation, ξ::Vec{rdim,Tξ}, cell_coordinates::AbstractVector{<:Vec{sdim, Tx}}) where {Tξ, Tx, rdim, sdim}
```
KnutAM, line 173: Since there is no information about the function interpolations to be mapped, I don't see why we should return a `MappingValues` here. 
```suggestion
function calculate_jacobian_and_spatial_coordinate(gip::ScalarInterpolation, ξ::Vec{rdim,Tξ}, cell_coordinates::AbstractVector{<:Vec{sdim, Tx}}) where {Tξ, Tx, rdim, sdim}
```

src/FEValues/GeometryMapping.jl
```diff
@@ -149,6 +166,25 @@ end
...
+    @boundscheck checkbounds(cell_coordinates, Base.OneTo(n_basefuncs))
+
+    fecv_J = zero(otimes_returntype(Vec{sdim,Tx}, Vec{rdim,Tξ}))
+    x = zero(Vec{sdim, Tx})
+    @inbounds for j in 1:n_basefuncs
+        dMdξ, M = reference_shape_gradient_and_value(gip, ξ, j)
+        x += M * cell_coordinates[j]
+        fecv_J += otimes_helper(cell_coordinates[j], dMdξ)
+    end
+    return MappingValues(fecv_J, nothing), x
```
KnutAM, line 184:
```suggestion
    return fecv_J, x
```

src/FEValues/common_values.jl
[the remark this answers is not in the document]
KnutAM: here

KnutAM (commented):
Long time since I looked at this (sorry!)
I think it would be good to move some unrelated changes to a separate PR, that will also keep the git blames easier to follow.

src/PointEvalHandler.jl
termi-official: The obvious one is (as indicated in the comment above) if you are close to the element boundary the line searcher gets triggered, so we can also support point finding for problematic cases where the evaluation of the ansatz functions outside the element gives you something unexpected. This also gets triggered if the Newton overshoots and you would move outside the element.

termi-official: Let me change it to a projected Newton then, makes probably more sense.

src/PointEvalHandler.jl
termi-official: Why?

src/FEValues/GeometryMapping.jl
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1017

src/PointEvalHandler.jl
KnutAM: Aha, thanks for explaining! 
So it is only to prevent getting outside? 
In that case, wouldn't it suffice to only check this during line-searching to simplify the code? E.g. 
```suggestion
        # If we are outside the element, reduce the step length until we are inside again
        while !check_isoparametric_boundaries(ref_shape, new_local_guess, sqrt(tol_norm)) && best_index < linesearch_max_substeps
            best_index += 1
            new_local_guess = local_guess - Δξ/2^(best_index-1)
        end
        local_guess = new_local_guess
        global_guess = spatial_coordinate(interpolation, new_local_guess, cell_coordinates)
        best_residual_norm = norm(global_guess - global_coordinate)
```
? 
To me, it looks like the current code will always run `linesearch_max_substeps`, and then just take the step with the lowest residual?

src/PointEvalHandler.jl
KnutAM: Seems easier to read when we have so many, instead of putting the `#= name =#` comment, and they are first kwarg, then arg, and then kwarg again later. 

We are also now defining the defaults differently for `PointEvalHandler` compared to `find_local_coordinate` (at least `linesearch_max_substeps` and `warn`), is this intended?

Perhaps it would be nice to define the defaults only in one location?

KnutAM: But if you disagree, can be discussed and potentially done separately! 
Don't let this stop the PR - it is just an internal detail after all, and I guess you intend the keywords to PointEvalHandler also as internal for now based on the comment?

src/PointEvalHandler.jl
termi-official: > To me, it looks like the current code will always run linesearch_max_substeps, and then just take the step with the lowest residual?
The one with the lowest residual, which is still inside the element.

Since I am not sure about the interaction with the convergence for your algorithm I think we should just stick to the normal Newton line searcher, which has linear convergence on the element boundary and essentially quadratic convergence inside the element. I think if we invest a bit more time we can even write down a proof that it the line searcher converges.

src/FEValues/GeometryMapping.jl
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1018

src/PointEvalHandler.jl
termi-official: I added a struct to gather these. Might also be interesting as an option to swap the solver.

src/PointEvalHandler.jl
KnutAM: I'm not suggestion to change the update algorithm, just the stopping criterion. 
But if you don't agree, then let's keep it and I can do a follow up PR to simplify at some point.

src/PointEvalHandler.jl
termi-official: Indeed, the kwargs are internals for now. We can make them later public once the interfaces settle

src/FEValues/common_values.jl
```diff
@@ -323,25 +323,47 @@ function_curl(fe_v::AbstractValues, q_point::Int, u::AbstractVector{<:Vec}) =
...
 
+"""
+    spatial_coordinate(ip::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec{sdim, T}})
+
+Compute the spatial coordinate in a given quadrature point. `cell_coordinates` contains the nodal coordinates of the cell.
+
+The coordinate is computed, using the geometric interpolation, as
+``\\mathbf{x} = \\sum\\limits_{i = 1}^n M_i (\\mathbf{\\xi}) \\mathbf{\\hat{x}}_i``
+"""
+function spatial_coordinate(interpolation::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec}) where {Tx, sdim}
```
KnutAM, line 356:
```suggestion
function spatial_coordinate(interpolation::ScalarInterpolation, ξ::Vec, cell_coordinates::AbstractVector{<:Vec})
```

src/PointEvalHandler.jl
```diff
@@ -89,54 +97,85 @@ function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid,
...
+cellcenter(::Type{<:RefHypercube{dim}}, _::Type{T}) where {dim, T} = zero(Vec{dim, T})
+cellcenter(::Type{<:RefSimplex{dim}}, _::Type{T}) where {dim, T} = Vec{dim, T}((ntuple(d->1/3, dim)))
+
+_solve_helper(A::Tensor{2,dim}, b::Vec{dim}) where {dim} = inv(A) ⋅ b
+_solve_helper(A::SMatrix{idim, odim}, b::Vec{idim,T}) where {odim, idim, T} = Vec{odim,T}(pinv(A) * b)
+
 # See https://discourse.julialang.org/t/finding-the-value-of-a-field-at-a-spatial-location-in-juafem/38975/2
-# TODO: should we make iteration params optional keyword arguments?
-function find_local_coordinate(interpolation, cell_coordinates::Vector{<:Vec{dim}}, global_coordinate::Vec{dim}; tol_norm = 1e-10) where dim
+function find_local_coordinate(interpolation::Interpolation{ref_shape}, cell_coordinates::Vector{<:Vec{sdim}}, global_coordinate::Vec{sdim}, strategy::NewtonLineSearchPointFinder; warn::Bool = false) where {sdim, ref_shape}
```
KnutAM, line 119: Just detail, but I think we only use `refshape` or `RefShape` in other places in the codebase. (Update required below as well)
```suggestion
function find_local_coordinate(interpolation::Interpolation{refshape}, cell_coordinates::Vector{<:Vec{sdim}}, global_coordinate::Vec{sdim}, strategy::NewtonLineSearchPointFinder; warn::Bool = false) where {sdim, refshape}
```

test/test_pointevaluation.jl
```diff
@@ -35,18 +43,59 @@ function scalar_field()
...
+    # @test_throws MethodError projector = L2Projector(ip_f, mesh; geom_ip=ip_g)
+    projector = L2Projector(ip_f, mesh)
+    projector_vals = project(projector, qp_vals, qr)
+
+    # set up PointEvalHandler and retrieve values
+    ph = PointEvalHandler(mesh, points)
+    @test all(x -> x !== nothing, ph.cells)
+
+    # vals = evaluate_at_points(ph, projector, projector_vals)
+    # @test f.(points) ≈ vals
```
KnutAM, line 83: Why the commented out tests?

KnutAM (approved):
A few more comments, but otherwise lgtm!

src/PointEvalHandler.jl
termi-official: Okay we got all kinds of combinations.
```
src/interpolations.jl
2:    Interpolation{ref_shape, order}()
4:Abstract type for interpolations defined on `ref_shape`
```
Is there an easy way to check such conventions in a pre-commit hook?

test/test_pointevaluation.jl
termi-official: They did not work. I fixed them with the last commit.

src/PointEvalHandler.jl
KnutAM: Haha, ok, then nvm. 
I guess there would be some way via the typos in #865, but not sure if it is worth the effort.

test/test_pointevaluation.jl
KnutAM: OK, maybe then a comment regarding the exact integration of a linear function `f(x)` for the quadratic interpolation with the chosen QR?

termi-official: What do you mean? We do not really integrate anything in the test itself, but use a setup which would be exact if we solve the problem and test if we can evaluate some function at the quadrature points with high precision.

KnutAM: Ah, on second thought, no I don't understand why you needed to make `f(x)` linear to make the test work then. Because the order shouldn't add up, we just need an interpolation which can represent `f(x)` exactly, so a quadratic `f(x)` should work?

The L2Projector does the integration I guess, so that's where you need sufficient accuracy to capture the nonlinearities in the test, and for that it should suffice to integrate the interpolation exactly...

termi-official: > Ah, on second thought, no I don't understand why you needed to make `f(x)` linear to make the test work then. Because the order shouldn't add up, we just need an interpolation which can represent `f(x)` exactly, so a quadratic `f(x)` should work?
> 
Quadratic f also works, but will be associated with larger error for points outside the ball of radius 1, because there are tiny errors in the found position which will be amplified.


> The L2Projector does the integration, so that's where you need sufficient accuracy to capture the nonlinearities in the test, and for that it should suffice to integrate the interpolation exactly...

Oh, right. :D

src/PointEvalHandler.jl
```diff
@@ -45,20 +46,27 @@ function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
...
-function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn) where {dim, T<:Real, C}
-
+function _get_cellcoords(points::AbstractVector{Vec{dim,T}}, grid::AbstractGrid, node_cell_dicts::Dict{C,Dict{Int, Vector{Int}}}, search_nneighbors, warn, strategy::NewtonLineSearchPointFinder) where {dim, T<:Real, C}
     # set up tree structure for finding nearest nodes to points
     kdtree = KDTree(reinterpret(Vec{dim,T}, getnodes(grid)))
     nearest_nodes, _ = knn(kdtree, points, search_nneighbors, true)
 
     cells = Vector{Union{Nothing, Int}}(nothing, length(points))
-    local_coords = Vector{Union{Nothing, Vec{dim, T}}}(nothing, length(points))
+    local_coords = Vector{Union{Nothing, Vec{1, T},Vec{2, T},Vec{3, T}}}(nothing, length(points))
```
fredrikekre, line 69: What is the reason for this change?

termi-official: Support for mixed-dimensional meshes. E.g. shells with beams.

fredrikekre: Okay, and if there is a field that lives on both domains then?

termi-official: I am not sure if I understand the question correctly. The point finder just gives you the reference coordinate and cell index for some spatial coordinate. If you have some spatial coordinate which lives on the boundary of two elements, then it is always up to the prayers of convergence which element is returned for now. Technically we need to distinguish this case from the case when the point is strictly in the interior of some element. But note that resolving this issue is orthogonal to the goal of this PR, which is about improving the handling of nonlinear geometries. Before this PR using either mixed-dimensional or embedded elements resulted in a cryptic error for users and it looks like changing this (and the solver to some pseudo-inverse instead of the actual inverse below) is sufficient to handle these cases. I can separate these 2 lines and the tests into a separate PR if you think it convolutes this PR too much

PR #825, opened 2023-10-13
Add docstring to SubDofHandler

KnutAM:


src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,14 @@ Access some grid representation for the dof handler.
 """
 get_grid(dh::AbstractDofHandler)
 
+"""
+    SubDofHandler(dh::AbstractDofHandler, cellset)
+
+Create an `sdh::SubDofHandler` for the cells in `cellset`. All cells in `cellset`
+must have the same type. Fields are added to the `sdh` by using [`add!`](@ref) the 
+same way as for the standard [`DofHandler`](@ref).
```
KnutAM, line 20:
```suggestion
same way as for the standard [`DofHandler`](@ref).
Close when all `SubDofHandler`s have been created and fields are added by calling 
[`close!`](@ref) on the parent dof handler, `dh`. 
```

src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,16 @@ Access some grid representation for the dof handler.
...
+Close when all `SubDofHandler`s have been created and fields are added by calling
```
KnutAM, line 21:
```suggestion
Close when all `SubDofHandler`s have been created and their fields have been added by calling 
```

src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,16 @@ Access some grid representation for the dof handler.
 """
 get_grid(dh::AbstractDofHandler)
 
+"""
+    SubDofHandler(dh::AbstractDofHandler, cellset)
+
+Create an `sdh::SubDofHandler` for the cells in `cellset`. All cells in `cellset`
```
termi-official, line 18:
```suggestion
Create a `sdh::SubDofHandler` for the cells in `cellset`. All cells in `cellset`
```

Just to be sure, are there any invariants on `cellset` which users need to be vary of?

src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,16 @@ Access some grid representation for the dof handler.
...
+must have the same type. Fields are added to the `sdh` by using [`add!`](@ref) the
```
termi-official, line 19: Since this is likely for aimed beginners here a maybe not so obvious question: What does "type" mean? I think we should be more specific.

src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,16 @@ Access some grid representation for the dof handler.
...
 get_grid(dh::AbstractDofHandler)
 
+"""
+    SubDofHandler(dh::AbstractDofHandler, cellset)
+
+Create an `sdh::SubDofHandler` for the cells in `cellset`. All cells in `cellset`
+must have the same type. Fields are added to the `sdh` by using [`add!`](@ref) the 
+same way as for the standard [`DofHandler`](@ref).
+[`close!`](@ref) should be called on the parent dof handler, `dh`, after
+all `SubDofHandler`s have been created and their fields have been added.
```
termi-official, line 22: Maybe we should add a very simple and concise example to the docstring. What do you think?

src/Dofs/DofHandler.jl
```diff
@@ -12,6 +12,16 @@ Access some grid representation for the dof handler.
 """
 get_grid(dh::AbstractDofHandler)
 
+"""
+    SubDofHandler(dh::AbstractDofHandler, cellset)
```
termi-official, line 16:
```suggestion
    SubDofHandler(dh::AbstractDofHandler, cellset) <: AbstractDofHandler
```
or not?

fredrikekre:
Nice. The docstring for the DofHandler can also be improved I think. I started on something here https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...fe/dh-docs some weeks ago, if you feel like including it here.

src/Dofs/DofHandler.jl
KnutAM: Solid reference: https://www.reddit.com/r/grammar/comments/jmrdr/quick_question_an_sd_card_or_a_sd_card/

Good point regarding the set. Currently, it fails at `convert`, e.g. for `UnitRange`, but I'm not sure if it is intended to allow e.g. a range, or different types of sets, in the future. 
So instead of changing the source code, I'll opt to make this a pure doc-pr, and just put `Set{Int}` in the docstring.

src/Dofs/DofHandler.jl
KnutAM: Good point, fixed!

src/Dofs/DofHandler.jl
KnutAM: Good idea! 
Done following @fredrikekre's `DofHandler` example!

src/Dofs/DofHandler.jl
KnutAM: I suggest not to for now, because e.g. `close!(sdh)` is not supported. 
Since this interface is not fully defined yet, but once this is settled it makes sense to document IMO.

KnutAM:
> Nice. The docstring for the DofHandler can also be improved I think. I started on something here https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...fe/dh-docs some weeks ago, if you feel like including it here.

Good idea - I've merged in your branch - makes sense to have these two docstrings fairly consistent with each other wrt. scope and cross-references.

KnutAM:
I thought of putting in a cross-ref to the Porous Media example, but if I understand correctly @kimauth will make a how-to for mixed grids which would be more suitable to link to, and suggest that the docstring will be updated in that future PR.

PR #830, opened 2023-10-19
Extensible InterpolationInfo

termi-official:
Towards #398 . I can also need this in FerriteViz.jl to provide a cleaner solution to visualize stress fields.

### DIFF
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -89,58 +89,44 @@ struct InterpolationInfo
     adjust_during_distribution::Bool
     n_copies::Int
     is_discontinuous::Bool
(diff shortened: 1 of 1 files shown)


termi-official:
Matrix-valued dof distribution and field evaluation seems to work. Proof of concept in https://github.com/Ferrite-FEM/FerriteViz.jl/pull/103/commits/536b4ef91980e0d456eacc1e5fd110defe3caeee . Gradients of matrix-valued stuff is not covered for now, because they require more work (and are not needed in FerriteViz for now)

fredrikekre (approved):

PR #831, opened 2023-10-23
Bump WriteVTK  compat

termi-official:


KnutAM:
I think this is already allowed, ref https://pkgdocs.julialang.org/v1/compatibility/#Compatibility (and docs/Manifest on master is 1.18) - or did I miss something?

termi-official:
Indeed. I only remembered this rule here https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0 . :)

PR #835, opened 2023-11-03
add isequal and hash methods to check tuple in BoundaryIndex Sets without type piracy 

koehlerson:
See https://julialang.slack.com/archives/C01R8JU271D/p1698848270253179 slackhole inc

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -765,8 +765,11 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #For (cellid, faceidx) in faceset
-        Base.in(v::Tuple{Int, Int}, s::Set{$INDEX}) = in($INDEX(v), s)
+        #necessary for (cellid, faceidx) in faceset
+        Base.isequal(x::$INDEX, y::$INDEX) = x.idx[1] == y.idx[1] && x.idx[2] == y.idx[2]
+        Base.isequal(x::Tuple{Int, Int}, y::$INDEX) = x[1] == y.idx[1] && x[2] == y.idx[2]
+        Base.isequal(y::$INDEX, x::Tuple{Int, Int}) = x[1] == y.idx[1] && x[2] == y.idx[2]
+        Base.hash(x::$INDEX,h::UInt) = Base.hash(x.idx,h)
     end
 end
 
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -765,11 +765,11 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #Necessary to check if, e.g. `(cellid, faceidx) in faceset`
-        Base.isequal(x::$INDEX, y::$INDEX) = x.idx[1] == y.idx[1] && x.idx[2] == y.idx[2]
+        # Necessary to check if, e.g. `(cellid, faceidx) in faceset`
+        Base.isequal(x::$INDEX, y::$INDEX) = x.idx == y.idx
         Base.isequal(x::Tuple{Int, Int}, y::$INDEX) = x[1] == y.idx[1] && x[2] == y.idx[2]
         Base.isequal(y::$INDEX, x::Tuple{Int, Int}) = x[1] == y.idx[1] && x[2] == y.idx[2]
-        Base.hash(x::$INDEX,h::UInt) = Base.hash(x.idx,h)
+        Base.hash(x::$INDEX, h::UInt) = hash(x.idx, h)
     end
 end
 
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -765,7 +765,7 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #necessary for (cellid, faceidx) in faceset
+        #Necessary to check if, e.g. `(cellid, faceidx) in faceset`
         Base.isequal(x::$INDEX, y::$INDEX) = x.idx[1] == y.idx[1] && x.idx[2] == y.idx[2]
         Base.isequal(x::Tuple{Int, Int}, y::$INDEX) = x[1] == y.idx[1] && x[2] == y.idx[2]
         Base.isequal(y::$INDEX, x::Tuple{Int, Int}) = x[1] == y.idx[1] && x[2] == y.idx[2]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -319,6 +319,10 @@ more discussion).
   **To upgrade** replace the quadrature rule passed to `FaceValues` with a
   `FaceQuadratureRule`.
 
+- Checking if a face `(ele_id, local_face_id) ∈ faceset` has been previously implemented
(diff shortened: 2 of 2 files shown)


termi-official (approved):

src/Grid/grid.jl
```diff
@@ -765,8 +765,11 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #For (cellid, faceidx) in faceset
-        Base.in(v::Tuple{Int, Int}, s::Set{$INDEX}) = in($INDEX(v), s)
+        #necessary for (cellid, faceidx) in faceset
```
KnutAM, line 768: When reading this I first thought `for` referred to a for loop, but I think the unpacking is done elsewhere.
```suggestion
        #Necessary to check if, e.g. `(cellid, faceidx) in faceset`
```

KnutAM (commented):
Nice! 

Are there any performance differences?

koehlerson:
> Nice!
> 
> Are there any performance differences?

master
```
julia> @benchmark in((1,2),testset)
BenchmarkTools.Trial: 10000 samples with 989 evaluations.
 Range (min … max):  48.218 ns … 69.210 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     48.507 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   48.717 ns ±  0.988 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▆▆█▇█▆▄                              ▁ ▁▁                   ▂
  ███████▆▃▃▁▄▄▅▃▅▄▅▄▃▁▅▃▃▄▃▁▄▄▅▄▄▁▅▆▇█████▇▇█▇██▆▇▇▇▆▇▇▆▆▆▆▆ █
  48.2 ns      Histogram: log(frequency) by time        53 ns <

```

PR
```
julia> @benchmark in((1,2),testset)
BenchmarkTools.Trial: 10000 samples with 994 evaluations.
 Range (min … max):  31.995 ns … 63.816 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     32.391 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   32.565 ns ±  1.520 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▃  █▃                                                       ▁
  █▅▃██▆▃▃▁▁▁▃▁▁▁▁▁▁▃▁▁▁▁▁▁▁▃▁▁▁▁▁▃▁▁▁▁▁▃▄▃▃▆▅▄▆▅▄▄▅▅▅▅▃▄▇▄▃▄ █
  32 ns        Histogram: log(frequency) by time      38.6 ns <

```

KnutAM (approved):
LGTM!

Perhaps add a changelog entry before merging

koehlerson:
I just noticed the following, which I encountered in the p4est branch as well. If we hash now `FaceIndex(1,2)` and `EdgeIndex(1,2)` the hash will be the same. This can be problematic if you have `Set{Union{FaceIndex,EdgeIndex}}` and do some set operations (which I think rely on the hashes and not only the `isequal` operation). In the p4est branch I just wrapped the `Tuple{Int,Int}` in a `Tuple{Int,Tuple{Int,Int}}` where the first int is sort of a category integer, but not sure if relevant

KnutAM:
Good point - did not think of this! 

I tried to understand what's going on, and AFAIU it works because `isequal(FaceIndex(1,2), EdgeIndex(1,2))` is `false`, such that the following check give `false` and the loop continues until eventually returning a negative index (indicating that the key wasn't found), for a set without `FaceIndex(1,2)` but with `EdgeIndex(1,2)`. 

https://github.com/JuliaLang/julia/blob/6d4f40957c6e68ecfd1dc13115beadff493406fd/base/dict.jl#L289

That being said, I think we need @fredrikekre or @KristofferC to confirm that this way is ok. The [rule for defining the hash](https://docs.julialang.org/en/v1/base/base/#Base.hash): `"Compute an integer hash code such that isequal(x,y) implies hash(x)==hash(y)"`, use *implies*, so to my understanding the implementation in this PR follows the docstring.

koehlerson:
yeah I'm just unsure if the statement needs to hold the other way around too. So should `hash(x) == hash(y)` imply `isequal(x,y)`. If so, then we should change it.

src/Grid/grid.jl
```diff
@@ -765,8 +765,11 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #For (cellid, faceidx) in faceset
-        Base.in(v::Tuple{Int, Int}, s::Set{$INDEX}) = in($INDEX(v), s)
+        #Necessary to check if, e.g. `(cellid, faceidx) in faceset`
+        Base.isequal(x::$INDEX, y::$INDEX) = x.idx[1] == y.idx[1] && x.idx[2] == y.idx[2]
+        Base.isequal(x::Tuple{Int, Int}, y::$INDEX) = x[1] == y.idx[1] && x[2] == y.idx[2]
+        Base.isequal(y::$INDEX, x::Tuple{Int, Int}) = x[1] == y.idx[1] && x[2] == y.idx[2]
+        Base.hash(x::$INDEX,h::UInt) = Base.hash(x.idx,h)
```
KnutAM, line 772: Sorry, didn't spot this in the first round. 
```suggestion
        Base.hash(x::$INDEX, h::UInt) = hash(x.idx, h)
```
Not required to remove the `Base.` in the last, not sure if it was intentional to highlight that we use the function from `Base`?

koehlerson: yeah, it was somewhat intentional to show that we forward to the `Base` function, but I can remove it!

KnutAM:
> yeah I'm just unsure if the statement needs to hold the other way around too. So should hash(x) == hash(y) imply isequal(x,y). If so, then we should change it.

At least the [docstring](https://docs.julialang.org/en/v1/base/base/#Base.isequal) for `isequal` is consistent with our interpretation: `"isequal(x,y) must imply that hash(x) == hash(y)."`

src/Grid/grid.jl
KnutAM: Fine to keep to IMO, makes it clear for the one reading the code!

termi-official:
> > yeah I'm just unsure if the statement needs to hold the other way around too. So should hash(x) == hash(y) imply isequal(x,y). If so, then we should change it.
> 
> At least the [docstring](https://docs.julialang.org/en/v1/base/base/#Base.isequal) for `isequal` is consistent with our interpretation: `"isequal(x,y) must imply that hash(x) == hash(y)."`

This does not imply the reverse.

src/Grid/grid.jl
```diff
@@ -765,8 +765,11 @@ for INDEX in (:VertexIndex, :EdgeIndex, :FaceIndex)
         #To be able to do a,b = faceidx
         Base.iterate(I::($INDEX), state::Int=1) = (state==3) ?  nothing : (I[state], state+1)
 
-        #For (cellid, faceidx) in faceset
-        Base.in(v::Tuple{Int, Int}, s::Set{$INDEX}) = in($INDEX(v), s)
+        #Necessary to check if, e.g. `(cellid, faceidx) in faceset`
+        Base.isequal(x::$INDEX, y::$INDEX) = x.idx[1] == y.idx[1] && x.idx[2] == y.idx[2]
```
fredrikekre, line 769:
```suggestion
        # Necessary to check if, e.g. `(cellid, faceidx) in faceset`
        Base.isequal(x::$INDEX, y::$INDEX) = x.idx == y.idx
```

fredrikekre (commented):
This is fine. You would still see strange things for e.g. `Set{Union{FaceIndex, Tuple{Int,Int}}}` where there is no way to differentiate `FaceIndex((1, 2))` and `(1, 2)`. I don't think such mixed sets are so common though.

Issue #844, opened 2023-11-16
Hyperelasticity example is not polyconvex


koehlerson:
After a conversation with my Prof. I noticed that we advertise a model that is not polyconvex as far as I understand.

$$
\Psi(\mathbf{C}) = \frac{\mu}{2} (I_1 - 3) - \mu \ln(J) + \frac{\lambda}{2} \ln(J)^2
$$

The last part is problematic since taking twice the derivative yields

$$
(1-\ln(J))\frac{1}{J^2}
$$

which can become smaller or equal 0 at $J=10$. I think a lot of people copy paste that function and it would be good to advertise a well posed function even tho it is the same function that is used in the FEniCS tutorial and maybe at other places too.

koehlerson:
![image](https://github.com/Ferrite-FEM/Ferrite.jl/assets/27497290/fff646a9-f9b3-41e5-b9ae-be59557247fe)

![image](https://github.com/Ferrite-FEM/Ferrite.jl/assets/27497290/09c13048-352c-4279-a85e-40c91f76aca6)

Here is a list with possible options and associated constraints/effects

koehlerson:
The one from Wikipedia is fine https://en.wikipedia.org/wiki/Neo-Hookean_solid

PR #846, opened 2023-11-16
change to well posed NeoHooke

koehlerson:
https://github.com/Ferrite-FEM/Ferrite.jl/issues/844

### DIFF
--- a/docs/src/literate-tutorials/hyperelasticity.jl
+++ b/docs/src/literate-tutorials/hyperelasticity.jl
@@ -74,7 +74,7 @@ using Ferrite, Tensors, TimerOutputs, ProgressMeter, IterativeSolvers
 # We use one possible, well-posed formulation from [Wikipedia](https://en.wikipedia.org/wiki/Neo-Hookean_solid)
 # !!! note
 #     The Neo-Hooke model is only a well defined terminology in the incompressible case.
-#     Thus, only $W(C)$ specifies the neo-Hookean behavior, the volume penalty $U(J)$ can vary in different formulations.
+#     Thus, only $W(\mathbf{C})$ specifies the neo-Hookean behavior, the volume penalty $U(J)$ can vary in different formulations.
 #     In order to obtain a well-posed problem, it is crucial to choose a convex formulation of $U(J)$.
 #     Other examples for $U(J)$ can be found, e.g. in [Hol:2000:nsm; Eq. (6.138)](@cite)
 #     ```math
@@ -82,7 +82,7 @@ using Ferrite, Tensors, TimerOutputs, ProgressMeter, IterativeSolvers
 #     ```
 #     where [SimMie:1992:act; Eq. (2.37)](@cite) published a non-generalized version with $\beta=-2$.
 #     This shows the possible variety of $U(J)$ while all of them refer to compressible neo-Hookean models.
-#     Sometimes the modified first invariant $\overbar{I}_1=\frac{I_1}{I_3^{1/3}}$ is used in $W((\mathbf{C})$ instead of $I_1$.
+#     Sometimes the modified first invariant $\overline{I}_1=\frac{I_1}{I_3^{1/3}}$ is used in $W(\mathbf{C})$ instead of $I_1$.
 # From the potential we obtain the second Piola-Kirchoff stress ``\mathbf{S}`` as
 #
 # ```math
--- a/docs/src/literate-tutorials/hyperelasticity.jl
+++ b/docs/src/literate-tutorials/hyperelasticity.jl
@@ -64,7 +64,7 @@ using Ferrite, Tensors, TimerOutputs, ProgressMeter, IterativeSolvers
 # We shall use a neo-Hookean model, where the potential can be written as
 #
 # ```math
-# \Psi(\mathbf{C}) = \frac{\mu}{2} (I_1 - 3) - \mu \ln(J) + \frac{\lambda}{2} \ln(J)^2,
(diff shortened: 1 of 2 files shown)


koehlerson:
image
looks quite the same, but `norm(u)` is slightly different, probably to be expected?

koehlerson:
alternatively in https://onlinelibrary.wiley.com/doi/full/10.1002/nme.4351 another formulation is used which is not normalized. The current one from Wikipedia is normalized (in the sense of $\psi(\boldsymbol{1})=0$)

KnutAM:
In true Ferrite-spirit: https://www.dealii.org/current/doxygen/deal.II/step_44.html#NeoHookeanmaterials and ref to Holzapfel (2001) therein :)
Perhaps nice to comment in the example that there are different possible formulations for the volumetric part (as is done in the deal.ii example)?

koehlerson:
Citing a 400 pages book by "see Holzapfel (2001)" is an engineering practice, I'll never understand. The reference is equation (6.138). Within the book, the actual reference is Simo and Miehe (1992) (https://www.sciencedirect.com/science/article/pii/004578259290170O equation 2.37) . The chapter of equation (6.138) talks about "ogden model for compressible rubber-like materials". I can add later a reference to Simo and Miehe for one alternative choice and maybe another statement that this is free of choice as long as it is strictly convex in $J$

KnutAM:
First of all: Sorry if I made you do a lot of digging now - that was not my intention! 

I think you should implement the (strictly convex, or polyconvex if that suffices) free energy that you prefer. (I agree that Psi(I)=0 is nice). I just thought that the following could be nice

1) Referring to where the model we use is taken from (doesn't have to be the original reference if well-established, could refer to an equation in a book)
2) Commenting that multiple options for the volumetric part exist/can be made (but additional examples/refs are not required IMO as it isn't a theory manual.)

koehlerson:
> First of all: Sorry if I made you do a lot of digging now - that was not my intention!

Sorry if the message was perceived that way. It was meant more in the direction that I'd rather see in our docs some precise citation, especially since this tutorial may be the template for beginners (including me back then). The rant was more about the general situation in engineering publications where I encounter often the same.

> 
> I think you should implement the (strictly convex, or polyconvex if that suffices) 

if it's strictly convex in $J$ then $\psi(\boldsymbol{F})$ becomes polyconvex w.r.t. $\boldsymbol{F}$ which is a key ingredient of well posedness. That's why you want the strictly convex term in $J$.

> 
>     1. Referring to where the model we use is taken from (doesn't have to be the original reference if well-established, could refer to an equation in a book)

Yeah that's fine for me. Will do that tomorrow.

> 
>     2. Commenting that multiple options for the volumetric part exist/can be made (but additional examples/refs are not required IMO as it isn't a theory manual.)

Yeah I was planning some `note !!!` or the like

lijas:
Perhaps this is a stupid questions, but does not chaning the potentional mean that it is no longar a Neo-hookean model? Or are there many types neo-hookean models? 
Also, the original formulation is only breaks down for really large deformation (J=10), right?

KnutAM:
> Perhaps this is a stupid questions, but does not chaning the potentional mean that it is no longar a Neo-hookean model? Or are there many types neo-hookean models?

To my understanding, the true/original "NeoHooke" model is incompressible, but can be extended to a compressible formulation for which there are different variants of the volumetric contribution.

lijas:
Ok thanks for clarifying. The original NeoHooke model would then require to be used with taylor-hood (displacement preasure) elements, I guess?

koehlerson:
> Ok thanks for clarifying. The original NeoHooke model would then require to be used with taylor-hood (displacement preasure) elements, I guess?

Neo Hooke is just $C_1 (I_1 - 3)$ and if you compute this without a pressure formulation you can increase the volume freely leading to many (infinite I guess) solutions. So either a pressure field is incorporated for the incompressible model or you add some well posed penalty terms as we have in the example here. But it's definitely worth to mention that there is no such thing as "the" compressible Neo Hooke model

termi-official (changes requested):
Hyperelasticity docs are broken on the deploy. I left a commen where this might come from. Can you check locally?

docs/src/literate-tutorials/hyperelasticity.jl
```diff
@@ -64,11 +64,23 @@ using Ferrite, Tensors, TimerOutputs, ProgressMeter, IterativeSolvers
...
-# \Psi(\mathbf{C}) = \frac{\mu}{2} (I_1 - 3) - \mu \ln(J) + \frac{\lambda}{2} \ln(J)^2,
+# \Psi(\mathbf{C}) = \underbrace{\frac{\mu}{2} (I_1 - 3)}_{W(C)} \underbrace{- {\mu} \ln(J) + \frac{\lambda}{2} (J - 1)^2}_{U(J)},
 # ```
 #
 # where ``I_1 = \mathrm{tr}(\mathbf{C})`` is the first invariant, ``J = \sqrt{\det(\mathbf{C})}``
 # and ``\mu`` and ``\lambda`` material parameters.
+# Note that we use here a compressible formulation of the original incompressible Neo Hooke model $W(F)$.
+# In order to compensate the compressible behavior, a volume penalty $U(J)$ is incorporated in $\psi$.
+# !!! note
+#    The Neo-Hooke model is only a well defined terminology in the incompressible case.
```
termi-official, line 75: I think you need one more space here for the indentation.

KnutAM (commented):
Great job fixing this and adding more details on neo-Hookean models with proper references!

PR #859, opened 2023-12-11
Use GeometryMapping in BCValues

### LINKED ISSUE #849: Use GeometryMapping in BCValues
The `GeometryMapping` in #764 should be reusable in `BCValues`. To avoid littering that PR with even more stuff, I'm leaving some preliminary and untested code here as a reference.

To be done in a separate PR after merging 764...

```julia
"""
    BCValues(func_interpol::Interpolation, geom_interpol::Interpolation, boundary_type::Union{Type{<:BoundaryIndex}})

`BCValues` stores the shape values at all faces/edges/vertices (depending on `boundary_type`) for the geometric interpolation (`geom_interpol`),
for each dof-position determined by the `func_interpol`. Used mainly by the `ConstraintHandler`.
"""
struct BCValues{V_GM, FQR} <: AbstractFaceValues
    geo_mapping::V_GM # AbstractVector{GeometryMapping}
    fqr::FQR          # FaceQuadratureRule
    current_face::ScalarWrapper{Int}
end

BCValues(func_interpol::Interpolation, geom_interpol::Interpolation, boundary_type::Type{<:BoundaryIndex} = Ferrite.FaceIndex) =
    BCValues(Float64, func_interpol, geom_interpol, boundary_type)

function BCValues(::Type{T}, func_interpol::Interpolation{refshape}, geom_interpol::Interpolation{refshape}, boundary_type::Type{<:BoundaryIndex} = Ferrite.FaceIndex) where {T,dim,refshape <: AbstractRefShape{dim}}
    # set up quadrature rules for each boundary entity with dof-positions
    # (determined by func_interpol) as the quadrature points
    interpolation_coords = reference_coordinates(func_interpol)

    qrs = QuadratureRule{refshape,T,dim}[]
    for boundarydofs in dirichlet_boundarydof_indices(boundary_type)(func_interpol)
        dofcoords = Vec{dim,T}[]
        for boundarydof in boundarydofs
            push!(dofcoords, interpolation_coords[boundarydof])
        end
        qrf = QuadratureRule{refshape,T}(fill(T(NaN), length(dofcoords)), dofcoords) # weights will not be used
        push!(qrs, qrf)
    end
    fqr = FaceQuadratureRule(qrs)
    geo_mapping = [GeometryMapping{0}(T, geom_interpol, qr) for qr in qrs]

    BCValues(geo_mapping, fqr, ScalarWrapper(1))
end

nfaces(bcv::BCValues) = length(bcv.geo_mapping)
getcurrentface(bcv) = bcv.current_face[]

function set_current_face!(bcv::BCValues, face_nr::Int)
    # Checking face_nr before setting current_face allows us to use @inbounds 
    # when indexing by getcurrentface(fv) in other places!
    checkbounds(Bool, 1:nfaces(bcv), face_nr) || throw(ArgumentError("Face index out of range."))
    bcv.current_face[] = face_nr
end

getnquadpoints(bcv::BCValues) = @inbounds getnquadpoints(bcv.fqr, getcurrentface(bcv))
```

KnutAM:
Closes #849 

Using `GeometryMapping` in `BCValues` slightly reduces code duplication and makes the similarity (and differences) between `BCValues` and `FaceValues` clearer. Not sure why faster on construction, but for `update!` it probably comes from the performance annotations used following the `FaceValues` setup with inbounds and inlines. 

Updated benchmark after #860, values from run 2 below

**Construction**
* PR: 18.499 ms (4266 allocations: 93.64 MiB)
* master: 20.720 ms (4258 allocations: 93.64 MiB)

**`update!(ch, t)`**
* PR: 333.900 μs (18 allocations: 1.47 KiB)
* master: 373.200 μs (20 allocations: 1.53 KiB)

<details> <summary> Benchmark script and results </summary>

```julia
using Ferrite, BenchmarkTools


function make_dh(grid)    
    dh = DofHandler(grid)
    add!(dh, :u, ip^2)
    add!(dh, :v, ip)
    close!(dh)
    return dh
end
dh = make_dh(grid)

function make_ch(dh, grid)
    ch = ConstraintHandler(dh)
    add!(ch, Dirichlet(:u, getfaceset(grid, "right"), Returns(zero(Vec{2}))))
    add!(ch, Dirichlet(:v, getfaceset(grid, "left"), Returns(0.0)))
    close!(ch)
    return ch
end

ch = make_ch(dh, grid)
@btime make_ch($dh, $grid)

update!(ch, 1.0)

@btime update!($ch, 2.0)

#= 
>>> PR
run 1
  18.414 ms (4266 allocations: 93.64 MiB)
  341.900 μs (18 allocations: 1.47 KiB)

run 2
  18.499 ms (4266 allocations: 93.64 MiB)
  333.900 μs (18 allocations: 1.47 KiB)

>>> master
run 1
    22.356 ms (4258 allocations: 93.64 MiB)
  367.800 μs (20 allocations: 1.53 KiB)

run 2
  20.720 ms (4258 allocations: 93.64 MiB)
  373.200 μs (20 allocations: 1.53 KiB)

=#
```

</details>


src/Dofs/ConstraintHandler.jl
```diff
@@ -419,8 +419,8 @@ function _update!(inhomogeneities::Vector{T}, f::Function, boundary_entities::Se
         reinit!(cc, cellidx)
 
         # no need to reinit!, enough to update current_entity since we only need geometric shape functions M
-        boundaryvalues.current_entity[] = entityidx
-
+        set_current_face!(boundaryvalues, entityidx)
+
```
termi-official, line 423:
```suggestion

```

src/FEValues/FaceValues.jl
```diff
@@ -158,15 +158,15 @@ function Base.show(io::IO, d::MIME"text/plain", fv::FaceValues)
...
+for each dof-position determined by the `func_interpol`. Used mainly by the `ConstraintHandler`.
 """
-struct BCValues{T}
-    M::Array{T,3}
-    nqp::Array{Int}
-    current_entity::ScalarWrapper{Int}
+struct BCValues{V_GM, FQR} <: AbstractFaceValues
+    geo_mapping::V_GM # AbstractVector{GeometryMapping}
+    fqr::FQR          # FaceQuadratureRule
+    current_face::ScalarWrapper{Int}
```
termi-official, line 169: This should also give the possibility for boundary conditions involving normals, right?

src/FEValues/FaceValues.jl
```diff
@@ -186,31 +186,22 @@ function BCValues(::Type{T}, func_interpol::Interpolation{refshape}, geom_interp
...
-    face = bcv.current_entity[]
-    @inbounds for i in 1:n_base_funcs
-        x += bcv.M[i,q_point,face] * xh[i] # geometric_value(fe_v, q_point, i) * xh[i]
-    end
-    return x
-end
+@inline get_geo_mapping(bcv::BCValues) = @inbounds bcv.geo_mapping[getcurrentface(bcv)]
+@inline geometric_value(bcv::BCValues, q_point::Int, base_func::Int) = geometric_value(get_geo_mapping(bcv), q_point, base_func)
+@inline getngeobasefunctions(bcv::BCValues) = getngeobasefunctions(get_geo_mapping(bcv))
+@inline getnquadpoints(bcv::BCValues) = @inbounds getnquadpoints(bcv.fqr, getcurrentface(bcv))
```
termi-official, line 207: Why are we selective about the `@inbounds` here?

termi-official (commented):
This breaks edge boundary conditions, right?

src/FEValues/FaceValues.jl
KnutAM: It would be possible to send the normal into the user-supplied function, so as long as all components are prescribed that would work. But, unfortunately, the normal is still not well-defined at cell vertices, so this would only be correct for smooth boundaries.

One way to implement this would be to store the prescribed dof, coordinate, and normal in a pre-calculated list, this would make it possible to define an averaged normal direction. But I'm not sure if this is wanted? (And would for sure be a different PR)

src/FEValues/FaceValues.jl
KnutAM: Because `getcurrentface(bcv)` is guaranteed to give a valid index due to the check in the `set_current_face!` function. 
The access with indices `q_point` and `base_func` in `geometric_value` is not safe, but the `@inline` allows the user to circumvent the boundscheck if desired.

KnutAM:
> This breaks edge boundary conditions, right?

That should work fine because we use the appropriate "dof-locations" based on `dirichlet_boundarydof_indices(boundary_type)(func_interpol)`. I didn't check this specifically, but I assume these cases are covered by the test suite.

src/Dofs/ConstraintHandler.jl
KnutAM: Thanks

KnutAM:
In view of #1152 I think this might be the wrong direction actually.

PR #860, opened 2023-12-11
Make close of ConstraintHandler 50x faster

KnutAM:
While making #859 I discovered that `close!(ch)` was quite slow. Did a profile and turns out most time was spent on `copy!(ch.free_dofs, setdiff(1:ndofs(ch.dh), ch.prescribed_dofs))`. For the benchmarked test case this PR reduces the construction of a ConstraintHandler from about 2 s to 0.04 s. 
(Take exact speedup with grain of salt since done on laptop using battery power)

```julia
using Ferrite

dh = DofHandler(grid)
add!(dh, :u, ip^2)
add!(dh, :v, ip)
close!(dh)

function make_ch(dh, grid)
    ch = ConstraintHandler(dh)
    add!(ch, Dirichlet(:u, getfaceset(grid, "right"), Returns(zero(Vec{2}))))
    add!(ch, Dirichlet(:v, getfaceset(grid, "left"), Returns(0.0)))
    close!(ch)
    return ch
end

ch = make_ch(dh, grid)
@time make_ch(dh, grid)

# PR
# run 1: 0.037628 seconds (4.26 k allocations: 93.643 MiB, 27.37% gc time)
# run 2: 0.027672 seconds (4.26 k allocations: 93.643 MiB)
  
# master 
# run 1: 1.989513 seconds (4.28 k allocations: 528.475 MiB, 2.04% gc time)
# run 2: 1.974789 seconds (4.28 k allocations: 528.475 MiB)
  
```

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -177,6 +177,23 @@ isclosed(ch::ConstraintHandler) = ch.closed[]
 free_dofs(ch::ConstraintHandler) = ch.free_dofs
 prescribed_dofs(ch::ConstraintHandler) = ch.prescribed_dofs
 
+# Equivalent to `copy!(out, setdiff(1:n_entries, diff))`, but requires that 
+# `issorted(diff)` and that `1 ≤ diff[1] ≤ diff[end] ≤ n_entries`
+function _sorted_setdiff!(out::Vector{Int}, n_entries::Int, diff::Vector{Int})
+    n_diff = length(diff)
+    resize!(out, n_entries - n_diff)
+    diff_ind = out_ind = 1
+    for i in 1:n_entries
+        if diff_ind ≤ n_diff && i == diff[diff_ind]
+            diff_ind += 1
+        else
+            out[out_ind] = i
+            out_ind += 1
+        end
+    end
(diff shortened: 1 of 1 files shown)


fredrikekre:
Nice, perhaps extract to a `sorted_setdiff!` helper?

KristofferC:
I wonder if there should exist a `free_dofs` variable computed at all? It seems wasteful to have a long array of numbers that is almost the dense set of 1:ndofs.

KnutAM:
If storing it is a problem, perhaps a freedof iterator can be created using similar logic to this PR? Potentially storing a `Vector{UnitRange{Int}}` for efficiency?
It is quite nice to have an easy way of doing e.g. `r[fdofs]` (even if not needed as it is used in the examples)

KristofferC:
> It is quite nice to have an easy way of doing e.g. r[fdofs] (even if not needed as it is used in the examples)

I guess I imagined something like https://github.com/JuliaData/InvertedIndices.jl where you would do `r[Not(cdofs)]`.

KnutAM:
Looks very neat - should for sure give the required functionality!
Nice to add to an example and docs if someone feels like removing `ch.free_dofs`!

Issue #848, opened 2023-11-29
Ginkgo support


termi-official:
Extension to allow the assembly into the matrix format from https://github.com/youwuyou/Ginkgo.jl

ShrutiRDalvi:
hello,
Is this issue solved? If not I'd like to work on it, please let me know how to go about it as I'm new to open source Thanks, a lot

termi-official:
I think @youwuyou started working on this. And maybe we should wait until https://github.com/Ferrite-FEM/Ferrite.jl/pull/864 is merged before this should be done, because it will be analogue to do.

Issue #628, opened 2023-03-22
GPU assembly example


termi-official:
It might be worth to explore assembling directly into `CuArray` and use AMGX to solve the resulting problem.

PR #864, opened 2023-12-23
Modularize assembly internals and add SparseMatrixCSR extension

termi-official:
Towards #848 and #628  . This should show how we could assemble into custom sparse matrix formats.

## TODO

* [x] Fix inhomogeneous boundary conditions
* [x] Add devdocs
* [x] Add test


ext/FerriteSparseMatrixCSR.jl
```diff
@@ -0,0 +1,80 @@
...
+                Ci += 1
+            elseif Kcol < dofs[Kecol]
+                # No match yet: advance the global matrix row pointer
+                Ci += 1
+            else # Kcol > dofs[Kecol]
+                # No match: no entry exist in the global matrix for this row. This is
+                # allowed as long as the value which would have been inserted is zero.
+                iszero(val) || _missing_sparsity_pattern_error(Krow, Kcol)
+                # Advance the local matrix row pointer
+                ci += 1
```
KnutAM, line 35: This is not covered by tests. Would it make sense to add a test with a specific field coupling (I believe that is when these conditionals become active) to check these corner cases?

ext/FerriteSparseMatrixCSR.jl
```diff
@@ -0,0 +1,80 @@
...
+            end
+        end
+        # Make sure that remaining entries in this column of the local matrix are all zero
+        for i in ci:maxlookups
+            if !iszero(Ke[Kerow, permutation[i]])
+                _missing_sparsity_pattern_error(Krow, sorteddofs[i])
+            end
```
KnutAM, line 42: Seems like also this loop is never active either in tests, which seems related to above.

src/Dofs/ConstraintHandler.jl
```diff
@@ -690,6 +661,45 @@ function apply!(KK::Union{SparseMatrixCSC,Symmetric}, f::AbstractVector, ch::Con
     end
 end
 
+# Generic version to compute "f -= K*inhomogeneities"
```
KnutAM, line 664: Nice to factor this out! Add to devdocs?

KnutAM (commented):
Just took a look as I was curious, nice work on factoring out some parts - this will make it easier to extend!

ext/FerriteSparseMatrixCSR.jl
termi-official: I am not sure what to do here tbh. Actually I would like to prefer to remove this case everywhere, because this is the case when the prealloacted sparsity matrix is "too small", but the values beyond its size is zero anyway. However, this would be technically a quite annoying breaking change for users running into this case.

ext/FerriteSparseMatrixCSR.jl
termi-official: I think I need @fredrikekre to help out here. The logic is the same as the for csc assembly, I just swapped the indices. I guess this is the same case as above, that we handle "too small sparse matrices"?

ext/FerriteSparseMatrixCSR.jl
KnutAM: I meant a test like this
```julia
grid = generate_grid(Quadrilateral, (10,10));
ip = Lagrange{RefQuadrilateral,1}();
dh = DofHandler(grid); add!(dh, :u, ip); add!(dh, :v, ip); close!(dh);
coupling = [true; false;; false; true];
K = create_sparsity_pattern(SparseMatrixCSR, dh; coupling);
Ke_rand = rand(ndofs_per_cell(dh,1), ndofs_per_cell(dh,1));
Ke_zero = zeros(ndofs_per_cell(dh,1), ndofs_per_cell(dh,1));
a = start_assemble(K);
assemble!(a, celldofs(dh,1), Ke_zero); # Should work
assemble!(a, celldofs(dh,1), Ke_rand); # Should fail
```

termi-official: Is this what you had in mind?

docs/src/devdocs/assembler.md
```diff
@@ -0,0 +1,37 @@
+# [Assembler](@id devdocs-assembler)
+
+The assembler handles the insertion of the element matrices and element vectors into the system matrix and right hand side. While the CSC and CSR formats are the most common sparse matrix formats in practice, users might want to have optimized custom matrix formats for their specific use-case. The default assembler [`AssemblerSparsityPattern`](@ref) should be able to handle most cases in practice. To support a custom format users have to dispatch the following functions:
+
+```@docs
+Ferrite._assemble_inner!
```
KristofferC, line 6: Feels weird to overload a banged function...

ext/FerriteSparseMatrixCSR.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function Ferrite.allocate_matrix(::Type{SparseMatrixCSR}, sp::AbstractSparsityPattern)
+    return allocate_matrix(SparseMatrixCSR{1,Float64,Int}, sp)
+end
+
+function Ferrite.allocate_matrix(::Type{MatrixType}, sp::AbstractSparsityPattern) where {Bi, Tv, Ti, MatrixType <: SparseMatrixCSR{Bi, Tv, Ti}}
+    # Allocate CSC first ...
+    K = allocate_matrix(SparseMatrixCSC{Tv, Ti}, sp)
+    # ... and transform to SparseMatrixCSR
+    return SparseMatrixCSR{Bi}(transpose(sparse(transpose(K))))
```
lijas, line 76: Will SparseMatrixCSR reuse the arrays from SparseMatrixCSC, or will it duplicate them? 
I think it is possible to create the SparseMatrixCSR matrix efficiently without an intermediate SparseMatrixCSC representation.

lijas (commented):
Nice!
It could be nice to use the SparseMatrixCSR in one of the tutorials also (for testing and showing user that it is possible)

fredrikekre:
I suppose in all(?) cases the solver will dictate what matrix you need to, or can, use. Is there any solver that can handle SparseMatrixCSR but not SparseMatrixCSC?

termi-official:
> Nice! It could be nice to use the SparseMatrixCSR in one of the tutorials also (for testing and showing user that it is possible)

I decided to not use it in one of the tutorials for now, because I didn't wanted to clutter any of these. It should also be clear how to use it from the docstring of start_assemble ?  Which test is missing in https://github.com/Ferrite-FEM/Ferrite.jl/pull/864/files#diff-7fa26eddc1fe6a8a5341e0fcdffb1d8c256e6a566819cfcf70b625ef80ab6abd ? (At least codecov says I have coverage).

termi-official:
> I suppose in all(?) cases the solver will dictate what matrix you need to, or can, use. Is there any solver that can handle SparseMatrixCSR but not SparseMatrixCSC?

If you talk about matrix formats, then e.g. Ginkgo and Hypre just have CSR formats (and no CSC).

ext/FerriteSparseMatrixCSR.jl
termi-official: IIUC for Bi=1 the arrays will be reused. All other cases will trigger allocs. And yes we should be able to optimize these further, but I leave these for later when I have more time to think how to write the algorithms down generically. This PR merely enables support in first place.

fredrikekre:
Yea but then you wouldn't go through SparseMatrixCSR.jl but directly to a HYPRE matrix or Ginkgo matrix?

termi-official:
> Yea but then you wouldn't go through SparseMatrixCSR.jl but directly to a HYPRE matrix or Ginkgo matrix?

Yes

docs/src/devdocs/assembly.md
```diff
@@ -1,5 +1,35 @@
...
+
+## Custom Assembler
+
+In case the default assembler is insufficient, users can implement a custom assemblers. For this, they can create a custom type and dispatch the following functions.
+
+```
+Ferrite.start_assemble!
+Ferrite.finish_assemble!
+Ferrite.assemble!
+```
```
termi-official, line 25: What is the correct way to communicate this? Adding `@docs` breaks the CI, because the doc string is already listed in references.

KnutAM: Is https://documenter.juliadocs.org/stable/man/syntax/#noncanonical-block what you want?

termi-official: You are right, thanks!

termi-official:
Thanks @KnutAM for the pointer on the canonical keyword. PR should be good to go now.

test/test_assembler_extensions.jl
```diff
@@ -0,0 +1,109 @@
...
+        J = [1, 2, 1, 2, 3, 2, 3]
+        V = zeros(7)
+        @test K == sparsecsr(I, J, V)
+        f = zeros(3)
+
+        # Check that incuding the ch doesnot mess up the pattern
+        ch = ConstraintHandler(dh)
+        add!(ch, Dirichlet(:u, getfacetset(grid, "left"), (x, t) -> 1))
+        close!(ch)
+        @test K == allocate_matrix(SparseMatrixCSR, dh, ch)
```
KnutAM, line 59: Same as above, perhaps introduce
```julia
function nonzero_indices_set(K::SparseMatrixCSR)
    cols = colvals(K);
    return Set((row, cols[i]) for row in 1:size(K, 1) for i in nzrange(K, row))
end
function nonzero_indices_set(K::AbstractSparseMatrixCSC)
    rows = rowvals(K);
    return Set((rows[i], col) for col in 1:size(K, 2) for i in nzrange(K, col))
end
```
to also allow comparing with the builtin assembly/sparsity?

test/test_assembler_extensions.jl
```diff
@@ -0,0 +1,109 @@
...
+        # Check if assembly works
+        assembler = start_assemble(K, f)
+        ke = [-1.0 1.0; 2.0 -1.0]
+        fe = [1.0, 2.0]
+        assemble!(assembler, [1, 2], ke, fe)
+        assemble!(assembler, [3, 2], ke, fe)
+        I = [1, 1, 2, 2, 2, 3, 3]
+        J = [1, 2, 1, 2, 3, 2, 3]
+        V = [-1.0, 1.0, 2.0, -2.0, 2.0, 1.0, -1.0]
+        finish_assemble(assembler)
```
KnutAM, line 70: Not required?
Alternatively,
```suggestion
        (K, f) === finish_assemble(assembler)
```

KnutAM (commented):
Really nice to make this more flexible!

Do I understand this PR correctly that the main purpose is to introduce the necessary abstractions to allow extensions to different (CSR) matrix formats, and that the SparseMatrixCSR extension is just an example? 

While testing to understand, I realized the following behavior,
```julia
julia> K = sparsecsr([2], [2], [2.])
2×2 SparseMatrixCSR{1, Float64, Int64} with 1 stored entry:
  [2, 2]  =  2.0

julia> K+K
2×2 Matrix{Float64}:
 0.0  0.0
 0.0  4.0
```
Which could lead to quite severe user issues if used in e.g. the transient heat tutorial [here](https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/transient_heat_equation/#Solution-of-the-system). So perhaps, instead of adding an extension that we have to maintain, could we add the implementation to the test suite instead? 

And when https://github.com/JuliaSparse/SparseArrays.jl/pull/546 is finalized, we could directly support most ops for abstract CSR inside Ferrite (but unless `SparseMatricesCSR` adopts the interface, we probably don't want to keep supporting that)?

termi-official:
Thanks for the review Knut!


> Do I understand this PR correctly that the main purpose is to introduce the necessary abstractions to allow extensions to different (CSR) matrix formats, and that the SparseMatrixCSR extension is just an example?

Almost. It **should** contain all changes to allow users to add support for any of their favorite serial matrix formats in their packages (BSR,CSR,CSC,...) and it adds helpers for CSR/CSC to do some of the stuff. The main blocker on master is right now that there is no easy way to add support for new matrix formats in the constraint handler and this PR introduces necessary hooks. I am not sure if I missed somthing more advanced than Dirichlet tho.


> So perhaps, instead of adding an extension that we have to maintain, could we add the implementation to the test suite instead?

Can elaborate why adding an extension should make us responsible for fixing bugs in upstream packages? If I would not add the CSR stuff here then there would be again the question why the refactor which I propose is necessary in first place...

> And when https://github.com/JuliaSparse/SparseArrays.jl/pull/546 is finalized, we could directly support most ops for abstract CSR inside Ferrite (but unless SparseMatricesCSR adopts the interface, we probably don't want to keep supporting that)?

I have essentially given up on the effort of establishing an interface for different matrix types centrally right now, as it does not seem to be something people actually want, if you look at all the discussion in e.g. slack or discourse. But to answer the question: Yes, that was my initial plan, to have interfaces to different matrix formats which we can adopt. I just started adding a custom intermediate layer in Thunderbolt to circumvent the lack of a well-defined interface.

test/test_assembler_extensions.jl
```diff
@@ -0,0 +1,109 @@
...
+        f0 = K0 * sol
+        f1 = K1 * sol
+        f2 = K2 * sol
+        apply!(K0, f0, ch)
+        apply!(K1, f1, ch)
+        apply!(K2, f2, ch)
+        @test K0 == K1
+        @test K1 == K2
+        @test f0 == f1
+        @test f1 == f2
```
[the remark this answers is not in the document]
termi-official: Why? This just test of `apply!` adopted from the CSC test battery here https://github.com/Ferrite-FEM/Ferrite.jl/blob/a503c0083eebb70a8b429b78cd6911d2bf283f58/test/test_constraints.jl#L1658-L1682 .

KnutAM:
> Can elaborate why adding an extension should make us responsible for fixing bugs in upstream packages?

No, I meant that it makes us responsible for maintaining the extension. And IIUC we don't actually have a use case to support `SparseMatricesCSR.SparseMatrixCSR` specifically (since AFAIU there are no solvers for that matrix type), it is an extra maintenance burden to support that one. Isn't our goal to support (or allow extensibility to support) formats with solvers, e.g. `Ginkgo` or `Hypre`?

Another option could be to add a how-to on how to support a new matrix format, if the interface is considered stable enough? (Or just mark as experimental/devdocs)

test/test_assembler_extensions.jl
```diff
@@ -0,0 +1,109 @@
...
+        dh = DofHandler(grid)
+        add!(dh, :u, Lagrange{RefLine, 1}())
+        close!(dh)
+
+        # Check if the matrix is correctly allocated
+        K = allocate_matrix(SparseMatrixCSR, dh)
+        I = [1, 1, 2, 2, 2, 3, 3]
+        J = [1, 2, 1, 2, 3, 2, 3]
+        V = zeros(7)
+        @test K == sparsecsr(I, J, V)
```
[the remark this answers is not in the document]
termi-official: Good point.

test/test_assembler_extensions.jl
```diff
@@ -0,0 +1,109 @@
...
+        Ke_zeros = zeros(ndofs_per_cell(dh, 1), ndofs_per_cell(dh, 1))
+        Ke_rand = rand(ndofs_per_cell(dh, 1), ndofs_per_cell(dh, 1))
+        dofs = celldofs(dh, 1)
+
+        for c1 in [true, false], c2 in [true, false], c3 in [true, false], c4 in [true, false]
+            coupling = [c1; c2;; c3; c4]
+            K = allocate_matrix(SparseMatrixCSR, dh; coupling)
+            a = start_assemble(K)
+            assemble!(a, dofs, Ke_zeros)
+            finish_assemble(a)
```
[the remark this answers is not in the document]
termi-official: Codecov will yell at me if I remove test coverage for this noop. :smile:

test/test_assembler_extensions.jl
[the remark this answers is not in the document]
KnutAM: Nvm, missed the `apply` lines :D

src/assembler.jl
```diff
@@ -118,6 +119,14 @@ function assemble!(a::COOAssembler{T}, rowdofs::AbstractVector{Int}, coldofs::Ab
...
 end
 
+"""
+    finish_assemble!(a::AbstractAssembler)
+
+Finalize the assembly. This step is e.g. necessary for some parallel matrix formats
+where a synchronization step is necessary.
+"""
+finish_assemble!(a::AbstractAssembler) = nothing
+
```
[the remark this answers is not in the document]
termi-official: The alternative is that all downstream package provide their own finalizer for all formats

termi-official:
> No, I meant that it makes us responsible for maintaining the extension

Right, as with all other extensions.


> AFAIU there are no solvers for that matrix type

IterativeSolvers.jl, Krylov.jl, KrylovKit.jl to name a few and we can implement how to cheaply convert the format to most direct solvers like Pardiso.jl , which require CSR. I also use it in Thunderbolt to wrap for my parallel CSR format and for parallel preconditioners therein.


> Isn't our goal to support (or allow extensibility to support) formats with solvers, e.g. Ginkgo or Hypre?

Yes. Should I add these also to the PR?


> Another option could be to add a how-to on how to support a new matrix format, if the interface is considered stable enough? 

I wanted to do this later after PRs to support the big packages are done, so we actually can figure out if it is stable enough.

src/Dofs/ConstraintHandler.jl
```diff
@@ -743,6 +714,50 @@ function apply!(KK::Union{SparseMatrixCSC, Symmetric}, f::AbstractVector, ch::Co
     return
 end
 
+"""
+    add_inhomogeneities!(K, f::AbstractVector, inhomogeneities::AbstractVector, prescribed_dofs::AbstractVector{<:Integer}, dofmapping::Dict{Int,Int})
+
+Compute "f -= K*inhomogeneities".
+By default this is a generic version via SpMSpV kernel.
+"""
+function add_inhomogeneities!(K, f::AbstractVector, inhomogeneities::AbstractVector, prescribed_dofs::AbstractVector{<:Integer}, dofmapping)
```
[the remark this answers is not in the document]
termi-official: Good catch. That makes sense from an interface point of view.

KnutAM:
> IterativeSolvers.jl, Krylov.jl, KrylovKit.jl to name a few

Aha, then I misunderstood the conversations here, https://github.com/Ferrite-FEM/Ferrite.jl/pull/864#issuecomment-2205365442, https://github.com/Ferrite-FEM/Ferrite.jl/pull/864#issuecomment-2205371751, https://github.com/Ferrite-FEM/Ferrite.jl/pull/864#issuecomment-2205393511, https://github.com/Ferrite-FEM/Ferrite.jl/pull/864#issuecomment-2205736667.

> Should I add these also to the PR?

No, that wasn't what I meant. Related to above, I just didn't know that `SparseMatricesCSR.jl` had that wide solver support as you say.

src/assembler.jl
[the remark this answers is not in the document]
KnutAM: With unrelated I didn't mean that we shouldn't introduce it, just that it could be a separate pr, explaining why we need it and how it generalizes code. 

E.g., to me, it seems logical that the interface matches that of `finish_assemble` (returning the matrix and vector), but perhaps this doesn't make sense for the distributed synchronization?

So to avoid stalling the rest of this PR due to my lack of understanding the exact use case, I meant to suggest to do discuss that separately and not to dismiss the idea - sorry if that wasn't clear!

test/test_assembler_extensions.jl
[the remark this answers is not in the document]
KnutAM: But then you could add my suggestion below in one place (using (K, f) === ), which actually tests the interface?

termi-official:
Probably to elaborate here, most of the iterative solvers packages literally just need to define matrix-vector products to be defined for the thing you shove into them.

src/Dofs/ConstraintHandler.jl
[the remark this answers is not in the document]
KnutAM: Out of curiosity, any reason not passing the constraint handler instead of inhomogeneities, prescribed_dofs, and dofmapping? (Esp. since dofmapping isn't always used).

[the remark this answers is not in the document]
termi-official: No specific reason. The internals are not consistent here. I changed it to take the ch for now to simplify forward compatibility.

test/test_assembler_extensions.jl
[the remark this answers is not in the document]
termi-official: I left `finish_assemble` as the exclusive finish for COO assemble, which seems to be some debug thing, and have `finish_assemble!` (which does not return any new matrix type) for the sync.

src/assembler.jl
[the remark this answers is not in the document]
termi-official: See e.g. https://github.com/Ferrite-FEM/FerriteDistributed.jl/blob/87b60e364455aaf927dedd10af899193223b6956/docs/src/literate/heat_equation_hypre.jl#L110 or https://github.com/JuliaHealth/Thunderbolt.jl/blob/d71f98c3cb286074625e1930a0b6306bc92e4b4e/src/discretization/operator.jl#L440 .

`finish_assemble` returning the matrix and vector is technical and I think just used for testing and debugging.

src/Dofs/ConstraintHandler.jl
```diff
@@ -743,6 +714,53 @@ function apply!(KK::Union{SparseMatrixCSC, Symmetric}, f::AbstractVector, ch::Co
...
 end
 
+"""
+    add_inhomogeneities!(f::AbstractVector, K::AbstractMatrix, ch::ConstraintHandler)
+
+Compute "f -= K*inhomogeneities".
+By default this is a generic version via SpMSpV kernel.
+"""
+function add_inhomogeneities!(f::AbstractVector, K::AbstractMatrix, inhomogeneities::AbstractVector, prescribed_dofs::AbstractVector{<:Integer}, dofmapping)
+    return f .-= K * sparsevec(prescribed_dofs, inhomogeneities, size(K, 2))
```
KnutAM, line 724: (See combined fix below)
```suggestion
    return mul!(f, K, sparsevec(prescribed_dofs, inhomogeneities, size(K, 2)), -1, 1)
```
Reduces allocations. Significant relative speed diff for `SparseMatrixCSC`, but for `SparseMatrixCSR`, the call is dominated by a very slow sparse matrix sparse vector multiplication, which I guess should be fixed upstream.

Also needed to add `mul!` to using list from LinearAlgebra.

KnutAM, line 724: And just combined fix for convenience. 
```suggestion
function add_inhomogeneities!(f::AbstractVector, K::AbstractMatrix, ch::ConstraintHandler)
    return mul!(f, K, sparsevec(ch.prescribed_dofs, ch.inhomogeneities, size(K, 2)), -1, 1)
```

KnutAM (commented):
Some details, but otherwise LGTM.

termi-official:
Thanks for the detailed review Knut!

KnutAM (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Next] - xxxx-xx-xx
+
+### Added
+- Support for directly assembling to SparseMatrixCSR (from SparseMatricesCSR.jl)
```
KnutAM, line 11:
```suggestion
- Support for directly assembling to `SparseMatrixCSR` (from `SparseMatricesCSR.jl`). ([#864])
```

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
...
+
+### Documentation updates
+- Extended assembly docs with information on how to support direct assembly into new matrix types.
```
KnutAM, line 14:
```suggestion
- Extended assembly docs with information on how to support direct assembly into new matrix types. ([#864])
```

KnutAM (commented):
Just some formatting (and we should run CI as it includes the formatting check) and the changelog is included in the docs.

PR #865, opened 2024-01-05
Add spellchecking action

KnutAM:


.github/workflows/SpellCheck.yml
```diff
@@ -0,0 +1,13 @@
...
+
+on: [pull_request]
+
+jobs:
+  typos-check:
+    name: Spell Check with Typos
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Actions Repository
+        uses: actions/checkout@v3
```
termi-official:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/SpellCheck.yml
```diff
@@ -0,0 +1,13 @@
...
+      - name: Check spelling
+        uses: crate-ci/typos@v1.16.3
```
termi-official:
```suggestion
        uses: crate-ci/typos@v1.44.0
```

termi-official (changes requested):

.github/workflows/SpellCheck.yml
henrij22: Hi, just dropping in, Why not just directly @v1, otherwise dependabot will have weekly PRs to update the version here :D 
https://github.com/crate-ci/typos/tags the tag seems to be updated regularly with each new version

termi-official: You have a good point here.

KnutAM: > Why not just directly @v1

Because then we get unrelated failures in PRs, when a spell-check is added to parts that were not touched by that PR.

.typos.toml
```diff
@@ -0,0 +1,9 @@
+[files]
+extend-exclude = ["*.svg", "*.toml"]
+
+[default.extend-words]
+ue = "ue"
+ba = "ba"
+nd = "nd"
+Numer = "Numer"
+Secrest = "Secrest"
```
termi-official:
```suggestion
Secrest = "Secrest"

```

.typos.toml
```diff
@@ -0,0 +1,21 @@
+[files]
+extend-exclude = ["*.svg", "*.toml"]
+
+[default.extend-words]
+ue = "ue"
+ba = "ba"
```
termi-official, line 6: What about these? I think "ue" should always be marked as code (i.e. `ue`). Not sure where "ba" appears.

.typos.toml
```diff
@@ -0,0 +1,21 @@
...
+nd = "nd"
+Numer = "Numer"
```
termi-official: Following what you have done below, this should go into the qrtables, right?

KnutAM: `ue` has 132 results in 14 files, so must be general. Not sure what you mean by mark as code? (`nd` (14, 4 files), `ba` only in two files, but still, valid variable names IMO).

PR #867, opened 2024-01-08
Export of discontinuous fields to VTK

KnutAM:
Export discontinuous fields to VTK - detects if a field is discontinuous and does it automatically.

TODO
- [x] Add tests using `apply_analytical!` to validate continuity when it should be
- [x] Document how to use this: #1161


KnutAM:
# Test case 
Generated to validate the discontineous, tests for different reference shapes will be needed before merge...

image

# DG tutorial (dg penalization variation)
The results looks quite strange, any idea if this is reasonable @AbdAlazezAhmed / @termi-official ? 

$\mu = 4.0 / h_e = (1 + O)^D / h_e$ (same result for $\mu = 2.0 / h_e $

image


$\mu = 1.0 / h_e $
image


$\mu = 0.5 / h_e $
image

termi-official:
> # Test case
> 
> Generated to validate the discontinuous, tests for different reference shapes will be needed before merge...
> 

Any reason not to use `apply_analytical!` for the first tests?

> # DG tutorial (dg penalization variation)
> 
> The results looks quite strange, any idea if this is reasonable @AbdAlazezAhmed / @termi-official ? (I'm wondering if something is off in the export, but test case looks correct)
> 

Looks correct for me. The parameter which you tune is basically a penalty on the jump in the gradients between the elements. Please notice in the first figure in the section above that your gradients are essentially continuous with large jumps between elements. Just look at the formula for 1D and I think you will understand what happens. :) Also, a too small value for the penalty parameter makes the forumlation unstable, which is why you see this weird checkerboarding pattern.

KnutAM:
> Any reason not to use apply_analytical! for the first tests?

Since it uses the nodal coordinate, the result wouldn't be discontinuous. But it is a good point adding that to validate the the result is continuous when it should!

src/Export/VTK.jl
```diff
@@ -27,13 +27,21 @@ end
 """
 struct VTKGridFile{VTK <: WriteVTK.DatasetFile}
     vtk::VTK
+    cellnodes::Vector{UnitRange{Int}}
```
termi-official, line 30: Maybe it is easier to maintain if we use a `Union{Vector{UnitRange{Int}},Nothing}` here to  distinguish between cases when `cellnodes` is accidentally (or intentionally) 0. One example here would be for MPI simulations we can run into the weird scenario that some processes have empty local grids. This might not be the only scenario tho.

termi-official (approved):
LGTM. CI failure is unrelated to this PR.

Issue #868, opened 2024-01-08
List Related Packages


cortner:
I wasted quite a bit of time trying to figure out how to do simple visualizations of Ferrite solutions. Eventually somebody on discourse pointed me to FerriteViz which is great. But neither the README nor the docs mention it. Maybe somebody who knows of all important packages related to Ferrite could list them somewhere?

KnutAM:
That's a good idea! Something like this: #869 ?

Also, whenever you have questions, don't hesitate to reach out on the Slack channel #ferrite-fem or on Discourse as you did!

KnutAM:
As discussed in #869, perhaps we could add an item under the code gallery where anyone who builds a package on top of Ferrite can make a PR to add a link to, and a short description of, their package.

termi-official:
> As discussed in #869, perhaps we could add an item under the code gallery where anyone who builds a package on top of Ferrite can make a PR to add a link to, and a short description of, their package.

Related to this: Should this only contain public packages or should we also showcase currently internally used packages which come with some publications (e.g. Maxi has one for convexification)? What about publication codes as in e.g. https://github.com/kimauth/Plastoxid.jl ?

I would also list work in progress packages, which have the actual goal to be developed long-term (as in e.g. FerriteAssembly.jl?).

DanielVandH:
It would probably be also useful to mention FerriteViz in the first tutorial, https://ferrite-fem.github.io/Ferrite.jl/dev/tutorials/heat_equation/. I initially thought the main way to visualise the solutions was with the vtk files based on the documentation there.

termi-official:
> It would probably be also useful to mention FerriteViz in the first tutorial, https://ferrite-fem.github.io/Ferrite.jl/dev/tutorials/heat_equation/.

cc @KnutAM what do you think?

> I initially thought the main way to visualise the solutions was with the vtk files based on the documentation there. (Precompilation of FerriteViz seems to be broken with Ferrite on master, but once it's fixed that would be good)

I have a PR which prepares next release here https://github.com/Ferrite-FEM/FerriteViz.jl/pull/103 . Master is just compatible with the last Ferrite release. Note that it does not have as many features as most VTK viewers.

KnutAM:
I'd propose a reference to the post-processing how to. With my proposed intro text here (https://github.com/Ferrite-FEM/Ferrite.jl/pull/875#discussion_r1651584377) I think this will be covered?

PR #869, opened 2024-01-08
Add list of registered packages in the Ferrite ecosystem

KnutAM:
Solves #868 

I limited to registered packages in this list that are made specifically (except Tensors) for Ferrite. I think that unregistered packages perhaps shouldn't be advertised as official (even though mentioning e.g. IGA.jl would be nice, what do you think @lijas?)
For related packages such as linear solvers and DifferentialEquations.jl, I suppose the tutorials take care of that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,9 @@ Please keep in mind that we are part of the Julia community and adhere to the
 
 ## Related packages
 The following registered packages are part of the `Ferrite.jl` ecosystem in addition to Ferrite itself:
-* [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl): [Makie.jl](https://docs.makie.org/stable/)-based visualization of Ferrite data
-* [FerriteGmsh.jl](https://github.com/Ferrite-FEM/FerriteGmsh.jl): Create, interact, and import [Gmsh](https://gmsh.info/) meshes into Ferrite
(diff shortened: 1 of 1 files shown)


cortner:
Thanks for putting this together. Personally I'd make the list as extensive as possible. You can always document which packages are recommended and which are experimental and caution is advised?

lijas:
With the current state of IGA.jl I dont think should be advertised. But maybe in the future when it becomes a bit more stable and I have registered it, we can add it to the list.

KnutAM:
> Personally I'd make the list as extensive as possible. You can always document which packages are recommended and which are experimental and caution is advised?

I agree that a more complete list would be nice. But since such a list would have to be dynamic as the status of packages changes it might quickly get outdated. Another argument is avoiding to characterize the quality/stability of other people's packages. 

On the current dev version, there is a [code gallery](https://ferrite-fem.github.io/Ferrite.jl/dev/gallery/) in the documentation. IMO this would be a nice place to showcase what has been done with Ferrite, including examples of packages built on top of Ferrite (e.g. stemming from a PhD project).

One option could be to add an item under the code gallery where anyone who builds a package on top of Ferrite can make a PR to add a link to, and a short description of, their package. I suggest leaving your original issue open to discuss that option, and then we can soon (once someone approves this PR) merge this one.

Issue #610, opened 2023-03-01
L2Projector for mixed grids


lijas:
It would be nice to be able to use the L2Projector if you have a domain with mixed celltypes, and you want to L2Project over the entire domain. Currently it is only possible to l2Project over one celltype at a time.

Possible api:

```
#init alt1:
projector = L2Projector(grid, set=mixedset) #Creates mass-matrix internaly

#init alt2:
projector = L2Projector(grid)
load_massmatrix!(projector, ip_hex, cellset1)
load_massmatrix!(projector, ip_tet, cellset2)

#...

# Create rhs:
load_forcevector!(projector, σqp1, qr_hex, cellset1)
load_forcevector!(projector, σqp2, qr_tet, cellset2)
σ_nodes = project(projector; project_to_nodes=true); 
```

termi-official:
I think it makes sense to merge the dof handlers first and then make one feature at a time consistent with mixed grids. What do you think?

KnutAM:
Are breaking changes required for this, or can this be done post 1.0?

PR #870, opened 2024-01-09
Fix bug in L2Projector

lijas:
I had a bug in the L2Projector when projecting the results to the nodes for a subset of cells in a mixed grid. Specifically the error is caused when calling `Ferrite.evaluate_at_grid_nodes(proj, point_vars)`

The error comes from that the L2Projector creates a DofHandler where not all cells belong to a subdofhandler (should this be allowed?). This in turns causes an error when ndofs_per_cell(dh, [cellid=1]) is called internally in L2Projector, where cell with id 1 does not belong to a subdofhandler.

There are probably multiple ways to fix this. I did not want to "leak" this issue to other parts of the code, so this PR creates a small hack in the L2Projector.

In the future we should probably improve the interface for the L2Projector anyways (#610 ), so maybe this is an ok fix for now?

```
using Ferrite

dim = 2
nodes = Node{dim, Float64}[]
push!(nodes, Node((0.0, 0.0)))
push!(nodes, Node((1.0, 0.0)))
push!(nodes, Node((2.0, 0.0)))
push!(nodes, Node((0.0, 1.0)))
push!(nodes, Node((1.0, 1.0)))
push!(nodes, Node((2.0, 1.0)))

cells = Ferrite.AbstractCell[]
push!(cells, Quadrilateral((1,2,5,4)))
push!(cells, Triangle((2,3,6)))
push!(cells, Triangle((2,6,5)))

quadset = 1:1
triaset = 2:3
mesh = Grid(cells, nodes)

# Triangle set
order = 2
ip = Lagrange{RefTriangle, order}()
ip_geom = Lagrange{RefTriangle, 1}()
qr = QuadratureRule{RefTriangle}(4)
nqp = getnquadpoints(qr)

qp_values_tria = [zeros(SymmetricTensor{2,2}, nqp) for _ in triaset]
#tria
proj = L2Projector(ip, mesh; geom_ip=ip_geom, set=triaset)
point_vars = project(proj, qp_values_tria, qr)

projection_at_nodes = evaluate_at_grid_nodes(proj, point_vars) #ERROR
```

Error:
```
ERROR: OutOfMemoryError()
Stacktrace:
  [1] Array
    @ ./boot.jl:477 [inlined]
  [2] Array
    @ ./boot.jl:486 [inlined]
  [3] zeros
    @ ./array.jl:584 [inlined]
  [4] zeros
    @ ./array.jl:580 [inlined]
  [5] CellCache(dh::DofHandler{2, Grid{2, Ferrite.AbstractCell, Float64}}, flags::UpdateFlags)
    @ Ferrite ~/Documents/juliapkg/Ferrite.jl/src/iterators.jl:65
  [6] CellIterator(gridordh::DofHandler{2, Grid{2, Ferrite.AbstractCell, Float64}}, set::Vector{Int64}, flags::UpdateFlags)
    @ Ferrite ~/Documents/juliapkg/Ferrite.jl/src/iterators.jl:274
  [7] CellIterator
    @ ~/Documents/juliapkg/Ferrite.jl/src/iterators.jl:265 [inlined]
  [8] _evaluate_at_grid_nodes!(data::Vector{SymmetricTensor{2, 2, Float64, 3}}, cv::CellValues{Ferrite.FunctionValues{1, Lagrange{RefTriangle, 2, Nothing}, Matrix{Float64}, Matrix{Vec{2, Float64}}, Matrix{Vec{2, Float64}}}, Ferrite.GeometryMapping{1, Lagrange{RefTriangle, 1, Nothing}, Matrix{Float64}, Matrix{Vec{2, Float64}}, Nothing}, QuadratureRule{RefTriangle, Float64, 2}, Vector{Float64}}, dh::DofHandler{2, Grid{2, Ferrite.AbstractCell, Float64}}, set::Vector{Int64}, u::Vector{SymmetricTensor{2, 2, Float64, 3}})
    @ Ferrite ~/Documents/juliapkg/Ferrite.jl/src/L2_projection.jl:273
  [9] _evaluate_at_grid_nodes(proj::L2Projector, vals::Vector{SymmetricTensor{2, 2, Float64, 3}}, #unused#::Val{false})
    @ Ferrite ~/Documents/juliapkg/Ferrite.jl/src/L2_projection.jl:267
 [10] evaluate_at_grid_nodes(proj::L2Projector, vals::Vector{SymmetricTensor{2, 2, Float64, 3}})
    @ Ferrite ~/Documents/juliapkg/Ferrite.jl/src/L2_projection.jl:240
 [11] top-level scope
    @ ~/Documents/juliapkg/Ferrite.jl/test/test_l2_projection.jl:184
```

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1019,8 +1019,9 @@ function _add!(ch::ConstraintHandler, pdbc::PeriodicDirichlet, interpolation::In
     # Dof map for mirror dof => image dof
     dof_map = Dict{dof_map_t,dof_map_t}()
 
-    mirror_dofs = zeros(Int, ndofs_per_cell(ch.dh))
-     image_dofs = zeros(Int, ndofs_per_cell(ch.dh))
+    n = ndofs_per_cell(ch.dh, first(face_map).mirror[1])
+    mirror_dofs = zeros(Int, n)
+    image_dofs  = zeros(Int, n)
     for face_pair in face_map
         m = face_pair.mirror
         i = face_pair.image
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -175,12 +175,14 @@ function ndofs_per_cell(dh::DofHandler)
     @assert length(dh.subdofhandlers) != 0
     return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
 end
-function ndofs_per_cell(dh::DofHandler, cell::Int)
+@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
     @boundscheck 1 <= cell <= getncells(get_grid(dh)) || throw(ArgumentError("cell = $cell ∉ 1:$(getncells(get_grid(dh)))"))
-    return @inbounds ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
+    sdh_idx = dh.cell_to_subdofhandler[cell]
+    @boundscheck 1 <= sdh_idx <= length(dh.subdofhandlers) || throw(ArgumentError("cell = $cell does no belong to any SubDofHandler"))
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[sdh_idx])
 end
-ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
-ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
+@inline ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
+@inline ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
 
 """
     celldofs!(global_dofs::Vector{Int}, dh::AbstractDofHandler, i::Int)
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -176,7 +176,7 @@ function ndofs_per_cell(dh::DofHandler)
     return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
 end
 function ndofs_per_cell(dh::DofHandler, cell::Int)
-    @boundscheck 1 <= cell <= getncells(get_grid(dh))
+    @boundscheck 1 <= cell <= getncells(get_grid(dh)) || throw(ArgumentError("cell = $cell ∉ 1:$(getncells(get_grid(dh)))"))
     return @inbounds ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
 end
 ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -175,14 +175,11 @@ function ndofs_per_cell(dh::DofHandler)
     @assert length(dh.subdofhandlers) != 0
     return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
 end
-@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
-    @boundscheck 1 <= cell <= getncells(get_grid(dh)) || throw(ArgumentError("cell = $cell ∉ 1:$(getncells(get_grid(dh)))"))
-    sdh_idx = dh.cell_to_subdofhandler[cell]
-    @boundscheck 1 <= sdh_idx <= length(dh.subdofhandlers) || throw(ArgumentError("cell = $cell does no belong to any SubDofHandler"))
-    return @inbounds ndofs_per_cell(dh.subdofhandlers[sdh_idx])
+function ndofs_per_cell(dh::DofHandler, cell::Int)
+    return ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
 end
-@inline ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
-@inline ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
+ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
+ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
 
 """
     celldofs!(global_dofs::Vector{Int}, dh::AbstractDofHandler, i::Int)
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -168,7 +168,14 @@ Return the number of degrees of freedom for the cell with index `cell`.
 
 See also [`ndofs`](@ref).
 """
-function ndofs_per_cell(dh::DofHandler, cell::Int=1)
+function ndofs_per_cell(dh::DofHandler)
+    if length(dh.subdofhandlers) > 1
+        error("There are more than one subdofhandler. Use `ndofs_per_cell(dh, cellid::Int)` instead.")
+    end
+    @assert length(dh.subdofhandlers) != 0
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
+end
+function ndofs_per_cell(dh::DofHandler, cell::Int)
     @boundscheck 1 <= cell <= getncells(get_grid(dh))
     return @inbounds ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
 end
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -388,7 +388,7 @@ to a Node.
 "Returns the number of nodes in the grid."
 @inline getnnodes(grid::AbstractGrid) = length(grid.nodes)
 "Returns the number of nodes of the `i`-th cell."
-@inline function nnodes_per_cell(grid::AbstractGrid) 
+function nnodes_per_cell(grid::AbstractGrid) 
     if !isconcretetype(getcelltype(grid))
         error("There are different celltypes in the `grid`. Use `nnodes_per_cell(grid, cellid::Int)` instead")
     end
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -388,7 +388,13 @@ to a Node.
 "Returns the number of nodes in the grid."
 @inline getnnodes(grid::AbstractGrid) = length(grid.nodes)
 "Returns the number of nodes of the `i`-th cell."
-@inline nnodes_per_cell(grid::AbstractGrid, i::Int=1) = nnodes(grid.cells[i])
+@inline function nnodes_per_cell(grid::AbstractGrid) 
+    if !isconcretetype(getcelltype(grid))
+        error("There are different celltypes in the `grid`. Use `nnodes_per_cell(grid, cellid::Int)` instead")
+    end
+    return nnodes(first(grid.cells))
+end
+@inline nnodes_per_cell(grid::AbstractGrid, i::Int) = nnodes(grid.cells[i])
 
 "Return the number type of the nodal coordinates."
 @inline get_coordinate_eltype(grid::AbstractGrid) = get_coordinate_eltype(first(getnodes(grid)))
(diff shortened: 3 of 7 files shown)


KnutAM:
I find the default `cellid=1` in `ndofs_per_cell(dh, cellid=1)` quite suspicious, and considered to propose removing before. Would be quite nice to have `ndofs_per_cell(dh)` error unless the grid type is concrete IMO.

lijas:
@kimauth convinced me to not go with the hacky solution :laughing: 

The new fix returns an errors if `ndofs_per_cell(dh)` is called when there are multiple subdofhandlers. This is techincally breaking, but in practice I dont think it will affect many users?

termi-official:
I think this is a valid breaking change, because it gives a clearer error to users if something fails (in contrast to the current interface).

KnutAM:
> The new fix returns an errors if ndofs_per_cell(dh) is called when there are multiple subdofhandlers. This is techincally breaking, but in practice I dont think it will affect many users?

Shouldn't be breaking since subdofhandlers aren't available on 0.3.14, and I guess `ndofs_per_cell(dh)` only worked for `dh::DofHandler`.

lijas:
Ugh, ran in to more problems because we are calling `cc = CellCache(ch.dh, UpdateFlags(; nodes=false, coords=false, dofs=true))` for mixed grid at some places in the code.

Please see the fix in the CellCache constructor and tell me if you think it is ok.

I also made same update for `nnodes_per_cell(dh)`

src/iterators.jl
```diff
@@ -51,17 +51,26 @@ struct CellCache{X,G<:AbstractGrid,DH<:Union{AbstractDofHandler,Nothing}}
...
 function CellCache(dh::DofHandler{dim}, flags::UpdateFlags=UpdateFlags()) where {dim}
-    N = nnodes_per_cell(get_grid(dh))
+    N = 0
+    if isconcretetype(getcelltype(get_grid(dh)))
+        N = nnodes_per_cell(get_grid(dh))
+    end
+    n = 0
+    if length(dh.subdofhandlers) == 1
+        n = ndofs_per_cell(dh)
+    end
```
KnutAM, line 71: You can also just set `N` and `n` equal to zero always, the vectors will be resized in the first reinit of the cache.

KnutAM, line 71: Or actually, since ndofs_per_cell is stored in the subdofhandler
```suggestion
    n = ndofs_per_cell(dh.subdofhandlers[1])
```

src/iterators.jl
```diff
@@ -51,17 +51,26 @@ struct CellCache{X,G<:AbstractGrid,DH<:Union{AbstractDofHandler,Nothing}}
...
     coords = zeros(Vec{dim,T}, N)
     return CellCache(flags, grid, ScalarWrapper(-1), nodes, coords, nothing, Int[])
 end
 
 function CellCache(dh::DofHandler{dim}, flags::UpdateFlags=UpdateFlags()) where {dim}
-    N = nnodes_per_cell(get_grid(dh))
+    N = 0
+    if isconcretetype(getcelltype(get_grid(dh)))
+        N = nnodes_per_cell(get_grid(dh))
+    end
```
KnutAM, line 67: And this will be resized anyway, but this way there would be no performance/allocation change for users of concrete grids. 
```suggestion
    N = nnodes_per_cell(get_grid(dh), 1)
```

lijas:
New error popping up :laughing: 

ndofs_per_cell(dh) was used when creating periodic constraints, which caused an error in my code with the new method for ndofs_per_cell(dh). I went through all the code to make sure ndofs_per_cell(dh) is not used anywhere else

src/Dofs/ConstraintHandler.jl
```diff
@@ -1019,8 +1019,9 @@ function _add!(ch::ConstraintHandler, pdbc::PeriodicDirichlet, interpolation::In
     # Dof map for mirror dof => image dof
     dof_map = Dict{dof_map_t,dof_map_t}()
 
-    mirror_dofs = zeros(Int, ndofs_per_cell(ch.dh))
-     image_dofs = zeros(Int, ndofs_per_cell(ch.dh))
+    n = ndofs_per_cell(ch.dh, first(face_map).mirror[1])
+    mirror_dofs = zeros(Int, n)
+    image_dofs  = zeros(Int, n)
```
termi-official, line 1024: I think this is not correct. Think e.g. about one side of the domain has a quad and the other side of the domain has a triangle.

lijas: The `collect_periodic_faces` only works when the mirror and image faces are the same celltype anyways, so I think this is fine.

KnutAM: I think this is fine too, but perhaps it is clearer (showing that we assume the same interpolation) if using 
```suggestion
    n = getnbasefunctions(interpolation)
    mirror_dofs = zeros(Int, n)
    image_dofs  = zeros(Int, n)
```
?

lijas:
Can we merge this?

src/Dofs/DofHandler.jl
```diff
@@ -168,8 +168,15 @@ Return the number of degrees of freedom for the cell with index `cell`.
...
-function ndofs_per_cell(dh::DofHandler, cell::Int=1)
-    @boundscheck 1 <= cell <= getncells(get_grid(dh))
+function ndofs_per_cell(dh::DofHandler)
+    if length(dh.subdofhandlers) > 1
+        error("There are more than one subdofhandler. Use `ndofs_per_cell(dh, cellid::Int)` instead.")
+    end
+    @assert length(dh.subdofhandlers) != 0
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
+end
+function ndofs_per_cell(dh::DofHandler, cell::Int)
```
KnutAM, line 178: AFAIK the current implementation would never elude the boundscheck, so I would suggest
```suggestion
@propagate_inbounds function ndofs_per_cell(dh::DofHandler, cell::Int)
```
or 
```suggestion
@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
```
or 
remove `@boundscheck` below. 

While I suppose that it is ok to rely on `dh.cell_to_subdofhandler[cell]` returning a valid index for `dh.subdofhandlers`, it is checked above for the other dispatch. 
In that respect, perhaps it would be better to just skip the boundscheck line, and remove the `@inbounds`?
Have you tried benchmarking if it makes a significant impact?

KnutAM (changes requested):
Looks good overall, but I think the `@inbounds` stuff needs some consideration.

src/Dofs/DofHandler.jl
lijas: One problem is that `dh.cell_to_subdofhandler[cell]` might return -1 when not all cells belong to a subdofhandler. This is extremely uncommon, but it was what caused the issue in L2Projector in the first case.

I dont really understand the reasoning in you comment (I dont know all the details of `@boundscheck` `@inbounds` etc)...What do you think about the latest commit?

Or is better to ignore bounds checking all together?
```
@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
    return ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
end
```

KnutAM: > ```julia
> @inline function ndofs_per_cell(dh::DofHandler, cell::Int)
>     return ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
> end
> ```

Yes, unless there is some major performance loss, I prefer this version (and can be optimized later with inbounds if required). With the latest commit, we anyways don't use inbounds on `dh.cell_to_subdofhandler[cell]`. 

(My reasoning is based on that inlining is required for `@boundscheck` to work as desired, at least if I understand the [docs](https://docs.julialang.org/en/v1/base/base/#Base.@boundscheck) correctly.)

src/Dofs/DofHandler.jl
```diff
@@ -168,12 +168,21 @@ Return the number of degrees of freedom for the cell with index `cell`.
...
+        error("There are more than one subdofhandler. Use `ndofs_per_cell(dh, cellid::Int)` instead.")
+    end
+    @assert length(dh.subdofhandlers) != 0
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
+end
+@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
+    @boundscheck 1 <= cell <= getncells(get_grid(dh)) || throw(ArgumentError("cell = $cell ∉ 1:$(getncells(get_grid(dh)))"))
+    sdh_idx = dh.cell_to_subdofhandler[cell]
+    @boundscheck 1 <= sdh_idx <= length(dh.subdofhandlers) || throw(ArgumentError("cell = $cell does no belong to any SubDofHandler"))
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[sdh_idx])
```
KnutAM, line 182:
```suggestion
function ndofs_per_cell(dh::DofHandler, cell::Int)
    return ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
```

src/Dofs/DofHandler.jl
```diff
@@ -168,12 +168,21 @@ Return the number of degrees of freedom for the cell with index `cell`.
...
 """
-function ndofs_per_cell(dh::DofHandler, cell::Int=1)
-    @boundscheck 1 <= cell <= getncells(get_grid(dh))
-    return @inbounds ndofs_per_cell(dh.subdofhandlers[dh.cell_to_subdofhandler[cell]])
+function ndofs_per_cell(dh::DofHandler)
+    if length(dh.subdofhandlers) > 1
+        error("There are more than one subdofhandler. Use `ndofs_per_cell(dh, cellid::Int)` instead.")
+    end
+    @assert length(dh.subdofhandlers) != 0
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[1])
```
KnutAM, line 176:
```suggestion
    return ndofs_per_cell(dh.subdofhandlers[1])
```

test/test_mixeddofhandler.jl
```diff
@@ -133,6 +133,7 @@ function test_2d_mixed_2_el()
     @test ndofs_per_cell(dh.subdofhandlers[1]) == 12
     @test ndofs_per_cell(dh, 2) == 9
     @test ndofs_per_cell(dh.subdofhandlers[2]) == 9
+    @test_throws ErrorException ndofs_per_cell(dh)
```
KnutAM, line 136:
```suggestion
    @test_throws ErrorException ndofs_per_cell(dh)
    @test_throws ErrorException Ferrite.nnodes_per_cell(grid)
```

src/Dofs/DofHandler.jl
```diff
@@ -168,12 +168,21 @@ Return the number of degrees of freedom for the cell with index `cell`.
...
+@inline function ndofs_per_cell(dh::DofHandler, cell::Int)
+    @boundscheck 1 <= cell <= getncells(get_grid(dh)) || throw(ArgumentError("cell = $cell ∉ 1:$(getncells(get_grid(dh)))"))
+    sdh_idx = dh.cell_to_subdofhandler[cell]
+    @boundscheck 1 <= sdh_idx <= length(dh.subdofhandlers) || throw(ArgumentError("cell = $cell does no belong to any SubDofHandler"))
+    return @inbounds ndofs_per_cell(dh.subdofhandlers[sdh_idx])
 end
-ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
-ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
+@inline ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
+@inline ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
```
KnutAM, line 185:
```suggestion
ndofs_per_cell(sdh::SubDofHandler) = sdh.ndofs_per_cell[]
ndofs_per_cell(sdh::SubDofHandler, ::Int) = sdh.ndofs_per_cell[] # for compatibility with DofHandler
```

src/Grid/grid.jl
```diff
@@ -388,7 +388,13 @@ to a Node.
 "Returns the number of nodes in the grid."
 @inline getnnodes(grid::AbstractGrid) = length(grid.nodes)
 "Returns the number of nodes of the `i`-th cell."
-@inline nnodes_per_cell(grid::AbstractGrid, i::Int=1) = nnodes(grid.cells[i])
+@inline function nnodes_per_cell(grid::AbstractGrid)
```
KnutAM, line 391:
```suggestion
function nnodes_per_cell(grid::AbstractGrid) 
```

lijas:
Thanks @KnutAM for the suggestions. I have to read up on all the details about boundschecking and `@inline` etc.

KnutAM (approved):

PR #872, opened 2024-01-11
CellMultiValues (new attempt)

KnutAM:
Supersedes #680 and #674, using the components from #764

- [x] Check performance
- [x] Update incompressible elasticity example
- [x] Add tests
- [x] Complete documentation

A couple of notes on the design
- Given a `cv::CellMultiValues`, `cv.u::FunctionValues`, which can be used just like a `CellValues` for all function-related queries, but not geometrical such as `getdetJdV`, these are called on `cv` itself. 
- The constructor accepts a named tuple of function interpolations. If two interpolations are identical, they will share the same `FunctionValues` object, saving both memory (bandwidth) and computations in reinit!. 
- `FunctionValues` subtypes `AbstractValues` to make all `function_value` and similar work. This implies that e.g. `spatial_coordinate(cv.u::FunctionValues, args...)` will give a method error on `getngeobasefunctions` instead of `spatial_coordinate`. We could add specializations if easier error messages are required. 

~It would be very convenient to also store the dof_range of each field, but having this as an input would be error-prone if the user provides the wrong order compared to what is stored in the `SubDofHandler`. If a convenience constructor similar to #806 would be used, this would be safe, but that should only be a high-level constructor interface. Ideas are welcome!~ (Edit: not generally applicable, so leave out for now)


src/FEValues/CellValues.jl
```diff
@@ -127,4 +127,110 @@ function Base.show(io::IO, d::MIME"text/plain", cv::CellValues)
     print(io, "\n Function interpolation: "); show(io, d, ip_fun)
     print(io, "\nGeometric interpolation: "); 
     sdim === nothing ? show(io, d, ip_geo) : show(io, d, ip_geo^sdim)
-end
\ No newline at end of file
+end
+
+"""
+    MultiCellValues([::Type{T},] quad_rule::QuadratureRule, func_interpols::NamedTuple, [geom_interpol::Interpolation])
```
termi-official, line 133: Maybe `CellMultiValues` for the name, because we have multiple values (in contrast to multiple cells)?

KnutAM:
I've renamed to `CellMultiValues` following @termi-official's suggestion. I find this name harder when coding, but that could be me having become accustomed to `MultiCellValues`. 
Another option could be `CellValuesGroup`

Otherwise, ready for review from my side if anyone wants to take a look!

termi-official (commented):
I left some comments.

Since I am struggling recently with the question: Do you think this should be still Ferrite core or in some package on top?

docs/src/literate-tutorials/incompressible_elasticity.jl
```diff
@@ -116,37 +117,29 @@ end;
...
-
+function assemble_up!(Ke, fe, cell, cellvalues, facevalues_u, grid, mp, ɛdev, t, dofrange_u, dofrange_p)
+    reinit!(cellvalues, cell)
     ## We only assemble lower half triangle of the stiffness matrix and then symmetrize it.
-    for q_point in 1:getnquadpoints(cellvalues_u)
-        for i in 1:n_basefuncs_u
-            ɛdev[i] = dev(symmetric(shape_gradient(cellvalues_u, q_point, i)))
+    for q_point in 1:getnquadpoints(cellvalues)
+        for i in keys(dofrange_u)
+            ɛdev[i] = dev(symmetric(shape_gradient(cellvalues[:u], q_point, i)))
```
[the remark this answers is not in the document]
KnutAM: Yes, for my test earlier this was no issues performance-wise. 
But I'll add a test to check for type stability!

docs/src/literate-tutorials/incompressible_elasticity.jl
```diff
@@ -116,37 +117,29 @@ end;
...
         end
-        dΩ = getdetJdV(cellvalues_u, q_point)
-        for i in 1:n_basefuncs_u
-            divδu = shape_divergence(cellvalues_u, q_point, i)
-            δu = shape_value(cellvalues_u, q_point, i)
-            for j in 1:i
-                Ke[BlockIndex((u▄, u▄), (i, j))] += 2 * mp.G * ɛdev[i] ⊡ ɛdev[j] * dΩ
+        dΩ = getdetJdV(cellvalues, q_point)
+        for (iᵤ, Iᵤ) in pairs(dofrange_u)
+            for (jᵤ, Jᵤ) in pairs(dofrange_u[1:iᵤ])
```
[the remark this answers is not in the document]
KnutAM: In my benchmarks, using block indices for the local matrices are always slow. 
Furthermore, it means that if you change the order the fields are added to your dofhandler, you'll silently get wrong results as opposed to using the dofrange directly.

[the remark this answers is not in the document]
termi-official: Note that the new code will still give you the wrong result in this case.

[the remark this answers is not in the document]
KnutAM: Did I miss an assumption of the ordering of the fields?

src/FEValues/CellValues.jl
```diff
@@ -128,4 +128,157 @@ function Base.show(io::IO, d::MIME"text/plain", cv::CellValues)
...
-end
\ No newline at end of file
+end
+
+"""
+    CellMultiValues([::Type{T},] quad_rule::QuadratureRule, func_interpols::NamedTuple, [geom_interpol::Interpolation])
+
+A `cmv::CellMultiValues` object generalizes the `CellValues` object to multiple fields. 
+In general, functions applicable to a `CellValues` associated with the function interpolation with `key` 
+can be called on `cmv[key]`, as `cmv[key] isa FunctionValues`.
```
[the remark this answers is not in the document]
KnutAM: Wouldn't that be for `typeof(cmv[key]) <: FunctionValues`?
In that case, I think it should be `cmv[key]::FunctionValues`, do you prefer that form?

docs/src/literate-tutorials/incompressible_elasticity.jl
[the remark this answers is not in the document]
KnutAM: Note from Slack communication: Yes, we assume u then p ordering when symmetrizing. 
This could potentially be moved to a "performance optimization" how-to to keep this first multifield tutorial as simple as possible.

docs/src/literate-tutorials/incompressible_elasticity.jl
termi-official, line 1: Should we also upgrade the other tutorials (Stokes, Navier-Stokes, Porous Media)?

src/FEValues/CellValues.jl
```diff
@@ -106,38 +106,207 @@ end
...
+end
+
+"""
+    CellMultiValues([::Type{T},] quad_rule::QuadratureRule, func_interpols::NamedTuple, [geom_interpol::Interpolation])
+
+A `cmv::CellMultiValues` is similar to a `CellValues` object, but includes values associated with multiple
+interpolations while sharing the same quadrature points and geometrical interpolation.
+
+In general, functions applicable to a `CellValues` associated with the function interpolation
+in `func_interpols` with `key::Symbol` can be called on `cmv[key]`, as `cmv[key] isa FunctionValues`.
```
termi-official: Prospectively for GPU support, can we ensure that `key` can be a simple index instead of a symbol? Or is named tuple already functional on GPUs? I would like to avoid that users must have multiple variants of their code.

src/FEValues/CellValues.jl
```diff
@@ -106,38 +106,207 @@ end
...
+and 1st order interpolations for the pressure, `p`, and temperature, `T`.
+```
+qr = QuadratureRule{RefQuadrilateral}(2)
+ip_geo = Lagrange{RefQuadrilateral,1}() # Optional
+ipu = Lagrange{RefQuadrilateral,2}()^2
+ipp = Lagrange{RefQuadrilateral,1}()
+ipT = Lagrange{RefQuadrilaterla,1}()
+cmv = CellMultiValues(qr, (u = ipu, p = ipp, T = ipT), ip_geo)
+```
+After reinitialization, the `cmv` can be used as, e.g.
```
termi-official: Maybe we can refer to the exact function here for new users. Something along these lines maybe?

```suggestion
After reinitialization via [`Ferrite.reinit!`](@ref), the `cmv` can be used as, e.g.
```

src/FEValues/FunctionValues.jl
```diff
@@ -43,7 +43,7 @@ for both the reference cell (precalculated) and the real cell (updated in `reini
 """
 FunctionValues
 
-struct FunctionValues{DiffOrder, IP, N_t, dNdx_t, dNdξ_t, d2Ndx2_t, d2Ndξ2_t}
+struct FunctionValues{DiffOrder, IP, N_t, dNdx_t, dNdξ_t, d2Ndx2_t, d2Ndξ2_t} <: AbstractValues
```
termi-official, line 59: While we are at it, I think it could help to understand the type hierarchy by adding a docstring for `AbstractValues` and extending https://ferrite-fem.github.io/Ferrite.jl/stable/devdocs/FEValues/ with FunctionValues

termi-official (commented):
I like it in principle. I went over everything and left some comments to clarify.

Can you also dump the benchmark script into the benchmarks folder somewhere (and we really should try to revive the benchmarks folder at some point for performance regression testing).

src/FEValues/CellValues.jl
KnutAM: I don't know for GPU, but that sounds like a pretty bad restriction if one cannot index named tuples (you should never need to loop over them I think in the element routines, and during reinit we use the unique tuple)?

docs/src/literate-tutorials/incompressible_elasticity.jl
KnutAM: Good point! 

Perhaps update e.g. the PorousMedia as this will simplify that, but leave it in Stokes/Navier-Stokes and comment that here we highlight that you can also use different `CellValues` (and vice-versa in PM / IncompElast refer to these for examples of using separate `CellValues`)?

src/FEValues/CellValues.jl
termi-official: You can loop over tuples, but your access must be compile-time determinable here to generate the GPU code. I am not sure if NamedTuple has the guarantee.

KnutAM: That this is type stable (via const prop) is tested

KnutAM: Is this test enough to make sure we are GPU safe here? 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/bdc4f52da9496fa38c18edeacb5a8ece52c41a62/test/test_cellvalues.jl#L310-L313

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -395,28 +390,28 @@ jac_sparsity = sparse(K);
...
-    ch::ConstraintHandler
-    dh::DofHandler
-    cellvalues_v::CellValues
-    u::Vector
+struct RHSparams{KT, CH, DH, CV, UT}
+    K::KT  # SparseMatrixCSC
+    ch::CH # ConstraintHandler
+    dh::DH # DofHandler
+    cv::CV # CellMultiValues
+    u::UT  # Vector
```
KnutAM, line 398: Making this concrete got runtime on my M3 from about 26s to about 21s

docs/src/literate-tutorials/porous_media.jl
```diff
@@ -364,12 +360,16 @@ function solve(dh, ch, domains; Δt = 0.025, t_total = 1.0)
...
 end;
 
 # Finally we call the functions to actually run the code
 dh, ch, domains = setup_problem()
-solve(dh, ch, domains);
+a = solve(dh, ch, domains);
+
+## test the result                 #src
+using Test                         #src
+@test norm(a) ≈ 1471.7903131811188 #src
```
KnutAM, line 372: Have validated that I get the same value on current master

henrij22:
Hi @KnutAM 
Thank you for your continous work on Ferrite. For my understanding, is this about having for example two different interpolations for one cellvalues object? Would this facilitate then Petrov-Galerkin methods, where we would use different interpolations for test and trial functions?

src/FEValues/FunctionValues.jl
KnutAM: Done

src/FEValues/CellValues.jl
termi-official: I played a bit around with named tuples and could not break GPU compilation, so let us proceed with the assumption that this API is safe for now. :)

docs/src/literate-tutorials/ns_vs_diffeq.jl
termi-official: Oopsie. Yea, I indeed messed up here. :D Still interesting to see that that the performance gain is "just" so marginal.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -395,28 +390,28 @@ jac_sparsity = sparse(K);
...
 end
 
-function navierstokes_rhs_element!(dvₑ, vₑ, cellvalues_v)
-    n_basefuncs = getnbasefunctions(cellvalues_v)
-    for q_point in 1:getnquadpoints(cellvalues_v)
-        dΩ = getdetJdV(cellvalues_v, q_point)
-        ∇v = function_gradient(cellvalues_v, q_point, vₑ)
-        v = function_value(cellvalues_v, q_point, vₑ)
+function navierstokes_rhs_element!(dvₑ, vₑ, cv)
+    n_basefuncs = getnbasefunctions(cv[:v])
```
termi-official: Just to make sure I get this correctly. There is no performance overhead associated to using `cv[:v]` directly over calling it ones and passing it into the functions, because it is guaranteed to be const-prop'able?

termi-official:
>Would getproperty be nicer than getindex? e.g. cv.u instead of cv[:u]?

No.

docs/src/literate-tutorials/ns_vs_diffeq.jl
KnutAM: Exactly. 
Added a check of this in the benchmark script.

fredrikekre:
Just thought it might hinder accidental use of something like
```
function foo(cv, field::Symbol)
    cvf = cv[field] # Might not be constant propagated all the way here
    # ...
end
```

KnutAM:
> Just thought it might hinder accidental use of something like

Yes, that would be a performance footgun! 

I'm working on refactor and by overloading `Base.propertynames` autocomplete works (cc @termi-official). Furthermore, by introducing a few more getters, we this can actually give better code-reuse internally :)

benchmark/benchmark-cellvalues.jl
```diff
@@ -0,0 +1,65 @@
+using Ferrite, BenchmarkTools
+
+function element_routine(cv::CellValues)
+    s = 0.0
+    for q_point in 1:getnquadpoints(cv)
+        for i in 1:getnbasefunctions(cv)
+            for j in 1:getnbasefunctions(cv)
+                s += norm(shape_value(cv, q_point, i)) * norm(shape_gradient(cv, q_point, j))
```
termi-official: I think this and below is missing the detJdV term for proper coverage.

KnutAM: Added (although not really for the type stability of accessing since this is called directly on the `CellMultiValues`)

docs/src/literate-tutorials/incompressible_elasticity.jl
```diff
@@ -23,7 +23,6 @@
 #md # The full program, without comments, can be found in the next
 #md # [section](@ref incompressible_elasticity-plain-program).
 using Ferrite, Tensors
-using BlockArrays, SparseArrays, LinearAlgebra
```
termi-official, line 26: Since you removed BlockArrays here, should we also remove it from the other examples and remove it from docs, as this PR essentially unlocks bypassing it efficiently?

Also, is the removal of the other ones considered safe or is this just a think which happens to work because of the include order of the examples?

KnutAM: With or without PR one can do without, but one good argument for keeping is demonstrating that we can work with general matrix types. However, having it in the first coupled problem isn't so nice since this is (in some benchmarks I ran at least) quite a lot slower than the dofrange approach and the complexity isn't required. 

So keeping it in the ns-tutorial is fine IMO, at least for now (and we also have the support for global block matrices)

termi-official: If there is a significant performance hit, then we should not recommend it through tutorials, right?

KnutAM: Depends if linear solve is dominating, it doesn't matter so much. 

But I suggest leaving this for a separate issue/pr in that case - this already has quite a lot of changes

src/FEValues/common_values.jl
```diff
@@ -72,6 +80,13 @@ this is the number for the current facet.
 """
 function getnquadpoints end
 
+"""
+    getnbasefunctions(fe_v::AbstractValues)
+
+Get the number of base functions for the interpolation in `fe_v`.
```
termi-official: I think this is not accurate as it kinda suggests there is only a single interpolation. Can you rephrase this to leave less room for interpretation?

termi-official (approved):
One last minor point and then it should be good to go. :) Thanks for taking care of this fine improvement @KnutAM and also thanks for suggesting the improvement to the design @fredrikekre !

fredrikekre:
This PR brings us very close to having this: https://github.com/Ferrite-FEM/Ferrite.jl/issues/676#issuecomment-1514542497

> MultiCellValues suggest that we have some value for multiple cells (as e.g. in a patch), right?

Maybe, perhaps there is a better word than "Multi" then?

termi-official:
> Or can it just be `CellValues` and all `CellValues` are considered "multiple" (just that the multiple can be 1 field)?

I also thought about this already. Any adverse side-effects which we should bother about by doing this?

KristofferC:
I guess it is a bit ugly to special-case all the `get_xxx(cv, qp, ...)` to be valid when there is only a single field?

src/FEValues/common_values.jl
KnutAM: Updated the docstring, to be clear that this doesn't apply to `fe_v` with multiple function interpolations, since that is not supported.

KristofferC:
I actually didn't know this was merged. That's great!

PR #875, opened 2024-01-15
Increase visibility on mismatch between dof ordering and node ordering

termi-official:
Title checks out.

### DIFF
--- a/docs/src/literate-howto/postprocessing.jl
+++ b/docs/src/literate-howto/postprocessing.jl
@@ -13,12 +13,28 @@
 #
 # ## Introduction
 #
-# After running a simulation, we usually want to visualize the results in different ways.
-# The `L2Projector` and the `PointEvalHandler` build a pipeline for doing so. With the `L2Projector`,
-# integration point quantities can be projected to the nodes. The `PointEvalHandler` enables evaluation of
-# the finite element approximated function in any coordinate in the domain. Thus with the combination of both functionalities,
-# both nodal quantities and integration point quantities can be evaluated in any coordinate, allowing for example
-# cut-planes through 3D structures or cut-lines through 2D-structures.
+# After running a simulation, we usually want to postprocess and visualize the results in
+# different ways. Ferrite provides several tools to facilitate these tasks:
+#
+#  - L2 projection of (discrete) data onto FE interpolations using the `L2Projector`
+#  - Evalutation of fields (solutions, projections, etc) at arbitrary, user-defined, points
+#    using the `PointEvalHandler`
+#  - Builtin functionality for exporting data (solutions, cell data, projections, etc) to
+#    the VTK format
+#  - [Makie.jl](https://docs.makie.org/) based plotting using
+#    [FerriteViz.jl](https://ferrite-fem.github.io/FerriteViz.jl/)
+#
+# This how-to demonstrates the VTK export, the `L2Projector` for projecting discrete
+# quadrature point data onto a continuous FE interpolation, and the `PointEvalHandler` for
+# evaluating the FE solution, and the projection, along a user-defined cut line through the
+# domain.
+
+# !!! warning "Custom visualization"
+#     A common assumption is that the numbering of degrees of freedom matche the numbering
+#     of the nodes in the grid. This is *NOT* the case in Ferrite. If the available tools
+#     don't suit your needs and you decide to "roll your own" visualization you need to be
+#     aware of this and take it into account. For the specific case of evaluating the
+#     solution at the grid nodes you can use [`evaluate_at_grid_nodes`](@ref).
 #
 # This example continues from the Heat equation example, where the temperature field was
 # determined on a square domain. In this example, we first compute the heat flux in each
@@ -130,12 +146,6 @@ Plots.plot(getindex.(points, 1), u_points, xlabel = "x (coordinate)", ylabel = "
 # Secondly, the horizontal heat flux (i.e. the first component of the heat flux vector) is plotted.
 Plots.plot(getindex.(points, 1), getindex.(q_points, 1), xlabel = "x (coordinate)", ylabel = "q_x (flux in x-direction)", label = nothing)
 # *Figure 4*: ``x``-component of the flux along the cut line from *Figure 2*.
-
-#md # ## What next?
-#md # For more complicated visualization workflows we recommend either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
-#md # or users should export the solution into vtk files and use e.g. [ParaView](https://www.paraview.org/), [Mayavi](https://docs.enthought.com/mayavi/mayavi/), ... .
-#md # It should be noted that the ordering of the DofHandler and the numbering of the nodes does not match, hence we cannot directly use solution
-#md # vectors to assign colors to discretizations.
 
 #md # ## [Plain program](@id postprocessing-plain-program)
 #md #
--- a/docs/src/literate-howto/postprocessing.jl
+++ b/docs/src/literate-howto/postprocessing.jl
@@ -132,7 +132,7 @@ Plots.plot(getindex.(points, 1), getindex.(q_points, 1), xlabel = "x (coordinate
 # *Figure 4*: ``x``-component of the flux along the cut line from *Figure 2*.
 
 #md # ## What next?
-#md # For more complicated visualization workflows we recomment either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
+#md # For more complicated visualization workflows we recommend either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
 #md # or users should export the solution into vtk files and use e.g. [ParaView](https://www.paraview.org/), [Mayavi](https://docs.enthought.com/mayavi/mayavi/), ... .
 #md # It should be noted that the ordering of the DofHandler and the numbering of the nodes does not match, hence we cannot directly use solution
 #md # vectors to assign colors to discretizations.
--- a/docs/src/literate-howto/postprocessing.jl
+++ b/docs/src/literate-howto/postprocessing.jl
@@ -131,6 +131,12 @@ Plots.plot(getindex.(points, 1), u_points, xlabel = "x (coordinate)", ylabel = "
 Plots.plot(getindex.(points, 1), getindex.(q_points, 1), xlabel = "x (coordinate)", ylabel = "q_x (flux in x-direction)", label = nothing)
 # *Figure 4*: ``x``-component of the flux along the cut line from *Figure 2*.
 
+#md # ## What next?
+#md # For more complicated visualization workflows we recomment either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
+#md # or users should export the solution into vtk files and use e.g. [ParaView](https://www.paraview.org/), [Mayavi](https://docs.enthought.com/mayavi/mayavi/), ... .
+#md # It should be noted that the ordering of the DofHandler and the numbering of the nodes does not match, hence we cannot directly use solution
+#md # vectors to assign colors to discretizations.
+
 #md # ## [Plain program](@id postprocessing-plain-program)
 #md #
 #md # Here follows a version of the program without any comments.
--- a/docs/src/literate-tutorials/heat_equation.jl
+++ b/docs/src/literate-tutorials/heat_equation.jl
@@ -66,10 +66,14 @@ cellvalues = CellValues(qr, ip);
 # our interpolation `ip` defined above.
 # Lastly we `close!` the `DofHandler`, it is now that the dofs are distributed
 # for all the elements.
-# Please note that the dof numbering does not follow the node numbering of the grid.
 dh = DofHandler(grid)
 add!(dh, :u, ip)
 close!(dh);
+
+# !!! warning "Numbering of degrees of freedom"
+#     A common assumption is that the numbering of degrees of freedom follows the global
+#     numbering of the nodes in the grid. This is *NOT* the case in Ferrite. For more
+#     details, see the [Ferrite numbering rules](@ref "Ordering-of-Dofs").
 
 # Now that we have distributed all our dofs we can create our tangent matrix,
 # using `allocate_matrix`. This function returns a sparse matrix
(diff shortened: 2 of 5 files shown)


docs/src/literate-howto/postprocessing.jl
```diff
@@ -129,6 +129,12 @@ Plots.plot(getindex.(points,1), u_points, xlabel="x (coordinate)", ylabel="u (te
 Plots.plot(getindex.(points,1), getindex.(q_points,1), xlabel="x (coordinate)", ylabel="q_x (flux in x-direction)", label=nothing)
 # *Figure 4*: ``x``-component of the flux along the cut line from *Figure 2*.
 
+#md # ## What next?
+#md # For more complicated visualization workflows we recomment either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
```
koehlerson, line 133:
```suggestion
#md # For more complicated visualization workflows we recommend either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
```

termi-official: Spell checker committed seppuku.

docs/src/literate-howto/postprocessing.jl
```diff
@@ -129,6 +129,12 @@ Plots.plot(getindex.(points,1), u_points, xlabel="x (coordinate)", ylabel="u (te
 Plots.plot(getindex.(points,1), getindex.(q_points,1), xlabel="x (coordinate)", ylabel="q_x (flux in x-direction)", label=nothing)
 # *Figure 4*: ``x``-component of the flux along the cut line from *Figure 2*.
 
+#md # ## What next?
+#md # For more complicated visualization workflows we recommend either using our visualization tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
+#md # or users should export the solution into vtk files and use e.g. [ParaView](https://www.paraview.org/), [Mayavi](https://docs.enthought.com/mayavi/mayavi/), ... .
+#md # It should be noted that the ordering of the DofHandler and the numbering of the nodes does not match, hence we cannot directly use solution
+#md # vectors to assign colors to discretizations.
```
KnutAM, line 136: The part on VTK is already mentioned [above](https://ferrite-fem.github.io/Ferrite.jl/previews/PR875/howto/postprocessing/#Exporting-to-VTK). Would it be better to rename this subsection and write something like
 ```suggestion
#md # ## Visualizing grid data in Julia
#md # To visualize the solution on the grid, note that the numbering of the degrees of
#md # freedom generated by the `DofHandler` does not match the grid node ordering. 
#md # To get data at the nodes, the function [`evaluate_at_grid_nodes`](@ref) can be used. 
#md # For more complicated visualization workflows in Julia we recommend using our visualization
#md # tool [FerriteViz.jl](https://github.com/Ferrite-FEM/FerriteViz.jl)
```
(Haven't checked that if the links work)

File: docs/src/literate-tutorials/heat_equation.jl [hunk 3]
```diff
@@ -66,6 +66,8 @@ cellvalues = CellValues(qr, ip);
 # our interpolation `ip` defined above.
 # Lastly we `close!` the `DofHandler`, it is now that the dofs are distributed
 # for all the elements.
+#md # !!! note
+#md #     The dof numbering does not follow the node numbering of the grid.
```
KnutAM, line 70: IMO the docstring + postprocessing howto is enought, and it perhaps not necessary to mention this here too?

KnutAM (commented):
Nice! 
Some suggestions. 

An alternative (or additional) solution if we have more of these questions could be to have a FAQ page that we can refer to.

docs/src/literate-howto/postprocessing.jl
termi-official: Not sure. Above we only refer to the export of the fluxes, where we don't have a good way to compute them correctly in ParaView.

docs/src/literate-tutorials/heat_equation.jl
termi-official: I honestly do not think that beginners read the doc strings when they start using Ferrite. Rather they start from some example and modify it for their needs. Hence I introduced this redundant piece of information to increase visibility.

docs/src/literate-howto/postprocessing.jl
KnutAM: Yes, but the regular export was already shown in the heat equation example that this how-to includes. In that case, perhaps it would make sense to modify that case above to say that one can then export the fluxes in addition to the temperature field, and modify that code block to 
```julia
vtk_grid("heat_equation", grid) do vtk
    vtk_point_data(vtk, dh, u)
    vtk_point_data(vtk, projector, q_projected, "q")
end;
```
if that resolves that. But I think for new users the vtk export option is quite visible since it is used in most (all?) examples.

docs/src/literate-tutorials/heat_equation.jl
termi-official: I would like to leave this undocumented, so users don't come to conclusions to exploit this structure (in case we want to change the dof numbering algorithm in the future).

docs/src/literate-howto/postprocessing.jl
termi-official: Yea, but still users come to the conclusion that they better write their own visualizer over and over again, hence I wanted to really highlight that it is easier to use what is there already. :)

docs/src/literate-howto/postprocessing.jl
KnutAM, line 1: Suggestion for the beginning of this file 
```julia 
# # [Post processing and visualization](@id howto-postprocessing)
# After running a simulation, we usually want to visualize the results in different ways.
# Ferrite.jl provides several tools to facilitate such tasks, and direct visualization of 
# solutions, cell data, and more by using the builtin vtk export or 
# our `Makie.jl`-based visualization tool `FerriteViz.jl`. 
# Note that if these tools don't suit your needs, and you want to create your own 
# visualization, the node numbers do not match the degrees of freedom numbering.
# You can use [`evaluate_at_grid_nodes`](@ref) to get the solution at the nodes. 
# 
# This how-to demonstrates
# * The basic vtk export functionality
# * The `L2Projector` for projecting quadrature point data to the FE interpolations 
# * The `PointEvalHandler` to evaluate this projection at user-defined points in the grid. 
# 
# The results from the heat equation tutorial are used as a basis for this demonstration.
```
And then skip the two first sentences after introduction. 

Also update the vtk block to the new syntax, and I think we should add the primary solution as well to the output.

docs/src/literate-howto/postprocessing.jl
```diff
@@ -129,6 +129,12 @@ Plots.plot(getindex.(points,1), u_points, xlabel="x (coordinate)", ylabel="u (te
...
+
```
KnutAM, line 137: Suggestion at top replaces this in my opinion. 
```suggestion
```

src/Dofs/DofHandler.jl
```diff
@@ -127,6 +127,9 @@ add!(dh, :u, ip_u)
 add!(dh, :p, ip_p)
 close!(dh)
 ```
+
+!!! note
```
KnutAM, line 131:
```suggestion
!!! note "Dof numbering"
```
(Or DoF numbering, but we are a bit inconsistent here...)

KnutAM (commented):
Took a second look at this, I think it makes sense to give a brief overview of the visualization options at the beginning of this how-to, before covering the details.

KristofferC:
Regarding a FAQ, I remember reading https://gds.blog.gov.uk/2013/07/25/faqs-why-we-dont-have-them/

KnutAM:
After seeing the latest issues with this, I've thought that it would probably be most clear for new users if we note this when you obtain the solution to the heat equation problem, not when distributing the dofs. I.e. adding a `!!! note` after `u = K\f` [here](https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/heat_equation/#Solution-of-the-system).

termi-official:
I assume we do not have any analytics to confirm that users actually visit (and read) the heat equation tutorial?

KnutAM:
> I assume we do not have any analytics to confirm that users actually visit (and read) the heat equation tutorial?

No. Just guess-work that pointing it out there would be more obvious, hence the question on Slack.

KnutAM:
What if we add it as a small code comment instead (to heat eq. and linear elasticity)? 
This way, anyone copying one of these basic examples will also get it even if the online tutorial isn't read. I'm quite sure most new users will start by running one of these tutorials.

utk11:
Hello, a example or note on how to get the dof number from a node number could be helpful so that getting the solution at that particular node from the global solution vector would be easy. 

Also how to get the dof number from a node number?😅😅

termi-official:
> Also how to get the dof number from a node number?😅😅

In general this assumption does not work, so we do not have utils for this. I will try to take a look into better infrastructure for isoparametric elements, but there are short term more important things to do on my end.

docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -312,6 +313,7 @@ assemble_external_forces!(f_ext, dh, getfacetset(grid, "top"), facetvalues, trac
 # To account for the Dirichlet boundary conditions we use the `apply!` function.
 # This modifies elements in `K` and `f`, such that we can get the
 # correct solution vector `u` by using solving the linear equation system $K_{ij} \hat{u}_j = f^\mathrm{ext}_i$,
+# Please note that the dof numbering does not follow the node numbering of the grid.
```
KnutAM, line 316:
```suggestion
# Please note that the dof numbering does not follow the node numbering of the grid, i.e. `(u[2 * i - 1], u[2 * i])` are not the displacements of node `i`. Use `evaluate_at_grid_nodes` to get the displacements ordered by node numbers instead.
```

docs/src/topics/fe_intro.md
```diff
@@ -72,13 +72,18 @@ set of *elements* or *cells*. We call this geometric discretization *grid* (or *
...
-as the *Galerkin finite element method*. Formally we write the evaluation of our approximations
-at a specific point $\mathbf{x}$ in our domain $\Omega$ as:
+*degree of freedom* (dof) $\hat{u}_i$. These are sometimes also called *weights* or *nodal values*.
+In general, the indices for the dofs do **not** coincide with the indices of the nodes of a grid,
+as we can use different shape functions for either. Hence, in Ferrite grid nodes are numbered
+differently than degrees of freedom by default.
+
+Note that *shape functions* are sometimes referred to as *basis functions* or *trial functions*,
+and instead of $\phi_i$ they are sometimes denoted $N_i$. In this example we choose to approximate
+the test function in the same way. This approach is known as the *(Bubnov-)Galerkin finite element
```
KnutAM, line 84:
```suggestion
the test function in the same way. This approach is known as the *Ritz-Galerkin finite element
```
Not sure, just you noted this somewhere else. Seems to me that Bubnov is a less restrictive classification, but in this example its even Ritz?

docs/src/topics/fe_intro.md
```diff
@@ -72,13 +72,18 @@ set of *elements* or *cells*. We call this geometric discretization *grid* (or *
 and denote it with $\Omega_h$. In this example the corners of the triangles are called
 *nodes*.
 
-Next we introduce the finite element approximation $u_\mathrm{h} \approx u$ as a sum of N nodal
+Next we introduce the finite element approximation $u_\mathrm{h} \approx u$ as a sum of (nodal)
 *shape functions*, where we denote each of these function by $\phi_i$ and the corresponding
```
KnutAM, line 76:
```suggestion
Next we introduce the finite element approximation $u_\mathrm{h} \approx u$ as linear combination of
*shape functions*, $\phi_i$, weighted by the corresponding *degree of freedoms* (dof) $\hat{u}_i$.
These are sometimes also called *weights* or *nodal values*, but the numbering of these do not correspond
to the node numbers in the grid in Ferrite. While such numbering is common in basic finite element codes,
Ferrite supports different approximations of the finite element fields and the geometry, prohibiting
such basic numbering. For more details, see the [Ferrite numbering rules](@ref "Ordering-of-Dofs").
```

docs/src/topics/fe_intro.md
```diff
@@ -72,13 +72,18 @@ set of *elements* or *cells*. We call this geometric discretization *grid* (or *
...
-*nodal values* $\hat{u}_i$. Note that *shape functions* are sometimes referred to as
-*basis functions* or *trial functions*, and instead of $\phi_i$ they are sometimes denoted $N_i$.
-In this example we choose to approximate the test function in the same way. This approach is known
-as the *Galerkin finite element method*. Formally we write the evaluation of our approximations
-at a specific point $\mathbf{x}$ in our domain $\Omega$ as:
+*degree of freedom* (dof) $\hat{u}_i$. These are sometimes also called *weights* or *nodal values*.
+In general, the indices for the dofs do **not** coincide with the indices of the nodes of a grid,
+as we can use different shape functions for either. Hence, in Ferrite grid nodes are numbered
+differently than degrees of freedom by default.
```
KnutAM, line 80: Included above
```suggestion
```

docs/src/topics/fe_intro.md
termi-official: I have always learned that
1. We call methods Ritz-Galerkin methods, if we utilize variational principles for energy minimization formulations
2. We call methods Bubnov-Galerkin methods, if we utilize approximate solutions of the weak form where the ansatz and test spaces coincide
3. We call methods Petrov-Galerkin methods, if <---....---> where the ansatz and test spaces are different

docs/src/topics/fe_intro.md
```diff
@@ -72,13 +72,18 @@ set of *elements* or *cells*. We call this geometric discretization *grid* (or *
...
-*shape functions*, where we denote each of these function by $\phi_i$ and the corresponding
-*nodal values* $\hat{u}_i$. Note that *shape functions* are sometimes referred to as
-*basis functions* or *trial functions*, and instead of $\phi_i$ they are sometimes denoted $N_i$.
-In this example we choose to approximate the test function in the same way. This approach is known
-as the *Galerkin finite element method*. Formally we write the evaluation of our approximations
-at a specific point $\mathbf{x}$ in our domain $\Omega$ as:
+Next we introduce the finite element approximation $u_\mathrm{h} \approx u$ as linear combination of
+*shape functions*, $\phi_i$, weighted by the corresponding *degree of freedoms* (dof) $\hat{u}_i$.
+These are sometimes also called *weights* or *nodal values*, but the numbering of these do not correspond
+to the node numbers in the grid in Ferrite. While such numbering is common in basic finite element codes,
```
termi-official, line 78:
```suggestion
Next we introduce the finite element approximation $u_\mathrm{h} \approx u$ as linear combination of
*shape functions* $\phi_i$. The corresponding weights are usually called *degree of freedoms* (dof) $\hat{u}_i$.
In different contexts $\hat{u}_i$ is also just called *weights* or *nodal values*. The numbering of the dofs does not correspond
to the node numbers in the grid in Ferrite. While such numbering is common in basic finite element codes,
```

KnutAM: One more iteration ...
```suggestion
Next, we introduce the finite element approximation $u_\mathrm{h} \approx u$ as linear combination of
*shape functions*, $\phi_i$. The corresponding weights are usually called *degree of freedoms* (dofs), $\hat{u}_i$.
Sometimes, the dofs are called *weights* or *nodal values*. In Ferrite, the numbering of the dofs does not correspond
to the node numbers in the grid. While such numbering is common in basic finite element codes,
```

docs/src/literate-howto/postprocessing.jl
termi-official: Is this still relevant?

termi-official:
> Regarding a FAQ, I remember reading https://gds.blog.gov.uk/2013/07/25/faqs-why-we-dont-have-them/

Going over this again, why is duplication a bad thing? Or can we agree that some duplication across the documentation can be helpful to maintain a clearer guidance through the topics?

docs/src/literate-howto/postprocessing.jl
KnutAM: I think that such an introduction statement is better than the "What next?" further down, but it is a matter of taste I guess.

KnutAM (approved):

PR #880, opened 2024-02-02
Add helper function compute_bouding_box 

lijas:
I think it can be nice to add some grid utils. In this PR I am adding `compute_bouding_box() `, which I have recently needed for some pre-processing stuff.

Note. The current code is type unstable for some reason. Both `max_vertex` and `min_vertex` are both `Core.Box`. Anyone understands why?

### DIFF
--- a/src/Ferrite.jl
+++ b/src/Ferrite.jl
@@ -99,6 +99,7 @@ include("Grid/topology.jl")
 include("Grid/utils.jl")
 include("Grid/grid_generators.jl")
 include("Grid/coloring.jl")
+include("Grid/grid_utils.jl")
 
 # Dofs
 include("Dofs/DofHandler.jl")
--- a/src/Grid/grid_utils.jl
+++ b/src/Grid/grid_utils.jl
@@ -1,0 +1,19 @@
+
+
+"""
+    compute_bounding_box(grid::AbstractGrid)
+
+Computes the bounding box for a given grid, based on its node coordinates. 
+Returns the minimum and maximum vertices of the bounding box.
+"""
+function compute_bounding_box(grid::AbstractGrid{dim}) where {dim}
+    T = get_coordinate_eltype(grid)
+    min_vertex = Vec{dim}(i->T(+Inf))
+    max_vertex = Vec{dim}(i->T(-Inf))
+    for node in getnodes(grid)
+        x = get_node_coordinate(node)
+        max_vertex = Vec{dim}(i -> max(x[i], max_vertex[i]))
+        min_vertex = Vec{dim}(i -> min(x[i], min_vertex[i]))
+    end
+    return min_vertex, max_vertex
+end
--- a/test/test_abstractgrid.jl
+++ b/test/test_abstractgrid.jl
@@ -82,6 +82,10 @@
     @test Ferrite.ndofs(dhs[1]) == Ferrite.ndofs(dhs[2])
     @test isapprox(u1,u2,atol=1e-8)
 
+    minv, maxv = Ferrite.compute_bounding_box(subtype_grid)
+    @test minv == Vec((-1.0,-1.0))
+    @test maxv == Vec((+1.0,1.0))
+
     colors1 = Ferrite.create_coloring(subtype_grid, alg = ColoringAlgorithm.WorkStream)
     colors2 = Ferrite.create_coloring(reference_grid, alg = ColoringAlgorithm.WorkStream)
     @test all(colors1 .== colors2)
--- a/test/test_grid_dofhandler_vtk.jl
+++ b/test/test_grid_dofhandler_vtk.jl
@@ -91,6 +91,10 @@ end
             rm(dofhandlerfilename*".vtu")
         end
 
+        minv, maxv = Ferrite.compute_bounding_box(grid)
+        @test minv == 2left
+        @test maxv == 2right
+    
     end
 
 end # of testset


KnutAM:
Interesting problem - don't know why it is type unstable but could solve it if Tensors would support map
```julia
Base.map(f, iters::T...) where {T<:AbstractTensor} = T(map(f, Tensors.get_data.(iters)...))
function compute_bounding_box(grid::AbstractGrid{dim}) where {dim}
    T = get_coordinate_eltype(grid)
    min_vertex = Vec{dim,T}(i->T(+Inf))
    max_vertex = Vec{dim,T}(i->T(-Inf))
    for node in getnodes(grid)
        x = get_node_coordinate(node)
        max_vertex = map(max, max_vertex, x)
        min_vertex = map(min, min_vertex, x)
    end
    return min_vertex, max_vertex
end
```
Somehow defining an output `tmp = max(max_vertex[1], x[1])` and only considering the first node, `tmp` was set as Any if max_vertex and min_vertex were updated and outputted, but type stable otherwise - so perhaps it is just inference giving up for some reason?

lijas:
I asked on slack helpdesk, and got this reply:

> 
> This will be slow, because Julia will need to pass the value boxed into the closure to make sure it gets the value of the right type.
> This is a bug in Julia, because type inference could be used to determine that the box is unnecessary, but currently in Julia, the compiler step that creates the box happens before type inference, so it can't be optimised away (or so I'm told - I don't understand it).
> The solution is:
> Either, typeassert your variables so that the box have less impact on performance (but it'll still allocate)
> Or, use a let statement

src/Grid/grid_utils.jl
```diff
@@ -0,0 +1,21 @@
+
+
+"""
+    compute_bounding_box(grid::AbstractGrid)
+
+Computes the bounding box for a given grid, based on its node coordinates. 
+Returns the minimum and maximum vertices of the bounding box.
+"""
+function compute_bounding_box(grid::AbstractGrid{dim}) where {dim}
```
fredrikekre, line 9: Perhaps:
```suggestion
function bounding_box(grid::AbstractGrid{dim}) where {dim}
```

src/Ferrite.jl
```diff
@@ -99,6 +99,7 @@ include("Grid/topology.jl")
 include("Grid/utils.jl")
 include("Grid/grid_generators.jl")
 include("Grid/coloring.jl")
+include("Grid/grid_utils.jl")
```
fredrikekre, line 102: Use the existing `Grid/utils.jl` file instead?

src/Grid/grid_utils.jl
```diff
@@ -0,0 +1,21 @@
...
+    T = get_coordinate_eltype(grid)
+    min_vertex = Vec{dim}(i->T(+Inf))
+    max_vertex = Vec{dim}(i->T(-Inf))
```
fredrikekre, line 12:
```suggestion
    min_vertex = Vec{dim}(i -> typemax(T))
    max_vertex = Vec{dim}(i -> typemin(T))
```

src/Grid/grid_utils.jl
```diff
@@ -0,0 +1,21 @@
...
+    for node in getnodes(grid)
```
fredrikekre, line 13: Might be useful to compute the bounding box for a subset, or a faceset etc, but can be added later.

test/test_abstractgrid.jl
```diff
@@ -82,6 +82,10 @@
     @test Ferrite.ndofs(dhs[1]) == Ferrite.ndofs(dhs[2])
     @test isapprox(u1,u2,atol=1e-8)
 
+    minv, maxv = Ferrite.compute_bounding_box(subtype_grid)
```
fredrikekre, line 85:
```suggestion
    minv, maxv = Ferrite.bounding_box(subtype_grid)
```

test/test_grid_dofhandler_vtk.jl
```diff
@@ -91,6 +91,10 @@ end
             rm(dofhandlerfilename*".vtu")
         end
 
+        minv, maxv = Ferrite.compute_bounding_box(grid)
```
fredrikekre, line 94:
```suggestion
        minv, maxv = Ferrite.bounding_box(grid)
```

fredrikekre (approved):

src/Grid/utils.jl
```diff
@@ -340,3 +340,25 @@ for (func,                   entity_f,            entity_t,     filter_f,
...
+"""
+    bounding_box(grid::AbstractGrid)
+
+Computes the bounding box for a given grid, based on its node coordinates. 
+Returns the minimum and maximum vertices of the bounding box.
+"""
+function bounding_box(grid::AbstractGrid{dim}) where {dim}
+    T = get_coordinate_eltype(grid)
+    min_vertex = Vec{dim}(i->typemin(T))
+    max_vertex = Vec{dim}(i->typemax(T))
```
fredrikekre, line 355:
```suggestion
    min_vertex = Vec{dim}(i->typemax(T))
    max_vertex = Vec{dim}(i->typemin(T))
```

src/Grid/utils.jl
```diff
@@ -340,3 +340,25 @@ for (func,                   entity_f,            entity_t,     filter_f,
     end
 end
 end
+
+
+
+"""
+    bounding_box(grid::AbstractGrid)
+
+Computes the bounding box for a given grid, based on its node coordinates.
```
termi-official, line 349:
```suggestion
Computes the axis-aligned bounding box for a given grid, based on its node coordinates. 
```

termi-official: Note that there are different types of bounding boxes.

src/Grid/utils.jl
```diff
@@ -340,3 +340,25 @@ for (func,                   entity_f,            entity_t,     filter_f,
...
 end
 end
+
+
+
+"""
+    bounding_box(grid::AbstractGrid)
+
+Computes the bounding box for a given grid, based on its node coordinates. 
+Returns the minimum and maximum vertices of the bounding box.
```
termi-official, line 350:
```suggestion
Returns the minimum and maximum vertex coordinates of the bounding box.
```
I think this is a bit more clear.

PR #888, opened 2024-02-29
Implement a proper sparsity pattern

fredrikekre:
This PR introduces `struct SparsityPattern` for a way to work with the pattern directly instead of producing a matrix directly.

Some advantages of this are:
 - Make it possible to add custom entries into the pattern (e.g. insert the necessary entries for a lagrange multiplier)
 - Size is no longer limited to `ndofs(dh) x ndofs(dh)` so it is easy to insert new dofs
 - Simpler to support new matrix types. Before this patch everything in the construction was tightly coupled to `SparseMatrixCSC` (e.g. how element connections and constraint condensation were implemented). With this patch everything is handled on the `SparsityPattern` level independent of what type of matrix will be instantiated in the end. All that is needed to support a new matrix type is now to write a `MatrixType(::SparsityPattern)` constructor.
 - Enables instantiation of multiple matrices based on the same pattern (e.g. stiffness and mass matrices).
 - Faster and lower memory usage (benchmarks incoming)


termi-official:
Supersedes #596 ?

docs/src/literate-tutorials/incompressible_elasticity.jl
```diff
@@ -258,7 +258,7 @@ function solve(ν, interpolation_u, interpolation_p)
     cellvalues_u, cellvalues_p, facevalues_u = create_values(interpolation_u, interpolation_p)
 
     ## Assembly and solve
-    K = create_sparsity_pattern(dh)
+    K = create_matrix(dh)
```
termi-official, line 261: Maybe we can showcase the use of the extension here?

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,41 @@
...
+If the solver only needs half the matrix there is no need to assemble both halves.
+For this purpose there is a [`create_symmetric_sparsity_pattern`](@ref) function that
+will only create the upper half of the matrix, and return a `Symmetric` wrapped
+`SparseMatrixCSC`.
+
+Given a `DofHandler` `dh` we can obtain the (symmetric) sparsity pattern as
+
+```julia
+K = create_matrix(dh)
+K = create_symmetric_sparsity_pattern(dh)
```
termi-official, line 23: I think this is not correct?

src/HeapAllocator.jl
```diff
@@ -0,0 +1,300 @@
...
+    return
+end
+
+# A page corresponds to a larger Libc.malloc call (MALLOC_PAGE_SIZE). Each page
+# is split into smaller blocks to minimize the number of Libc.malloc/Libc.free
+# calls.
+mutable struct Page
+    const ptr::SizedPtr{UInt8} # malloc'd pointer
+    const blocksize::UInt      # blocksize for this page
+    const freelist::BitVector  # block is free/used
```
termi-official, line 26: I guess we discard support for 1.6 after this PR is merged?

src/HeapAllocator.jl
termi-official, line 1: Will this module become a custom package?

src/Dofs/sparsity_pattern.jl
```diff
@@ -102,38 +415,164 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+                    # Insert col as a non zero index for this row
+                    add_entry!(sp, row, col)
+                end
+            end
+        end
+    end
+    # 2. Make sure diagonal entries are included
+    for d in 1:ndofs(dh)
+        add_entry!(sp, d, d)
+    end
```
termi-official, line 452: Why? For example many saddle point problems do not need these for the block 2,2.

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,396 @@
...
+"""
+    add_entry!(sp::AbstractSparsityPattern, row::Int, col::Int)
+
+Add an entry to the sparsity pattern `sp` at row `row` and column `col`.
+"""
+add_entry!(sp::AbstractSparsityPattern, row::Int, col::Int)
+
+# This is necessary to avoid warning about not importing Base.eachrow when
+# adding docstring before the definitions further down.
+function eachrow end
```
termi-official, line 36: Is there a reason why we don't dispatch Base.eachrow?

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,396 @@
...
+
+# This is implemented as an extension in ext/FerriteBlockArrays.jl. This method exist to
+# give a nice error message. Keep the signature in sync with the constructor in
+# ext/FerriteBlockArrays.jl (just a tad less specific here).
+function BlockSparsityPattern(::AbstractVector{<:Integer})
+    msg = "Ferrite's block matrix functionality depends on the package BlockArrays to " *
+          "be installed and loaded into the session. Install BlockArrays " *
+          "(`pkg> add BlockArrays`) and load it (`using BlockArrays`) to enable."
+    error(msg)
+end
```
termi-official, line 186: OT: Should I adopt this design for the GPU PR?

docs/src/reference/sparsity_pattern.md
```diff
@@ -0,0 +1,80 @@
...
+ 4. Instantiate sparse matrices from the sparsity pattern using [`create_matrix`](@ref)
+
+For example, the steps above could look as follows when creating a `SparseMatrixCSC` matrix:
+```julia
+dh = DofHandler(...)
+ch = ConstraintHandler(...)
+# 1. Create empty pattern
+sparsity_pattern = SparsityPattern(ndofs(dh), ndofs(dh))
+# 2. Add entries
+create_sparsity_pattern!(sparsity_pattern, dh)
```
termi-official, line 20: Maybe we should mention around here how to manually add entries (via `Ferrite.add_entry!`).

test/test_dofs.jl
termi-official, line 1: Maybe we should add test coverage manually adjusting the sparsity pattern via `add_entry!`?

termi-official (commented):
Thanks for this one Fredrik! It looks really awesome. :) I roughly went over the implementation and have some things which I like to discuss.

Since I could not annotate the files: I think you missed updating the the dg example and the integration test.

docs/src/literate-tutorials/incompressible_elasticity.jl
KnutAM: Following the discussion https://github.com/Ferrite-FEM/Ferrite.jl/pull/872#discussion_r1502656633 - I think it would be better to add such a showcase in a how-to or another tutorial, to avoid complicating the first tutorial on multi-field problems.

termi-official: I really have memory like a goldfish haha

src/Dofs/sparsity_pattern.jl
KristofferC: Many sparse solvers need them.

termi-official: Sparse solvers as in direct sparse solvers?

docs/src/literate-tutorials/incompressible_elasticity.jl
fredrikekre: The block array extension? I don't think it make sense to use that unless you have a solver that takes advantage over easy access to individual blocks.

docs/src/topics/sparse_matrix.md
fredrikekre: Yea, missed this. Actually I removed symmetric support on the pattern level, but this is one point to discuss.

src/HeapAllocator.jl
fredrikekre: Yea, eventually I think that would be useful, but can simmer here for a while first :)

src/Dofs/sparsity_pattern.jl
fredrikekre: This isn't new, we already added the diagonal explicitly. `apply!` actually requires it too, since we insert a dummy equation on constrained dofs. Even if that wasn't the case it doesn't really cost anything to have it.

src/Dofs/sparsity_pattern.jl
fredrikekre: It's not really the same concept; `Base.eachrow` is for matrices and gives the full row, this just give the stored entries of the row. We can call it something else inside of Ferrite though.

src/Dofs/sparsity_pattern.jl
fredrikekre: It is a bit of a hack to export a struct from an extension, not sure if it is a good idea and might move the blocksparsity pattern inside of Ferrite (but keep the matrix instantiation in the extension).

test/test_dofs.jl
fredrikekre: Yea I haven't really written tests for the new stuff, focused on getting old tests to pass.

src/Dofs/sparsity_pattern.jl
```diff
@@ -102,38 +415,164 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
-"""
-    _add_cross_coupling(coupling_sdh, dof_i, dof_j, cell_field_dofs, neighbor_field_dofs, i, j, sym, keep_constrained, ch, cnt, I, J)
+function _create_sparsity_pattern!(
+        sp::AbstractSparsityPattern, dh::DofHandler, ch::Union{ConstraintHandler, Nothing},
+        keep_constrained::Bool,
+        coupling::Union{Vector{<:AbstractMatrix{Bool}}, Nothing},
+        cross_coupling::Union{AbstractMatrix{Bool}, Nothing},
+        topology,
+    )
+    # 1. Add all connections between dofs for every cell while filtering based
```
lijas, line 425: Maybe factor this out similar to `cross_element_coupling!` so it is possible to call it from user code building custom pattern?

src/Dofs/sparsity_pattern.jl
termi-official: I see. Technically we should also be able to just instantiate a normal matrix with a block pattern. Or should this functionality be rather moved into `close!(::DofHandler, ::DofStrategy)``?

src/HeapAllocator.jl
```diff
@@ -0,0 +1,308 @@
+module HeapAllocator
+
+@eval macro $(Symbol("const"))(field)
+    if VERSION >= v"1.8.0-DEV.1148"
+        Expr(:const, esc(field))
+    else
+        return esc(field)
+    end
+end
```
KnutAM, line 9: Worth putting in Compat.jl? 
Ref: https://github.com/JuliaLang/Compat.jl/issues/800

src/Dofs/sparsity_pattern.jl
```diff
@@ -102,38 +401,164 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
-Mutates `I, J` to account for cross-element coupling by calling [`_add_cross_coupling`](@ref).
-Returns the updated value of `cnt`.
+    cross_element_coupling!(
+        sp::SparsityPattern, dh::DofHandler, ch::Union{ConstraintHandler, Nothing},
+        topology::ExclusiveTopology, keep_constrained::Bool,
+        cross_coupling::AbstractMatrix{Bool},
+    )
 
-Used internally for sparsity patterns with cross-element coupling.
+TODO: Expose to users?
```
termi-official, line 551: I would lean towards answering this with a "yes" to give users more fine grained control (and modularity) over the sparsity pattern. And maybe we should discuss the name again in this case.

termi-official:
> Before cleaning up and finishing this PR there are some things I would like input on:
> 
>     1. `create_sparsity_pattern(dh::DofHandler, ...)` still exist but now return a sparsity pattern (and not the matrix). If you don't update your code there will be an error when you try `start_assemble` which we can intercept and print a nice message (you might run into a `MethodError` earlier though). I think this is an OK change.
> 

I think this is a good idea not only for this reason. New users might not understand the difference between sparsity pattern and the actual sparse matrix and might try to pass the pattern to the assembler anyway.

Alternatively we can also not error in this case and just assemble into a standard matrix format, which can be queried by `finalize_assemble(assembler)`.

>     2. The new function `create_matrix` can be used as a direct replacement to `create_sparsity_pattern` so when you don't care about modifying the pattern and just want a `SparseMatrixCSC` you can use `create_matrix(dh)`.
> 

I think this is a great convenience for the large portion of our user base.

>     3. `create_matrix` can also now be used to pass the type of sparse matrix you want, e.g. `create_matrix(SparseMatrixCSC{Float32, Int}, sparsity_pattern)`. Many matrix types are not so nice to write like this though (e.g. `SparseMatrixCSR` have a first type parameter which corresponds to the indexing....) and for a block matrix you need something like `BlockMatrix{Float64, Matrix{SparseMatrixCSC{Float64, Int}}` to fully specify the type. Of course we can make `create_matrix(BlockMatrix, sparsity_pattern)` work and fill in the defaults, but if you don't want the defaults it will be pretty clunky. One idea is to provide aliases in Ferrite, e.g. `Ferrite.MatrixTypes.CSRMatrix{Tv, Ti}`. Thoughts?

I think I am not understanding the issue here. If we want to have a very specific matrix, then we should be specific here. Not sure if `Ferrite.MatrixTypes` is a better way to go here, because it makes extensions a bit more clunky.

KnutAM:
Nice work @fredrikekre! 

1. I agree that this is a good change, and think it is reasonable that `start_assemble(create_sparsity_pattern(dh))` errors.
2. `create_matrix(dh, [ch])` is nice!
   IMO also supporting `create_matrix(MatrixType, dh, [ch])` would make sense (didn't check if already supported)
3. I think adding matrix specification conveniences should be postponed until we gather some experience with the new setup. If some cases are frequently used, it could make sense to provide shortcuts

KnutAM:
Perhaps a stupid question, but would it not make sense to support `sp = create_sparsity_pattern(MatrixType, dh, ch)` to allow easy use for extensions such as BlockSparsityPattern?

fredrikekre:
> I think I am not understanding the issue here. If we want to have a very specific matrix, then we should be specific here. Not sure if Ferrite.MatrixTypes is a better way to go here, because it makes extensions a bit more clunky.

The issue is that it looks ugly to have something like
```julia
A = create_matrix(BlockMatrix{Float64, Matrix{SparseMatrixCSR{1, Float64, Int64}}}, sparsity_pattern)
```
But with aliases we could have e.g.
```julia
A = create_matrix(MatrixTypes.BlockMatrix{MatrixTypes.CSRMatrix{Float64, Int}}, sparsity_pattern)
```
A method for that would be defined in an extension still. The point is that type parameters isn't always "user-friendly" so with MatrixTypes we could normalize it a bit. But we can do this later as Knut points out.

> Perhaps a stupid question, but would it not make sense to support sp = create_sparsity_pattern(MatrixType, dh, ch) to allow easy use for extensions such as BlockSparsityPattern?

The matrix type doesn't necessarily map directly to one pattern type. And if you want a specific pattern type, why not call that constructor  directly (`create_sparsity_pattern!(MyPattern(...), dh, ch)`)?.

As it is right now, `DofHandler` defaults to `SparsityPattern` which then defaults to `SparseMatrixCSC`.

KnutAM:
> why not call that constructor directly (create_sparsity_pattern!(MyPattern(...), dh, ch))?.

If `MyPattern === BlockSparsityPattern` it wouldn't be available as it is defined in the extension, wasn't that the reason for the hack earlier, which then could be avoided?

fredrikekre:
BlockSparsityPattern is in Ferrite now, but even with the hack you could still call the constructor (but get an error telling you to load the needed package which you would have to do anyway in order to pass the matrix type?).

docs/src/topics/sparse_matrix.md
```diff
@@ -1,6 +1,170 @@
...
+    the matrix directly. For example, most examples in this documentation don't deal with
+    the sparsity pattern explicitly.
+
+
+### Increasing the sparsity
+
+By default when creating a sparsity pattern it is assumed that a DoF couple with all other
+DoFs in the same element.
+
+is created such that DoFs couple with all DoFs in the same
```
KristofferC, line 145: Remove?

termi-official:
Note that the assembler itself is still tightly coupled to SparseMatrixCSC.

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+are discussed.
+
+```@contents
+Pages = ["sparse_matrix.md"]
+Depth = 2:2
+```
+
+## Sparsity pattern
+
+The sparse structure of the linear system depends on many factors such as e.g. the weak
```
KnutAM, line 14: I'm a bit torn here since the derivative is not always the definition of the matrix (but implicitly assumed in the text below for e.g. reference to Poisson's equation). Yet, I find it logical introduce the coupling via the derivative, e.g.

----------
In general, a finite element problem requires solving an equation system
```math
\boldsymbol{r}(\boldsymbol{a}) = 0
```
to find $\boldsymbol{a}$. For nonlinear problems when using the Newton-Raphson method, 
or for linear problems, a system matrix is constructed as 
```math
K_{ij} = \frac{\partial r_i}{\partial a_i}
```
Each entry, $r_i$, in $\boldsymbol{r}$, typically depend on only a few of the unknowns, $a_j$, such that many entries in $K_{ij}$ will be zero. This so-called sparse structure of the linear system, $K_{ij}$, depends on many factors such as e.g. the weak ...

----------

With the current text, it is for example not obvious that $K_{ij}$ can be zero while $K_{ji}$ is not. (I.e. which is nonzero "DoFs `i` and `j` couple"). 

(Not sure if it makes sense to discuss cases when using alternative techniques for defining the stiffness matrix via quasi-newton methods here, perhaps just mention that when discussing the `coupling` keyword?)

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+```
+
+Moreover, if the problem is solved with periodic boundary conditions, for example by constraining the
+value on the right side to the value on the left side, there will be additional couplings.
+In the example above, this means that DoF 4 should be equal to DoF 1. Since DoF 4 is
+constrained it has to be eliminated from the system. Existing entries that include DoF 4 are
+`(3, 4)`, `(4, 3)`, and `(4, 4)`. Given the simple constraint in this case we can simply
+replace DoF 4 with DoF 1 in these entries and we end up with entries `(3, 1)`, `(1, 3)`, and
+`(1, 1)`. This results in two new entries: `(3, 1)` and `(1, 3)` (entry `(1, 1)` is already
+included).
```
KnutAM, line 77: While this is a very nice introductory text, I think it hides the important things below.
Perhaps add at least the example in a collapsible block?

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+incremental construction efficient it is necessary to use a dynamic data structure which
+allow for fast insertions.
+
+The sparsity pattern also serves as a "matrix builder". When all entries are inserted into
+the sparsity pattern the dynamic data structure is typically converted, or "compressed",
+into a sparse matrix format such as e.g. the [*compressed sparse row
+(CSR)*](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format))
+format or the [*compressed sparse column
+(CSC)*](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_column_(CSC_or_CCS))
+format, which is the default sparse matrix type implemented in the SparseArrays standard
```
KnutAM, line 93:
```suggestion
format, where the latter is the default sparse matrix type implemented in the SparseArrays standard
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Basic sparsity patterns construction in Ferrite
+
+Working with the sparsity pattern explicitly is in many cases not necessary. For basic
+usage (e.g. when only one matrix needed, when no customization of the pattern is
+required, etc) there exist convenience methods of [`create_matrix`](@ref) that return
+the matrix directly. For example, most examples in this documentation don't deal with
+the sparsity pattern explicitly.
+    
+### Costum sparsity patterns construction in Ferrite
```
KnutAM, line 110:
```suggestion
### Custom sparsity patterns construction in Ferrite
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+
+In more advanced cases, we need more fine grained control of the sparsity pattern. The 
+following steps are typically taken when constructing a sparsity pattern in Ferrite:
+
+ 1. *Initialize an empty pattern.* This can be done by either using the
+    [`init_sparsity_pattern(dh)`](@ref) function or by using a constructor directly.
+    `init_sparsity_pattern` will return a default pattern type that is compatible with the
+    DofHandler. In some cases you might require another type of pattern (for example a
+    blocked pattern) and in that case you can use the constructor directly.
```
KnutAM, line 119:
```suggestion
    blocked pattern, see [Blocked sparsity pattern](@ref)) and in that case you can use the constructor directly.
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+required, etc) there exist convenience methods of [`create_matrix`](@ref) that return
+the matrix directly. For example, most examples in this documentation don't deal with
+the sparsity pattern explicitly.
+    
+### Costum sparsity patterns construction in Ferrite
+
+In more advanced cases, we need more fine grained control of the sparsity pattern. The 
+following steps are typically taken when constructing a sparsity pattern in Ferrite:
+
+ 1. *Initialize an empty pattern.* This can be done by either using the
```
KnutAM, line 115:
```suggestion
 1. **Initialize an empty pattern:** This can be done by either using the
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+    [`init_sparsity_pattern(dh)`](@ref) function or by using a constructor directly.
+    `init_sparsity_pattern` will return a default pattern type that is compatible with the
+    DofHandler. In some cases you might require another type of pattern (for example a
+    blocked pattern) and in that case you can use the constructor directly.
+
+ 2. *Add entries to the pattern.* There are a number of functions that add entries to the
```
KnutAM, line 121:
```suggestion
 2. **Add entries to the pattern:** There are a number of functions that add entries to the
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+       interfaces between elements (e.g. for discontinuous Galerkin methods).
+     - [`Ferrite.add_entry!`](@ref) adds a single entry to the pattern. This can be used if
+       you need to add custom entries that are not covered by the other functions.
+
+ 3. *Condense the pattern.* This is done with the [`condense_sparsity_pattern!`](@ref)
+    function and adds entries required from constraints and boundary conditions in the
+    ConstraintHandler. Note that condensing the pattern *must* be done as the last operation
+    on the pattern, e.g. after adding all other entries.
+
+ 4. *Instantiate the matrix*. A sparse matrix can be created from the sparsity pattern using
```
KnutAM, line 137:
```suggestion
 4. **Instantiate the matrix:** A sparse matrix can be created from the sparsity pattern using
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+     - [`cross_element_coupling!`](@ref) adds entries for couplings between the DoFs in
+       neighboring elements. These entries are required when integrating along internal
+       interfaces between elements (e.g. for discontinuous Galerkin methods).
+     - [`Ferrite.add_entry!`](@ref) adds a single entry to the pattern. This can be used if
+       you need to add custom entries that are not covered by the other functions.
+
+ 3. *Condense the pattern.* This is done with the [`condense_sparsity_pattern!`](@ref)
+    function and adds entries required from constraints and boundary conditions in the
+    ConstraintHandler. Note that condensing the pattern *must* be done as the last operation
+    on the pattern, e.g. after adding all other entries.
```
KnutAM, line 135: As discussed on Slack...
```suggestion
 2. **Add entries to the pattern:** This can conveniently be done by using [`create_sparsity_pattern!`](@ref). If a finer-grained control is required, there are a number of functions that add entries to the
    pattern:
     - [`cell_couplings!`](@ref) adds entries for all couplings between the DoFs
       within each element. These entries correspond to assembling the standard element
       matrix and is thus almost always required.
     - [`cross_element_coupling!`](@ref) adds entries for couplings between the DoFs in
       neighboring elements. These entries are required when integrating along internal
       interfaces between elements (e.g. for discontinuous Galerkin methods).
     - [`Ferrite.add_entry!`](@ref) adds a single entry to the pattern. This can be used if
       you need to add custom entries that are not covered by the other functions.
     - [`condense_sparsity_pattern!`](@ref) adds entries required from constraints and boundary conditions in the
    ConstraintHandler. Note that condensing the pattern *must* be done as the last operation
    on the pattern, e.g. after adding all other entries. 
```

KnutAM: With the extra comments that I think `constraint_couplings!` is a better name than `condense_sparsity_pattern!`. 

Side-note: Sounds like having to call `condense_sparsity_pattern!` last can easily lead to user bugs. Would it make sense to add a `condensed::Ref{Bool}` (or `ScalarWrapper`) in `SparsityPattern` to allow the other methods error if trying to add entries after condensation?

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+ 4. *Instantiate the matrix*. A sparse matrix can be created from the sparsity pattern using
+    [`create_matrix`](@ref), see [Instantiating the sparse matrix](@ref) below for more
+    details.
+
+!!! note "Swiss army knife"
+    The [`create_sparsity_pattern!`](@ref) function is something of a swiss army knife since
+    it can do more than just adding the basic element couplings as described above. For
+    example, when passed the correct arguments, it can also do the job of
+    `cross_element_coupling!` and `condense_sparsity_pattern!`. Refer to the API reference
+    for details.
```
KnutAM, line 146: To be removed if adopting my suggestion with `cell_couplings!` above

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+
+Discuss the `keep_constrained` keyword argument.
+
+
+
+
+## Creating the sparsity pattern
+
+Given a `DofHandler` we can obtain the corresponding sparse matrix by using the
+[`create_sparsity_pattern`](@ref) function. This will setup a `SparseMatrixCSC`
```
KnutAM, line 178:
```suggestion
[`create_matrix`](@ref) function. This will setup a `SparseMatrixCSC`
```
?

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+
+This is the most general case and results in a dense
+matrix. However, in many cases the sparsity can be increased by reducing the number of DoFs
+that couple with each other. This can be done by using the `coupling` keyword argument in
+the `create_sparsity_pattern` function. This keyword argument takes a function that returns
+a boolean value indicating if two DoFs should couple or not. The function should take two
+arguments, the first being the element index and the second being the DoF index within the
+element. The function should return `true` if the DoFs should couple and `false` otherwise.
```
KnutAM, line 164: This sentence doesn't make sense to me. 
I would have expected that `f(i, j)` should say if DoF `i` couples to DoF `j`? But here it sounds like the `cellid` is expected, but perhaps I'm just confused with array element vs finite element?

Wasn't there an option to give the field coupling via a `Matrix{Bool}` before?

Side-note: Probably giving `uncoupled::Vector{Pair{Symbol,Symbol}}` would be more user-friendly to remove couplings instead, but can be done later...

docs/src/topics/sparse_matrix.md
KnutAM: Unresolved this since this looks inconsistent here, shouldn't it at least be `create_symmetric_matrix(dh)`, or removed if not used anymore?

KnutAM (commented):
Very nice description (and work on this PR, looking forward to the benchmarks 🚀)

Added some suggestions also to make it faster to get to the main points at the end of the docs!

docs/src/topics/sparse_matrix.md
fredrikekre: This paragraph is Copilot hallucinations I forgot to remove haha

termi-official:
I have another consideration which came to my mind.

Is `create_sparsity_pattern!` really the name which we want here? I would argue that we might want to have a `create_default_sparsity_pattern!` function, which calls internally something like `add_H1_sparsity_pattern` to creates default sparsity pattern for standard H1 functions. Then we could factor out the face-coupling pattern (for DG) into a function `add_DG_sparsity_pattern`. This way might be able to modularize the sparsity pattern construction a bit for different use-cases. What do you think?

docs/src/topics/sparse_matrix.md
lijas: If you dont call the condense function last, I think you will at least get an error in the assembly-routine.

I think the name change make sense!

termi-official: > With the extra comments that I think `constraint_couplings!` is a better name than `condense_sparsity_pattern!`.
> 

The process is called condensation, right? So why should we use a different name for this?

> Side-note: Sounds like having to call `condense_sparsity_pattern!` last can easily lead to user bugs. Would it make sense to add a `condensed::Ref{Bool}` (or `ScalarWrapper`) in `SparsityPattern` to allow the other methods error if trying to add entries after condensation?

I think this is a really good idea.

KnutAM:
> Is `create_sparsity_pattern!` really the name which we want here?

To me, it makes sense to have this function to do the "right thing" (such that it will work) given the inputs. I thought that if you give it a dofhandler with DG-interpolations it will call `cross_element_coupling!`? And if you give a constrant handler, it will call `constraint_couplings!` etc. (But maybe I misunderstood your explanation earlier @fredrikekre?)

For advanced use-cases (for example if you have additional info that can increase the sparsity), I think the separation with the extra functions are really nice, but I don't think we need to add extra convenience for many different cases (at least not yet).

fredrikekre:
Yea, right now `create_sparsity_pattern!` is a bit of a do-it-all function

termi-official:
> > Is `create_sparsity_pattern!` really the name which we want here?
> 
> To me, it makes sense to have this function to do the "right thing" (such that it will work) given the inputs. I thought that if you give it a dofhandler with DG-interpolations it will call `cross_element_coupling!`? And if you give a constrant handler, it will call `constraint_couplings!` etc. (But maybe I misunderstood your explanation earlier @fredrikekre?)
> 
This only happens if you also pass the topology. You cannot differentiate between DG and "normal" P0 on interpolation level, as the coupling is a property of the weak form and not of the interpolation.

docs/src/topics/sparse_matrix.md
KnutAM: > The process is called condensation, right? So why should we use a different name for this?

Sorry, missed this comment @termi-official: My logic is that the condensation happens when you call `apply!`, but this step only adds additional entries due to the couplings introduced by the affine constraints.

termi-official: No problem. I see. I still think users reading `constraint_couplings` would expect something different happening. What about `add_constraint_pattern!` or even something like this?

KnutAM: Yes, probably good to be consistent with using `add_XY!` or `XY!` here, and since 
`add_entry!` is kindof needed perhaps having add makes sense. For the `couplings` part, was just to be consistent with `cross_element_coupling!`. 
So the options could be
* `add_cell_pattern!`
* `add_cross_element_pattern!` (cell?)
* `add_entry!`
* `add_constraint_pattern!`
or 
* `add_cell_couplings!`
* `add_cross_element_couplings!` (cell?)
* `add_entry!`
* `add_constraint_couplings!`

I think I prefer `couplings`, as this is more logical for the others IMO.

termi-official: For me `pattern` makes sense too for the suffix since we insert some kind of pattern to the sparsity pattern.

What about `cross_element` -> `interface`?

fredrikekre: Perhaps
```julia
insert_cell_couplings!(...)
insert_interface_couplings!(...)
insert_constraint_couplings!(...)
insert_entry!(...)
```
or
```julia
create_cell_couplings!(...)
create_interface_couplings!(...)
create_constraint_couplings!(...)
create_coupling!(...)
```
or
```julia
add_cell_couplings!(...)
add_interface_couplings!(...)
add_constraint_couplings!(...)
add_coupling!(...)
```

fredrikekre:
Any thoughts on https://github.com/Ferrite-FEM/Ferrite.jl/pull/888#discussion_r1635597188? Perhaps also change `create_matrix` -> `alloc_matrix` or `allocate_matrix` to make it more clear that there is no assembly happening?

termi-official:
> Any thoughts on [#888 (comment)](https://github.com/Ferrite-FEM/Ferrite.jl/pull/888#discussion_r1635597188)? Perhaps also change `create_matrix` -> `alloc_matrix` or `allocate_matrix` to make it more clear that there is no assembly happening?

I am not sure if it is really some ambiguity in the word `create` here. We can also argue for alloc/allocate/instantiate or similar. I have no preference here either.

docs/src/topics/sparse_matrix.md
fredrikekre: A typo I think, but any of the suggestions could work with `_coupling(s)` -> `entries` too.

lijas:
I think `add_*` is better than `create_*`.
I am not sure about `create_matrix` or `allocate_matrix`. Both are fine with me.

KnutAM:
I agree with @lijas that `add` is better and somehow consistent with `add!` used elsewhere in the code base. 

I find that `allocate_matrix` is a better name, but it is also not very important.

KristofferC:
`allocate` heavily suggests that the thing is "non-initialized" and has to be further "processed". So that might make sense in this case.

CHANGELOG.md
```diff
@@ -244,6 +244,22 @@ more discussion).
   ```
 
 
+- **Sparsity pattern and global matrix construction**: since there is now explicit support
+  for working with the sparsity pattern before instantiating a matrix the function
+  `create_sparsity_pattern` have been repurposed. To simply recover the old functionaly that
```
KristofferC, line 249:
```suggestion
  `create_sparsity_pattern` has been repurposed. To recover the old functionality that
```

CHANGELOG.md
```diff
@@ -317,6 +333,17 @@ more discussion).
   a given grid (based on its node coordinates), and returns the minimum and maximum vertices
   of the bounding box. ([#880][github-880])
 
+- Support for working with sparsity patterns have been added. This means that Ferrite
```
KristofferC, line 336:
```suggestion
- Support for working with sparsity patterns has been added. This means that Ferrite
```

CHANGELOG.md
```diff
@@ -317,6 +333,17 @@ more discussion).
...
+  exposes the intermediate "state" between the DofHandler and the instantiated matrix as the
+  new struct `SparsityPattern`. This make it possible to insert custom equations or
+  couplings in the pattern before instantiating the matrix. The function
+  `create_sparsity_pattern` have been repurposed to return a `SparsityPattern` instead of a
+  `SparseMatrixCSC`. The new function `allocate_matrix` is used to instantiate the matrix
```
KristofferC, line 341:
```suggestion
  `SparseMatrixCSC`. The new function `allocate_matrix` is instead used to instantiate the matrix
```

CHANGELOG.md
```diff
@@ -317,6 +333,17 @@ more discussion).
...
+  instead. Refer to the documentation for more details. ([#888][github-888])
```
KristofferC, line 342:
```suggestion
  . Refer to the documentation for more details. ([#888][github-888])
```

CHANGELOG.md
```diff
@@ -317,6 +333,17 @@ more discussion).
...
+
+  **To upgrade**: if you simply want to recover the old functionality and don't need to work
```
KristofferC, line 344:
```suggestion
  **To upgrade**: if you want to recover the old functionality and don't need to work
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,191 @@
...
+Pages = ["sparse_matrix.md"]
+Depth = 2:2
+```
+
+## Sparsity pattern
+
+The sparse structure of the linear system depends on many factors such as e.g. the weak
+form, the discretization, and the choice of interpolation(s). In the end it boils down to
+how the degrees of freedom (DoFs) *couple* with each other. The most common reason that two
+DoFs couple is because they simply belong to the same element. Note, however, that this is
```
KristofferC, line 17:
```suggestion
DoFs couple is because they belong to the same element. Note, however, that this is
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,191 @@
...
+not guaranteed to result in a coupling since it depends on the specific weak form that is
+being discretized, see [XXX](@ref). Boundary conditions and constraints can also result in
```
KristofferC, line 19: TODO here

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,191 @@
...
+additional DoF couplings.
+
+If DoFs `i` and `j` couple, then the computed value in the eventual matrix will be
+*nonzero*. In this case the entry `(i, j)` should be included in the sparsity pattern.
```
KristofferC, line 23: Is it worth saying something like "structurally non-zero"? Like, it could happen for various reasons that it becomes zero anyway.

src/Ferrite.jl
```diff
@@ -139,6 +139,7 @@ include("Dofs/DofHandler.jl")
 include("Dofs/ConstraintHandler.jl")
 include("Dofs/apply_analytical.jl")
 include("Dofs/sparsity_pattern.jl")
+include("Dofs/bsp.jl")
```
termi-official, line 142:
```suggestion
include("Dofs/block_sparsity_pattern.jl")
```

ext/FerriteBlockArrays.jl
```diff
@@ -1,60 +1,59 @@
...
-    if !(size(B, 1) == size(B, 2) == ndofs(dh))
-        error("size of input matrix ($(size(B))) does not match number of dofs ($(ndofs(dh)))")
-    end
-    K = create_sparsity_pattern(dh, ch; kwargs...)
-    ax = axes(B)
-    for block_j in blockaxes(B, 2), block_i in blockaxes(B, 1)
-        range_j = ax[2][block_j]
-        range_i = ax[1][block_i]
-        B[block_i, block_j] = K[range_i, range_j]
+"""
```
termi-official, line 23: What about BlockedMatrix?

docs/src/reference/sparsity_pattern.md
```diff
@@ -0,0 +1,57 @@
...
+See the topic section on [Sparsity pattern and sparse matrices](@ref topic-sparse-matrix).
+
+## Sparsity patterns
+
+### `AbstractSparsityPattern`
+
+The following applies to all subtypes of `AbstractSparsityPattern`:
+
+```@docs
+AbstractSparsityPattern
```
termi-official, line 13: Is this type really ment to be exported?

docs/src/topics/sparse_matrix.md
termi-official: This point also picked my attention. I would not even to down to nonlinear problems here. The point is really about linear operators, which either appear naturally as part of a PDE description (see e.g. Navier-Stokes example or Poisson problem or even linear elasticity) or in the linearization (e.g. second variation in energy-based elasticity). Hence the sparse matrix is not directly about linear systems, but about the representation of the discretization of an operator which allows e.g. efficient linear solves and fast simulations. So I have something into this direction in my head:
```suggestion
The sparse structure of linear operators in the context of the simulation of partial differential equations depends on many factors such as e.g. the weak
```
but it needs further refinement. Does this make sense for you guys?

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,165 @@
...
+can also result in additional DoF couplings.
+
+If DoFs `i` and `j` couple, then the computed value in the eventual matrix will be
+*structurally nonzero*[^1]. In this case the entry `(i, j)` should be included in the
+sparsity pattern. Conversely, if DoFs `i` and `j` *don't* couple, then the computed value
+will be *zero*. In this case the entry `(i, j)` should *not* be included in the sparsity
+pattern since there is no need to allocate memory for entries that will be zero.
+
+The sparsity, i.e. the ratio of zero-entries to the total number of entries, is often *very*
+high and taking advantage of this results in huge savings in terms of memory. For example,
```
termi-official, line 29: Maybe we should be a bit more specific here to not mislead users accidentally.
```suggestion
The sparsity, i.e. the ratio of zero-entries to the total number of entries, is for most practical problems using low-order interpolations *very*
high for  and taking advantage of this results in huge savings in terms of memory. For example,
```
Note that this statement is not true anymore for e.g. high order Ansatz or interpolations spanning multiple elements.

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,165 @@
...
+    1. Since DoF 4 is constrained it has to be eliminated from the system. Existing entries
+    that include DoF 4 are `(3, 4)`, `(4, 3)`, and `(4, 4)`. Given the simple constraint in
+    this case we can simply replace DoF 4 with DoF 1 in these entries and we end up with
+    entries `(3, 1)`, `(1, 3)`, and `(1, 1)`. This results in two new entries: `(3, 1)` and
+    `(1, 3)` (entry `(1, 1)` is already included).
+
+## Creating sparsity patterns
+
+Creating a sparsity pattern can be quite expensive if not done properly and therefore
+Ferrite provides efficient methods and datastructures for this. In general the sparsity
```
termi-official, line 85:
```suggestion
Ferrite provides efficient methods and data structures for this. In general the sparsity
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,165 @@
...
+allow for fast insertions.
+
+The sparsity pattern also serves as a "matrix builder". When all entries are inserted into
+the sparsity pattern the dynamic data structure is typically converted, or "compressed",
+into a sparse matrix format such as e.g. the [*compressed sparse row
+(CSR)*](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format))
+format or the [*compressed sparse column
+(CSC)*](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_column_(CSC_or_CCS))
+format, where the latter is the default sparse matrix type implemented in the SparseArrays
+standard library. These matrix formats allow for fast linear algebra operations, such as
```
termi-official, line 97: Maybe we should just add a link to make sure this is clear for new users which are not too familiar with the ecosystem.
```suggestion
format, where the latter is the default sparse matrix type implemented in the [SparseArrays
standard library](https://sparsearrays.juliasparse.org/dev/). These matrix formats allow for fast linear algebra operations, such as
```

docs/src/topics/sparse_matrix.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Blocked sparsity pattern
+
+!!! todo
+    Discuss `BlockSparsityPattern` and `BlockArrays` extension.
+
+## Instantiating the sparse matrix
+
+!!! todo
+    Write text.
```
termi-official, line 165: Will this be handled in a separate PR or do you intend to add it here?

ext/FerriteBlockArrays.jl
termi-official, line 1: Should we modify one of the examples to use show how to use this feature, or add a how-to guide? I am not sure how easy users will find this feature and what the implications for them are.

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
+"""
+    AbstractSparsityPattern
+
+Supertype for sparsity pattern implementations, e.g. [`SparsityPattern`](@ref) and
+[`BlockSparsityPattern`](@ref).
+"""
+abstract type AbstractSparsityPattern end
+
+"""
+    n_rows(sp::AbstractSparsityPattern)
```
termi-official, line 14: Is there any reason to use a new convention for the function naming here?(xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/741) Maybe `getnrows`, `numrows`, or at least `nrows` to be somewhat consistent in the naming?

Analogue for the 2 functions below.

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
+        sp::AbstractSparsityPattern, dh::DofHandler,
+        ch::Union{ConstraintHandler, Nothing} = nothing;
+        keep_constrained::Bool = true,
+        coupling::Union{AbstractMatrix{Bool}, Nothing} = nothing,
+        interface_coupling::Union{AbstractMatrix{Bool}, Nothing} = nothing,
+        topology = nothing,
+    )
+    # Argument checking
+    isclosed(dh) || error("the DofHandler must be closed")
+    if n_rows(sp) < ndofs(dh) || n_cols(sp) < ndofs(dh)
```
termi-official, line 232: Out of curiosity, why is this `<` and not `!=`?

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
+#  - `dh`: the DofHandler for which to generate the pattern
+#  - `ch`: the ConstraintHandler. Used to add the non-zero entries resulting from
+#    `AffineConstraint`s (and `PeriodicDirichlet` constraints, which uses `AffineConstraint`s
+#    internally). See [`add_constraint_entries!`](@ref).
+# # Keyword arguments
+#  - `keep_constrained`: whether or not entries for constrained DoFs should be kept
+#    (`keep_constrained = true`) or eliminated (`keep_constrained = false`) from the sparsity
+#    pattern. `keep_constrained = false` requires passing the ConstraintHandler.
+#  - `coupling`: the coupling between fields/components within each cell. By default
+#    (`coupling = nothing`) it is assumed that all DoFs in each cell couple with each other.
```
termi-official, line 261: Debris?

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
+ - `coupling`: the coupling between fields/components within each cell. By default
+   (`coupling = nothing`) it is assumed that all DoFs in each cell couple with each other.
+"""
+function add_cell_entries!(
+        sp::AbstractSparsityPattern,
+        dh::DofHandler, ch::Union{ConstraintHandler, Nothing} = nothing;
+        keep_constrained::Bool = true, coupling::Union{AbstractMatrix{Bool}, Nothing} = nothing,
+    )
+    # Expand coupling from nfields × nfields to ndofs_per_cell × ndofs_per_cell
+    # TODO: Perhaps this can be done in the loop over SubDofHandlers instead.
```
termi-official, line 288: The resulting pattern should be equivalent in this case anyway, right?

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
+    ```julia
+    sp = init_sparsity_pattern(dh)
+    create_sparsity_pattern!(sp, dh)
+    K = allocate_matrix(sp)
+    M = allocate_matrix(sp)
+    ```
+    instead of
+    ```julia
+    K = allocate_matrix(dh)
+    M = allocate_matrix(dh)
```
termi-official, line 425: What about
```julia
    K = allocate_matrix(dh)
    M = deepcopy(K)
```
?

src/Dofs/sparsity_pattern.jl
```diff
@@ -1,83 +1,473 @@
...
     sz == size(coupling, 2) || error("coupling not square")
-    sym && (issymmetric(coupling) || error("coupling not symmetric"))
 
     # Return one matrix per (potential) sub-domain
     outs = Matrix{Bool}[]
     field_dims = map(fieldname -> n_components(dh, fieldname), dh.field_names)
 
-    for sdh in dh.subdofhandlers
-        out = zeros(Bool, ndofs_per_cell(sdh), ndofs_per_cell(sdh))
+    for fh in dh.subdofhandlers
```
termi-official, line 465: Why the rename here? :)

src/Dofs/sparsity_pattern.jl
```diff
@@ -102,38 +492,139 @@ function _coupling_to_local_dof_coupling(dh::DofHandler, coupling::AbstractMatri
...
+    # Insert new entries into the sparsity pattern
+    for (row, colidxs) in sp′
+        # TODO: Extract row here and just insert_sorted
+        for col in colidxs
+            add_entry!(sp, row, col)
+        end
+    end
+
+    # Release the memory in the temporary heap
+    HeapAllocator.free(heap)
```
termi-official, line 601: Since I just see it: Can't we automate this so this does not accidentally leak memory if e.g. one of the methods above throws an exception or errors? I have finalizers in mind (https://docs.julialang.org/en/v1/base/base/#Base.finalizer) but I am not sure if this works as intended.

src/Dofs/sparsity_pattern.jl
termi-official: If the concept is different, then maybe we should rename the function? `row_iterator` maybe?

src/exports.jl
```diff
@@ -116,14 +116,23 @@ export
...
-    create_sparsity_pattern,
-    create_symmetric_sparsity_pattern,
     dof_range,
     renumber!,
     DofOrder,
     evaluate_at_grid_nodes,
     apply_analytical!,
 
+# Sparsity pattern
+    AbstractSparsityPattern,
```
termi-official, line 126: Do we want to export this one?

test/test_sparsity_patterns.jl
```diff
@@ -0,0 +1,332 @@
...
+    if !(1 <= row <= tp.nrowscols[1] && 1 <= col <= tp.nrowscols[2])
+        error("out of bounds")
+    end
+    r = tp.data[row]
+    k = searchsortedfirst(r, col)
+    if k == lastindex(r) + 1 || r[k] != col
+        insert!(r, k, col)
+    end
+    return
+end
```
termi-official, line 23: Maybe we should make this a how-to?

termi-official (commented):
Looks really cool already! Found some things which I discuss or which I think needs some clarification.

src/Dofs/sparsity_pattern.jl
KristofferC: Deepcopy should pretty much never be used.

termi-official: Why? Don't we want a fully recursive copy of the sparse matrix with a duplication of the internal arrays here?

KristofferC:
I am a bit uneasy about the HeapAllocator stuff. While I understand it was fun to write is there really no way whatever this is used for that cannot be used in pure, safe Julia?

src/Dofs/sparsity_pattern.jl
fredrikekre: `copy` is enought here and would work for perfectly fine for `SparseMatrixCSC` but doesn't work in general (e.g. no copy for HYPREMatrix etc). I can add a note.

KnutAM: > Deepcopy should pretty much never be used.

@fredrikekre has told me the same, but is this documented somewhere in the julia ecosystem?

KristofferC: > Why?

See e.g. https://github.com/JuliaLang/julia/issues/42796#issuecomment-951232853

fredrikekre:
## PR one matrix

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc    |
|-----------|-----------|----------|-------|----------|----------|
|  10x10x10 |     29114 |     0.28 |   20% |      226 |  223 MiB |
|  20x20x20 |    216024 |     1.74 |    9% |      901 |  1.4 GiB |
|  30x30x30 |    710734 |     5.57 |  4.5% |   2.62 k |  4.3 GiB |
|  40x40x40 |   1663244 |     13.9 |  7.5% |   5.87 k |  9.9 GiB |
|  50x50x50 |   3223554 |     30.6 |  9.8% |  11.11 k | 19.2 GiB |
|  60x60x60 |   5541664 |     65.3 | 12.7% |  18.84 k | 32.8 GiB |
|  70x70x70 |   8767574 |    120.1 | 18.4% |  29.54 k | 51.7 GiB |

## master one matrix

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc     |
|-----------|-----------|----------|-------|----------|-----------|
|  10x10x10 |     29114 |     0.22 | 41.5% |   3.09 k |   608 MiB |
|  20x20x20 |    216024 |      1.1 | 15.9% |   3.09 k |   4.1 GiB |
|  30x30x30 |    710734 |      3.7 | 14.1% |   3.09 k |  12.9 GiB |
|  40x40x40 |   1663244 |      9.7 | 10.2% |   3.10 k |  30.4 GiB |
|  50x50x50 |   3223554 |     38.4 |  6.2% |   3.10 k |  58.9 GiB |
|  60x60x60 |   5541664 |     75.6 |  5.8% |   3.10 k | 101.4 GiB |
|  70x70x70 |   8767574 |    177.7 |  7.4% |   3.10 k | 171.9 GiB |


## PR two matrices

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc    |
|-----------|-----------|----------|-------|----------|----------|
|  10x10x10 |     29114 |      0.3 | 17.9% |      234 |  320 MiB |
|  20x20x20 |    216024 |      1.9 |  8.6% |      909 |  2.1 GiB |
|  30x30x30 |    710734 |      6.1 |  5.3% |   2.62 k |  6.6 GiB |
|  40x40x40 |   1663244 |     14.8 |  8.1% |   5.87 k | 15.4 GiB |
|  50x50x50 |   3223554 |     33.7 |  9.0% |  11.12 k | 29.6 GiB |
|  60x60x60 |   5541664 |     72.6 | 12.8% |  18.85 k | 50.8 GiB |
|  70x70x70 |   8767574 |    134.9 | 17.8% |  29.55 k | 80.1 GiB |

## master two matrices

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc     |
|-----------|-----------|----------|-------|----------|-----------|
|  10x10x10 |     29114 |     0.34 | 30.2% |   3.16 k |   1.2 GiB |
|  20x20x20 |    216024 |     2.27 | 22.2% |   3.16 k |   8.3 GiB |
|  30x30x30 |    710734 |     7.44 | 15.6% |   3.17 k |  25.8 GiB |
|  40x40x40 |   1663244 |    20.85 | 11.6% |   3.18 k |  60.8 GiB |
|  50x50x50 |   3223554 |    80.63 |  7.2% |   3.18 k | 117.8 GiB |
|  60x60x60 |   5541664 |   155.89 |  6.7% |   3.18 k | 202.7 GiB |

(70x70x70 OOM killed)


## Code for PR

```julia
using Ferrite

n = 10

const grid = generate_grid(Hexahedron, (n, n, n))
const top = ExclusiveTopology(grid)
const dh = DofHandler(grid)
add!(dh, :u, Lagrange{RefHexahedron, 2}()^3)
add!(dh, :p, Lagrange{RefHexahedron, 1}())
close!(dh)

const ch = ConstraintHandler(dh)
const facets = collect_periodic_facets(grid)
add!(ch, PeriodicDirichlet(:u, facets))
close!(ch)

function single_matrix(dh, ch)
    dsp = init_sparsity_pattern(dh)
    add_cell_entries!(dsp, dh, ch)
    add_constraint_entries!(dsp, ch)
    K = allocate_matrix(dsp)
    return K
end

function double_matrix(dh, ch)
    dsp = init_sparsity_pattern(dh)
    add_cell_entries!(dsp, dh, ch)
    add_constraint_entries!(dsp, ch)
    K = allocate_matrix(dsp)
    M = allocate_matrix(dsp)
    return K, M
end

@time single_matrix(dh, ch)
@time double_matrix(dh, ch)
```

## Code for master

```julia
using Ferrite

n = 10

const grid = generate_grid(Hexahedron, (n, n, n))
const top = ExclusiveTopology(grid)
const dh = DofHandler(grid)
add!(dh, :u, Lagrange{RefHexahedron, 2}()^3)
add!(dh, :p, Lagrange{RefHexahedron, 1}())
close!(dh)

const ch = ConstraintHandler(dh)
const facets = collect_periodic_facets(grid)
add!(ch, PeriodicDirichlet(:u, facets))
close!(ch)

function single_matrix(dh, ch)
    return K
end

function double_matrix(dh, ch)
    M = create_sparsity_pattern(dh, ch)
    return K, M
end

@time single_matrix(dh, ch)
@time double_matrix(dh, ch)
```

ext/FerriteBlockArrays.jl
fredrikekre: I think you would just create a normal matrix and then block it yourself then.

src/Dofs/sparsity_pattern.jl
fredrikekre: You can create a pattern with ndofs+x rows/cols and have entries from 1:ndofs inserted in the first block.

src/Dofs/sparsity_pattern.jl
fredrikekre: Yes, just a bit unnecessary to expand them into a collection instead of on-demand.

src/Dofs/sparsity_pattern.jl
fredrikekre: Yea I don't think this is necessary now after my simplifications.

fredrikekre:
> I am a bit uneasy about the HeapAllocator stuff. While I understand it was fun to write is there really no way whatever this is used for that cannot be used in pure, safe Julia?

I simplified it greatly and in particular removed the manual malloc and pointer computation. Now it is just a way to batch allocate. It is very similar to what Knut did in #974 except that it batches based on size so that it becomes a bit easier to reuse memory. I think we can merge the two implementations in the future.

ext/FerriteBlockArrays.jl
fredrikekre: Yea if someone has a good example where a block-based solver is useful that would indeed be nice.

KnutAM:
Thanks for the comprehensive benchmarks!

* Is there anything that can be done about the performance regression for one-matrix systems below 40^3. I think this size (1.6 M dofs) is quite common in Ferrite, and e.g. mesh adaptivity could suffer here? (Is there a difference in 2d with a less dense sparsity pattern?)
* Out of curiosity: Do you know why the number of allocs on master barely changes when creating one or two matrices. From the code I would expect the exact double  as the alloc size does.

And another question, why is the sparsity pattern row-major? Seems like this adds a small perf-hit with the extra buffer for standard CSC matrices?

fredrikekre:
> Is there anything that can be done about the performance regression for one-matrix systems below 40^3. I think this size (1.6 M dofs) is quite common in Ferrite, and e.g. mesh adaptivity could suffer here? (Is there a difference in 2d with a less dense sparsity pattern?)

It is a trade-off between memory and time -- master uses quite a lot more memory. In my experience that have been the limiting factor. Compared to a linear solve of this size I am not too worried. I suppose also one should factor in that we now have quite a lot more flexibility. We can keep the old memory-hungry alg for `allocate_matrix(dh)` but I don't think it is worth the duplicated code. I can check 2D too.

> Do you know why the number of allocs on master barely changes when creating one or two matrices. 

That does look a bit strange, I will have a look again.

fredrikekre:
> That does look a bit strange, I will have a look again.

I didn't run the function first so almost all the allocations are from compilation. I can update the table later.

KnutAM:
> We can keep the old memory-hungry alg for allocate_matrix(dh) but I don't think it is worth the duplicated code.

I also don't think that is worth it (and there is no doubt that this PR is an improvement, was more curious if something could be done to get the same perf).

fredrikekre:
Without compilation:

 ## PR one matrix

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc    |
|-----------|-----------|----------|-------|----------|----------|
| 10x10x10 | 29114 | 0.2 s | 0.0% | 226 | 223.204 MiB |
| 20x20x20 | 216024 | 1.7 s | 3.0% | 901 | 1.354 GiB |
| 30x30x30 | 710734 | 5.8 s | 6.9% | 2615 | 4.318 GiB |
| 40x40x40 | 1663244 | 14.6 s | 8.2% | 5865 | 9.985 GiB |
| 50x50x50 | 3223554 | 30.6 s | 10.0% | 11107 | 19.177 GiB |

## master one matrix

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc     |
|-----------|-----------|----------|-------|----------|-----------|
| 10x10x10 | 29114 | 0.1 s | 0.0% | 74 | 608.068 MiB |
| 20x20x20 | 216024 | 1.2 s | 21.7% | 75 | 4.234 GiB |
| 30x30x30 | 710734 | 3.9 s | 12.2% | 76 | 12.882 GiB |
| 40x40x40 | 1663244 | 16.5 s | 7.2% | 81 | 30.423 GiB |
| 50x50x50 | 3223554 | 42.2 s | 6.0% | 82 | 58.908 GiB |


## PR two matrices

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc    |
|-----------|-----------|----------|-------|----------|----------|
| 10x10x10 | 29114 | 0.3 s | 0.0% | 234 | 320.632 MiB |
| 20x20x20 | 216024 | 1.9 s | 5.4% | 909 | 2.058 GiB |
| 30x30x30 | 710734 | 6.2 s | 7.6% | 2623 | 6.628 GiB |
| 40x40x40 | 1663244 | 15.6 s | 8.3% | 5873 | 15.382 GiB |
| 50x50x50 | 3223554 | 37.7 s | 10.4% | 11115 | 29.627 GiB |

## master two matrices

| Mesh size | ndofs(dh) | time [s] | %gc   | n allocs | alloc     |
|-----------|-----------|----------|-------|----------|-----------|
| 10x10x10 | 29114 | 0.3 s | 18.3% | 148 | 1.188 GiB |
| 20x20x20 | 216024 | 2.3 s | 25.5% | 150 | 8.468 GiB |
| 30x30x30 | 710734 | 8.1 s | 14.0% | 152 | 25.764 GiB |
| 40x40x40 | 1663244 | 28.9 s | 9.7% | 162 | 60.846 GiB |
| 50x50x50 | 3223554 | 87.6 s | 7.0% | 164 | 117.816 GiB |

PR #889, opened 2024-03-01
Update some missing Lagrange docs

termi-official:


KnutAM (approved):
Nice! 
Just some formatting fix suggestions while we are at it...

docs/src/topics/degrees_of_freedom.md
```diff
@@ -25,8 +25,8 @@ need to specify number of components for the field. Here we add a vector field `
 (2 components for a 2D problem) and a scalar field `:p`.
 
 ```@example dofs
-add!(dh, :u, Lagrange{2,RefTetrahedron,1}()^2)
-add!(dh, :p, Lagrange{2,RefTetrahedron,1}())
+add!(dh, :u, Lagrange{RefTriangle,1}()^2)
+add!(dh, :p, Lagrange{RefTriangle,1}())
```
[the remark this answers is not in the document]
termi-official: Do we have this documented somehwere? And do you think adding a formatter might help out for such issues?

docs/src/topics/grid.md
```diff
@@ -86,7 +86,7 @@ In this space a local numbering of nodes and faces exists, i.e.
 The example shows a local face ID ordering, defined as:
 
 ```julia
-faces(::Lagrange{2,RefCube,1}) = ((1,2), (2,3), (3,4), (4,1))
+faces(::Lagrange{RefQuadrilateral,1}) = ((1,2), (2,3), (3,4), (4,1))
 ```
 
 Other face ID definitions [can be found in the src files](https://github.com/Ferrite-FEM/Ferrite.jl/blob/8224282ab4d67cb523ef342e4a6ceb1716764ada/src/interpolations.jl#L154) in the corresponding `faces` dispatch.
```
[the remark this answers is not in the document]
termi-official: No, but we do not have docs for these yet. #689 maybe? Also, this doc must be specified further. I just remembered that `faces` does not exist anymore for interpolations (and that this doc string was a bit misleading in first place).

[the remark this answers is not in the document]
KnutAM: > `faces` does not exist anymore for interpolations

What do you mean? On master e.g. `Ferrite.faces(Lagrange{RefQuadrilateral,1}())` works fine?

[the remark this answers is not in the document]
termi-official: https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/deprecations.jl#L8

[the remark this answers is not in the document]
KnutAM: It is really tricky with those deprecations since we don't see anything unless running tests (or explicitly asking for depwarn) - since many users will just run scripts, would it perhaps be better to give an explicit warning?

Sorry for spamming this PR, but just to save the idea in case this could be a valuable pattern, what about 
```julia
function foo(::AbstractFloat)
    println("Allowed")
end
@generated function foo(v::Integer)
    @warn "Calling foo(::Integer) is deprecated"
    return :(println("Was allowed before"))
end
```
Giving 
```
julia> foo(1.)
Allowed

julia> foo(1)
┌ Warning: Calling foo(::Integer) is deprecated
└ @ Main REPL[11]:2
Was allowed before

julia> foo(1)
Was allowed before
```
CC @fredrikekre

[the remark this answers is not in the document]
termi-official: Okay let us do this specific change in a different PR since I really just want to update the docs here.

termi-official:
I noticed that on master some things can be now done in a better way (i.e. without relying on assumptions taken between the interpolation and cell) so I updated it. Also some dispatches were missing, which I added for completeness of the API.

termi-official:
I just saw that #935 has quite a bit of overlap here, so we need to decide on a merge order.

src/Grid/grid.jl
```diff
@@ -131,6 +157,11 @@ reference_facets(::Type{<:AbstractRefShape})
 @inline reference_facets(refshape::Type{<:AbstractRefShape{2}}) = reference_edges(refshape)
 @inline reference_facets(refshape::Type{<:AbstractRefShape{3}}) = reference_faces(refshape)
 
+@inline reference_faces(::AbstractCell{refshape}) where refshape <:AbstractRefShape = reference_faces(refshape)
+@inline reference_edges(::AbstractCell{refshape}) where refshape <:AbstractRefShape = reference_edges(refshape)
+@inline reference_vertices(::AbstractCell{refshape}) where refshape <:AbstractRefShape = reference_vertices(refshape)
+@inline reference_facets(::AbstractCell{refshape}) where refshape <:AbstractRefShape = reference_facets(refshape)
```
KnutAM, line 163: These seem untested, I don't think we use them, but rather use `reference_faces(refshape)` etc. directly, see e.g. the `vertices` function. 
If introduced here, should be used there, but not really needed if only required there?

Still, if we want this, then
```suggestion
@inline reference_faces(::AbstractCell{refshape}) where refshape = reference_faces(refshape)
@inline reference_edges(::AbstractCell{refshape}) where refshape = reference_edges(refshape)
@inline reference_vertices(::AbstractCell{refshape}) where refshape = reference_vertices(refshape)
@inline reference_facets(::AbstractCell{refshape}) where refshape = reference_facets(refshape)
```
(already restricted in `AbstractCell` [ref](https://github.com/Ferrite-FEM/Ferrite.jl/blob/c33a191f618fa9dfb2d5b5efb3c6eae8000ba6d1/src/Ferrite.jl#L60))

docs/src/devdocs/elements.md
```diff
@@ -2,17 +2,20 @@
...
+Ferrite.reference_vertices(::Ferrite.AbstractRefShape)
 Ferrite.vertices(::Ferrite.AbstractCell)
+Ferrite.reference_edges(::Ferrite.AbstractRefShape)
 Ferrite.edges(::Ferrite.AbstractCell)
 Ferrite.reference_faces(::Ferrite.AbstractRefShape)
 Ferrite.faces(::Ferrite.AbstractCell)
+Ferrite.reference_facets(::Ferrite.AbstractRefShape)
+Ferrite.facets(::Ferrite.AbstractCell)
 Ferrite.geometric_interpolation(::Ferrite.AbstractCell)
 ```
```
KnutAM, line 20: Actually, this isn't true AFAIU.
It suffices to implement `get_node_ids(c::MyCellType)` and `geometric_interpolation`. The part regarding reference shapes should be moved to reference shape devdocs, and stated here that this refers to a new cell for existing reference shape. 

```suggestion
**Move to refshapes devdocs**
### Required methods to implement for a new `AbstractRefShape`
```@docs
Ferrite.reference_vertices
Ferrite.reference_edges(::Ferrite.AbstractRefShape)
Ferrite.reference_faces(::Ferrite.AbstractRefShape)
```
which automatically defines 
```@docs
Ferrite.reference_facets(::Ferrite.AbstractRefShape)
```


### Required methods to implement for all subtypes of `AbstractCell` to define a new element

```@docs
Ferrite.get_node_ids
geometric_interpolation
```

### Automatically defined methods using the above required methods
```@docs
Ferrite.vertices(::Ferrite.AbstractCell)
Ferrite.edges(::Ferrite.AbstractCell)
Ferrite.faces(::Ferrite.AbstractCell)
Ferrite.facets(::Ferrite.AbstractCell)
```

KnutAM (approved):
LGTM if if removing changes in grid.md as discussed on Slack, and fixes the comments

docs/src/devdocs/elements.md
termi-official: Oh right, I totally misread the headline.

src/Grid/grid.jl
```diff
@@ -157,18 +157,19 @@ reference_facets(::Type{<:AbstractRefShape})
...
 
 """
     geometric_interpolation(::AbstractCell)::Interpolation
+    geometric_interpolation(::Type{AbstractCell})::Interpolation
 
 Each `AbstractCell` type has a unique geometric interpolation describing its geometry.
 This function returns that interpolation.
 """
-geometric_interpolation(::AbstractCell)
+geometric_interpolation(::T) where T <: AbstractCell = geometric_interpolation(T)
```
KnutAM, line 172: Missing test

docs/src/devdocs/elements.md
```diff
@@ -8,22 +8,35 @@ by the associated reference element.
 ### Required methods to implement for all subtypes of `AbstractCell` to define a new element
 
 ```@docs
-Ferrite.reference_vertices(::Ferrite.AbstractRefShape)
+Ferrite.get_node_ids
+```
+
+Please note that if your cell holds a single tuple called `nodes` to carry the node numbers, then
+this method will work automatically. Almost all functions work automatically with the information
+provided by the reference element.
```
KnutAM, line 16: Already mentioned in the docstring
```suggestion
```

docs/src/devdocs/elements.md
```diff
@@ -8,22 +8,35 @@ by the associated reference element.
...
+
+Please note that if you want to implement a custom element which does not have a reference shape,
+then the relevant functions below need to be dispatched.
```
KnutAM, line 19: I don't think this is possible, `AbstractCell{RT}` requires `RT<:AbstractRefShape`? 

```suggestion
```

src/Grid/grid.jl
```diff
@@ -131,13 +157,19 @@ reference_facets(::Type{<:AbstractRefShape})
 @inline reference_facets(refshape::Type{<:AbstractRefShape{2}}) = reference_edges(refshape)
 @inline reference_facets(refshape::Type{<:AbstractRefShape{3}}) = reference_faces(refshape)
 
+@inline reference_faces(::AbstractCell{refshape})    where refshape = reference_faces(refshape)
+@inline reference_edges(::AbstractCell{refshape})    where refshape = reference_edges(refshape)
+@inline reference_vertices(::AbstractCell{refshape}) where refshape = reference_vertices(refshape)
```
KnutAM, line 162: Use [here](https://github.com/Ferrite-FEM/Ferrite.jl/pull/889/files#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90L154-L171)?

src/Grid/grid.jl
```diff
@@ -131,13 +157,19 @@ reference_facets(::Type{<:AbstractRefShape})
...
+@inline reference_facets(::AbstractCell{refshape})   where refshape = reference_facets(refshape)
```
KnutAM, line 163: Missing test

docs/src/devdocs/elements.md
termi-official:
```
struct MyFancyDynamicCell
    nodes::Vector{Int}
end
```
?

KnutAM: But that won't fit in the grid which requires
```
struct MyFancyDynamicCell <: AbstractCell{RefShape}
    nodes::Vector{Int}
end
```
?

KnutAM: And I guess lot of the codebase assumes that the refshape can be determined only by the type, so not sure how an arbitrary polygon element would work. Do we already have an issue for this? Otherwise, that would be good.

termi-official: But I would leave this issue for later, as it does not have any priority for me.

src/Grid/grid.jl
KnutAM: Forgot about this one! 
I added test and this dispatch in #981. So no need to add test here if you're ok with 981 (this can be merged before or after, slight merge conflict but easy to resolve)

KnutAM: Resolved as 981 is merged

docs/src/devdocs/elements.md
termi-official: Should we then include a note that we require subtyping for our Grid? (Possibly separate PR)

src/Grid/grid.jl
termi-official: You mean update the linked code lines to use this? I think the code above is fine as it is tho. The functions here are mere convenience for cases where the access to the cell type might not that straight forward. :)

docs/src/devdocs/elements.md
KnutAM: I think you already included the requirement of an AbstractCell [here](https://github.com/Ferrite-FEM/Ferrite.jl/pull/889/files#diff-b331ab3d590056374ff36772c42e8ee13544e3ce67db0ab6a3369da27e8e6659R5)?

src/Grid/grid.jl
KnutAM: Sure, but when we have them I think it makes sense to use them internally as well? 
```diff
- function vertices(c::AbstractCell{RefShape}) where RefShape
+ function vertices(c::AbstractCell)
    ns = get_node_ids(c)
-   return map(i -> ns[i], reference_vertices(RefShape))
+   return map(i -> ns[i], reference_vertices(c))
end
```
Is kind of nicer too, and will automatically add test coverage for these methods...

termi-official:
Example to be reinserted separately.

Now to a full example. Let us reconstruct the local facets of a grid boundary with the described machinery to see how everything
fits together:
```julia
julia> function recompute_facetset_on_boundary(cells, global_facets)
           # Facets in 2D are edges and have two vertices
           d = Dict{NTuple{2, Int}, FacetIndex}()
           for (c, cell) in enumerate(cells) # c is global cell number
               # Grab all local facets
               local_facets = Ferrite.reference_facets(cell)
               for (f, facet) in enumerate(local_facets) # f is local facet number
                   # Translate into global nodes
                   global_facet = ntuple(i-> cell.nodes[facet[i]], length(facet))
                   d[global_facet] = FacetIndex(c, f)
               end
           end

           facets = Vector{FacetIndex}()
           for facet in global_facets
               # lookup the element, local facet combination for this facet
               push!(facets, d[facet])
           end

           return facets
       end

julia> recompute_facetset_on_boundary(cells, facets)
Vector{FacetIndex} with 2 elements:
  FacetIndex((2, 2))
  FacetIndex((4, 2))
```

Note that this example can be simplified by directly using `Ferrite.facets`:

```julia
julia> function recompute_facetset_on_boundary_alternative(cells, global_facets)
           # Facets in 2D are edges and have two vertices
           d = Dict{NTuple{2, Int}, FacetIndex}()
           for (c, cell) in enumerate(cells) # c is global cell number
               # Grab all local facets
               for (f, global_facet) in enumerate(Ferrite.facets(cell)) # f is local facet number
                   d[global_facet] = FacetIndex(c, f)
               end
           end

           facets = Vector{FacetIndex}()
           for facet in global_facets
               # lookup the element, local facet combination for this facet
               push!(facets, d[facet])
           end

           return facets
       end

julia> recompute_facetset_on_boundary_alternative(cells, facets)
Vector{FacetIndex} with 2 elements:
  FacetIndex((2, 2))
  FacetIndex((4, 2))
```

src/Grid/grid.jl
```diff
@@ -65,6 +75,20 @@ corresponds to the local index into this tuple.
...
+    reference_edges(::AbstractRefShape)
+    reference_edges(::AbstractCell)
+
+Returns a tuple of 2-tuples containing the ordered local node indices corresponding to
+the vertices that define an *oriented edge*.
+
+An *oriented edge* is an edge with the first node having the lowest local index and the other
+node the secon dindex.
+
+Note that the vertices are sufficient to define a face uniquely.
```
KnutAM, line 88:
```suggestion
Returns a tuple of 2-tuples containing the local node indices corresponding to
the vertices that define the edge. The ordering determines the positive direction of this edge.
```

src/Grid/grid.jl
```diff
@@ -56,6 +56,16 @@ nedges(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_edges(T))
 nfaces(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_faces(T))
 nfacets(  ::Type{T}) where {T <: AbstractRefShape} = length(reference_facets(T))
 
+
+"""
+    reference_vertices(::AbstractRefShape)
+    reference_vertices(::AbstractCell)
+
+Returns a tuple of integers containing the ordered local vertex indices corresponding to
+the corners or endpoints of an element.
```
KnutAM, line 65:
```suggestion
Returns a tuple of integers containing the local node indices corresponding to
the vertices (i.e. corners or endpoints) of the cell.
```

src/Grid/grid.jl
```diff
@@ -65,6 +75,20 @@ corresponds to the local index into this tuple.
...
 
+"""
+    reference_edges(::AbstractRefShape)
+    reference_edges(::AbstractCell)
+
+Returns a tuple of 2-tuples containing the ordered local node indices corresponding to
+the vertices that define an *oriented edge*.
+
+An *oriented edge* is an edge with the first node having the lowest local index and the other
+node the secon dindex.
```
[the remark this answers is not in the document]
KnutAM: Same for existing docs for faces, but this docstring isn't part of this PR so can be done elsewhere. However, I cannot understand the definition of an "oriented face" there...

src/Grid/grid.jl
```diff
@@ -78,6 +102,7 @@ edges(::AbstractCell)
 
 """
     reference_faces(::AbstractRefShape)
```
KnutAM, line 104:
```suggestion
    reference_faces(::Type{<:AbstractRefShape})
```

src/Grid/grid.jl
```diff
@@ -65,6 +75,20 @@ corresponds to the local index into this tuple.
...
+
+Returns a tuple of 2-tuples containing the ordered local node indices corresponding to
+the vertices that define an *oriented edge*.
+
+An *oriented edge* is an edge with the first node having the lowest local index and the other
+node the secon dindex.
+
+Note that the vertices are sufficient to define a face uniquely.
+"""
+reference_edges(::Union{AbstractRefShape, AbstractCell})
```
KnutAM, line 90:
```suggestion
reference_edges(::Union{Type{<:AbstractRefShape}, AbstractCell})
```

src/Grid/grid.jl
```diff
@@ -65,6 +75,20 @@ corresponds to the local index into this tuple.
 """
 vertices(::AbstractCell)
 
+"""
+    reference_edges(::AbstractRefShape)
```
KnutAM, line 79:
```suggestion
    reference_edges(::Type{<:AbstractRefShape})
```

src/Grid/grid.jl
```diff
@@ -56,6 +56,16 @@ nedges(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_edges(T))
...
 
+
+"""
+    reference_vertices(::AbstractRefShape)
+    reference_vertices(::AbstractCell)
+
+Returns a tuple of integers containing the ordered local vertex indices corresponding to
+the corners or endpoints of an element.
+"""
+reference_vertices(::Union{AbstractRefShape, AbstractCell})
```
KnutAM, line 67:
```suggestion
reference_vertices(::Union{Type{<:AbstractRefShape}, AbstractCell})
```

src/Grid/grid.jl
```diff
@@ -56,6 +56,16 @@ nedges(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_edges(T))
 nfaces(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_faces(T))
 nfacets(  ::Type{T}) where {T <: AbstractRefShape} = length(reference_facets(T))
 
+
+"""
+    reference_vertices(::AbstractRefShape)
```
KnutAM, line 61:
```suggestion
    reference_vertices(::Type{<:AbstractRefShape})
```

src/Grid/grid.jl
```diff
@@ -87,7 +112,7 @@ nodes spanning such that the normal to the face is pointing outwards.
 
 Note that the vertices are sufficient to define a face uniquely.
 """
-reference_faces(::AbstractRefShape)
+reference_faces(::Union{AbstractRefShape, AbstractCell})
```
KnutAM, line 115:
```suggestion
reference_faces(::Union{Type{<:AbstractRefShape}, AbstractCell})
```

docs/src/devdocs/reference_cells.md
```diff
@@ -15,3 +15,19 @@ Ferrite.RefTetrahedron
 Ferrite.RefHexahedron
 Ferrite.RefPrism
 ```
+
+### Required methods to implement for all subtypes of `AbstractRefShape` to define a new reference element
+
+```@docs
+Ferrite.geometric_interpolation(::Ferrite.AbstractRefShape)
```
KnutAM, line 22: I don't think this is possible? 
```suggestion
```

docs/src/devdocs/reference_cells.md
```diff
@@ -15,3 +15,19 @@ Ferrite.RefTetrahedron
...
+Ferrite.reference_vertices(::Ferrite.AbstractRefShape)
+Ferrite.reference_edges(::Ferrite.AbstractRefShape)
+Ferrite.reference_faces(::Ferrite.AbstractRefShape)
```
KnutAM, line 25:
```suggestion
Ferrite.reference_vertices(::Type{<:Ferrite.AbstractRefShape})
Ferrite.reference_edges(::Type{<:Ferrite.AbstractRefShape})
Ferrite.reference_faces(::Type{<:Ferrite.AbstractRefShape})
```

docs/src/devdocs/reference_cells.md
```diff
@@ -15,3 +15,19 @@ Ferrite.RefTetrahedron
...
+```
+
+which automatically defines
+
+
+```@docs
+Ferrite.reference_facets(::Ferrite.AbstractRefShape)
```
KnutAM, line 32:
```suggestion
Ferrite.reference_facets(:Type{<:Ferrite.AbstractRefShape})
```

KnutAM (commented):
Spotted a few more things when going through carefully

docs/src/topics/grid.md
```diff
@@ -78,10 +78,29 @@ Facet- and vertexsets are represented by `OrderedSet{<:BoundaryIndex}`, where `B
...
+
+The example shows a local edge ordering, defined as:
+
+```julia
+Ferrite.reference_facets(RefQuadrilateral) = ((1,2), (2,3), (3,4), (4,1))
+```
+
+Other facet definitions can be found in the src file `src/Grid/grid.jl` in the corresponding dispatches for [`reference_facets`](@ref Ferrite.reference_facets). Furthermorem you can query specific information about subentities via [`reference_vertices`](@ref Ferrite.reference_vertices), [`reference_edges`](@ref Ferrite.reference_edges) and [`reference_faces`](@ref Ferrite.reference_faces).
+
+
```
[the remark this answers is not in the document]
termi-official: Right, this is some merge oopsie. Let me fix that.

src/Grid/grid.jl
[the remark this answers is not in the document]
termi-official: xref https://github.com/Ferrite-FEM/Ferrite.jl/pull/689/files#r1394394569 and the other threads in the PR for discussion on this matter. We really should try to merge this.

But right, that is not the formal definition and it does not match master. It was ment to be a digestible one for users. Let me propose something different for the edge then.

The formal idea for the orientation of faces and edges is as follows. You have an ordered tuple (which is the Tuple type in Julia), which can be cyclically permutated and reversed. This set of nodes contains all admissible orientations for an edge/face. Hence the return value for `faces` and `edges` is contained in this set. This is in contrast to some "unoriented" face/edge, which is a unique representation, i.e. for us the return value of sortface/sortedge.

src/Grid/grid.jl
termi-official:
```suggestion
Returns a tuple of 2-tuples containing the ordered local node indices (corresponding to the vertices) that define an edge.

Note that the vertices are sufficient to define an edge uniquely.
```

docs/src/topics/grid.md
```diff
@@ -78,10 +78,29 @@ Facet- and vertexsets are represented by `OrderedSet{<:BoundaryIndex}`, where `B
...
 The highlighted facets, i.e. the two edges from node ID 3 to 6 and from 6 to 9, on the right hand side of our test mesh can now be described as
 
 ```julia
-boundary_facets = [(3, 6), (6, 9)]
+julia> boundary_facets = [
+           (3,6),
+           (6,9)
+       ]
```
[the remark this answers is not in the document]
termi-official: The diff to baseline is really just the rename from faces to boundary_facets of this line. So in https://github.com/Ferrite-FEM/Ferrite.jl/commit/6dfa0e46c63e8c6dcc466568dfcbf2555108df37 the collapse to a single line was an oopsie?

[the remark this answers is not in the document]
KnutAM: No, I did that on purpose, since it was just to state how it is defined, and I found that spending 4 lines on that part attracts more attention than it should. This could even have been done in the text since the `boundary_facets` variable isn't used later.

src/Grid/grid.jl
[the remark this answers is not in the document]
KnutAM: Yes, I rely on this ordering in #798 e.g. [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/714d67c483d292f8244b53d990e0ac9ae0df2c54/src/interpolations.jl#L1755-L1780), where I have to be careful to define the sign of the base functions positive or negative depending on the definition of the local edge direction, and during reinit! check if the global edge direction is the same or reversed, in which case the base function sign is flipped (in `apply_mapping!`). 

So my take-away if I don't miss anything, is that it doesn't matter how we define the local order, but it must match the defintions of the interpolation. 

I think users will find it a bit strange (even if I see the logic), that a sorted entity is not ordered, while the non-sorted one is :) But I still think that is the best wording we can use.

src/Grid/grid.jl
KnutAM: I think the phrase "up to its orientation" isn't so clear (could be since I'm not a native English speaker) (does it include the orientation or not?)

In general I'm not sure about the last sentence, since without the global node indices, the edge isn't uniquely defined? So doesn't this statement rather belong to `edges` and not `reference_edges`? 

I think we should write that `edges(c::AbstractCell) returns the global node indices corresponding to the local node indices of c's reference shape's vertices`, and in `edges` state that `The global node indices of the vertices are sufficient to determine both the edge and its direction uniquely`.

src/Grid/grid.jl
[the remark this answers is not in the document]
termi-official: Do we have the exact invariant to be fullfilled between the canonical geometric interpolation of a cell and the reference cell documented (and tested) somewhere? (xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/558)

docs/src/topics/grid.md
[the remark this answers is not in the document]
termi-official: Indeed. However, this is an artifact of moving the reference geometry topic out of the grid topic.

src/Grid/grid.jl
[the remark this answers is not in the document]
KnutAM: I don't think we even define the coordinates of the reference shapes directly anywhere, they are all implicitly defined via the interpolations defined on that reference cell. 

I'm not sure what assumptions would be broken if defining a new interpolation with something like 
```julia
MyTestIP{RefShape, order, IP<:ScalarInterpolation{RefShape, order}} <: ScalarInterpolation{RefShape, order}
    ip::IP
end
getnbasefunctions(ip) = getnbasefunctions(ip.ip)
reference_shape_value(ip::MyTestIP, x, i) = reference_shape_value(ip.ip, x, getnbasefunctions(ip) + 1 - i)
```
To use a different ordering (also updating the `dof_indices` functions). 
Would be interesting to test...

src/Grid/grid.jl
KnutAM: > Do we have some reference in the docs what positive/negative orientation means in this context?

No, since that PR isn't merged I don't think so. But sure, the definition is the same, and my logic to how this should work in Ferrite (with #798) is that 
* The dof-distriubtion takes care of the "dof location" for each base functions. E.g. if the edge is reversed, we also need to reverse the positions (this seemed to work out of the box which was nice, but I guess expected since this is the same problem for standard nodal interoplations). 
* The `reinit!` takes care of flipping the sign of the vectorial base functions if the edge is reversed. Similarly for faces, we need to determine if the face is positive or negative, and flip the sign accordingly. But the rotation doesn't matter for `reinit!`, as the location has already been taken care of by the dof-distribution. 


Another side-note, now we should rename this to `EdgeOrientationInfo` after redefining edges :)

KnutAM: > Maybe the missing information here is that my wording is "oriented" for the "not sorted" entities as they are. Hence the reference edge would always be oriented, but not necessarily unique (and we also cannot determine this with the information at hand). The physical edge (queried via edges would then still be oriented and we can query details on its orientation, which allows us to define some new entity, which is still an edge through sortedge. I am sure that we are missing this important piece of information in the doc strings, which is why I wanted to add it.

As I wrote above, I think "ordered" => "not sorted" and "sorted" => "unique entity description" is clear. I would rather favor adding this description to `edges` and `faces`, rather than `reference_edges` and `reference_faces`, since that is where we have this information. Then it should be explained that `reference_*` provide the local vertices that are used to obtain this information from cells sharing the same reference shape.

termi-official: Okay, I put a bit more time into this and my proposal is as follows. In this PR we do
```suggestion
Returns a tuple of 2-tuples containing the ordered local node indices (corresponding to the vertices) that define an edge.
```
and, in a separate PR, we add a subsection on the orientation stuff into the Grid topic guide together with renaming SurfaceOrientation and PathOrientationInfo to FaceOrientationInfo and EdgeOrientationInfo. The orientation API should then be made be public, but marked experimental for now, as users could use this stuff for custom dof handlers, custom elements and custom processing routines.

KnutAM: Great! 

> In this PR we do

The above suggested change for `reference_edges` is good. 

Could do the equivalent change `reference_faces` here. 
For `reference_facets`, I guess we should just refer to `reference_edges` and `reference_faces`? Something like 
```
Returns a tuple of tuples which define a facet, see `reference_edges` and `reference_faces` for further details.
``` 

> The orientation API should then be made be public

Not sure. I thought `SurfaceOrientation` isn't properly working yet? What's the advantage of making public but experimental over simply keeping in devdocs? But anyways, let's discuss that in the relevant (at minimum, renaming) pr!

termi-official: I can implement SurfaceOrientation if you need it. This is the easy part, no worries. The "hard" part is high order 3D ansatz functions, because you need to think about the permutation map for all combinations of admissible surface orientations.

KnutAM: I don't need it now at least, I also don't see a need to rush this to public api as long as it is documented in devdocs.

src/Grid/grid.jl
```diff
@@ -120,26 +135,32 @@ facets(::AbstractCell)
...
+@inline reference_faces(::AbstractCell{refshape})    where refshape = reference_faces(refshape)
+@inline reference_edges(::AbstractCell{refshape})    where refshape = reference_edges(refshape)
+@inline reference_vertices(::AbstractCell{refshape}) where refshape = reference_vertices(refshape)
+@inline reference_facets(::AbstractCell{refshape})   where refshape = reference_facets(refshape)
+
 """
-    geometric_interpolation(::AbstractCell)
-    geometric_interpolation(::Type{<:AbstractCell})
+    geometric_interpolation(::AbstractCell)::ScalarInterpolation
+    geometric_interpolation(::Type{AbstractCell})::ScalarInterpolation
```
KnutAM, line 158: Residue from merging I guess
```suggestion
    geometric_interpolation(::Type{<:AbstractCell})::ScalarInterpolation
```

src/Grid/grid.jl
```diff
@@ -120,26 +135,32 @@ facets(::AbstractCell)
...
 
 Each `AbstractCell` type has a unique geometric interpolation describing its geometry.
 This function returns that interpolation, which is always a scalar interpolation.
 """
-geometric_interpolation(cell::AbstractCell) = geometric_interpolation(typeof(cell))
+geometric_interpolation(::T) where T <: AbstractCell = geometric_interpolation(T)
```
KnutAM, line 163: Residue from merging I guess
```suggestion
geometric_interpolation(cell::AbstractCell) = geometric_interpolation(typeof(cell))
```

KnutAM (approved):
Nice!

A lot of discussion for a small docs PR :) 
But I think we ended up with some quite important improvements / clarifications in the end!

docs/src/devdocs/elements.md
```diff
@@ -2,24 +2,39 @@
...
 
 ### Required methods to implement for all subtypes of `AbstractCell` to define a new element
 
 ```@docs
+Ferrite.get_node_ids
 Ferrite.vertices(::Ferrite.AbstractCell)
+Ferrite.reference_edges(::Ferrite.AbstractRefShape)
 Ferrite.edges(::Ferrite.AbstractCell)
 Ferrite.reference_faces(::Ferrite.AbstractRefShape)
 Ferrite.faces(::Ferrite.AbstractCell)
```
KnutAM, line 16:
```suggestion
```

PR #890, opened 2024-03-06
[X-PR] p4est/p8est face transformation fix

termi-official:
To not mess around with the PR's state here is a separate PR which proposes a fix for the face transformation. Probably some downstream implementations will break and will require further intervention.

# TODO

* [x] Check creategrid again and add errors for things that still go wrong
* [x] Make codecov happy by adding more tests
* [x] Proper citations

### DIFF
--- a/test/test_p4est.jl
+++ b/test/test_p4est.jl
@@ -176,7 +176,7 @@ end
     @test Ferrite.transform_face(adaptive_grid, FaceIndex(4,3), o) == OctantBWG(0,(0,-8))
 
     (grid_new, hnodes) = Ferrite.creategrid(adaptive_grid)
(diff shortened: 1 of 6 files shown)


test/test_p4est.jl
```diff
@@ -247,6 +325,24 @@ end
             @test all(Ferrite.morton.(tree.leaves,l,5) == collect(1:2^(3*l)))
         end
     end
+
+    # Reproducer test for Fig.3
```
koehlerson, line 329:
```suggestion
    # Reproducer test for Fig.3 BWG 11
```
since there are already parts which reference the other algorithmic paper. let me know if I can review

Issue #897, opened 2024-04-03
Topology optimization: more robust


MaxVZIKM:
It was observed, that in the code example for topology optimization the update of the densities only depends on \eta. Correct is to multiply eta with the iterations needed for stability n_j. Otherwise, a behavior similar to checkerboarding may be observed for larger 3D problems. The density simply is updated too fast. Thus, with this issue the example is more robust.

PR #898, opened 2024-04-03
Update topology_optimization.jl

MaxVZIKM:
Multiplication of the number of iterations needed for stability (n_j) with the viscosity (\eta) within the function update_density. If n_j is bigger than 1, behavior similar to checkerboarding could be observed otherwise. This solves issue #897.


KristofferC:
The animation of the example should probably be updated to match this as well then?

koehlerson:
I guess there's something missing in the theory section of the example then?

PR #905, opened 2024-04-25
Explicitly import all names

fredrikekre:


src/Ferrite.jl
```diff
@@ -1,14 +1,27 @@
 module Ferrite
-using Reexport
+
+using Reexport: @reexport
 @reexport using Tensors
```
KnutAM, line 4: Doesn't this cause implicit imports from Tensors? (the WriteVTK reexport is removed in #692 )

src/Ferrite.jl
```diff
@@ -1,14 +1,27 @@
...
 @reexport using WriteVTK
 
+using Base:
+    @propagate_inbounds
+using EnumX:
+    EnumX, @enumx
 using LinearAlgebra
```
termi-official, line 11: Since I cannot find it in the Julia docs let me quickly ask here. Is the ":" optional?

fredrikekre: No, just mistake:
```suggestion
using LinearAlgebra:
```

src/Ferrite.jl
fredrikekre: Yes, but usage of them would be flagged by https://github.com/ericphanson/ExplicitImports.jl. I didn't add a check to CI because there are some false detections (https://github.com/ericphanson/ExplicitImports.jl/issues/34, https://github.com/ericphanson/ExplicitImports.jl/issues/35, https://github.com/ericphanson/ExplicitImports.jl/issues/36).

I wonder if we should stop reexporting Tensors though. I think in most examples we do `using Ferrite, Tensors` but it is pretty fundamental, and I don't see much of a risk of reexporting since we are "in control" of that package too.

KnutAM: Aha, ok. 
I think not reexporting is fine (even good, because Tensors isn't a "base" package IMO for Ferrite, but a standalone thing).

On the other hand, that is perhaps a non-necessary breaking change since we are in control as you say (even if we did have a problem with `curl` (#442), but if CI is added that would catch such cases then AFAIU).

PR #906, opened 2024-04-26
Explicitly import all names 2

KnutAM:
xref #905 

Just noticed these when merging #692 

Most important was the `import Base: push!`, but figured preferences should also be imported in `src/Ferrite.jl` instead of in the utils

I guess `using Preferences: Preferences` is equivalent to `import Preferences`, but I changed it just to be consistent with the remaining statements.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -1,8 +1,7 @@
 Base.@deprecate_binding DirichletBoundaryConditions ConstraintHandler
 Base.@deprecate_binding DirichletBoundaryCondition Dirichlet
 
-import Base: push!
-@deprecate push!(dh::AbstractDofHandler, args...) add!(dh, args...)
+@deprecate Base.push!(dh::AbstractDofHandler, args...) add!(dh, args...)
 
 @deprecate vertices(ip::Interpolation) vertexdof_indices(ip) false
 @deprecate faces(ip::Interpolation) facedof_indices(ip) false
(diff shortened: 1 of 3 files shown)


src/deprecations.jl
```diff
@@ -1,8 +1,7 @@
 Base.@deprecate_binding DirichletBoundaryConditions ConstraintHandler
 Base.@deprecate_binding DirichletBoundaryCondition Dirichlet
 
-import Base: push!
-@deprecate push!(dh::AbstractDofHandler, args...) add!(dh, args...)
+@deprecate Base.push!(dh::AbstractDofHandler, args...) add!(dh, args...) false
```
fredrikekre, line 4: I have a vague recollection that this doesn't work which is why there was an explicit import before.

KnutAM: Yes, seems like the required export_old is not available on 1.6
I guess we could change to
``` 
function Base.push!(dh::AbstractDofHandler, args...) 
    @warn "push!(dh::DofHandler, args...) is deprecated, use add!(dh, args...) instead" maxlog=1
    return add!(dh, args...)
end
```
what do you think?

fredrikekre: What was the problem with the existing code? I think for 1.0 we will just make all of these errors and then we can use the `Base.push!` pattern.

KnutAM: > What was the problem with the existing code?

Just that `push!` could be overloaded anywhere in the codebase without explicitly calling `Base.push!` (such a well-known function, so shouldn't be a problem, but AFAIU that was the reason for preferring e.g. `using Base: push!` over `import Base: push!`?

But if we remove it, then it doesn't matter and I'll close this PR :)

PR #907, opened 2024-04-29
Update constraints.md

koehlerson:


KnutAM (approved):
LGTM - I got the same failure in #692 on Mac. 
Would be good if someone with a mac could check it locally.

Issue #899, opened 2024-04-10
Coupling shells to solid faces


lijas:
It is quite uncommon, but sometimes you want to add a "skin" of shell elements on the face of a solid. One example of this which I have encountered, is to model sandwich composites with shell elements on the top and bottom surfaces of a solid foam core.

At the moment, it is not possible to this for higher order shells/solids. The reason for this is because shells (e.g Quadrilaterals) do not share a consistent definition of faces and edges with solids (e.g hexaherons). Note that it works for linear shells/solids since dofs are only distributed on nodes in this case.

MWE:
```
# Ferrite grid: Beam attached to face
# 3 ____ 4  4
# |      |  |   
# |      |  | (Beam attached to face)    
# 1 ____ 2  2

dim = 2
grid = generate_grid(Quadrilateral, (1,1))
line1 = Line((2,4))
grid = Grid([grid.cells[1], line1], grid.nodes)

order = 2
ip_quads = Lagrange{RefQuadrilateral, order}()#^dim
ip_shell = Lagrange{RefLine, order}()

dh = DofHandler(grid)
sdh_quads = SubDofHandler(dh, Set(1))
add!(sdh_quads, :u, ip_quads)
sdh_coh = SubDofHandler(dh, Set(2))
add!(sdh_coh, :u, ip_shell)
close!(dh)

dofsquad = zeros(Int, ndofs_per_cell(dh, 1))
dofsbeam = zeros(Int, ndofs_per_cell(dh, 2))

celldofs!(dofsquad, dh, 1)
celldofs!(dofsbeam, dh, 2)
@show dofsbeam # dofsbeam = [2, 3, 10]
```
As you can see, the dofs on the shell (or beam since this in 2d), contains a 10th dofs, but it should be the 6th dofs shared with the quad. Note that #789 would solve this (however with some annoying breaking changes)

termi-official:
Why does the mentioned PR solve the issue? We would have to make the "1D cell dofs" also edge dofs to make it work here, right?

lijas:
I dont really understand what you mean by "1D cell dofs", but a line element would basically be considered as an edge, so if a Line element are attached to the facet (edge) of a quad, the dofhandler would distribute the dofs correctly

termi-official:
Ah, sorry, I missed part of the diff. (This one:)
```patch
-facedof_indices(::Lagrange{RefLine,2}) = ((1,), (2,))
[...]

Issue #843, opened 2023-11-14
Revive mixed dimensional relations in topology


termi-official:
The initial version of our topology was able to handle mixed dimensional constraints, but it was removed to obtain type stability in https://github.com/Ferrite-FEM/Ferrite.jl/pull/753 which made the topology construction a bit faster . The idea is for the recovery of this feature basically 6 additional collections for neighborhoods:
* 3D -> 2D
* 2D -> 3D
* 3D -> 1D
* 1D -> 3D
* 2D -> 1D
* 1D -> 2D
And to have queries by co-dimension which directly works on these arrays. This should be enough for type stability.

termi-official:
I think with the new face and edge API we can query mixed-dimensional relations in a type stable way with the queries on FaceIndex and EdgeIndex.

Minimal test:

```julia
using Ferrite
cells = [
    Hexahedron((1, 2, 3, 4, 5, 6, 7, 8)),
    Quadrilateral((3, 2, 9, 10)),
    ]
nodes = [Node(coord) for coord in zeros(Vec{3,Float64}, 10)]
grid = Grid(cells, nodes)

topo = ExclusiveTopology(grid)
@assert any(.!isempty.(topo.edge_edge_neighbor))
```

termi-official:
Another tricky corner case to be handled is to get non-trivial vertex-vertex neighborhoods right.
```
using Ferrite
grid = generate_grid(Hexahedron, (2, 1, 1), Vec{3}((-2.5, -2.5, -1.0)), Vec{3}((2.5, 2.5, 1.0)))
mixed_grid = Grid([grid.cells..., Line((1,8))], grid.nodes, cellsets=Dict("A" => OrderedSet([1]), "B" => OrderedSet([2]), "C" => OrderedSet([3])))
topo = ExclusiveTopology(mixed_grid)
```

Issue #595, opened 2023-02-03
Extract boundary entities


termi-official:
Following the discussion in #565 it is helpful to add utilities to extract boundary entities (faces, edges, ...). Boundary faces can be identified as a function of the grid and a topology. So maybe something like `addboundaryfaceset!(grid, topology, "bottom" x -> norm(x[3]) ≈ 0.0);` makes sense.

graphitical:
Hi Dennis, the `topology` you list here, would that be an `ExclusiveTopology` or something else?

termi-official:
Yes. Currently only the `ExclusiveTopology` is implemented in Ferrite. Note that the topology interface is still rather experimental. The topology is decoupled from the mesh, because this allows to have different neighborhood construction and search algorithms, depending on the specific use-case of a user.

graphitical:
Thanks. Would the topology argument you list originally be necessary? I'm a bit lost on how all the pieces work together. It seems like `grid` and the defining boundary function should contain all the information necessary to create a boundary.

termi-official:
The topology here is just ment to help to identify which entities are not in the interior of the grid. Note that the grid has basically all the information that is also in the topology. We precompute topological information from the grid to speed up queries like for example "Who is my neighbor on face 1", which is stored in such topology data structures. This makes boundary queries easy, because a boundary face can be defined as a face without neighbor face in a suitable topology. A solution which does not use any topology data structure is also fine and possible.

graphitical:
I think I see. 
So you were originally suggesting creating a new type of topology (i.e., `ExteriorTopology`) which would contain all the topological information for all the exterior entities.  

The intent of this would be to reduce the search space for checking if a node satisfies the query function.

Is that correct?

This is my understanding because `ExclusiveTopology` seems to contain topological information for the **entire** grid so that obviously contains a lot of points we don't want to check against.

koehlerson:
I don't think this is exactly what Dennis meant. You can just use the `ExclusiveTopology` and query that by 

```
julia> grid = generate_grid(Quadrilateral,(100,100))
Grid{2, Quadrilateral, Float64} with 10000 Quadrilateral cells and 10201 nodes

julia> topology = ExclusiveTopology(grid);

julia> findall(isempty,topology.face_neighbor)
400-element Vector{CartesianIndex{2}}:
 CartesianIndex(1, 1)
 CartesianIndex(2, 1)
 CartesianIndex(3, 1)
[...]

PR #914, opened 2024-05-14
Proper codim and dim entities

### LINKED ISSUE #901: Adopt proper names for entities with different dimensions and codimensions
Many issues seem to stem from not having a clear distinction between dimension and codimension. 

Adopting the correct names (ref [defelement](https://defelement.com/ciarlet.html)) in Ferrite would be a major change, especially since users would have to change from `FaceValues` / `faceset` / `FaceIndex` to `FacetValues` / `facetset` / `FacetIndex`. 

However, it seems to have the benefit of making the code more understandable and extensible as we add more advanced features. It is notable that at least `fenics` adopts this defintion. 

Ref discussions in #899, #767, #789 (and probably more). 

Most user calls would occur during construction, so that a regular warning / upgrade error would be feasible for standard users. IMO this should happen before releasing 1.0 or not at all. 

Edit: I earlier mistakenly assumed these names were part of the Ciarlet (1978) definitions, but couldn't find any references to this, and AFAIU now that only refers to the triplet $(\mathcal{R},\ \mathcal{V}, \mathcal{L})$ for the reference shape, interpolation space, and unknowns/nodal values which we implicitly already have. Changed the name to reflect this.

KnutAM:
As agreed with @lijas, this PR continues from his work in #789, but introduces facets on the user-facing side. See #789 for an extended discussion. 

The defintions in this PR are
* `face`: Always 2d entity
* `edge`: Always 1d entity
* `vertex`: Always 0d entity
* `facet`: Always codim 1 entity relative the cell's reference dimension.

With this PR, all code with grids with a single reference dimension, the code changes from using `face` to using `facet`. `FacetIndex` and `FacetValues` are introduced (this also includes embedded elements).

## FEValues
Renaming
`FaceValues` -> `FacetValues`
`FaceIterator` -> `FacetIterator`
`FaceQuadratureRule` -> `FacetQuadratureRule`

## Grid sets
An important change is then that the `Grid` struct looses `facesets` and `edgesets`, which are replaced by `facetsets`. The removal is motivated by avoiding confusion between `facetset` and `faceset`, and for current examples `facesets` and `edgesets` are not needed. To not remove the options, the following functions are introduced 
* `create_vertexset`, `create_edgeset`, `create_faceset`, and `create_facetset`: `(::AbstractGrid, ::Function; all=true)`
* `create_boundaryvertexset`, `create_boundaryedgeset`, `create_boundaryfaceset`, and `create_boundaryfacetset`: `(::AbstractGrid, ::ExclusiveTopology, ::Function; all=true)`

So that sets can be created outside the grid. (`add*!` functions are available for `vertex` and `facet`). 

To avoid too many issues in dependent packages, `addfaceset!` automatically translates the given `Set{FaceIndex}` to `Set{FacetIndex}` and adds it, warning the users to change their code. Similar for creating a `Grid` and giving a `faceset` but not a `facetset`. 

## Topology
`get_facet_facet_neighborhood` is defined, which gives the neighborhood depending on the reference dimension of the grid. If the grid has cells with different reference dimensions, an `ArgumentError` is thrown, asking the users to access the `vertex_vertex_neighbor`, `edge_edge_neighbor`, or `face_face_neighbor` fields explicitly. 

`getneighborhood` can be querried with `FacetIndex` only if all cells have the same reference dimension. If not, an `ArgumentError` is thrown, asking users to use `VertexIndex`, `EdgeIndex`, or `FaceIndex` explicitly. 

## Extract Boundary entities (#595)
For the code in #595, I struggled a bit to update it, and found it easier to rewrite using the `boundaryfunction` instead of `@eval` function generation. This also means that `getfaceedges`, `getfacevertices`, `getedgevertices` are not used, and these are only in devdocs. So I commented out these for now along with the tests. Let me know if these should be kept (@AbdAlazezAhmed / @termi-official)?

## Breaking dependent packages
`FerriteGmsh` and `FerriteMeshParser` break with this PR. The computational_homogenization and porous media examples include the required method redefinitions that should be included there. In addition, these packages should update to not use `FaceIndex`. Once I get feedback on this PR, I'll add PRs to those for updating them. If desired, supporting both the old and new version will be possible. 

This should close #901 and #899. 
Not tested, but at minimum it simplifies solving #843 without adding new fields to `ExclusiveTopology`. 

## TODO
- [ ] Open PRs in `FerriteGmsh.jl` and `FerriteMeshParser.jl`
- [x] Fix formatting: https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#discussion_r1604508540
- [ ] Fix test coverage / delete unused methods
- [x] Fix https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#discussion_r1604685335
- [x] Move definitions: https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#discussion_r1604697572
- [x] Fix issue with unordered dofs on faces, https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#discussion_r1605583870


termi-official:
> ## Topology
> 
> `get_facet_facet_neighborhood` is defined, which gives the neighborhood depending on the reference dimension of the grid. If the grid has cells with different reference dimensions, an `ArgumentError` is thrown, asking the users to access the `vertex_vertex_neighbor`, `edge_edge_neighbor`, or `face_face_neighbor` fields explicitly.
> 
> `getneighborhood` can be querried with `FacetIndex` only if all cells have the same reference dimension. If not, an `ArgumentError` is thrown, asking users to use `VertexIndex`, `EdgeIndex`, or `FaceIndex` explicitly.

That should fix the type stability issue.

> ## Extract Boundary entities (#595)
> 
> For the code in #595, I struggled a bit to update it, and found it easier to rewrite using the `boundaryfunction` instead of `@eval` function generation. This also means that `getfaceedges`, `getfacevertices`, `getedgevertices` are not used, and these are only in devdocs. So I commented out these for now along with the tests. Let me know if these should be kept (@AbdAlazezAhmed / @termi-official)?

We use something similar in the p4est PR. However, we need to iterate over the tests as you said.

> ## Breaking dependent packages
> 
> `FerriteGmsh` and `FerriteMeshParser` break with this PR. The computational_homogenization and porous media examples include the required method redefinitions that should be included there. In addition, these packages should update to not use `FaceIndex`. Once I get feedback on this PR, I'll add PRs to those for updating them. If desired, supporting both the old and new version will be possible.
> 

FerriteViz and FerriteDistributed will also break. To fix the latter I will need more time. xref https://github.com/Ferrite-FEM/FerriteDistributed.jl/pull/39

KnutAM:
> FerriteViz and FerriteDistributed will also break.

Thanks for pointing out. I hope that for at least FerriteDistributed this PR will make things simpler!

> We use something similar in the p4est PR. However, we need to iterate over the tests as you said.

Thanks for the heads-up. I couldn't find that, could you point me to where?

fredrikekre:
Is #789 supposed to be merged first?

KnutAM:
> Is https://github.com/Ferrite-FEM/Ferrite.jl/pull/789 supposed to be merged first?

Would be ok for me, but sounds like double work since tests aren't passing in #789 yet. 
This PR follows our discussions from a few weeks ago and reverts a few changes from 789. 
@lijas and I decided that I should give that implementation a try to see how it works out. IMO it looks good considering that tests are passing, and my (biased) view is that this is a good way forward.

termi-official:
> > FerriteViz and FerriteDistributed will also break.
> 
> Thanks for pointing out. I hope that for at least FerriteDistributed this PR will make things simpler!
> 

I think it will help both packages in the long run. :) 

> > We use something similar in the p4est PR. However, we need to iterate over the tests as you said.
> 
> Thanks for the heads-up. I couldn't find that, could you point me to where?

Essentially you can find the baseline topology logic for a hypercube (d=2,3 only) here https://github.com/Ferrite-FEM/Ferrite.jl/blob/47c966ef27bb7c87c7bcce2a8fd79a065de07040/src/Adaptivity/BWG.jl#L1791-L1876

src/Grid/grid.jl
```diff
@@ -355,7 +422,30 @@ This function takes the local vertex representation (a `VertexIndex`) and looks
...
+@inline getdim(g::AbstractGrid) = getsdim(g) # TODO: Deprecate
+
+"""
+    get_reference_dimensionality(grid::AbstractGrid) -> Union{Int, Symbol}
+
+Get information about the reference dimensions of the cells in the grid. 
+If all cells have the same reference dimension, `rdim::Int` is returned. 
+Otherwise, the `Symbol` `:mixed` is returned indicating a mixed-rdimensionality grid.
+"""
+get_reference_dimensionality(g::AbstractGrid) = _get_reference_dimensionality(getcells(g))
```
lijas, line 435: `getrdim() `?

src/Dofs/ConstraintHandler.jl
```diff
@@ -906,12 +906,12 @@ function filter_dbc_set(grid::AbstractGrid, fhset::AbstractSet{Int}, dbcset::Abs
 end
 
 struct PeriodicFacePair
-    mirror::FaceIndex
-    image::FaceIndex
+    mirror::Tuple{Int,Int}
```
lijas, line 909: I guess these can be `FacetIndex` now (was not possible in #789)

lijas (commented):
Nice work! I think there are both benefits and drawbacks with #789 and this PR, but I definitely see the appeal with this approach. So lets go with this PR over #789.

@termi-official or @koehlerson, do you think you this will work nicely with your adaptive stuff? Would it make sense to try this PR with your p4est stuff to see if you run in to any problems, before merging?

One opinion I have, which I might be in the minority of having, is to use  `boundary` instead of `facet` (e.g `getboundaryset(grid, "right")`, `BoundaryValues`). I think it is more natural/common to use `boundary` (we typically say boundary-integrals for both 2d and 3d problems), where as I never heard anyone use facets until recently.

src/Grid/grid.jl
KnutAM: Would be ok. Had that first, then `getrdims()::Tuple` since it could be different. But never needed the information about which rdims are in the grid if more than one, so ended with returning `:mixed` for that case. I didn't want to error on :mixed, and calling `rdim` getting a Symbol with just the information was confusing to me (but potentially only to me).

fredrikekre:
> One opinion I have, which I might be in the minority of having, is to use boundary instead of facet (e.g getboundaryset(grid, "right"), BoundaryValues). I think it is more natural/common to use boundary (we typically say boundary-integrals for both 2d and 3d problems), where as I never heard anyone use facets until recently.

I think in most cases in the PR where "facet" is used in the PR it is for a single element. I guess a boundary is a bit of a higher term, i.e. a boundary consist of many (element) facets. In this case it could make sense to use `boundaryset` (although it would just be boundary then I guess since it is a set of facets...).

KnutAM:
> Essentially you can find the baseline topology logic for a hypercube (d=2,3 only) here 

Thanks! Aha, so the current functions are not used, then I think it makes sense to remove them since they become obsolete with this PR and are not public.

src/Dofs/DofHandler.jl
```diff
@@ -773,8 +757,33 @@ function sortface_fast(face::Tuple{Int,Int,Int,Int})
...
+    sortfacet_fast(facet::NTuple{N, Int})
+
+Returns the unique representation of the `facet` by sorting its node indices. 
+Dispatches on `sortedges_fast` or `sortfaces_fast` depending on `N`
+"""
+function sortfacet_fast end 
+
+# Vertex
+sortfacet(facet::Tuple{Int}) = facet
+sortfacet_fast(facet::Tuple{Int}) = (facet, nothing)
```
koehlerson, line 778: was this covered bevor in tests when it was named `sortface_fast` ? I'm just wondering about the codecov complains

src/Grid/grid.jl
```diff
@@ -47,6 +47,8 @@ getrefshape(::AbstractCell{refshape}) where refshape = refshape
 nvertices(c::AbstractCell) = length(vertices(c))
 nedges(   c::AbstractCell) = length(edges(c))
 nfaces(   c::AbstractCell) = length(faces(c))
+nvertices(::Type{T}) where {T <: AbstractRefShape} = length(reference_vertices(T))
+nedges(   ::Type{T}) where {T <: AbstractRefShape} = length(reference_edges(T))
```
koehlerson, line 51: seems to be missing in the tests according to codecov

src/Grid/grid.jl
```diff
@@ -898,3 +972,20 @@ function OrientationInfo(surface::NTuple{N, Int}) where N
     end
     return OrientationInfo(flipped, shift_index)
 end
+
+# Dispatches for facets
+
+@inline facets(c::AbstractCell{<:AbstractRefShape{1}}) = map(i -> (i,), vertices(c)) # facet always tuple of tuple
+@inline facets(c::AbstractCell{<:AbstractRefShape{2}}) = edges(c)
+@inline facets(c::AbstractCell{<:AbstractRefShape{3}}) = faces(c)
```
koehlerson, line 980: should be moved directly under or above the `edge`,`face`,`vertices` definition of `AbstractCell` imho

src/Grid/topology.jl
```diff
@@ -70,9 +70,22 @@ mutable struct ExclusiveTopology <: AbstractTopology
     edge_edge_neighbor::Matrix{EntityNeighborhood{EdgeIndex}}
     # lazy constructed face topology
     face_skeleton::Union{Vector{FaceIndex}, Nothing}
+    edge_skeleton::Union{Vector{EdgeIndex}, Nothing}
+    vertex_skeleton::Union{Vector{VertexIndex}, Nothing}
```
koehlerson, line 74: I don't think a vertex skeleton is useful. It's jused the unique nodes - higher order nodes, but I guess that's the trade off for one dimension if one wants to do an integral over the boundary between two elements

koehlerson (commented):
+- missing tests if codecov is right

docs/src/literate-tutorials/plasticity.jl
```diff
@@ -139,13 +139,13 @@ end
 function create_values(interpolation)
     ## setup quadrature rules
     qr      = QuadratureRule{RefTetrahedron}(2)
-    face_qr = FaceQuadratureRule{RefTetrahedron}(3)
+    facet_qr = FacetQuadratureRule{RefTetrahedron}(3)
 
-    ## cell and facevalues for u
+    ## Cell and FacetValues for u
```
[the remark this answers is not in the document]
KnutAM: Thanks, find-replace error :)

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -193,6 +193,21 @@ using Test #src
 # the [`FerriteGmsh`](https://github.com/Ferrite-FEM/FerriteGmsh.jl) package:
 
 using FerriteGmsh
+
+# Overload for specific elements in this tutorial.
```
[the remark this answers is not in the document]
KnutAM: My plan was to make a pr which will support both 0.3 and master beforehand instead, if you agree. Since it is nice with a FerriteGmsh that works with both, and avoid having to checkout FerriteGmsh.jl with a specific version for working with Ferrite#master.

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -193,6 +193,21 @@ using Test #src
...
+function FerriteGmsh.tofacesets(boundarydict::Dict{String,Vector}, elements::Vector{Triangle})
+    faces = Ferrite.facets.(elements)
+    facesets = Dict{String,Set{FaceIndex}}()
+    for (boundaryname, boundaryfaces) in boundarydict
+        facesettuple = Set{FaceIndex}()
```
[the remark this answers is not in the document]
KnutAM: This is just a workaround since `Grid(cells, nodes; facesets)` works with a warning if `facetset` is not given. In FerriteGmsh this would be different. 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/f8c67406bffa90afcaa92817fbdea7d161d11850/src/Grid/grid.jl#L390-L398

src/Dofs/DofHandler.jl
KnutAM: Good spot, this is wrong. 
Functions definitions should be reversed.

src/Grid/topology.jl
KnutAM: Yes, it is used by e.g. the `InterfaceIterator` in 1d.

src/Dofs/DofHandler.jl
KnutAM: I guess, because `sortface` and `sortedge` is used in the dof distribution.
The only place we use `sortfacet` is https://github.com/Ferrite-FEM/Ferrite.jl/blob/f8c67406bffa90afcaa92817fbdea7d161d11850/src/Dofs/ConstraintHandler.jl#L1355-L1368

KnutAM: Looks like only `sortfacet_fast` will be required actually...

KnutAM:
I'm considering defining generic functions for `<entity>` =  `faces`, `edges`, and `vertices` using the `reference_<entity>` functions, see start here. 
https://github.com/Ferrite-FEM/Ferrite.jl/compare/kam/codim...kam/codim_refentity?expand=1

Current master: https://github.com/Ferrite-FEM/Ferrite.jl/blob/9fd9d34cf4f4200c6bcff4b68bfa9cc9873ed893/src/Grid/grid.jl#L100-L104 
and
https://github.com/Ferrite-FEM/Ferrite.jl/blob/9fd9d34cf4f4200c6bcff4b68bfa9cc9873ed893/src/Grid/grid.jl#L190-L196

Current PR hard codes this: https://github.com/Ferrite-FEM/Ferrite.jl/blob/6fc8fc479f381989585847b3b9bba282d044347b/src/Grid/grid.jl#L255-L269

Left separate to avoid creating another large diff since some are currently reviewing. Please object if someone doesn't like this. Otherwise, I will finish that and merge in during the weekend.

src/Grid/utils.jl
```diff
@@ -1,348 +1,3 @@
-"""
```
fredrikekre, line 1: Where did this file go?

src/Dofs/DofHandler.jl
```diff
@@ -718,9 +705,6 @@ For more details we refer to [1] as we follow the methodology described therein.
     !!!TODO Investigate if we can somehow pass the interpolation into this function in a typestable way.
 """
 @inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int,Int}, orientation::SurfaceOrientationInfo, adjust_during_distribution::Bool)
-    if adjust_during_distribution && length(dofs) > 1
-        error("Dof distribution for interpolations with multiple dofs per face not implemented yet.")
```
fredrikekre, line 722: Is this check moved elsewhere?

fredrikekre (approved):
Very nice. I wonder if it would be worth making all renames into  it's own commit so that it can be ignored in git blame (https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt). Perhaps it wouldn't be too difficult to gather files with only renames in a separate commit but let renames in e.g. interpolation.jl where there are other changes too stay in the "real" commit. Perhaps not worth it.

termi-official:
I am currently not having time for the review, but if this one lingers around for some more time I will definitely take a look.

src/Grid/utils.jl
KnutAM: Replaced by code in [grid.jl](https://github.com/Ferrite-FEM/Ferrite.jl/blob/6fc8fc479f381989585847b3b9bba282d044347b/src/Grid/grid.jl#L676-L716)

Could make sense to move all functionality relating to adding / creating sets in the grid to utils. Can do that before merging.

Ref for motivation: *Extract Boundary entities* in https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#issue-2295919300 and discussions https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#issuecomment-2111881613 -> https://github.com/Ferrite-FEM/Ferrite.jl/pull/914#issuecomment-2112837346

src/exports.jl
```diff
@@ -86,19 +87,17 @@ export
...
-    getfaceset,
-    getedgeset,
+    getfacetset,
     getvertexset,
     get_node_coordinate,
     getcoordinates,
     getcoordinates!,
     onboundary,
     nfaces,
+    nfacets,
```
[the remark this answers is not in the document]
KnutAM: Yes, probably good. 
Perhaps neither should be exported?

src/Dofs/DofHandler.jl
```diff
@@ -701,7 +688,7 @@ Here the unique representation is the sorted node index tuple.
 Note that in 3D we only need indices to uniquely identify a face,
 so the unique representation is always a tuple length 3.
 """
-sortface_fast(face::Tuple{Int,Int}) = sortedge_fast(face) # Face in 2D is the same as edge in 3D.
+function sortface_fast end
```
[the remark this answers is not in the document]
KnutAM: Introduced in #627 (An then `sorface_fast` was added in https://github.com/Ferrite-FEM/Ferrite.jl/pull/753 being the same as the old `sortface`). 

`sortface` returns an additional orientation info. 

IMO `sortface` should always call `sortface_fast`, but that's a separate issue...

src/Dofs/DofHandler.jl
KnutAM: Good spot! Not AFAIK. 
@lijas do you know more?

src/Dofs/DofHandler.jl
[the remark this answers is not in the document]
termi-official: Yea this is a naming issue. `sort*_fast` materializes a unique identifier for an entity and `sort*` materializes the unique identifier+orientation metadata. See https://github.com/Ferrite-FEM/FerriteDistributed.jl/pull/39/files#diff-046950d759ac820d5fb4bd9a9aee475ad4213cbc1ad6f0fc204c05914fc3864b for some work on this.

src/Grid/grid.jl
```diff
@@ -355,7 +449,30 @@ This function takes the local vertex representation (a `VertexIndex`) and looks
...
+function _get_reference_dimensionality(cells::AbstractVector{<:AbstractCell})
+    # Could make fast-path for eltype being union of cells with different rdims, but @KristofferC recommends against that,
+    # https://discourse.julialang.org/t/iterating-through-types-of-a-union-in-a-type-stable-manner/58285/3
+    # Note, this function is inherently type-instable.
+    rdims = Set{Int}()
+    for cell in cells
+        push!(rdims, getdim(cell))
+    end
+    length(rdims) == 1 && return first(rdims)
+    return :mixed
```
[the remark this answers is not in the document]
KnutAM: OK, did that instead now.
(Don't really understand why though, I thought returning `:mixed` (and dispatching on `Val{:mixed}` was more readable than `nothing` / `Val{nothing}`).

test/runtests.jl
```diff
@@ -17,7 +17,7 @@ if HAS_EXTENSIONS && MODULE_CAN_BE_TYPE_PARAMETER
     import Metis
 end
 
-const RUN_JET_TESTS = VERSION >= v"1.9"
+const RUN_JET_TESTS = false#VERSION >= v"1.9"
```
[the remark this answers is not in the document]
KnutAM: Is the new solution (removing "JET" from "extras" and `add`ing it only if `RUN_JET_TESTS`) ok? 
Couldn't figure out another way to add it without nightly failing since JET won't precompile on nightly...

src/Dofs/DofHandler.jl
KnutAM: Fixed by checking reference dim of cell. This should be safe in all cases except two cells on top of eachother, which is the same limitation as the check on master. Diff showing fix: https://github.com/Ferrite-FEM/Ferrite.jl/pull/914/commits/5223729c4a0b0683e331f0535602d0125c3cb52c

src/Grid/grid.jl
KnutAM: Renamed to `get_reference_dimension`

docs/src/literate-gallery/helmholtz.jl
```diff
@@ -80,14 +80,14 @@ end;
...
-dbc = Dirichlet(:u, union(getfaceset(grid, "top"), getfaceset(grid, "right")), (x,t) -> u_ana(x))
+dbc = Dirichlet(:u, union(getfacetset(grid, "top"), getfacetset(grid, "right")), (x,t) -> u_ana(x))
 add!(dbcs, dbc)
 close!(dbcs)
 update!(dbcs, 0.0)
 
 K = create_sparsity_pattern(dh);
 
-function doassemble(cellvalues::CellValues, facevalues::FaceValues,
+function doassemble(cellvalues::CellValues, facetvaluesFacetValues,
```
fredrikekre, line 90:
```suggestion
function doassemble(cellvalues::CellValues, facetvalues::FacetValues,
```

src/Grid/grid.jl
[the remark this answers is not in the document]
fredrikekre: I thought this was user-facing and didn't see it was used for Val-dispatch.

docs/src/literate-gallery/helmholtz.jl
KnutAM: Thanks - fixed!

src/Grid/grid.jl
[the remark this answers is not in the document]
KnutAM: OK, reverted and clarified in docstring that it is internal.

src/Grid/grid.jl
KnutAM: Solved (with nice redundant code reduction too :))

fredrikekre:
Squashed and wrote a commit message locally. Merged to master in https://github.com/Ferrite-FEM/Ferrite.jl/commit/2ca4f941f76c2d350d485c337f1d892ab25de312 (thought it could be useful to keep the commit history around here instead of first pushing the squashed-commit to this branch).

I know there are some more todos but this PR was already huge and touches a lot of code so the risk of merge conflicts was pretty big.

Thanks @lijas for pushing for this initially and @KnutAM for finishing the PR!

KnutAM:
Thanks for the quick review and actions!

> could be useful to keep the commit history around

Good idea!

> I know there are some more todos but this PR was already huge and touches a lot of code so the risk of merge conflicts was pretty big.

Agreed, I'll open separate (smaller) prs for those!

PR #916, opened 2024-05-16
Generalize CSC assembler

termi-official:
These changes should be sufficient to allow assembly into CSC matrix types other than SparseMatrixCSC and the Symmtric counterpart.

Also a step towards shaping the interface for #628 , since resize is not allowed on the GPU.

## TODOs
* [x] Fix CI
* [x] Interface considerations

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -144,13 +144,13 @@ matrix_handle(a::SymmetricCSCAssembler) = a.K.data
 vector_handle(a::AbstractCSCAssembler) = a.f
 
 """
-    start_assemble(K::AbstractSparseMatrixCSC;            fillzero::Bool=true, maxcelldofs_hint::Int=0) -> CSCAssembler
-    start_assemble(K::AbstractSparseMatrixCSC, f::Vector; fillzero::Bool=true, maxcelldofs_hint::Int=0) -> CSCAssembler
+    start_assemble(K::AbstractSparseMatrixCSC;            fillzero::Bool=true) -> CSCAssembler
+    start_assemble(K::AbstractSparseMatrixCSC, f::Vector; fillzero::Bool=true) -> CSCAssembler
 
 Create a `CSCAssembler` from the matrix `K` and optional vector `f`.
 
-    start_assemble(K::Symmetric{AbstractSparseMatrixCSC};                 fillzero::Bool=true, maxcelldofs_hint::Int=0) -> SymmetricCSCAssembler
-    start_assemble(K::Symmetric{AbstractSparseMatrixCSC}, f::Vector=Td[]; fillzero::Bool=true, maxcelldofs_hint::Int=0) -> SymmetricCSCAssembler
+    start_assemble(K::Symmetric{AbstractSparseMatrixCSC};                 fillzero::Bool=true) -> SymmetricCSCAssembler
+    start_assemble(K::Symmetric{AbstractSparseMatrixCSC}, f::Vector=Td[]; fillzero::Bool=true) -> SymmetricCSCAssembler
 
 Create a `SymmetricCSCAssembler` from the matrix `K` and optional vector `f`.
 
@@ -160,8 +160,6 @@ necessary for efficient matrix assembly. To assemble the contribution from an el
 
 The keyword argument `fillzero` can be set to `false` if `K` and `f` should not be zeroed
 out, but instead keep their current values.
-The keyword argument `maxcelldofs_hint` can be set to the largest number celldofs over all cells
-to eliminate runtime allocations.
 """
 start_assemble(K::Union{AbstractSparseMatrixCSC, Symmetric{<:Any,<:AbstractSparseMatrixCSC}}, f::Vector; fillzero::Bool)
 
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -1,12 +1,19 @@
-struct Assembler{T}
+abstract type AbstractSparseAssembler end
+abstract type AbstractSparseCSCAssembler <: AbstractSparseAssembler end
+
+"""
+This assembler creates a COO (**coo**rdinate format) representation of a sparse matrix during
+assembly and converts it into a `SparseMatrixCSC{T}` on finalization.
+"""
+struct COOAssembler{T}
     I::Vector{Int}
     J::Vector{Int}
     V::Vector{T}
 end
 
-Assembler(N) = Assembler(Float64, N)
+COOAssembler(N) = COOAssembler(Float64, N)
 
-function Assembler(::Type{T}, N) where T
+function COOAssembler(::Type{T}, N) where T
     I = Int[]
     J = Int[]
     V = T[]
@@ -14,11 +21,11 @@ function Assembler(::Type{T}, N) where T
     sizehint!(J, N)
     sizehint!(V, N)
 
-    Assembler(I, J, V)
+    COOAssembler(I, J, V)
 end
 
 """
-    start_assemble([N=0]) -> Assembler
+    start_assemble([N=0]) -> COOAssembler
 
 Create an `Assembler` object which can be used to assemble element contributions to the
 global sparse matrix. Use [`assemble!`](@ref) for each element, and [`finish_assemble`](@ref),
@@ -33,24 +40,28 @@ as described in the [manual](@ref man-assembly).
     the same pattern. See the [manual section](@ref man-assembly) on assembly.
 """
 function start_assemble(N::Int=0)
-    return Assembler(N)
+    return start_assemble(Float64, N)
+end
+
+function start_assemble(t::Type{T}, N::Int=0) where T
+    return COOAssembler(t, N)
 end
 
 """
-    assemble!(a::Assembler, dofs, Ke)
+    assemble!(a::COOAssembler, dofs, Ke)
 
 Assembles the element matrix `Ke` into `a`.
 """
-function assemble!(a::Assembler{T}, dofs::AbstractVector{Int}, Ke::AbstractMatrix{T}) where {T}
+function assemble!(a::COOAssembler{T}, dofs::AbstractVector{Int}, Ke::AbstractMatrix{T}) where {T}
     assemble!(a, dofs, dofs, Ke)
 end
 
 """
-    assemble!(a::Assembler, rowdofs, coldofs, Ke)
+    assemble!(a::COOAssembler, rowdofs, coldofs, Ke)
 
 Assembles the matrix `Ke` into `a` according to the dofs specified by `rowdofs` and `coldofs`.
 """
-function assemble!(a::Assembler{T}, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix{T}) where {T}
+function assemble!(a::COOAssembler{T}, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix{T}) where {T}
     nrows = length(rowdofs)
     ncols = length(coldofs)
 
@@ -67,12 +78,12 @@ function assemble!(a::Assembler{T}, rowdofs::AbstractVector{Int}, coldofs::Abstr
 end
 
 """
-    finish_assemble(a::Assembler) -> K
+    finish_assemble(a::COOAssembler) -> K
 
 Finalizes an assembly. Returns a sparse matrix with the
-assembled values. Note that this step is not necessary for `AbstractSparseAssembler`s.
+assembled values. Note that this step is not necessary for the standard `AbstractSparseAssembler`s.
 """
-function finish_assemble(a::Assembler)
+function finish_assemble(a::COOAssembler)
     return sparse(a.I, a.J, a.V)
 end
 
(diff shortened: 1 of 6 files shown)


src/assembler.jl
```diff
@@ -99,20 +100,20 @@ Return a reference to the underlying matrix/vector of the assembler.
...
     f::Vector{Tv}
     permutation::Vector{Int}
     sorteddofs::Vector{Int}
 end
 
-function Base.show(io::IO, ::MIME"text/plain", a::Union{AssemblerSparsityPattern,AssemblerSymmetricSparsityPattern})
+function Base.show(io::IO, ::MIME"text/plain", a::AbstractSparseCSCAssembler)
     print(io, typeof(a), " for assembling into:\n - ")
     summary(io, a.K)
     if !isempty(a.f)
```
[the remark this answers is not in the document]
termi-official: Nice catch! Thanks. The functions already exist. :)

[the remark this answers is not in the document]
KnutAM: But e.g. `matrix_handle` returns the data stored inside the `Symmetric` wrapper, so I thought that was a different interface.

src/assembler.jl
```diff
@@ -164,22 +166,33 @@ stiffness (and force) in `A`, given the element degrees of freedom `dofs`.
...
-@propagate_inbounds function assemble!(A::AssemblerSymmetricSparsityPattern, dofs::AbstractVector{Int}, Ke::AbstractMatrix, fe::AbstractVector)
+@propagate_inbounds function assemble!(A::SymmetricCSCAssembler, dofs::AbstractVector{<:Integer}, Ke::AbstractMatrix, fe::AbstractVector)
     _assemble!(A, dofs, Ke, fe, true)
 end
 
-@propagate_inbounds function _assemble!(A::AbstractSparseAssembler, dofs::AbstractVector{Int}, Ke::AbstractMatrix, fe::AbstractVector, sym::Bool)
+@propagate_inbounds function _sortdofs_for_assembly(A::AbstractSparseAssembler, dofs::AbstractVector)
+    ld = length(dofs)
+    permutation = A.permutation
+    sorteddofs = A.sorteddofs
```
KnutAM, line 187: Here as well (but since you nicely extracted this function, could just dispatch on the union of the existing assemblers?)

src/assembler.jl
```diff
@@ -164,22 +166,33 @@ stiffness (and force) in `A`, given the element degrees of freedom `dofs`.
...
+    resize!(sorteddofs, ld)
+    copyto!(sorteddofs, dofs)
+    sortperm2!(sorteddofs, permutation)
+    return sorteddofs, permutation
+end
+
+@propagate_inbounds function _assemble!(A::AbstractSparseCSCAssembler, dofs::AbstractVector{<:Integer}, Ke::AbstractMatrix, fe::AbstractVector, sym::Bool)
     ld = length(dofs)
     @boundscheck checkbounds(Ke, keys(dofs), keys(dofs))
     if length(fe) != 0
```
KnutAM, line 198: And below here

src/assembler.jl
termi-official: What exactly is the issue here and below?

KnutAM: That the input is `AbstractSparseAssembler`, but specific fields are used (i.e. the implementer of an `AT <:  AbstractSparseAssembler` must provide these specific fields. 
In this case, I guess the function is specific to `Union{CSCAssembler, SymmetricCSCAssembler}`, and I would think it makes more sense to dispatch on that? 

A more general question perhaps for my understanding, what methods do you expect to redefine for a custom `AbstractSparseCSCAssembler`?

src/assembler.jl
```diff
@@ -99,59 +100,60 @@ Return a reference to the underlying matrix/vector of the assembler.
...
+        summary(io, f)
     end
 end
 
-matrix_handle(a::AssemblerSparsityPattern) = a.K
-matrix_handle(a::AssemblerSymmetricSparsityPattern) = a.K.data
-vector_handle(a::Union{AssemblerSparsityPattern, AssemblerSymmetricSparsityPattern}) = a.f
+matrix_handle(a::AbstractSparseCSCAssembler) = a.K
+matrix_handle(a::SymmetricCSCAssembler) = a.K.data
+vector_handle(a::AbstractSparseCSCAssembler) = a.f
```
KnutAM, line 128:
```suggestion
vector_handle(a::Union{CSCAssembler, SymmetricCSCAssembler}) = a.f
```

src/assembler.jl
termi-official: I see it as follows: We provide sensible defaults for these dispatches and if you want to do something different, then you can do so.

koehlerson: This is also the same for AbstractGrid and brings the advantage that you only need to dispatch the specific things that actually changed

KnutAM: I guess we have different opinions here then :) I see that it makes it "easier" to hack, but direct field access interfaces instead of accessor functions makes renaming a field a breaking change. I now recall that this was discussed before, ref https://github.com/Ferrite-FEM/Ferrite.jl/pull/655#issuecomment-1490372711 regarding the Grid/DofHandler, but I don't think a consensus was found.

KnutAM: IMO we should
* Dispatch `show`, `vector_handle`, and `_sortdofs_for_assembly` on `Union{SymCSCAssembler, CSCAssembler}`
* Use `vector_handle` in `_assemble!`
So that if you bring your own `assembler::AbstractCSCAssembler` that should use `_assemble!`, you must define `vector_handle`, `matrix_handle`, and `_sortdofs_for_assembly`

termi-official: A good example of why we prefer this is it given in FerriteDistributed
https://github.com/Ferrite-FEM/FerriteDistributed.jl/blob/main/src/NODDofHandler.jl#L31-L108
https://github.com/Ferrite-FEM/FerriteDistributed.jl/blob/main/src/NODDofHandler.jl#L127-L143
Although it is not as extreme for the assembler

KnutAM: AFAIU most of those won't work with https://github.com/Ferrite-FEM/Ferrite.jl/commit/7e8a57150738094bb951d6e672fdeed205d0a1ff due to the subdofhandler, and thus wouldn't be redundant? Which is my point, that this approach quickly leads to breaking the code. 

But just to be clear - I strongly support writing our functions to support abstract types. I just think there should be functions then and not field access, because functions are much easier to overload / replace when changes are made.

koehlerson: Just another example, if we would consistently rewrite the `AbstractGrid` dispatches to use accessor functions, then those https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L385-L467 would need to be dispatched for https://github.com/Ferrite-FEM/Ferrite.jl/blob/731dfbb0f0733d943de90a55c0a7c31b82704d2e/src/Adaptivity/ncgrid.jl which only adds a lot of redundant code. I think this issue was discussed back then when we started to use some other grid types. We also have in the docs the statement that one can rely on the fallback `AbstractGrid` dispatches in case specific parts of the datastructure don't change: https://ferrite-fem.github.io/Ferrite.jl/stable/manual/grid/#AbstractGrid

KnutAM: I started generalizing this in https://github.com/Ferrite-FEM/Ferrite.jl/pull/662/files#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90. 
I would expect that you should only need to define 
```julia
get_coordinate_type
getcells
getnodes
getcellsets
getnodesets
getfacesets
getedgesets
getvertexsets
```
To make an equivalent grid as the current, potentially with more features such as the ncgrid. 

> We also have in the docs the statement that one can rely on the fallback AbstractGrid dispatches in case specific parts of the datastructure don't change: https://ferrite-fem.github.io/Ferrite.jl/stable/manual/grid/#AbstractGrid

"In case that certain structures are preserved from the Ferrite.Grid type, you don't need to dispatch on your own type, but rather rely on the fallback AbstractGrid dispatch."

Oh, I was unaware of this statement. This is a pretty strong statement in the docs, which will mean any internal change to the grid has to be considered a breaking change. Do we really want to promise this with 1.0? (I mean, even if it works and we decide to keep allowing it, I think we should add a disclaimer to that statement that such changes will not be considered breaking)

koehlerson: > 
> ```julia
> get_coordinate_type
> getcells
> getnodes
> getcellsets
> getnodesets
> getfacesets
> getedgesets
> getvertexsets
> ```
> 
> To make an equivalent grid as the current, potentially with more features such as the ncgrid.

Yeah that's the point, I don't need to define any of those, since the structure is preserved for those functions. https://github.com/Ferrite-FEM/Ferrite.jl/blob/731dfbb0f0733d943de90a55c0a7c31b82704d2e/src/Adaptivity/ncgrid.jl is the complete implementation of the datastructure, there is no additional code to it, which wouldn't be the case if we use the approach of: https://github.com/Ferrite-FEM/Ferrite.jl/pull/662/files#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90

> Do we really want to promise this with 1.0? (I mean, even if it works and we decide to keep allowing it, I think we should add a disclaimer to that statement that such changes will not be considered breaking)

Disclaimer is fine for me. I'm not sure if either @KristofferC or @fredrikekre convinced me back then that the fallback dispatches are a good idea or maybe I'm mixing up memories

termi-official: > AFAIU most of those won't work with [7e8a571](https://github.com/Ferrite-FEM/Ferrite.jl/commit/7e8a57150738094bb951d6e672fdeed205d0a1ff) due to the subdofhandler, and thus wouldn't be redundant? Which is my point, that this approach quickly leads to breaking the code.
> 
> But just to be clear - I strongly support writing our functions to support abstract types. I just think there should be functions then and not field access, because functions are much easier to overload / replace when changes are made.

(Wrong commit?) But okay, let us discuss API design then for a moment, because this is what this comment (and linked ones) boil down to. 

You are right, breaking internals is breaking for subpackages which depend on the internals. And there is no good around it. Every change, even a simple bug fix, can be potentially breaking. My point is that the subpackages like for example FerriteDistributed should not diverge from the Ferrite codebase. We changed the features for the dof handler and the distributed dof handler is not updated yet. So if FerriteDistributed would be registered and we would tag a new release of Ferrite with the breaking internal change, then the current release of FerriteDistributed it would be simply not compatible with the release. Users will probably expect that if they pull the new Ferrite release with FerriteDistributed, that the new features will integrate, which we cannot guarantee in general. In the worst case this will lead to a silent failure.

To give another example: The node ordering for our elements is technically an internal invariant and should not matter for the user. However, changing it breaks the mesh readers and exporters. Simply because we need to rely on some internal assumptions. 

I think this makes the point clear that breaking internals will require checking downstream packages anyway if we rely in one way or another on internal details.

However, in software development the pure renaming of member variables almost never happens. Furthermore, I would argue that it is sensible to assume that specific sets of similar data types share similar members within a framework, especially if they do essentially the same thing in a slightly different way which we cannot easily express in the algorithms.

KnutAM: (The commit link was copied from FerriteDistributed.jl's readme stating which Ferrite version it should work with)

Of course this doesn't prevent issues with downstream packages, and as you say, there isn't an easy way around. My point was just that having these functions can reduce the risk somewhat (and make fixing things easier because it encourages using these functions more), at IMO a low cost of duplicating functions accessing the fields. One example of a typical rename would be to put part of the functionality in a sub-struct as we did for `CellValues` recently. 

But at the end of the day, this is a design choice. I commented because I thought this was just a mistake, but if not and it can be argued then for the assembler this is a minor issue.

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
+
+"""
+Assembler for sparse matrix with CSC storage type.
+"""
+struct CSCAssembler{Tv,Ti,MT<:AbstractSparseMatrixCSC{Tv,Ti}} <: AbstractSparseCSCAssembler
+    K::MT
     f::Vector{Tv}
     permutation::Vector{Int}
     sorteddofs::Vector{Int}
 end
```
KnutAM, line 132: Since we are renaming and introducing new types here, would it make sense to get rid of the symmetric type, seems like some code could be replaced which AFAIU is only used to get a bool when calling `_assemble!`? E.g. 
```suggestion
struct CSCAssembler{Tv, Ti, MT} <: AbstractSparseCSCAssembler
    K::MT
    f::Vector{Tv}
    permutation::Vector{Int}
    sorteddofs::Vector{Int}
    function CSCAssembler(K::Union{DT, Symmetric{Tv, TT}}, f::Vector{Tv}, permutation::Vector{Int}, sorteddofs::Vector{Int}) where {Tv, Ti, DT<:AbstractSparseMatrixCSC{Tv,Ti}}
        return new{Tv, Ti, typeof(K)}(K, f, permutation, sorteddofs)
    end
end
symmetric_pattern(::CSCAssembler) = false
symmetric_pattern(::CSCAssembler{<:Any, <:Any, <:Symmetric}) = true
```

with 
```julia
matrix_handle(K::Symmetric) = K.data
matrix_handle(K::AbstractMatrix) = K

matrix_handle(a::AbstractSparseAssembler) = matrix_handle(get_matrix(a))
```
alternatively 
```julia
matrix_handle(a::AbstractSparseAssembler) = symmetric_pattern(a) ? get_matrix(a).data : get_matrix(a)
```

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
-vector_handle(a::Union{AssemblerSparsityPattern, AssemblerSymmetricSparsityPattern}) = a.f
+matrix_handle(a::AbstractSparseCSCAssembler) = a.K
+matrix_handle(a::SymmetricCSCAssembler) = a.K.data
+vector_handle(a::AbstractSparseCSCAssembler) = a.f
 
 """
-    start_assemble(K::SparseMatrixCSC;            fillzero::Bool=true) -> AssemblerSparsityPattern
-    start_assemble(K::SparseMatrixCSC, f::Vector; fillzero::Bool=true) -> AssemblerSparsityPattern
+    start_assemble(K::AbstractSparseMatrixCSC;            fillzero::Bool=true) -> CSCAssembler
+    start_assemble(K::AbstractSparseMatrixCSC, f::Vector; fillzero::Bool=true) -> CSCAssembler
```
KnutAM, line 160:
```suggestion
    start_assemble(K::AbstractSparseMatrixCSC;            fillzero::Bool=true, sizehint::Int=0) -> CSCAssembler
    start_assemble(K::AbstractSparseMatrixCSC, f::Vector; fillzero::Bool=true, sizehint::Int=0) -> CSCAssembler
```
Or do you intend to keep this as private API for now? (In that case, perhaps add a note about this?) If public, perhaps `workspace_sizehint` is better even if longer, not to confuse with the sizehint supplied to `start_assemble(N)`?

src/assembler.jl
```diff
@@ -243,8 +292,8 @@ function _missing_sparsity_pattern_error(Krow::Int, Kcol::Int)
         "$(Kcol)] is missing in the sparsity pattern. Make sure you have called `K = " *
         "create_sparsity_pattern(dh)` or `K = create_sparsity_pattern(dh, ch)` if you " *
         "have affine constraints. This error might also happen if you are using " *
-        "`::AssemblerSparsityPattern` in a threaded assembly loop (you need to create an " *
-        "`assembler::AssemblerSparsityPattern` for each task)."
+        "`::CSCAssembler` in a threaded assembly loop (you need to create an " *
+        "`assembler::CSCAssembler` for each task)."
```
KnutAM, line 296:
```suggestion
        "the assembler in a threaded assembly loop (you need to create one " *
        "`assembler` for each task)."
```

or pass the type into this function, but seems overkill to me.

src/assembler.jl
```diff
@@ -287,6 +336,11 @@ end
 
 # Sort utilities
 
+"""
+    sortperm2!(data::AbstractVector, permutation::AbstractVector)
+
+Sort the input vector inplace and compute the corresponding permutation.
+"""
```
KnutAM, line 343: devdocs?

src/assembler.jl
```diff
@@ -164,22 +196,38 @@ stiffness (and force) in `A`, given the element degrees of freedom `dofs`.
...
+@propagate_inbounds function assemble!(A::SymmetricCSCAssembler, dofs::AbstractVector{<:Integer}, Ke::AbstractMatrix, fe::AbstractVector)
     _assemble!(A, dofs, Ke, fe, true)
 end
 
-@propagate_inbounds function _assemble!(A::AbstractSparseAssembler, dofs::AbstractVector{Int}, Ke::AbstractMatrix, fe::AbstractVector, sym::Bool)
+"""
+    _sortdofs_for_assembly(A::AbstractSparseAssembler, dofs::AbstractVector) -> sorteddofs, permutation
+
+Sorts the dofs into a separate buffer and returns it together with a permutation vector.
+"""
```
KnutAM, line 218: devdocs

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
-
 """
     matrix_handle(a::AbstractSparseAssembler)
     vector_handle(a::AbstractSparseAssembler)
 
-Return a reference to the underlying matrix/vector of the assembler.
+Return a reference to the underlying matrix/vector of the assembler used during
+assembly operations.
 """
 matrix_handle, vector_handle
```
KnutAM, line 110: devdocs?

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
-struct AssemblerSparsityPattern{Tv,Ti} <: AbstractSparseAssembler
-    K::SparseMatrixCSC{Tv,Ti}
+"""
+    get_matrix(a::AbstractSparseAssembler)
+    get_vector(a::AbstractSparseAssembler)
+
+Return a reference to the matrix into which we assemble.
+The returned object might differ from [`matrix_handle`](@ref).
+"""
+get_matrix, get_vector
```
KnutAM, line 119: devdocs?

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
+
+get_vector(a::AbstractSparseAssembler) = a.f
+get_matrix(a::AbstractSparseAssembler) = a.K
+
+"""
+Assembler for sparse matrix with CSC storage type.
+"""
```
KnutAM, line 126: Docs?

src/assembler.jl
```diff
@@ -89,71 +100,92 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
     f::Vector{Tv}
     permutation::Vector{Int}
     sorteddofs::Vector{Int}
 end
-struct AssemblerSymmetricSparsityPattern{Tv,Ti} <: AbstractSparseAssembler
-    K::Symmetric{Tv,SparseMatrixCSC{Tv,Ti}}
+
+"""
+Assembler for symmetric sparse matrix with CSC storage type.
+"""
```
KnutAM, line 136: I would say docs, but I rather prefer merging with `CSCAssembler`

src/assembler.jl
KnutAM: I think @fredrikekre wanted to remove the symmetric assembler type anyways, and with this we maintain all functionality AFAIU.

KnutAM (commented):
I think this is shaping up nicely. Some suggestions / discussion points which would be good to decide on before releasing 1.0. 

The add to docs/devdocs comments can of course wait until the other details are finalized, just added it as reminders.

src/assembler.jl
termi-official: Oh, I forgot to push the file haha

termi-official:
Thanks for the review Knut!

I would like to suggest that the removal of the symmetric API should go into a separate PR.

src/assembler.jl
```diff
@@ -89,71 +100,94 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
 
-`AssemblerSparsityPattern` and `AssemblerSymmetricSparsityPattern` allocate workspace
+`CSCAssembler` and `SymmetricCSCAssembler` allocate workspace
 necessary for efficient matrix assembly. To assemble the contribution from an element, use
 [`assemble!`](@ref).
 
 The keyword argument `fillzero` can be set to `false` if `K` and `f` should not be zeroed
 out, but instead keep their current values.
+The keyword argument `maxcelldofs_hint` can be set to the largest number celldofs over all cells
+to eliminate runtime allocations.
```
KnutAM, line 176: I guess technically they could still happen due to the resize, just unlikely? Perhaps "reduce" or "reduce the chance of"

src/assembler.jl
```diff
@@ -1,24 +1,31 @@
-struct Assembler{T}
+abstract type AbstractSparseAssembler end
+abstract type AbstractSparseCSCAssembler <: AbstractSparseAssembler end
+
+"""
+This assembler creates a COO representation of a sparse matrix during
+assembly and converts it into a `SparseMatrixCSC{T}` on finalization.
+"""
+struct COOAssembler{T}
```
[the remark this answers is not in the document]
KnutAM: If useful it should be kept.

KnutAM (approved):
LGTM. 
I like the name-changes, but perhaps there should be a `@deprecate_binding` or something @fredrikekre ? 
I just think we are at a point now that we need to reduce annoyance to users where possible during the upgrade.

src/assembler.jl
termi-official: If you set it to the described value you should have a hard guarantee that resize is basically a noop.

KnutAM: Seems hard to trigger at least, somehow filling the values seems to be required to be even more safe, but I don't know the internals good enough to say what is happening. 
```julia
vals = Float64[];
resize!(vals, 100_000_000);
@time resize!(vals, 1);
# 0.000011 seconds
GC.gc();
@time resize!(vals, 100_000_000);
# 0.000010 seconds
@time fill!(vals, 1.0);
# 0.491245 seconds (133 allocations: 6.391 KiB, 2.23% compilation time)
@time resize!(vals, 100_000_000_0);
# 1.977564 seconds (1 allocation: 0 bytes)
```
where the last one is just interesting, perhaps a bug in reporting...

termi-official: cc @fredrikekre can you bring light into the dark here? From my understanding of the `resize!` docs we have a guarantee that once we have an vector with $n$ elements consecutive resize calls will not allocate as long as the resize is to at most $n$ elements.

fredrikekre: I think currently there is no shrinking of the backing memory if that's what you are asking. Is the potential runtime resize problematic though? Perhaps this kwarg isn't necesary.

termi-official: This is prospective for GPU assembly where we are not allowed to runtime allocate memory.

termi-official:
This should be ready again.

src/assembler.jl
```diff
@@ -89,71 +100,94 @@ Assembles the element residual `ge` into the global residual vector `g`.
...
+    return CSCAssembler(K, f, zeros(Int,maxcelldofs_hint), zeros(Int,maxcelldofs_hint))
 end
-function start_assemble(K::Symmetric{T,<:SparseMatrixCSC}, f::Vector=T[]; fillzero::Bool=true) where T
+function start_assemble(K::Symmetric{T,<:SparseMatrixCSC}, f::Vector=T[]; fillzero::Bool=true, maxcelldofs_hint::Int=0) where T
     fillzero && (fillzero!(K); fillzero!(f))
-    return AssemblerSymmetricSparsityPattern(K, f, Int[], Int[])
+    return SymmetricCSCAssembler(K, f, zeros(Int,maxcelldofs_hint), zeros(Int,maxcelldofs_hint))
 end
 
+finish_assemble(a::AbstractSparseCSCAssembler) = get_matrix(a)
```
KnutAM, line 189: Why is this needed? Not covered by tests and I think this shouldn't be encouranged since it doesn't "do" anything.

src/assembler.jl
```diff
@@ -1,24 +1,31 @@
...
+assembly and converts it into a `SparseMatrixCSC{T}` on finalization.
+"""
+struct COOAssembler{T}
     I::Vector{Int}
     J::Vector{Int}
     V::Vector{T}
 end
 
-Assembler(N) = Assembler(Float64, N)
+COOAssembler(N) = COOAssembler(Float64, N)
```
KnutAM, line 14: We are adding defaults both here and for `start_assemble`, I guess it suffices to add this for `start_assemble`? Alternatively, should we not have the `start_assemble` function for this case, and rather have to explicityly define a `COOAssembler` when required? Seems reasonable to me to request this type explicitly in those special cases?

KnutAM (commented):
Some more comments when looking back on this. But still overall a good generalization IMO.

src/assembler.jl
termi-official: I need this interface for making virtually any assembly format efficient that is not vanilla CSR or CSC. E.g. EA or ParCSR, which need a synchronization phase after all data is there (instead of syncing potentially each element individually).

src/assembler.jl
termi-official: I think from the interface point of view the start_assemble makes more sense here, but I do not have a strong opinion. @fredrikekre ?

src/assembler.jl
KnutAM: It is not clear what this should return if you have a matrix + a vector to assemble for example?

I would find a two step version more logical, I.e. `finish_assemble!(a); K = get_matrix(a)`. (Not sure what the finish function should be called). Perhaps `materialize!` (or `close!`)

src/assembler.jl
```diff
@@ -33,24 +40,28 @@ as described in the [manual](@ref man-assembly).
     the same pattern. See the [manual section](@ref man-assembly) on assembly.
 """
 function start_assemble(N::Int=0)
-    return Assembler(N)
+    return start_assemble(Float64, N)
```
KnutAM, line 43:
```suggestion
    return COOAssembler(N)
```

fredrikekre (approved):

src/assembler.jl
```diff
@@ -33,24 +40,28 @@ as described in the [manual](@ref man-assembly).
...
+end
+
+function start_assemble(t::Type{T}, N::Int=0) where T
+    return COOAssembler(t, N)
```
KnutAM, line 47: Just noticed, change to
```suggestion
function start_assemble(::Type{T}, N::Int=0) where T
    return COOAssembler(T, N)
```
?

KnutAM: Well, too slow...

fredrikekre: Reworked in  #1058

PR #924, opened 2024-05-19
Deprecate `onboundary`, remove `boundary_matrix` property from `Grid`

fredrikekre:
This patch deprecates the `onboundary` function which checks whether an element has any facet alone an external boundary. This functionality relied on the `boundary_matrix` property in `Grid`, but this was only populated correctly by the builtin grid generators (and not e.g. for meshes parsed from other formats).

Checking `onboundary` was just a shortcut to avoid some extra work in the facetset containment check which you always had to do anyways, but since this is just a tuple hash it isn't expensive anyway.

In addition, nowadays there are utilities such as `FacetIterator` which let's you explicitly loop over the facet sets you want directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,6 +385,17 @@ more discussion).
   now you can still use them by prefixing `Ferrite.`, e.g. `Ferrite.getweights`.)
   ([#754][github-754])
 
+- The `onboundary` function (and the associated `boundary_matrix` property of the `Grid`
+  datastructure) have been removed ([#924][github-924]). Instead of first checking
+  `onboundary` and then check whether a facet belong to a specific facetset, check the
+  facetset directly. For example:
+  ```diff
+  - if onboundary(cell, local_face_id) && (cell_id, local_face_id) in getfacesets(grid, "traction_boundary")
+  + if (cell_id, local_face_id) in getfacesets(grid, "traction_boundary")
+       # integrate the "traction_boundary" boundary
+    end
+  ```
+
 ### Fixed
 
 - Benchmarks now work with master branch. ([#751][github-#751], [#855][github-#855])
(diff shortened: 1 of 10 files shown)


termi-official:
Should we also update the examples then to use the FaceIterator instead of locally checking for all faces if we are on the boundary? I think especially if we have multiple small sets on the boundary there could be quite a bit of performance overhead.

lijas (approved):

CHANGELOG.md
```diff
@@ -385,6 +385,16 @@ more discussion).
   now you can still use them by prefixing `Ferrite.`, e.g. `Ferrite.getweights`.)
   ([#754][github-754])
 
+- The `onboundary` function have been removed ([#924][github-924]). Instead of first
```
KnutAM, line 388:
```suggestion
- The `onboundary` function and the associated `boundary_matrix` in the `Grid` datastructure have been removed ([#924][github-924]). Instead of first
```

fredrikekre:
> Should we also update the examples then to use the FaceIterator instead of locally checking for all faces if we are on the boundary?

You mean for the case where you previously could guard many face-set checks behind `onboundary`?
```
if onboundary(...)
    if # boundary 1
        # do things
    end
    if # boundary 2
        # do things
    end
end
```
I guess then it would be preferred to loop explicitly over the sets, but I think the current examples are fine.

fredrikekre:
I think it would be nice with a function that gave you the `FacetIndex`s, for example `facetindices`, then you could replace
```diff
- for local_facetindex in 1:nfacets(cell)
-     if (cellid(cell), local_facetindex) in getfacetset(grid, "boundary")
+ for facetindex in facetindices(cell)
+     if facetindex in getfacetset(grid, "boundary")
          # do something
      end
  end
```

PR #926, opened 2024-05-20
Shell on solid-face test

### LINKED ISSUE #899: Coupling shells to solid faces

lijas:
Test that the dofhandler distributes the dofs correctly when a shell (or beam) element is attached to the face of a solid.

Closes #899

### DIFF
--- a/test/test_dofs.jl
+++ b/test/test_dofs.jl
@@ -679,3 +679,78 @@ end
     K_default = create_sparsity_pattern(dh)
     @test K == K_cont == K_default
 end
+
+
+@testset "shell on solid face" begin
+
+    # Node numbering:
+    # 3 ____ 4  4
+    # |      |  |   
+    # |      |  | (Beam attached to facet)    
+    # 1 ____ 2  2
+
+    dim = 2
+    grid = generate_grid(Quadrilateral, (1,1))
+    line1 = Line((2,4))
+    grid = Grid([grid.cells[1], line1], grid.nodes)
+
+    order = 2
+    ip_solid = Lagrange{RefQuadrilateral, order}()#^dim
+    ip_shell = Lagrange{RefLine, order}()
+
+    dh = DofHandler(grid)
+    sdh_solid = SubDofHandler(dh, Set(1))
+    add!(sdh_solid, :u, ip_solid)
+    sdh_shell = SubDofHandler(dh, Set(2))
+    add!(sdh_shell, :u, ip_shell)
+    close!(dh)
+
+    dofsquad = zeros(Int, ndofs_per_cell(dh, 1))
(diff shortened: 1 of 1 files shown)


fredrikekre:
Nice that no changes were required haha

termi-official:
Since I will need this in the not-so-far future, is there a good reference which explains how to couple the elements consistently? And can we add the theory to the docs?

KnutAM:
> Since I will need this in the not-so-far future, is there a good reference which explains how to couple the elements consistently? And can we add the theory to the docs?

I feel that I don't understand the problem/question (i.e. how it could be done inconsistently): if two elements share the same entity (face/edge/vertex), they will be coupled. Of course a write-up of how the dof-distribution works in devdocs/topic guide could be nice!

termi-official:
E.g. if you have a shell element with rotational and displacment degrees of freedom coupled to a solid element (straight line) for which  you just enforce that the displacement is the same, then the shell can freely rotate on the corner of the element, so you also have to put constraints on the rotational dofs.

KnutAM:
Aha, now I understand the problem, but I don't know the answer.

koehlerson:
Just don't is the answer 🙃

PR #927, opened 2024-05-20
Faster ExclusiveTopology construction

KnutAM:
Had this as notes lingering around after seeing #617, nothing urgent or major.
Timings should be verified by someone since this is so heavy on GC it seems to be quite unreliable to time it and profiling changes from run to run...

Somewhat representative btime runs give 
```julia
grid = generate_grid(Hexahedron, (40, 40, 40)) # more stable timings than 3x80
@btime ExclusiveTopology($grid);
528.067 ms (3582741 allocations: 378.69 MiB) # PR
580.268 ms (4094719 allocations: 464.89 MiB) # master
```

### DIFF
--- a/src/Grid/topology.jl
+++ b/src/Grid/topology.jl
@@ -107,9 +107,10 @@ function _num_shared_vertices(cell_a::C1, cell_b::C2) where {C1, C2}
 end
 
 function _exclusive_topology_ctor(cells::Vector{C}, vertex_cell_table::Array{Set{Int}}, vertex_table, face_table, edge_table, cell_neighbor_table) where C <: AbstractCell
+    cell_neighbor_ids = Set{Int}()
     for (cell_id, cell) in enumerate(cells)
         # Gather all cells which are connected via vertices
-        cell_neighbor_ids = Set{Int}()
+        empty!(cell_neighbor_ids)
         for vertex ∈ vertices(cell)
             for vertex_cell_id ∈ vertex_cell_table[vertex]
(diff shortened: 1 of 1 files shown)


src/Grid/topology.jl
```diff
@@ -107,17 +107,18 @@ function _num_shared_vertices(cell_a::C1, cell_b::C2) where {C1, C2}
 end
 
 function _exclusive_topology_ctor(cells::Vector{C}, vertex_cell_table::Array{Set{Int}}, vertex_table, face_table, edge_table, cell_neighbor_table) where C <: AbstractCell
+    cell_neighbor_ids = Set{Int}()
```
KristofferC, line 110: Is there an advantage to making this ordered so the order in the `cell_neighbor_table[cell_id]` will not be "random"? Maybe it doesn't matter.

termi-official:
I cannot see any significant difference between the PR and master unfortunately.


Master `966.810 ms (4094719 allocations: 464.89 MiB)`
PR `956.878 ms (3582741 allocations: 378.69 MiB)`

So a about 1% difference on my machine. We need to think about how to increase data locality and reduce allocations here if we want to see a significant speedup.

src/Grid/topology.jl
termi-official: Probably this does not matter, because we do not enforce any ordering and locality properties on our grids in general.

KnutAM:
Thanks for checking it!

PR #934, opened 2024-05-21
Add `make clean` target to docs/Makefile

fredrikekre:
Sometimes old generated files that stick around in a local repo can trip up Documenter. This patch adds a `make clean` target which removes files that are generated during the doc build.


KnutAM:
I guess it works if `make` is installed via e.g. MinGW (haven't tried), easier to run via WSL in that case. But maybe I'm just misunderstanding how to use it :)

Yes, `run(`git ls-files`)` works

Issue #931, opened 2024-05-21
consistent docs with entity renaming


koehlerson:
with https://github.com/Ferrite-FEM/Ferrite.jl/pull/914 some docs are off
- [ ] https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/src/Grid/grid.jl#L281-L308
- [ ] https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/docs/src/topics/grid.md

maybe more, feel free to add items to the list

KnutAM:
Thanks, I'll start adressing these now that the mesh parsers have fixes!

PR #935, opened 2024-05-21
Document reference shape entity naming

KnutAM:
* Added a reference shape topic documentation without exposing new API keeping numbering in devdocs. 
* Updated the grid topic docs, moving parts to reference shape. 
* For the grid, removed the `compute_faceset` example as this was outdated and replaced it with a line showing how to get the unique facet representation as I found that clearer. Let me know if you disagree!

Solves part 2 of #931


KnutAM:
Didn't change too much for FerriteGmsh.jl, I think that needs some updates wrt supported cell types @koehlerson? I also think its code needs updates there (e.g. the use of the old Cell struct for the serendipity elements, but didn't find time yet to look into that...)

docs/make.jl
```diff
@@ -62,6 +62,7 @@ bibtex_plugin = CitationBibliography(
         "Topic guides" => [
             "Topic guide overview" => "topics/index.md",
             "topics/fe_intro.md",
+            "topics/reference_shapes.md",
```
fredrikekre, line 65: There is `devdocs/reference_cell.md` where we refer to these as "reference cells" rather than "reference shapes". I think they are always synonyms in our case though, right?

KnutAM: Yes. But they are `AbstractRefShape`, so shape makes more sense to me here? What do you think?

Of course, per definition now the `Cell` is the `RefShape<:AbstractRefShape` since it is codim 0 ...

termi-official: Note that there is still a slight difference (which we might want to point out at least in the devdocs). The reference shape is for us really the integration domain in standardized coordinates, which the cell is technically the reference domain mapped into the physical domain by connecting it with a geometric Ansatz function and associated coordinates.

KnutAM: Good point, 
do you think the current [topics doc](https://ferrite-fem.github.io/Ferrite.jl/previews/PR935/topics/reference_shapes/) is off in this respect?

termi-official: A thing which I spotted is that vertex refers to point, not to corner.

KnutAM: Thanks!

> A thing which I spotted is that vertex refers to point, not to corner.

Yes, chose avoiding corner since calling the endpoints of a `Line` for corner sounds strange to me.

> No, but it does not explain it specifically either.

Good, that should be described in either FEValues or Grid IMO.

termi-official: > Thanks!
> 
> > A thing which I spotted is that vertex refers to point, not to corner.
> 
> Yes, chose avoiding corner since calling the endpoints of a `Line` for corner sounds strange to me.
> 

Good point. But this also creates an ambiguity with the nodes used to describe the nonlinearity of some cells. Any idea if we can resolve this?

> > No, but it does not explain it specifically either.
> 
> Good, that should be described in either FEValues or Grid IMO.

Actually this should be in the docs for the cells/elements maybe? And then the other docs can refer to this. This is especially important for new users to understand that we can decouple the geometry interpolation from the solution interpolation, and that this trick comes with some issues e.g. for visualization in ParaView. Probably this train of thought should be put into some docs PR along other missing things.

KnutAM: OK, I added a short note about the difference between a node and a vertex (this could be improved with a link if we have another place to describe the node concept). 

Especially with #798, there will be a need to explain non-nodal interpolations.

But I'll suggest leaving that for later. I think the docs in this PR is required before 1.0, but the others can be done later.

KnutAM: > reference cells" rather than "reference shapes"

For interpolations we also refer to them as "reference elements" 😆

termi-official: Was probably me haha

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+| Entity   | Description |
+| :------- | :---------- |
+| `Vertex` | 0-dimensional points in the reference shape. |
+| `Edge`   | 1-dimensional entities connecting two vertices. |
+| `Face`   | 2-dimensional entities whose boundary is defined by 3 or more edges. |
+| `Volume` | 3-dimensional entity enclosed by faces. |
+
+Note that a `Node` in Ferrite is not the same as a `Vertex`.
+Vertices denote endpoints of edges, while nodes may also be be located in the middle
+of edges (e.g. for a `QuadraticLine` cell).
```
termi-official, line 27:
```suggestion
Note that a `Node` in Ferrite is not the same as a `Vertex`.
Vertices denote endpoints of lines and the corners shapes with dimension larger than one, while nodes may also be located in the interior of edges, faces and volumes, acting as control points for the (geometric) non-linearity. Think for example about a `QuadraticLine` cell which has one node at the center of an edge and two more nodes located at the vertices/endpoints.
```

I think we can expand on this a bit, since this is really important to understand for users. I would suggest something in this direction, but it needs further refinement I think.

fredrikekre: I thought that for interpolations we considered the middle point as a vertex and that node was just "points in a grid"...

termi-official: It is a point in the grid because we enforce nodal interpolations for the geometry. :P

KnutAM: I think of (maybe influenced by @termi-official) that we have two types of nodes
* Geometric nodes: Points in the grid (technically support points for the geometric interpolation).
* Algebraic nodes: Support points for nodal interpolations (e.g. Lagrange interpolations, but not Nedelec), as referred to [here](https://finite-element.github.io/L2_fespaces.html#ciarlet-s-finite-element). The coordinates of these nodes are not explicitly defined in the grid, but obtained by mapping their location in the reference point by the geometric interpolation of the cell. 

I've thought that it would make sense to put this in a separate doc. I don't want to put the emphasis here, since that is too hidden and takes away focus from the main point (dealing with reference shapes, and nodes haven't been introduced yet). 

I'll open a separate issue where we could list which terms could be defined on a "defintions" page, which would be good to have IMO.

termi-official: > I think of (maybe influenced by @termi-official) that we have two types of nodes
> 
>     * Geometric nodes: Points in the grid (technically support points for the geometric interpolation).
> 
>     * Algebraic nodes: Support points for nodal interpolations (e.g. Lagrange interpolations, but not Nedelec), as referred to [here](https://finite-element.github.io/L2_fespaces.html#ciarlet-s-finite-element). The coordinates of these nodes are not explicitly defined in the grid, but obtained by mapping their location in the reference point by the geometric interpolation of the cell.
> 

*achievement unlocked* :sunglasses: 

> 
> I've thought that it would make sense to put this in a separate doc. I don't want to put the emphasis here, since that is too hidden and takes away focus from the main point (dealing with reference shapes, and nodes haven't been introduced yet).
> 
> I'll open a separate issue where we could list which terms could be defined on a "defintions" page, which would be good to have IMO.

Makes sense. Should we then just revert the change here and add to the issue that we might want to put a reference in this page to the explanation for the nodes?

KnutAM: I think the short note here is fine for now, and once we have a good list we can shorten it more or even remove it completely and rather link to the definitions page. 

> I'll open a separate issue

#944

docs/make.jl
KnutAM: I don't want to add a lot of rename to this PR, but focus on getting the new notation documented. IMO we should go over the docs and use "reference shape" consistently in another PR. I'll open (yet another issue) for that, so that we don't forget (also a nice first-time pr if we reach consensus in the issue discussion)

docs/src/topics/index.md
```diff
@@ -6,6 +6,8 @@ This is an overview of the *topic guides*.
 ```@contents
 Pages = [
     "fe_intro.md",
+    "reference_shapes",
+    "FEValues",
```
fredrikekre, line 10: Need `.md`?

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+
+| Entity   | Description |
+| :------- | :---------- |
+| `Vertex` | 0-dimensional points in the reference shape. |
+| `Edge`   | 1-dimensional entities connecting two vertices. |
+| `Face`   | 2-dimensional entities whose boundary is defined by 3 or more edges. |
+| `Volume` | 3-dimensional entity enclosed by faces. |
+
+Note that a `Node` in Ferrite is not the same as a `Vertex`.
+Vertices denote endpoints of edges, while nodes may also be be located in the middle
```
KnutAM, line 26:
```suggestion
Vertices denote endpoints of edges, while nodes may also be located in the middle
```

docs/make.jl
termi-official: Sure, let us split these issues into smaller ones and just keep track of them like you propose. Thanks.

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Entity naming
+Ferrite denotes the entities of a reference shape as follows
+
+| Entity   | Description |
+| :------- | :---------- |
+| `Vertex` | 0-dimensional points in the reference shape. |
+| `Edge`   | 1-dimensional entities connecting two vertices. |
+| `Face`   | 2-dimensional entities whose boundary is defined by 3 or more edges. |
+| `Volume` | 3-dimensional entity enclosed by faces. |
```
fredrikekre, line 23:
```suggestion
| Vertex | 0-dimensional entity in the reference shape. |
| Edge   | 1-dimensional entity connecting two vertices. |
| Face   | 2-dimensional entity enclosed by edges. |
| Volume | 3-dimensional entity enclosed by faces. |
```

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+| :------- | :---------- |
+| `Cell`   | *0-codimensional* entity, i.e. the same as the reference shape. |
+| `Facet`  | *1-codimensional* entity defining the boundary of cells. |
+
+Standard use cases mostly deal with these codimensional entities,
+such as [`CellValues`](@ref) and [`FacetValues`](@ref).
+
+!!! note "Definition of codimension"
+    In Ferrite, *codimension* is defined relative to the reference dimension of the specific entity.
+    Note that other Finite Element codes may define it differently
```
fredrikekre, line 43:
```suggestion
    Note that other finite element codes may define it differently
```

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+Ferrite denotes the entities of a reference shape as follows
+
+| Entity   | Description |
+| :------- | :---------- |
+| `Vertex` | 0-dimensional points in the reference shape. |
+| `Edge`   | 1-dimensional entities connecting two vertices. |
+| `Face`   | 2-dimensional entities whose boundary is defined by 3 or more edges. |
+| `Volume` | 3-dimensional entity enclosed by faces. |
+
+Note that a `Node` in Ferrite is not the same as a `Vertex`.
```
fredrikekre, line 25:
```suggestion
Note that a node in Ferrite is not the same as a vertex.
```

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+such as [`CellValues`](@ref) and [`FacetValues`](@ref).
+
+!!! note "Definition of codimension"
+    In Ferrite, *codimension* is defined relative to the reference dimension of the specific entity.
+    Note that other Finite Element codes may define it differently
+    (e.g. relative the highest reference dimension in the grid).
+
+## Entity numbering
+Each reference shape defines the numbering of its `Vertex`, `Edge`, and `Face` entities,
+where the `Edge` and `Face` entities are defined from their `Vertex` numbers.
```
fredrikekre, line 48:
```suggestion
Each reference shape defines the numbering of its vertices, edges, and faces entities,
where the edge and face entities are defined from their vertex numbers.

!!! note
    The numbering and identification of entities is (mostly) for internal use and typically
    not something users of Ferrite need to interact with.
```

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+Ferrite.reference_vertices(RefQuadrilateral)
+```
+and its edges are then defined as
+```@example
+using Ferrite #hide
+Ferrite.reference_edges(RefQuadrilateral)
+```
+where the numbers refer to the vertex number.
+Finally, this reference shape is 2-dimensional, so it only has a single `face`,
+corresponding to the `Cell`,
```
fredrikekre, line 68:
```suggestion
Finally, this reference shape is 2-dimensional, so it only has a single face,
corresponding to the cell itself,
```

docs/src/topics/reference_shapes.md
```diff
@@ -0,0 +1,85 @@
...
+```@example
+using Ferrite #hide
+Ferrite.reference_faces(RefQuadrilateral)
+```
+also defined in terms of its vertices.
+
+As this is a 2-dimensional reference shape, the `facets` are the `edges`, i.e.
```
fredrikekre, line 75:
```suggestion
As this is a 2-dimensional reference shape, the facets are the edges, i.e.
```

Issue #941, opened 2024-05-23
fix topology getneighborhood call in 2D


koehlerson:
The renaming of the entities leads to a wrong code branch for `getneighborhood(top,grid,facetidx)` if the grid is in 2D. This is due to the fact that the `facetidx` is converted to an `edgeidx` and then an edge neighborhood reconstruction is built which was only designed for the three dimensional case.

https://github.com/Ferrite-FEM/Ferrite.jl/blob/68e1ab15bea4618f76b9ed1d850e2ce33375e266/src/Grid/topology.jl#L282-L292

What should be done is this:

https://github.com/Ferrite-FEM/Ferrite.jl/blob/68e1ab15bea4618f76b9ed1d850e2ce33375e266/src/Grid/topology.jl#L259-L263

but with the `edge_edge` table in 2D. Some dispatch distinction for 2D and 3D for an `EdgeIndex` should be enough. Some regression testing should be included. It could be that the first, wrong code branch returns the same as the intended one, problem for me is that there are assumptions about the grid type that I don't fulfill in the p4est branch, and that's why I noticed this. It should be changed anyway for performance reasons.

koehlerson:
Context for the first code branch is that some edge neighborhood is reconstructed, since the table only stores the exclusively connected element via the edge. However, there are more elements connected by this edge (but also by a face)

KnutAM:
I'm not sure if I understand when this will fail. Which assumptions are you talking about? 
Do you have a specific case showing the that we could test for? 

Regarding the performance, that is clear
```julia
grid = generate_grid(Quadrilateral, (2,2));
top = ExclusiveTopology(grid);
```
Latest release (btiming: `55.127 ns (2 allocations: 128 bytes)`)
```julia
julia> getneighborhood(top, grid, FaceIndex(1, 2), true)
2-element Vector{FaceIndex}:
 FaceIndex((2, 4))
 FaceIndex((1, 2))
```
Master (btiming: `363.636 ns (19 allocations: 1.50 KiB)`)
```julia
julia> getneighborhood(top, grid, FacetIndex(1, 2), true)
2-element Vector{EdgeIndex}:
 EdgeIndex((2, 4))
 EdgeIndex((1, 2))
```

koehlerson:
> I'm not sure if I understand when this will fail. Which assumptions are you talking about? Do you have a specific case showing the that we could test for?

[...]

PR #942, opened 2024-05-23
add dispatch for 2D edgeidx topology

koehlerson:
Fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/941. I don't know how to test if the correct codepath is tracked, since the outcome is the same for both code paths.


KnutAM:
Is this more than an optimization for grids with `sdim=2` that there it is assumed that each `edge` is only shared between 2 cells? 

I.e. how does this play with embedded shells, when we have `AbstractGrid{3}` but `rdim=2`? Would that lead to the same error that you had for the AMR?

koehlerson:
In future aspects, this is also a bugfix. However, we don't support currently mixed dimensional types in the topology. This was excluded for performance reasons, but can be brought back by adding mixed spatial tables https://github.com/Ferrite-FEM/Ferrite.jl/issues/843

termi-official:
> In future aspects, this is also a bugfix. However, we don't support currently mixed dimensional types in the topology. This was excluded for performance reasons, but can be brought back by adding mixed dimension tables #843

Technically speaking, the latest changes should allow type stable queries on mixed dimensional grids.

KnutAM:
> In future aspects, this is also a bugfix.

I'm not following this though, because with the changes in #914, the topology is now completely independent of the spatial dimension of the grid. So AFAIU, the only point of this change is to speed up the computation when we can assume that each edge is only shared between two cells. 
This assumption would be broken in the future when supporting embedded elements, since we could then have a membrane (Line) element B(2-3) like this 
```julia
4 -- 3 3 3 -- 6
| A  | | | C  |
1 -- 2 2 2 -- 5
```
(The same would of course apply to faces, and that needs to be generalized to work on mixed grids, AFAIU the opposite of what this PR does to faces)

Actually, the above example works on master but gives the wrong neighborhood with this branch
```julia
julia> nodes = [Node(Float64.(x)) for x in ((0,0), (1,0), (1,1), (0,1), (2,0), (2,1))];

julia> cells = [Quadrilateral((1, 2, 3, 4)), Line((2, 3)), Quadrilateral((2, 5, 6, 3))];

julia> grid = Grid(cells, nodes);

julia> top = ExclusiveTopology(grid);

julia> getneighborhood(top, grid, EdgeIndex(2,1), true)
3-element Vector{EdgeIndex}:
 EdgeIndex((3, 4))
 EdgeIndex((1, 2))
 EdgeIndex((2, 1))

julia> getneighborhood(top, grid, EdgeIndex(1,2), true)
3-element Vector{EdgeIndex}:
 EdgeIndex((3, 4))
 EdgeIndex((2, 1))
 EdgeIndex((1, 2))

# By purpose not allowed for grids with mixed reference dimensions:
julia> getneighborhood(top, grid, FacetIndex(1,2), true)
ERROR: ArgumentError: getneighborhood with FacetIndex is is only supported for grids containing cells with a common reference dimension.
...
```

KnutAM:
What could be possible though, is instead of dispatching on the `sdim` would be to querry the reference dimension of the grid, something like this perhaps 
```julia
function getneighborhood(top::ExclusiveTopology, grid::AbstractGrid, edgeidx::EdgeIndex, include_self=false)
    if get_reference_dimension(grid) == 2
        if include_self
        return [top.edge_edge_neighbor[edgeidx[1],edgeidx[2]].neighbor_info; edgeidx]
        else
            return top.edge_edge_neighbor[edgeidx[1],edgeidx[2]].neighbor_info
        end
    else
        # generic code as before
    end
end
```
`get_reference_dimension` will return `:mixed` if there are different reference dimensions. However, it is only "free of cost" to call for concrete cell types (or union cell types with the same reference dimension). So in that case, it might make more sense to call this during `ExclusiveTopology` construction, and save it. (For type stability could perhaps save it as `Symbol(get_reference_dimension)`)

koehlerson:
> > In future aspects, this is also a bugfix.
> 
> I'm not following this though, because with the changes in #914, the topology is now completely independent of the spatial dimension of the grid. So AFAIU, the only point of this change is to speed up the computation when we can assume that each edge is only shared between two cells. This assumption would be broken in the future when supporting embedded elements, since we could then have a membrane (Line) element B(2-3) like this
> 
> ```julia
> 4 -- 3 3 3 -- 6
> | A  | | | C  |
> 1 -- 2 2 2 -- 5
> ```
> 
> (The same would of course apply to faces, and that needs to be generalized to work on mixed grids, AFAIU the opposite of what this PR does to faces)
> 
> Actually, the above example works on master (shown below) but gives the wrong neighborhood with this branch
> 
> ```julia
> julia> nodes = [Node(Float64.(x)) for x in ((0,0), (1,0), (1,1), (0,1), (2,0), (2,1))];
> 
> julia> cells = [Quadrilateral((1, 2, 3, 4)), Line((2, 3)), Quadrilateral((2, 5, 6, 3))];
> 
> julia> grid = Grid(cells, nodes);
> 
> julia> top = ExclusiveTopology(grid);
> 
> julia> getneighborhood(top, grid, EdgeIndex(2,1), true)
> 3-element Vector{EdgeIndex}:
>  EdgeIndex((3, 4))
>  EdgeIndex((1, 2))
>  EdgeIndex((2, 1))
> 
> julia> getneighborhood(top, grid, EdgeIndex(1,2), true)
> 3-element Vector{EdgeIndex}:
>  EdgeIndex((3, 4))
>  EdgeIndex((2, 1))
>  EdgeIndex((1, 2))
> 
> # By purpose not allowed for grids with mixed reference dimensions:
> julia> getneighborhood(top, grid, FacetIndex(1,2), true)
> ERROR: ArgumentError: getneighborhood with FacetIndex is is only supported for grids containing cells with a common reference dimension.
> ...
> ```

That example works because of the happy coincidence that the code branch does the correct thing even though it was intended for something different. This doesn't mean that mixed dimensional grids work in general on master.

From the test suite (currently commented out due to unsupport of mixed dimensional grids)
```julia
  cells = [
        Hexahedron((1, 2, 3, 4, 5, 6, 7, 8)),
        Hexahedron((11, 13, 14, 12, 15, 16, 17, 18)),
        Quadrilateral((2, 9, 10, 3)),
        Quadrilateral((9, 11, 12, 10)),
        ]
    nodes = [Node(coord) for coord in zeros(Vec{3,Float64}, 18)]
    grid = Grid(cells, nodes)
    topology = ExclusiveTopology(grid)
    getneighborhood(top,grid,FaceIndex(3,4)) # returns nothing, should be EdgeIndex(1,2)
 ```

**Comment by koehlerson** [comment]
If you are sure that the code branch always does the correct thing for every possible mixed dimensional grid we want to support, then it's okay to leave it as is and close the PR. I'm just not sure that's the case since I wrote it with something different in mind

**Comment by KnutAM** [comment]
> If you are sure that the code branch always does the correct thing for every possible mixed dimensional grid we want to support, then it's okay to leave it as is and close the PR. I'm just not sure that's the case since I wrote it with something different in mind

No, I'm not sure :) However, it seems to me (with the above example) that this PR doesn't make it more correct, or do I miss something? As I understand, the 3d code does not assume that there is only one edge, while the 2d code does.

Another option, for solving the issue with AMR, could even be to dispatch only the case with `FacetIndex` to the special branch, since when called with `FacetIndex`, it is guaranteed to not be embedded.

File: src/Grid/topology.jl [hunk 1]
```diff
@@ -263,6 +263,14 @@ function getneighborhood(top::ExclusiveTopology, grid::AbstractGrid, faceidx::Fa
     end
 end
 
+function getneighborhood(top::ExclusiveTopology, grid::AbstractGrid{2}, edgeidx::EdgeIndex, include_self=false)
```
termi-official, line 266: I guess it comes down to this
```suggestion
function getneighborhood(top::ExclusiveTopology, grid::AbstractGrid{2,<: Union{<:Ferrite.AbstractCell{<:Ferrite.AbstractRefShape{2}}}}, edgeidx::EdgeIndex, include_self=false)
```

koehlerson:
It makes it more correct in the sense of restoring exactly the code branch that has been called before https://github.com/Ferrite-FEM/Ferrite.jl/pull/914 I will go in the AMR branch probably with `get_facet_facet_neighborhood` since I just need the tables and nothing reconstructed there

KnutAM:
> It makes it more correct in the sense of restoring exactly the code branch that has been called before https://github.com/Ferrite-FEM/Ferrite.jl/pull/914 I will go in the AMR branch probably with get_facet_facet_neighborhood since I just need the tables and nothing reconstructed there.

In that case we should error on `ExclusiveTopology` (as before 914) until #843 is fixed, which sounds like a good solution to me. With that additional change, I think this PR good! But of course, considering the discussion I wouldn't count on it to stay after 843, but I could be wrong here!

Concretely, I suggest changing https://github.com/Ferrite-FEM/Ferrite.jl/blob/cd82d86621d97ecf17cb2e25b6c7cdedeec4305a/src/Grid/topology.jl#L247

to 

```julia

function ExclusiveTopology(grid::AbstractGrid)
    get_reference_dimension(grid) === :mixed && throw(ArgumentError("Exclusive topology is currently not supported for grids containing cells with different reference dimensions"))
    return ExclusiveTopology(getcells(grid)) 
end
```

KnutAM (approved):
Sorry for misunderstanding this earlier, after working more with the topology, I think I understand the problem :)

In #974 I enforce rdim=sdim for all cells, so IMO we can skip that here and do it there instead, and this can from my perspective be merged as-is!

PR #943, opened 2024-05-23
replace getdim with getrefdim/getspatialdim and getfielddim by n_components

### LINKED ISSUE #936: getdim consistency
Currently we have 
```julia
[1] getdim(g::Ferrite.AbstractGrid)
[2] getdim(::Union{Ferrite.AbstractCell{refshape}, Type{<:Ferrite.AbstractCell{refshape}}}) where {refdim, refshape<:Ferrite.AbstractRefShape{refdim}}
[3] getdim(::Interpolation{shape}) where {dim, shape<:Ferrite.AbstractRefShape{dim}}
```
from `methods(Ferrite.getdim)`, but only [3] is documented. [2,3] returns the reference dim and [1] returns `g's` spatial dimension. 

I would suggest replacing [1] by `getsdim(::AbstractGrid)` (public, not exported)

For consistency, `getrdim` would make sense for [2,3], but probably an uncessary breaking change and ok if we don't mix definitions. 

(After #914, we have the internal `Ferrite.get_reference_dimension(::AbstractGrid)`, but this may also return `:mixed` if there are cells with different rdims in the grid, so probably not good as public api, and can in any case be needed later). 

Thoughts?

KnutAM:
Closes #936

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -225,7 +225,7 @@ end
     getfieldnames(dh::DofHandler)
     getfieldnames(sdh::SubDofHandler)
 
-Return a vector with the unique names of all fields. The order is the sam eas the order in
+Return a vector with the unique names of all fields. The order is the same as the order in
 which they were originally added to the (Sub)DofHandler. Can be used as an iterable over all
 the fields.
 """
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -141,7 +141,14 @@ function Base.show(io::IO, mime::MIME"text/plain", dh::DofHandler)
     else
         println(io, "  Fields:")
         for fieldname in getfieldnames(dh)
-            println(io, "    ", repr(fieldname), ", dim: ", n_components(dh, fieldname))
+            ip = getfieldinterpolation(dh, find_field(dh, fieldname))
+            if ip isa ScalarInterpolation
+                field_type = "scalar"
+            elseif ip isa VectorInterpolation
+                _getvdim(::VectorInterpolation{vdim}) where vdim = vdim
+                field_type = "Vec{$(_getvdim(ip))}"
+            end
+            println(io, "    ", repr(fieldname), ", ", field_type)
         end
     end
     if !isclosed(dh)
@@ -239,8 +246,8 @@ its index (see [`find_field`](@ref)) or its name.
 """
 function n_components(dh::DofHandler, field_idxs::NTuple{2, Int})
     sdh_idx, field_idx = field_idxs
-    fielddim = n_components(dh.subdofhandlers[sdh_idx], field_idx)
-    return fielddim
+    n = n_components(dh.subdofhandlers[sdh_idx], field_idx)
+    return n
 end
 n_components(dh::DofHandler, name::Symbol) = n_components(dh, find_field(dh, name))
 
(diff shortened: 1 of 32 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -141,7 +141,7 @@ function Base.show(io::IO, mime::MIME"text/plain", dh::DofHandler)
     else
         println(io, "  Fields:")
         for fieldname in getfieldnames(dh)
-            println(io, "    ", repr(fieldname), ", dim: ", getfielddim(dh, fieldname))
+            println(io, "    ", repr(fieldname), ", dim: ", n_components(dh, fieldname))
```
fredrikekre, line 144: I guess this should be split into e.g. "scalar field" instead of `dim: 1` then?

fredrikekre (approved):

KnutAM:
I think this can be merged as-is, despite the issues below (just causes same docstring to show up in different places in the docs)

However, filtering away the `getrefdim`'s docstring for `Interpolation` and `AbstractRefShape` input doesn't work... Seems like a julia bug

```julia
julia> @doc Ferrite.getrefdim
  Ferrite.getrefdim(RefShape::Type{<:AbstractRefShape})

  Get the dimension of the reference shape

  Ferrite.getrefdim(::Interpolation)

  Return the dimension of the reference element for a given interpolation.

  Ferrite.getrefdim(cell::AbstractCell)
  Ferrite.getrefdim(::Type{<:AbstractCell})

  Get the reference dimension of the cell, i.e. the dimension of the cell's reference shape.

julia> @doc Ferrite.getrefdim(::Interpolation)
  Ferrite.getrefdim(RefShape::Type{<:AbstractRefShape})

  Get the dimension of the reference shape

  Ferrite.getrefdim(::Interpolation)

  Return the dimension of the reference element for a given interpolation.

julia> @doc Ferrite.getrefdim(::Ferrite.AbstractCell)
  Ferrite.getrefdim(RefShape::Type{<:AbstractRefShape})

  Get the dimension of the reference shape

  Ferrite.getrefdim(::Interpolation)

  Return the dimension of the reference element for a given interpolation.

  Ferrite.getrefdim(cell::AbstractCell)
  Ferrite.getrefdim(::Type{<:AbstractCell})

  Get the reference dimension of the cell, i.e. the dimension of the cell's reference shape.

julia> @doc Ferrite.getrefdim(::Type{<:Ferrite.AbstractRefShape})
  Ferrite.getrefdim(RefShape::Type{<:AbstractRefShape})

  Get the dimension of the reference shape

  Ferrite.getrefdim(::Interpolation)

  Return the dimension of the reference element for a given interpolation.

julia> typespecs = dict[Base.Docs.Binding(Ferrite, :getrefdim)].order;

julia> typespecs = dict[Base.Docs.Binding(Ferrite, :getrefdim)].order
3-element Vector{Type}:
 Union{Tuple{Type{<:Ferrite.AbstractRefShape{rdim}}}, Tuple{rdim}} where rdim
 Union{Tuple{Interpolation{RefShape}}, Tuple{RefShape}} where RefShape
 Tuple{Ferrite.AbstractCell}
```

where 
* `Union{Tuple{Interpolation{RefShape}}, Tuple{RefShape}} where RefShape` 
* `Union{Tuple{Type{<:Ferrite.AbstractRefShape{rdim}}}, Tuple{rdim}} where rdim`
matches any input, as `RefShape` or `rdim`, respectively, can be anything. 

The offending function definitions are 
* `@inline getrefdim(::Interpolation{RefShape}) where RefShape = getrefdim(RefShape)`
* `getrefdim(::Type{<:AbstractRefShape{rdim}}) where rdim = rdim`

PR #938, opened 2024-05-22
Higher order gradients in CellValues

lijas:
New version of #891 so that I can see preview of docs. I also needed to rebase on master

Implementation of higher order gradients in CellValues. This was quite easy to do thanks to Knuts new FunctionValues and GeometryMapping.

- [x] Tests for FaceValues
- [x] Update changelog
- [x] Type stable CellValues constructor (?)

PR #947, opened 2024-05-24
Simplify and speedup shape_gradient_and_value for embedded

KnutAM:
Noticed this when reviewing #938 

```julia
ip = Lagrange{RefQuadrilateral,1}()^3; ξ = rand(Vec{2});
@btime Ferrite.shape_gradient_and_value($ip, $ξ, 1);
# PR:     15.030 ns (0 allocations: 0 bytes)
# master: 26.459 ns (1 allocation: 64 bytes)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.3.14"
 
 [deps]
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
+ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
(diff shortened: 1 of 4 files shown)


termi-official:
Since this is already done I am okay with merging this, but shouldn't we rather try to finalize https://github.com/Ferrite-FEM/Tensors.jl/pull/188#issuecomment-1713268821 ?

termi-official (approved):

fredrikekre:
Need to re-resolve the doc manifest to include ForwardDiff as a dependency.

KnutAM:
> but shouldn't we rather try to [Tensors.jl#188](https://github.com/Ferrite-FEM/Tensors.jl/pull/188#issuecomment-1713268821)

I would love to, but need quite some input on https://github.com/Ferrite-FEM/Tensors.jl/pull/211 before doing a managable implementation of https://github.com/Ferrite-FEM/Tensors.jl/pull/188 (and that will probably be after Ferrite 1.0)

I simply did this to check my understanding of the code, since @lijas re-uses this code in https://github.com/Ferrite-FEM/Ferrite.jl/pull/938. So perhaps not clear from the PR, but since I then had this code the main merit is to simplify for others to see what is happening here, instead of relying on Tensor internals. (On the other hand, this code was probably beneficial for generalizing to the 3rd order case, but will look at that now if a similiar simplification can be suggested in 938).

PR #948, opened 2024-05-25
Fix and test write_cellset

KnutAM:
Writing all cellsets was not covered by the tests and contained a bug

### DIFF
--- a/test/test_vtk_export.jl
+++ b/test/test_vtk_export.jl
@@ -42,8 +42,8 @@
         end
     end
     @testset "write_cellset" begin
-        # sha1 tested in test_grid_dofhandler_vtk.jl, this just smoke tests that we get the same
-        # when taking all cellsets in the grid
+        # More tests in `test_grid_dofhandler_vtk.jl`, this just validates writing all sets in the grid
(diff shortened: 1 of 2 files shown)


test/test_vtk_export.jl
```diff
@@ -41,4 +41,22 @@
...
+        # More tests in `test_grid_dofhandler_vtk.jl`, this just validates writing all sets in the grid
+        # which is not tested there, see https://github.com/Ferrite-FEM/Ferrite.jl/pull/948
+        mktempdir() do tmp
+            grid = generate_grid(Quadrilateral, (2,2))
+            addcellset!(grid, "set1", 1:2)
+            addcellset!(grid, "set2", 1:4)
+            manual = joinpath(tmp, "manual")
+            auto = joinpath(tmp, "auto")
+            VTKFile(manual, grid) do vtk
+                Ferrite.write_cellset(vtk, grid, keys(Ferrite.getcellsets(grid)))
```
KnutAM, line 54: I'm not sure if this is the best test, since the order is not part of the API, and while this test works now, it would fail if we decide to e.g. sort the keys in `write_cellset`. 
I'd be happy to hear if someone has ideas for a better test!

PR #949, opened 2024-05-26
WIP: L2Projector on mixed grids

KnutAM:
A proposal to solve #610, which mimicks the `DofHandler` syntax. 
To make this convenient, the following breaking changes are introduced.

* When projecting on a subset, the data should be **indexed with the cellnr and not the index in the set**. To avoid requiring passing a `Vector` with length equal to all cells, passing the quadrature point data of type `T` as `AbstractDict{Int, AbstractVector{T}}` is supported. 
* `L2Projector` no longer supports getting the geometric interpolation, will always take the one from the cell type (I don't think this is even breaking, because anything else would probably have failed) 

Syntax for mixed grids 
```julia
proj = L2Projector(grid)
add!(proj, ip_tria, set_tria)
add!(proj, ip_quad, set_quad)
close!(proj)

projected = project(proj, data, [qr_tria, qr_quad])
```


src/L2_projection.jl
```diff
@@ -71,21 +67,22 @@ function _mass_qr(::Lagrange{shape, 2}) where {shape <: RefSimplex}
...
-    println(io, "  geometric interpolation: ", proj.geom_ip)
-end
-
-function _assemble_L2_matrix(fe_values, set, dh)
-
-    n = Ferrite.getnbasefunctions(fe_values)
+function _assemble_L2_matrix(dh::DofHandler, qrs_lhs::Vector{<:QuadratureRule})
     M = create_symmetric_sparsity_pattern(dh)
     assembler = start_assemble(M)
+    for (sdh, qr_lhs) in zip(dh.subdofhandlers, qrs_lhs)
```
termi-official, line 73: How does this ensure that the pairing is correct? Think about having 2 subdomains and the 2nd subdomain has triangles and quads, so we have 3 sdhs. If I understand this correctly, then this code will fail if we want to do an L2 projection on the second subdomain.

lijas: Then `qrs_lhs` would include three quadrature rules, one for each subdofhandler?

src/L2_projection.jl
```diff
@@ -71,21 +67,22 @@ function _mass_qr(::Lagrange{shape, 2}) where {shape <: RefSimplex}
...
+        ip_fun = only(sdh.field_interpolations)
+        ip_geo = default_interpolation(getcelltype(sdh))
+        cv = CellValues(qr_lhs, ip_fun, ip_geo)
```
lijas, line 76: Would it possible to pass `update_detJdV=true` and `update_gradient = false` here ?

src/L2_projection.jl
```diff
@@ -160,68 +153,75 @@ The order of the returned data correspond to the order of the `L2Projector`s int
...
+end
+function project(p::L2Projector, vars::AbstractMatrix, qr_rhs)
     # TODO: Random access into vars is required for now, hence the collect
     return project(p, collect(eachcol(vars)), qr_rhs)
 end
 
-function _project(vars, proj::L2Projector, fe_values::AbstractValues, M::Integer, ::Type{T}) where {T}
+function _project(proj::L2Projector, qrs_rhs::Vector{<:QuadratureRule}, vars::Union{AbstractVector, AbstractDict}, M::Integer, ::Type{T}) where T
+    f = zeros(ndofs(proj.dh), M)
+    for (sdh, qr_rhs) in zip(proj.dh.subdofhandlers, qrs_rhs)
```
lijas, line 177: Maybe add a checks here that checks that length(qrs_rhs) as number of length(subdofhandle)r (and the same `refshape`)? CellValues will probable throw an error anyways, but could be nice to catch it earlier.

test/test_l2_projection.jl
```diff
@@ -121,7 +120,12 @@ function test_projection_mixedgrid()
 
     quadset = 1:1
     triaset = 2:3
-    mesh = Grid(cells, nodes)
+    return Grid(cells, nodes), quadset, triaset
```
lijas, line 123: Should we add a test where we have a third cell set which is excluded from the L2-projection? It seems like it should work, but would be nice test.

lijas (commented):
Nice! I like the api, and the code even looks simpler now.

src/L2_projection.jl
KnutAM: I think the updated syntax solves this better (see updated comment)

termi-official: Okay, probably a different issue than the one resolved in this PR. I was wondering about an L2 projection on true subdomains with mixed grids (and a mixed grid on the full domain).

lijas (approved):

Issue #950, opened 2024-05-27
default_interpolation -> geometric_interpolation ? 


KnutAM:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/41cfff526912c64ebe41db83cbe7315e454d6660/src/Grid/grid.jl#L134-L139

To me this function name doesn't reflect the docstring (which describes what the function actually does IMO).
Wouldn't it be better to call this `geometric_interpolation`, otherwise I find it confusing as `default` indicates that there are other alternative interpolations for a given cell? 

Or are there any cases that I'm not aware of when different geometric interpolations can be used for the same cell? If not, I'd suggest changing this in 1.0. 

(The old name could be kept to avoid making things more breaking for users, but perhaps removed from API. This would then be to just get a default for the function interpolation on a cell, which I guess the function was originally intended for).

KnutAM:
When? 
```
julia> using Ferrite
julia> grid = generate_grid(QuadraticQuadrilateral, (2,2));
julia> cv = CellValues(QuadratureRule{RefQuadrilateral}(2), Lagrange{RefQuadrilateral,1}(), Lagrange{RefQuadrilateral,1}());

julia> x = getcoordinates(grid, 1);

julia> reinit!(cv, x)
ERROR: ArgumentError: the number of (geometric) base functions (4) does not match the number of coordinates in the vector (9). Perhaps you forgot to use an appropriate geometric interpolation when creating FE values? See https://github.com/Ferrite-FEM/Ferrite.jl/issues/265 for more details.
```
Or is there another context you refer to?

termi-official:
> https://github.com/Ferrite-FEM/Ferrite.jl/blob/41cfff526912c64ebe41db83cbe7315e454d6660/src/Grid/grid.jl#L134-L139
> 
> To me this function name doesn't reflect the docstring (which describes what the function actually does IMO). Wouldn't it be better to call this `geometric_interpolation`, otherwise I find it confusing as `default` indicates that there are other alternative interpolations for a given cell?
> 
> Or are there any cases that I'm not aware of when different geometric interpolations can be used for the same cell? If not, I'd suggest changing this in 1.0.
> 
> (The old name could be kept to avoid making things more breaking for users, but perhaps removed from API. This would then be to just get a default for the function interpolation on a cell, which I guess the function was originally intended for).

[...]

PR #953, opened 2024-05-27
Rename default_interpolation -> geometric_interpolation

KnutAM:
Ref discussion in #950 

* Exports and documents `geometric_interpolation`
* "soft"-deprecates `default_interpolation` with single warning per session
* Clarifies docstring: Each cell has a unique geometric interpolation, and `geometric_interpolation` returns this. 

Note that `geometric_interpolation(::AbstractValues)` also exists, but this also returns the geometric interpolation, so I think this is fine. 

03039fe263cc7505fc5b50e8af04347844f19c3c contains the pure rename, other fine-tunings etc in remaining commits in case we should add this to git-blame ignored commits, @fredrikekre

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -372,3 +372,8 @@ end
 
 getdim(args...) = error("`Ferrite.getdim` is deprecated, use `getrefdim` or `getspatialdim` instead")
 getfielddim(args...) = error("`Ferrite.getfielddim(::AbstractDofHandler, args...) is deprecated, use `n_components` instead")
+
+function default_interpolation(cell::AbstractCell)
+    @warn "Ferrite.default_interpolation is deprecated, use the exported `geometric_interpolation` instead" maxlog=1
+    return geometric_interpolation(cell)
+end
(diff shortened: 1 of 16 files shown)


fredrikekre:
> https://github.com/Ferrite-FEM/Ferrite.jl/commit/03039fe263cc7505fc5b50e8af04347844f19c3c contains the pure rename, other fine-tunings etc in remaining commits in case we should add this to git-blame ignored commits, @fredrikekre

I think it is fine to squash this into one commit, but it made it very easy to review so that was nice.

src/deprecations.jl
```diff
@@ -372,3 +372,8 @@ end
 
 getdim(args...) = error("`Ferrite.getdim` is deprecated, use `getrefdim` or `getspatialdim` instead")
 getfielddim(args...) = error("`Ferrite.getfielddim(::AbstractDofHandler, args...) is deprecated, use `n_components` instead")
+
+function default_interpolation(cell::AbstractCell)
```
termi-official, line 376: The deprecation is on the wrong dispatch. This should be `default_interpolation(cell_type::Type{<:AbstractCell)`or something. See the remaining code in the PR.

KnutAM: Thanks, the docstring is also wrong!
```julia
help?> geometric_interpolation
search: geometric_interpolation

  Each AbstractCell type has a unique geometric interpolation describing its geometry. This function returns that
  interpolation.
```

PR #954, opened 2024-05-28
Run ExplicitImport.jl's checks on CI

fredrikekre:


src/Export/VTK.jl
```diff
@@ -49,7 +49,7 @@ end
 Base.close(vtk::VTKFile) = WriteVTK.vtk_save(vtk.vtk)
 
 function Base.show(io::IO, ::MIME"text/plain", vtk::VTKFile)
-    open_str = WriteVTK.isopen(vtk.vtk) ? "open" : "closed"
+    open_str = isopen(vtk.vtk) ? "open" : "closed"
```
KnutAM, line 52: Are these mistakes now detected automatically?

.github/workflows/Check.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches: ["master"]
+
+jobs:
+
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
```
KnutAM, line 14: Why is there a specific commit here and not `@v3.0.1` ?

KnutAM (approved):
Changes in code looks good to me, just some curiosity questions.

src/Export/VTK.jl
fredrikekre: Yes :)

.github/workflows/Check.yml
fredrikekre: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

PR #955, opened 2024-05-28
Changelog entry for geometric_interpolation

### LINKED ISSUE #950: default_interpolation -> geometric_interpolation ?

KnutAM:
Changelog for #953, closes #950.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -317,6 +317,10 @@ more discussion).
   a given grid (based on its node coordinates), and returns the minimum and maximum vertices
   of the bounding box. ([#880][github-880])
 
+- A new function, `geometric_interpolation`, is exported, which gives the geometric interpolation
+  for each cell type. This is equivalent to the deprecated `Ferrite.default_interpolation` function.
+  ([#953][github-953])
+
 ### Changed
 
 - `create_sparsity_pattern` now supports cross-element dof coupling by passing kwarg
@@ -432,6 +436,8 @@ more discussion).
 
 - `transform!` have been deprecated in favor of `transform_coordinates!`.
   ([#754][github-754])
+
+- `Ferrite.default_interpolation` has been deprecated in favor of `geometric_interpolation` [#953][github-953]
 
 ### Removed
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -433,6 +437,8 @@ more discussion).
 - `transform!` have been deprecated in favor of `transform_coordinates!`.
   ([#754][github-754])
 
+- `Ferrite.default_interpolation` has been deprecated in favor of `geometric_interpolation` [#953][github-953]
```
fredrikekre, line 440: I believe `default_interpolation` doesn't exist in the previous release(?) so if you want to mention it here perhaps mention that it has only been on master for quite some time but not in a release

KnutAM: Its available in the latest released version. Are you perhaps thinking about `default_geometric_interpolation(::Interpolation)`?

KnutAM: But only documented in devdocs: https://ferrite-fem.github.io/Ferrite.jl/stable/devdocs/elements/#Ferrite.default_interpolation-Tuple{Ferrite.AbstractCell}

fredrikekre: Hmm, why are these not the same function then? They do the same thing or?

KnutAM: No, because `default_geometric_interpolation` gives the linear interpolation version of the given function interpolation, independently if the supplied function interpolation is higher order or linear. Used to get the default geometric interpolation for the FEValues (#695). 

With #806 we would use `geometric_interpolation` such that it always is correct, since passing the dofhandler gives access to the cell type. 

But I guess the default will in most cases work - if I recall correctly when we discussed this I think the argument was that it is more common to have a quadratic interpolation on a linear grid, than to use a quadratic grid.

fredrikekre: Yea okay, that make sense.

fredrikekre, line 440:
```suggestion
- `Ferrite.default_interpolation` has been deprecated in favor of `geometric_interpolation`. ([#953][github-953])
```

PR #958, opened 2024-05-28
Matrix-valued interpolations

termi-official:
Foundation for #398 . Also, we basically emulate this in FerriteViz.jl when visualizing stress and in the AMR error estimation when computing the stress error. However, I would like to keep this API internal for now.

## TODO

* [ ] Tests


KnutAM:
Very cool to be able to work directly with tensorial fields! 
However, would it make sense to rather talk about tensors instead of just a matrix, we could do e.g. 
```julia
abstract type TensorInterpolation{TB, refshape, iporder} <: Interpolation{refshape, iporder, Nothing} end

struct TensorizedInterpolation{TB, refshape, iporder, SI <: ScalarInterpolation{refshape, iporder}} <: TensorInterpolation{TB, refshape, iporder}
    ip::SI
    function TensorizedInterpolation{TB}(ip::ScalarInterpolation{refshape, iporder}) where {TB<:AbstractTensor{<:Any, <:Any, <:Any}, refshape, iporder}
        return new{TB, refshape, iporder, typeof(ip)}(ip)    
    end
end
```
where `TB` is always `Tensors.get_base(TT::Type{<:AbstractTensor})`. 

This would allow
```
julia> ip = Lagrange{RefTriangle, 1}();

julia> TensorizedInterpolation{Tensor{2,3}}(ip)
TensorizedInterpolation{Tensor{2, 3}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())

julia> TensorizedInterpolation{SymmetricTensor{2,2}}(ip)
TensorizedInterpolation{SymmetricTensor{2, 2}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())
```
This allows re-using e.g. `Tensors.n_components` and even allows us to initialize the correct type of the shape value with e.g. `TB{Float64}`

This will also play nicely with mixed tensors, i.e.
```
julia> struct MixedTensor{order, dims, T, M} <: AbstractTensor{order, dims, T}
           data::NTuple{M, T}
       end

julia> TensorizedInterpolation{MixedTensor{2,(2,3)}}(ip)
TensorizedInterpolation{MixedTensor{2, (2, 3)}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())
```

termi-official:
> Very cool to be able to work directly with tensorial fields! However, would it make sense to rather talk about tensors instead of just a matrix, we could do e.g.
> 
> ```julia
> abstract type TensorInterpolation{TB, refshape, iporder} <: Interpolation{refshape, iporder, Nothing} end
> 
> struct TensorizedInterpolation{TB, refshape, iporder, SI <: ScalarInterpolation{refshape, iporder}} <: TensorInterpolation{TB, refshape, iporder}
>     ip::SI
>     function TensorizedInterpolation{TB}(ip::ScalarInterpolation{refshape, iporder}) where {TB<:AbstractTensor{<:Any, <:Any, <:Any}, refshape, iporder}
>         return new{TB, refshape, iporder, typeof(ip)}(ip)    
>     end
> end
> ```
> 
> where `TB` is always `Tensors.get_base(TT::Type{<:AbstractTensor})`.
> 
> This would allow
> 
> ```
> julia> ip = Lagrange{RefTriangle, 1}();
> 
> julia> TensorizedInterpolation{Tensor{2,3}}(ip)
> TensorizedInterpolation{Tensor{2, 3}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())
> 
> julia> TensorizedInterpolation{SymmetricTensor{2,2}}(ip)
> TensorizedInterpolation{SymmetricTensor{2, 2}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())
> ```
> 
> This allows re-using e.g. `Tensors.n_components` and even allows us to initialize the correct type of the shape value with e.g. `TB{Float64}`
> 
> This will also play nicely with mixed tensors, i.e.
> 
> ```
> julia> struct MixedTensor{order, dims, T, M} <: AbstractTensor{order, dims, T}
>            data::NTuple{M, T}
>        end
> 
> julia> TensorizedInterpolation{MixedTensor{2,(2,3)}}(ip)
> TensorizedInterpolation{MixedTensor{2, (2, 3)}, RefTriangle, 1, Lagrange{RefTriangle, 1, Nothing}}(Lagrange{RefTriangle, 1}())
> ```

Indeed. This is one of the reasons why I would like to keep the API internal for now.I first need to wrap my head around a generic indexing logic and I need to think of a way to properly generalize operations like curl and divergence to be useful by default and to be customizeable if the user works with a slightly different definition.

src/interpolations.jl
```diff
@@ -1535,6 +1536,95 @@ reference_coordinates(ip::VectorizedInterpolation) = reference_coordinates(ip.ip
...
+end
+
+# vdim1 != vdim2 != refdim
+function shape_gradient_and_value(ipm::MatrixizedInterpolation{vdim1, vdim2, shape}, ξ::V, I::Int) where {vdim1, vdim2, refdim, shape <: AbstractRefShape{refdim}, T, V <: Vec{refdim, T}}
+    tosmat(v::Tensor{2}) = SMatrix{vdim1,vdim2}((v...,))
+    tosmat(v::SMatrix) = v
+    tosvec(v::Vec) = SVector((v...,))
+    tovec(sv::SVector) = Vec((sv...))
+    val = shape_value(ipm, ξ, I)
+    grad = ForwardDiff.jacobian(sv -> tosmat(shape_value(ipm, tovec(sv), I)), tosvec(ξ))
```
KnutAM, line 1614: I don't think this will do the right thing unfortunately.
But for these matrix interpolations, it probably is better to calculate the gradients wrt. the scalar base-interpolation, and then just insert the values in the right places (and zeros elsewhere).

KnutAM (commented):
Hm, ok. I'm not sure if I fully understand the purpose? If this is only internal because we expect it to change, then it seems like it could be tested outside as well? I don't see any code that isn't just overloading with new types? 

Or is the purpose to move functionality from FerriteViz to Ferrite?

termi-official:
> If this is only internal because we expect it to change, then it seems like it could be tested outside as well?

What do you mean by outside? If you propose to solve a matrix-valued problem, then I do not have the time currently to implement it. 

> I don't see any code that isn't just overloading with new types?
> 
> Or is the purpose to move functionality from FerriteViz to Ferrite?

We use something very similar to this PR in FerriteViz and this is beneficial to simplify ZZ error estimators. In the future I want to also highlight how we can solve matrix-valued problems (e.g. DG for elasticity) with the API. However, Fredrik repeatedly told me to make smaller PRs to make them more reviewable, which I can fully understand. So here is it. Does this explain it?

src/interpolations.jl
termi-official: Do you have any suggestion how we can cover this case in testing (in addition to the vectorial case)?

KnutAM: I think @lijas solution [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/cf97ee510936259a22ab0e9dee0e423f918112e6/test/test_interpolations.jl#L225-L234) was quite nice, just having the option of calling that function explicitly as setup [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/cf97ee510936259a22ab0e9dee0e423f918112e6/src/interpolations.jl#L1533-L1540). 

Maybe combined with calling something like 
```julia
for Is in 1:getnbasefunctions(ipm.ip)
    g, v = shape_gradient_and_value(ipm.ip, ξ, Is)
    Im = (Is - 1)*getnbasefunctions(ipm.ip)
    for i in 1:vdim1, j in 1:vdim2
        Im += 1
        G, V = shape_gradient_and_value(ipm.ip, ξ, Im)
        @test G[i,j,:] \approx g
        @test V[i,j] \approx v
    end
end
```

PR #961, opened 2024-05-30
use concrete types in `MaterialState` in topology example

KristofferC:
```julia
julia> @time topopt(0.03, 0.5, 60, "large_radius"; output=:false);
# before
#  10.199698 seconds (419.33 M allocations: 19.371 GiB, 23.67% gc time, 0.05% compilation time)
# after
#  6.227020 seconds (170.44 M allocations: 12.847 GiB, 14.20% gc time)
```


termi-official:
This seems to be a very common mistake which I see over and over again in user codes. Would it make sense to compile some performance FAQ or is there a better way to show our users how to do this correctly?

KristofferC:
Yeah, it's pretty annoying, 

Having a performance tips section in the same way as the Julia manual has it seems like a good idea to me.

PR #962, opened 2024-05-30
Rename reference and references

### LINKED ISSUE #945: Docs naming for Reference vs References
Currently we have the *Reference* containing a list of our methods and *References* containing a citation list. We should rename these to be more clear.

termi-official:
Resolves #945


KnutAM (approved):
Nice, I think even without the naming conflict, API reference is much clearer!

PR #963, opened 2024-05-30
cache neighborhood information in topology example

KristofferC:
`getneighborhood` is a quite expensive function that allocates quite a bit and it is called a huge number of times in this example. The neighborhood information is the same in every iteration though so it makes sense to cache it. I don't think it makes the example code much uglier and it is probably more instructive.

```julia
@time topopt(0.03, 0.5, 60, "large_radius"; output=:false);
# Before
#  6.227020 seconds (170.44 M allocations: 12.847 GiB, 14.20% gc time)
# After
#  2.443282 seconds (2.40 M allocations: 2.076 GiB, 1.12% gc time)
```

After this the profile look "clean" (assembly + linear solving is dominating) and the GC time is at least not 10+%.

### DIFF
--- a/docs/src/literate-gallery/topology_optimization.jl
+++ b/docs/src/literate-gallery/topology_optimization.jl
@@ -185,19 +185,19 @@ function compute_densities(states, dh)
 end
 #md nothing # hide
 
-# Now we calculate the Laplacian. For this purpose, we will later create the grid topology of
-# the grid by using the function `ExclusiveTopology`. Then we iterate through each face of each element,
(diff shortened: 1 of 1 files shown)


KristofferC:
Good to go? This includes https://github.com/Ferrite-FEM/Ferrite.jl/pull/961

KnutAM (approved):
LGTM, nice to have it run faster :) 

It is still a bit decieving that a nonlinear routine is used, when the problem is linear just to update the state in the second iteration, but that is another issue.

PR #964, opened 2024-05-30
Expose some missing docstrings to devdocs

termi-official:
Found that some of the docstrings were not exposed to the docs.

### DIFF
--- a/docs/src/devdocs/FEValues.md
+++ b/docs/src/devdocs/FEValues.md
@@ -21,7 +21,7 @@ Ferrite.BCValues
 
 ## Internal utilities
 ```@docs
-Ferrite.embedded_det
+Ferrite.embedding_det
 Ferrite.shape_value_type
 Ferrite.shape_gradient_type
 ```
--- a/docs/src/devdocs/FEValues.md
+++ b/docs/src/devdocs/FEValues.md
@@ -8,6 +8,7 @@
     * [`FacetValues`](@ref)
     * [`BCValues`](@ref)
   * [`PointValues`](@ref)
+  * [`InterfaceValues`](@ref)
 
 
 ## Internal types
@@ -16,6 +17,13 @@ Ferrite.GeometryMapping
 Ferrite.MappingValues
 Ferrite.FunctionValues
 Ferrite.BCValues
+```
+
+## Internal utilities
+```@docs
+Ferrite.embedded_det
+Ferrite.shape_value_type
+Ferrite.shape_gradient_type
 ```
 
 ## Custom FEValues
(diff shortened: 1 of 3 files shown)


docs/src/devdocs/FEValues.md
```diff
@@ -18,6 +19,13 @@ Ferrite.FunctionValues
 Ferrite.BCValues
 ```
 
+## Internal utilities
+```@docs
+Ferrite.embedded_det
```
KnutAM, line 24: Missing? 
https://ferrite-fem.github.io/Ferrite.jl/previews/PR964/devdocs/FEValues/#Internal-utilities

docs/src/devdocs/elements.md
```diff
@@ -25,9 +25,12 @@ Ferrite.get_coordinate_eltype(::Ferrite.AbstractGrid)
 Ferrite.get_coordinate_eltype(::Node)
 Ferrite.toglobal
 Ferrite.sortface
+Ferrite.sortface_fast
 Ferrite.sortedge
+Ferrite.sortedge_fast
```
KnutAM, line 30:
```suggestion
Ferrite.sortface
Ferrite.sortface_fast
Ferrite.sortedge
Ferrite.sortedge_fast
Ferrite.sortfacet_fast
```

These docstrings seems a bit out of date after #914 but I can update them in a separate PR.

docs/src/devdocs/FEValues.md
```diff
@@ -8,6 +8,7 @@
     * [`FacetValues`](@ref)
     * [`BCValues`](@ref)
```
KnutAM, line 9:
```suggestion
    * [`BCValues`](@ref Ferrite.BCValues)
```
Will this work? Link isn't working at the moment.

KnutAM (approved):
LGTM after fixing the missing `embedded_det` and adding `sortfacet_fast`. Link would be nice to fix as well, but can also be done elsewhere.

docs/src/devdocs/elements.md
termi-official: I am pretty sure that we should not use `sortfacet_fast` right now. This will break queries on 4D simplices (which is still on my todo list) in a very nasty way.

docs/src/devdocs/FEValues.md
termi-official: Have we renamed this function? :D

docs/src/devdocs/elements.md
KnutAM: > 4D simplices (which is still on my todo list)

But please don't, that will probably break a lot of other assumptions too 😆

termi-official: When I developed the prototype the main problem was that Tensors.jl just works up to dim=3. Other than that (for linear ansatz) the generalization is straight forward for most of the codebase.

PR #973, opened 2024-06-04
fix type error in grid generators and add some test functions for gri…

Abdelrahman912:
### Summary
This PR fixes type conversion issues in the `generate_grid` function in `grid_generators.jl` and adds test coverage for all grid generators.

### Tests
- Added new test cases in `test/test_grid_generators.jl`.
- Verified that all tests pass locally.


test/test_grid_generators.jl
termi-official, line 1: Can you add regression checks to see if the grid entries have the correct type?

src/Grid/grid_generators.jl
```diff
@@ -69,7 +69,7 @@ end
 
 function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
       for i in 0:ny-1
-        ratio_bounds = i / (ny-1)
+        ratio_bounds = convert(eltype(LL), (i / (ny-1)))
```
termi-official, line 72: Can you leave a comment at the beginning of the function why this is done? It might not be obvious at first sight and someone might want to revert this.

termi-official:
Thanks for putting this together @Abdelrahman912 ! I left a few minor comments. It should be good to go afterwards.

src/Grid/grid_generators.jl
```diff
@@ -69,7 +69,7 @@ end
 
 function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
```
KnutAM, line 70:
```suggestion
function _generate_2d_nodes!(nodes::Vector{Node{2, T}}, nx, ny, LL, LR, UR, UL) where T
```

src/Grid/grid_generators.jl
```diff
@@ -69,7 +69,7 @@ end
...
       for i in 0:ny-1
-        ratio_bounds = i / (ny-1)
+        ratio_bounds = convert(eltype(LL), (i / (ny-1)))
```
KnutAM, line 78:
```suggestion
        ratio_bounds = convert(T, i) / (ny-1))

```

src/Grid/grid_generators.jl
```diff
@@ -78,7 +78,7 @@ function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
         y1 = LR[2] * (1 - ratio_bounds) + ratio_bounds * UR[2]
 
         for j in 0:nx-1
-            ratio = j / (nx-1)
+            ratio = convert(eltype(LL), j / (nx-1))
```
KnutAM, line 90:
```suggestion
            ratio = convert(T, j) / (nx-1)
```

src/Grid/grid_generators.jl
```diff
@@ -284,9 +284,9 @@ function generate_grid(::Type{Pyramid}, nel::NTuple{3,Int}, left::Vec{3,T}=Vec{3
 
     #Center node in each "voxel"
     for k in 1:nel_z, j in 1:nel_y, i in 1:nel_x
-        midx = 0.5(coords_x[i+1] + coords_x[i])
-        midy = 0.5(coords_y[j+1] + coords_y[j])
-        midz = 0.5(coords_z[k+1] + coords_z[k])
+        midx = convert(eltype(left),0.5(coords_x[i+1] + coords_x[i]))
+        midy = convert(eltype(left),0.5(coords_y[j+1] + coords_y[j]))
+        midz = convert(eltype(left),0.5(coords_z[k+1] + coords_z[k]))
```
KnutAM, line 301:
```suggestion
        midx = convert(T, coords_x[i+1] + coords_x[i]) / 2
        midy = convert(T, coords_y[j+1] + coords_y[j]) / 2
        midz = convert(T, coords_z[k+1] + coords_z[k]) / 2
```

KnutAM (commented):
Nice 🎉 
Just two small drive-by comments from me
1) I think it is more logical to take the type parameter from the node than the left point (even if this is equivalent, but seeing that requires looking at the call site)
2) I think it is better to convert and then do division by integer. For `Float32` it doesn't matter (only perf), but for higher precision floats it makes a difference

src/Grid/grid_generators.jl
Abdelrahman912: I agree with this solution, taking the type from node is more descriptive than using left

test/test_grid_generators.jl
```diff
@@ -0,0 +1,82 @@
...
+        @test all(typeof(cell) == C for cell in grid.cells)
+    end
+    
+    # Testing Line grid
+    nel = (10,)
+    left = Vec{1,T}((-1.0,))
+    right = Vec{1,T}((1.0,))
+    grid = generate_grid(Line, nel, left, right)
+    @test isa(grid, Grid{1, Line, T})
+    check_grid_types(grid)
```
KnutAM, line 25: This checks the same as `@test isa(grid, Grid{1, Line, T})`, so shouldn't be required.

test/test_grid_generators.jl
```diff
@@ -0,0 +1,82 @@
...
+        # Check cells
+        @test all(typeof(cell) == C for cell in grid.cells)
+    end
+    
+    # Testing Line grid
+    nel = (10,)
+    left = Vec{1,T}((-1.0,))
+    right = Vec{1,T}((1.0,))
+    grid = generate_grid(Line, nel, left, right)
+    @test isa(grid, Grid{1, Line, T})
```
KnutAM, line 24: These tests are also nice to put into a loop over the cell types you want to test
```suggestion
    for CT in (Line, QuadraticLine, Quadrilateral, ...)
        rdim = Ferrite.getrefdim(CT)
        nels = ntuple(i -> 2, rdim)
        left = - ones(Vec{rdim,T}
        right =  ones(Vec{rdim,T}
        grid = generate_grid(CT, nels, left, right)
        @test isa(grid, Grid{rdim, CT, T})
    end
```

test/test_grid_generators.jl
```diff
@@ -0,0 +1,82 @@
+using Test
+using Ferrite
+using Tensors
```
KnutAM, line 3: Not required here (included in `runtests.jl`)
```suggestion
```

src/Grid/grid_generators.jl
```diff
@@ -284,9 +293,12 @@ function generate_grid(::Type{Pyramid}, nel::NTuple{3,Int}, left::Vec{3,T}=Vec{3
...
     for k in 1:nel_z, j in 1:nel_y, i in 1:nel_x
-        midx = 0.5(coords_x[i+1] + coords_x[i])
-        midy = 0.5(coords_y[j+1] + coords_y[j])
-        midz = 0.5(coords_z[k+1] + coords_z[k])
+        # midx = 0.5(coords_x[i+1] + coords_x[i]) # old code, this will be always Float64 regardless of the element type of left
+        # midy = 0.5(coords_y[j+1] + coords_y[j]) # old code, this will be always Float64 regardless of the element type of left
+        # midz = 0.5(coords_z[k+1] + coords_z[k]) # old code, this will be always Float64 regardless of the element type of left
+        midx = convert(T, coords_x[i+1] + coords_x[i]) / 2
+        midy = convert(T, coords_y[j+1] + coords_y[j]) / 2
+        midz = convert(T, coords_z[k+1] + coords_z[k]) / 2
```
KnutAM, line 301: On a second look, this convert doesn't seem to be required now that we divide by 2 instead of multiplying with 0.5, 
```suggestion
        midx = (coords_x[i+1] + coords_x[i]) / 2
        midy = (coords_y[j+1] + coords_y[j]) / 2
        midz = (coords_z[k+1] + coords_z[k]) / 2
```

KnutAM (commented):
Nice, just left some comments on the testing

src/Grid/grid_generators.jl
```diff
@@ -284,9 +293,12 @@ function generate_grid(::Type{Pyramid}, nel::NTuple{3,Int}, left::Vec{3,T}=Vec{3
 
     #Center node in each "voxel"
     for k in 1:nel_z, j in 1:nel_y, i in 1:nel_x
-        midx = 0.5(coords_x[i+1] + coords_x[i])
-        midy = 0.5(coords_y[j+1] + coords_y[j])
-        midz = 0.5(coords_z[k+1] + coords_z[k])
+        # midx = 0.5(coords_x[i+1] + coords_x[i]) # old code, this will be always Float64 regardless of T
```
KristofferC, line 296: Just delete these lines?

test/test_grid_generators.jl
```diff
@@ -0,0 +1,24 @@
...
+        @test isa(grid, Grid{rdim, CT, T})
+    end
+   
+end
+
+# Run tests for different floating-point types
+@testset "Generate Grid Tests" begin
+    test_generate_grid(Float64)
+    test_generate_grid(Float32)
+    test_generate_grid(Float16)
```
KristofferC, line 23: I think it is fine to only test with two here? Does the third case add anything?

Abdelrahman912: It really doesn't at all, but I added it for the sake of completeness only, anyways, I will remove it

src/Grid/grid_generators.jl
```diff
@@ -67,9 +67,10 @@ function generate_grid(::Type{QuadraticLine}, nel::NTuple{1,Int}, left::Vec{1,T}
     return Grid(cells, nodes, facetsets=facetsets)
 end
 
-function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
+function _generate_2d_nodes!(nodes::Vector{Node{2, T}}, nx, ny, LL, LR, UR, UL) where T
       for i in 0:ny-1
-        ratio_bounds = i / (ny-1)
+
```
KnutAM, line 72:
```suggestion
```

src/Grid/grid_generators.jl
```diff
@@ -78,7 +79,8 @@ function _generate_2d_nodes!(nodes, nx, ny, LL, LR, UR, UL)
         y1 = LR[2] * (1 - ratio_bounds) + ratio_bounds * UR[2]
 
         for j in 0:nx-1
-            ratio = j / (nx-1)
+
```
KnutAM, line 82:
```suggestion
```

test/test_grid_generators.jl
```diff
@@ -0,0 +1,23 @@
+# Helper function to test grid generation for a given floating-point type
+function test_generate_grid(T::Type)
+
+    # Define the cell types to test
+    cell_types = [Line, QuadraticLine, Quadrilateral,QuadraticQuadrilateral,Triangle,QuadraticTriangle,Hexahedron,Wedge,Pyramid,Tetrahedron,SerendipityQuadraticHexahedron]
```
KnutAM, line 5:
```suggestion
    cell_types = [
        Line, QuadraticLine,
        Quadrilateral, QuadraticQuadrilateral, Triangle, QuadraticTriangle,
        Hexahedron, Wedge, Pyramid, Tetrahedron, SerendipityQuadraticHexahedron]
```

KnutAM (commented):
Great work on your first PR @Abdelrahman912 🚀 

Content-wise this looks good to me now, just some formatting fixes as commented below, and the pre-commit action should pass. See https://github.com/Ferrite-FEM/Ferrite.jl/pull/928 for detail on how to do that automatically.

PR #974, opened 2024-06-05
ExclusiveTopology with ArrayOfVectorViews backend

KnutAM:
See discussion in #965, this PR introduces a cleaned-up version. 

Most important changes in this PR compared to master are:
* Storage from `Array{EntityNeighborhood{T}, N}` -> `ArrayOfVectorViews{T, N} <: AbstractArray{"View{Vector{T}}", N}`
* `getcells` dispatches removed (since these would now act on `Vector{<:BoundaryIndex}`). In any case, `getcells` normally returns `::AbstractCell`s, but in topology they return `::Int`s on mater, if any, `cellid` should be overloaded).
* `EntityNeighborhood` is gone

Compared to #965, the support for hashed indexing is removed and only `ExclusiveTopology` is implemented. Before finalizing, I would like some feedback from the topology side, and then I'll add

- [x] Separate tests for `ArrayOfVectorViews`
- [x] Documentation of `ArrayOfVectorViews`
- [x] Add warning for non-conforming grids, ref https://github.com/Ferrite-FEM/Ferrite.jl/issues/941#issuecomment-2162836723 (and for embedded grids)
- [x] Update docstrings for `ExclusiveTopology`
- [x] Changelog

Let me know if more details are required for the initial review. 

Note that in relation to #942, this PR explicitly errors if trying to construct `ExclusiveTopology` with grids containing embedded cells, thus avoiding silent bugs.

**Benchmarks**

```julia
using Ferrite, BenchmarkTools
function sum_indices(nbh::AbstractMatrix)
    v = 0
    for nbs in nbh
        for n in nbs
            v += n[1] + n[2]
        end
    end
    return v
end

function sum_indices(top::ExclusiveTopology)
    v  = sum_indices(top.vertex_vertex_neighbor)
    v += sum_indices(top.edge_edge_neighbor)
    v += sum_indices(top.face_face_neighbor)
    return v
end
function sum_skeleton_facet_neighbors(top, grid)
    skeleton = Ferrite.facetskeleton(top, grid)
    v = 0
    for facetidx in skeleton
        v += sum(n[1] + n[2] for n in getneighborhood(top, grid, facetidx); init = 0)
    end
    return v
end

grid = generate_grid(Hexahedron, (10, 10, 10));
top = ExclusiveTopology(grid); # Compilation

@btime sum_indices($top); 
# Master: 55.700 μs (0 allocations: 0 bytes)
#     PR: 36.000 μs (0 allocations: 0 bytes)
Ferrite.facetskeleton(top, grid); # Compilation

grid = generate_grid(Hexahedron, (100, 100, 100));
@time top = ExclusiveTopology(grid);
# Master: 27.150434 seconds (64.58 M allocations: 7.195 GiB, 72.43% gc time)
#     PR:  3.799651 seconds (45 allocations: 1.895 GiB, 13.84% gc time)
@time Ferrite.facetskeleton(top, grid);
# Master: 0.143506 seconds (2 allocations: 46.234 MiB)
#     PR: 0.032492 seconds (2 allocations: 46.234 MiB)
@btime sum_skeleton_facet_neighbors($top, $grid);
# Master: 59.683 ms (0 allocations: 0 bytes)
#     PR: 21.887 ms (0 allocations: 0 bytes)
```
(PR refers to b7e2add)


src/ArrayOfVectorViews.jl
```diff
@@ -0,0 +1,85 @@
+struct ArrayOfVectorViews{T, N} <: AbstractArray{SubArray{T, 1, Vector{T}, Tuple{UnitRange{Int64}}, true}, N}
+    indices::Array{UnitRange{Int}, N}
```
KristofferC, line 2: Are these always "dense" so that you will have

```
[1:2, 3:6, 7:15, ...]
```

in that case it is unnecessary to store the second number for all of these?

KnutAM: Currently they are not guaranteed to be, that would require another sorting during compress, but could be a good optimization solving both the lower index storage requirement, and speeding up iteration somewhat. (If sizehint provided to the `ConstructionBuffer` is large enough, they are already "dense" after compressing)

KristofferC: My guess is that it is worth it, especially for those cases where the "inner" arrays are short.

KnutAM: Yes, but that feels like an optimization we can do later? 
From the outside, everything (including type parameters) will remain the same...

KristofferC: Sure

KnutAM: > (If sizehint provided to the ConstructionBuffer is large enough, they are already "dense" after compressing)

My comment above was wrong, that is only true if also started `add!`ed in that order, so some signficant sorted is required because normally this is not the case.

KnutAM:
c276cb2 brings construction in benchmark further down from 
`5.527448 seconds (58 allocations: 2.107 GiB, 9.91% gc time)`
to 
`4.217167 seconds (45 allocations: 1.895 GiB, 12.49% gc time)`

src/ArrayOfVectorViews.jl
termi-official, line 1: Can we wrap this into a module and export only the stuff which we need?

src/Grid/topology.jl
```diff
@@ -1,4 +1,4 @@
-
+include("../ArrayOfVectorViews.jl")
```
termi-official, line 1: This should go into `src/Ferrite.jl` right?

src/Grid/topology.jl
```diff
@@ -86,11 +135,22 @@ function _get_facet_facet_neighborhood(::ExclusiveTopology, #=rdim=#::Val{:mixed
...
-    print(io, "  Face neighbors: $(size(topology.face_face_neighbor))\n")
-    println(io, "  Edge neighbors: $(size(topology.edge_edge_neighbor))")
+# Guess of how many neighbors depending on grid dimension and index type.
+# This is just a performance optimization, and a good default is sufficient.
+_getsizehint(::AbstractGrid{3}, ::Type{FaceIndex}) = 1 # 2
+_getsizehint(::AbstractGrid, ::Type{FaceIndex}) = 0 # No faces exists in 2d or lower dim
+_getsizehint(::AbstractGrid{dim}, ::Type{EdgeIndex}) where dim = 1 #dim^2
+_getsizehint(::AbstractGrid{dim}, ::Type{VertexIndex}) where dim = 1 # 2^dim
+_getsizehint(::AbstractGrid{1}, ::Type{CellIndex}) = 2
+_getsizehint(::AbstractGrid{2}, ::Type{CellIndex}) = 12
```
termi-official, line 145: Maybe we should add an additional layer to also dispatch on the cell type to not overestimate too much. Where do you have the numbers from?

src/ArrayOfVectorViews.jl
KnutAM: Yes, I think that makes sense!

src/Grid/topology.jl
KnutAM: Yes, left it temporarily here but should probably move it. Will do when putting into a module!

src/Grid/topology.jl
KnutAM: Yes, the commented versions are from hypercubes if I recall correctly, but an overestmate for the exclusive topology. 

Not sure if it is worthwhile to make this general since it isn't super-sensitive in my benchmarks (of course it matters, but now allocations are a quite small fraction of the time). So could even be better to simplify further and hard-code in `ExclusiveTopology`, since the same sizehints won't be valid for e.g. `CoverTopology` (didn't think of this when starting to implement this)

KnutAM:
Thanks for the comments @termi-official - fixed the include and making into submodule. 

Do you or @koehlerson have any comments on the breaking changes that topology outputs are views of a vector instead of `EntityNeighborhood`. Is this ok?

koehlerson:
> Thanks for the comments @termi-official - fixed the include and making into submodule.
> 
> Do you or @koehlerson have any comments on the breaking changes that topology outputs are views of a vector instead of `EntityNeighborhood`. Is this ok?

No that's fine and the `EntityNeighborhood` thing can be removed. I'll go through the changes soon :)

KnutAM:
> No that's fine and the EntityNeighborhood thing can be removed. I'll go through the changes soon :)

Great, thanks! That was my major concern with this PR.

KristofferC:
FWIW (without having done any detailed review) I'm fine merging this, it clearly is principally a lot better than status quo and the benchmarks suggest it is also a lot better in practice. Great job @KnutAM.

koehlerson (commented):
Code looks fine to me, but the docs are not synced to the code right now. 

~~- [ ] adjust `ExclusiveTopology` struct docs~~
~~-[ ]  devdocs for the collection of views~~
~~-[ ] changelog entry~~

Sorry just noticed your task list now. Where are the docs of the `ArrayOfVectorViews`? I think it could be nice to have them in the devdocs

src/Grid/topology.jl
```diff
@@ -106,156 +165,118 @@ function _num_shared_vertices(cell_a::C1, cell_b::C2) where {C1, C2}
...
+function build_cell_neighbor(grid, cells, vertex_to_cell; ncells)
+    # In this case, we loop over the cells in order and all all neighbors at once.
+    # Then we can create ArrayOfVectorViews directly without the CollectionsOfViews.ConstructionBuffer
+    sizehint = _getsizehint(grid, CellIndex)
+    data = empty!(Vector{Int}(undef, ncells * sizehint))
 
-ExclusiveTopology(grid::AbstractGrid) = ExclusiveTopology(getcells(grid))
+    indices = Vector{Int}(undef, ncells + 1)
+    # cell_neighbor_ids = Set{Int}() #TODO: Remove, just show old version to make review easier
+    cell_neighbor_ids = Int[] # Faster than Set in combination with ∈ || check below
```
KnutAM, line 244:
```suggestion
    cell_neighbor_ids = Int[]
```

src/Grid/topology.jl
```diff
@@ -14,6 +13,7 @@ the given entity is included in the returned list as well.
 """
 getneighborhood
 
+#TODO: Remove
```
KnutAM, line 16:
```suggestion
#TODO: To be removed
```
You mean like this, or to remove the comment? 
Currently, `EntityNeighborhood` is still used by the `vertex_star_stencils` code. 

I guess it would make sense to change the code for this as well to return an `ArrayOfVectorViews` instead of `Dict: global_vertex_nr -> Vector{EdgeIndex}`? I didn't so far to minimize the diff, and I don't know exactly the use-cases for that function.

KnutAM:
Thanks for the review - then I'll try to wrap it up during this week :)

I also just discussed with @fredrikekre on Slack, and we were wondering when the `include_self` keyword is used. Currently, this requires allocating a new vector (still returning a view for type-stability), potential improvements could be 

* Remove the `include_self`, leave that to the user since "self" is available anyways
* Include self as the first index in the stored `ArrayOfVectorViews`, such that `views(vals[index], 2:end)` can be returned if `include_self=false`, and `vals[index]` if `include_self=true`. But this will change the behavior if direct field access is used. 

Not required to do in this PR, but any of these changes must happen before 1.0...

src/Grid/topology.jl
koehlerson: `vertex_star_stencils` is, if I remember correctly, giving you all vertices that are connected by an edge. It can be used for mixing FEM and FDM approaches as in e.g. the topology optimization, but there the FDM operator acts on the "elements", whereas you can change that to operate on the nodes (which I think is more accurate). Would be nice to remove that. Is it currently really returning an `EdgeIndex` vector?

koehlerson:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/a10097a99cb1736999cc1bf685eaa032331fd93a/src/Grid/topology.jl#L271

e.g. here, the self reference is computed and either included or not.

termi-official:
Maxi is correct, I need it e.g. for the distributed stuff and it is a nice convenience. Note that you can make this non-allocating by adding the self-reference during construction and return 1:N for include_self=true and 1:N-1 for include_self=false.

termi-official:
> Not required to do in this PR, but any of these changes must happen before 1.0...

Why? The topology API is experimental and we can just add a note that we might break things here in minor releases, like we do for other stuff too.

KnutAM:
> Note that you can make this non-allocating by adding the self-reference during construction and return 1:N for include_self=true and 1:N-1 for include_self=false.

Yes, that was option 2 above right? But I'm not sure what that will break for anyone using direct field access (not calling the neighborhood), since the returned indices will then include self?

KnutAM:
> FWIW (without having done any detailed review) I'm fine merging this, it is clearly principally a lot better than the status quo and the benchmarks suggest it is also a lot better in practice. Great job @KnutAM.

Thanks - part of the solution was yours though :)

src/Grid/topology.jl
KnutAM: Posting a quick function to get some size info for later use
```julia
function analyze_top_sizes(grid)
    top = ExclusiveTopology(grid);
    for key in fieldnames(typeof(top))
        nbh = getfield(top, key)
        nbh isa Ferrite.ArrayOfVectorViews || continue
        max_size = maximum(length, nbh; init=0)
        avg_size = sum(length, nbh; init=0)/length(nbh)
        println(key)
        println("  max: ", max_size, " entities")
        println("  avg: ", avg_size, " entities")
    end
end
```

src/Grid/topology.jl
KnutAM: Yes, a view of a `Vector{EdgeIndex}`.

docs/src/devdocs/special_datastructures.md
```diff
@@ -0,0 +1,14 @@
+# Special data structures
+
+## `ArrayOfVectorViews`
```
KristofferC, line 3: Could the obvious question of "why not use ArraysOfArrays" be answered shortly in this section as well?

KnutAM:
This is now ready from my perspective.
Decided to revert my attempts to include self in the saved data, simply because this broke a lot of tests and I don't have time to investigate and fix all of those. 

This should also just be an optimization and as @termi-official pointed out, the topology is considered experimental. That should probably be clarified in the docs in a separate PR before 1.0, since IMO the goal should be that 1.0 is stable unless otherwise clarly stated.

termi-official:
> This should also just be an optimization and as @termi-official pointed out, the topology is considered experimental. That should probably be clarified in the docs in a separate PR before 1.0, since IMO the goal should be that 1.0 is stable unless otherwise clarly stated.

We clarified it already when we developed it. See https://ferrite-fem.github.io/Ferrite.jl/dev/topics/grid/#Topology . I remember that we had it in the doc string too at some point, but looks like this information has been lost in the development. Where else would you put it?

KnutAM:
Aha, I hadn't read that. I'll just put a note in the docstring then. 
(I read the one on [release](https://ferrite-fem.github.io/Ferrite.jl/stable/reference/grid/#Topology) which gave me the impression that only ´getneighborhood` was not stable API).

KnutAM:
When rechecking the timings, I saw some worse performance now. 
So took @termi-official's suggestion and optimized the sizehints

<details> <summary> Script </summary> 

```julia
function analyze_top_sizes(grid; show_avg=false)
    top = ExclusiveTopology(grid);
    for key in fieldnames(typeof(top))
        nbh = getfield(top, key)
        nbh isa Ferrite.ArrayOfVectorViews || continue
        max_size = maximum(length, nbh; init=0)
        avg_size = sum(length, nbh; init=0)/length(nbh)
        if show_avg
            println(key)
            println("  max: ", max_size, " entities")
            println("  avg: ", avg_size, " entities")
        else
            println(key, ": ", max_size)
        end
    end
end

function print_sizes(n=10, celltypes=[Line, Triangle, Quadrilateral, Tetrahedron, Hexahedron])
   for CT in celltypes
       grid = generate_grid(CT, ntuple(_->n, Ferrite.getrefdim(CT)))
       println(CT)
       analyze_top_sizes(grid)
       println()
   end
end
```
</details> 

Max number of entities for different generated grids

| field                   | Line | Triangle | Quadrilateral | Tetrahedron | Hexahedron |
| ----------------------- | ---- | -------- | ------------- | ----------- | ---------- |
| vertex_to_cell          | 2    | 6        | 4             | 24          | 8          |
| cell_neighbor           | 2    | 12       | 8             | 70          | 26         |
| face_face_neighbor      | 0    | 0        | 0             | 1           | 1          |
| edge_edge_neighbor      | 0    | 1        | 1             | 3           | 1          |
| vertex_vertex_neighbor  | 1    | 3        | 1             | 13          | 1          |

KnutAM:
#988 should be fixed after merging this PR to avoid annoying conflicts.

termi-official:
Resolves #617 ?

src/CollectionsOfViews.jl
```diff
@@ -0,0 +1,153 @@
...
+    data = Vector{T}(undef, data_length)
+    data_index = 1
+    for (idx, ar) in pairs(b.indices)
+        copyto!(data, data_index, b.data, ar.start, ar.ncurrent)
+        indices[lin_idx[idx]] = data_index
+        data_index += ar.ncurrent
+    end
+    indices[length(indices)] = data_index
+    resize!(b.data, 0); sizehint!(b.data, 0)
+    isa(b.indices, Vector) && (resize!(b.indices, 0); sizehint!(b.indices, 0))
```
termi-official, line 130: Can we add some comment here why everything is set to zero length?

src/Grid/topology.jl
```diff
@@ -14,65 +13,90 @@ the given entity is included in the returned list as well.
...
+- `cell_neighbor::AbstractArray{AbstractVector{Int}, 1}`:            cellid to all connected cells
+- `face_neighbor::AbstractArray{AbstractVector{FaceIndex}, 2}`:      `face_neighbor[cellid,   local_face_id]`   -> neighboring faces
+- `edge_neighbor::AbstractArray{AbstractVector{EdgeIndex}, 2}`:      `edge_neighbor[cellid,   local_edge_id]`   -> neighboring edges
+- `vertex_neighbor::AbstractArray{AbstractVector{VertexIndex}, 2}`:  `vertex_neighbor[cellid, local_vertex_id]` -> neighboring vertices
+- `face_skeleton::Union{Vector{FaceIndex}, Nothing}`:     List of unique faces in the grid given as `FaceIndex`
+- `edge_skeleton::Union{Vector{EdgeIndex}, Nothing}`:     List of unique edges in the grid given as `EdgeIndex`
+- `vertex_skeleton::Union{Vector{VertexIndex}, Nothing}`: List of unique vertices in the grid given as `VertexIndex`
+
+!!! warning "Limitations"
+    Non-conforming grids will silently not work.
```
termi-official, line 37: Something in this direction maybe? Not sure how to be clear and concise here.
```suggestion
    The implementation only works with conforming grids, i.e. grids without "hanging nodes". Non-conforming grids will give unexpected results.
```

src/Grid/topology.jl
```diff
@@ -14,65 +13,90 @@ the given entity is included in the returned list as well.
...
+    Grids with embedded cells (different reference dimension compared
```
termi-official, line 38: Just for me, this is mainly a matter of not having test coverage right now?

src/Grid/topology.jl
```diff
@@ -86,12 +110,31 @@ function _get_facet_facet_neighborhood(::ExclusiveTopology, #=rdim=#::Val{:mixed
...
+    isconcretetype(CT) && return _getsizehint(getrefshape(CT)(), IDX)
+    rdim = get_reference_dimension(g)::Int
+    return _getsizehint(RefSimplex{rdim}(), IDX) # Simplex is "worst case", used as default.
 end
+_getsizehint(::AbstractRefShape,    ::Type{FaceIndex}) = 1 # Always 1 or zero if not mixed rdim
+
+_getsizehint(::AbstractRefShape{1}, ::Type{EdgeIndex}) = 1
+_getsizehint(::AbstractRefShape{2}, ::Type{EdgeIndex}) = 1
+_getsizehint(::AbstractRefShape{3}, ::Type{EdgeIndex}) = 3 # Number for RefTetrahedron
+_getsizehint(::RefHexahedron,       ::Type{EdgeIndex}) = 1 # Optim for RefHexahedron
```
termi-official, line 126: Can you add some commen on why this is optimal for hex? I am not seeing it directly, sorry.

src/Grid/topology.jl
```diff
@@ -106,164 +149,125 @@ function _num_shared_vertices(cell_a::C1, cell_b::C2) where {C1, C2}
...
-    max_edges = 0
-    if isconcretetype(celltype)
-        max_vertices = nvertices(cells[1])
-        max_faces = nfaces(cells[1])
-        max_edges = nedges(cells[1])
+"Return the highest number of vertices, edges, and faces per cell"
+function _max_nentities_per_cell(cells::Vector{C}) where C
+    if isconcretetype(C)
+        cell = first(cells)
+        return nvertices(cell), nedges(cell), nfaces(cell)
```
termi-official, line 156: Since this is a corner case: It is expected that 1D elements dispatch on `nfaces` then?

src/CollectionsOfViews.jl
KnutAM: Does this comment explain it well enough? 
```suggestion
    # Since user-code in "push_at_index!" have access to `b`, setting dimensions to zero here
    # allows GC:ing the data in `b` even in cases when the compiler cannot guarantee that 
    # it is unreachable.
    resize!(b.data, 0); sizehint!(b.data, 0)
    isa(b.indices, Vector) && (resize!(b.indices, 0); sizehint!(b.indices, 0))
```

test/test_grid_dofhandler_vtk.jl
```diff
@@ -321,36 +328,38 @@ end
...
+    @test topology.edge_edge_neighbor[3,1] == [EdgeIndex(1,3)]
+    @test topology.edge_edge_neighbor[3,2] == [EdgeIndex(4,4)]
+    @test topology.edge_edge_neighbor[3,3] == [EdgeIndex(5,1)]
+    @test topology.edge_edge_neighbor[4,1] == [EdgeIndex(2,3)]
+    @test topology.edge_edge_neighbor[4,3] == [EdgeIndex(6,1)]
+    @test topology.edge_edge_neighbor[4,4] == [EdgeIndex(3,2)]
+    @test topology.edge_edge_neighbor[5,1] == [EdgeIndex(3,3)]
+    @test topology.edge_edge_neighbor[5,2] == [EdgeIndex(6,4)]
+    @test topology.edge_edge_neighbor[5,3] == Ferrite.BoundaryIndex[]
+    @test topology.edge_edge_neighbor[5,4] == Ferrite.BoundaryIndex[]
```
termi-official, line 356: So the type instability is not fixed here?
```suggestion
    @test topology.edge_edge_neighbor[5,3] == Ferrite.EdgeIndex[]
    @test topology.edge_edge_neighbor[5,4] == Ferrite.EdgeIndex[]
```

(also below)

termi-official (commented):
Great improvement on the performance and I think the data structure will also help in the long run. I left a few comments.

src/Grid/topology.jl
KnutAM: I don't think it is easy to see due to the exclusive nature, I could add a ref to https://github.com/Ferrite-FEM/Ferrite.jl/pull/974#issuecomment-2182897133 if you think that helps?

src/Grid/topology.jl
KnutAM: For embedded cells I think there will be correctness issues, since the algorithms are not written for that in mind, AFAIU from the discussion in #942 and related issue.

src/Grid/topology.jl
KnutAM: This is covered by `reference_faces(::Type{RefLine}) = ()` (which gets called by `faces` getting called by `nfaces`). So it is not required to explicitly dispatch `nfaces` for a new 1D element.

test/test_grid_dofhandler_vtk.jl
KnutAM: I was not aware of this being due to type-instabilities, there should be no such for this test for sure! 
But this is indeed a strange test at least now, since e.g. 
```julia
julia> String[] == Int[]
true
```
I guess we should rather test
```suggestion
    @test isempty(topology.edge_edge_neighbor[5,3])
    @test isempty(topology.edge_edge_neighbor[5,4])
```

src/Grid/topology.jl
termi-official: Ohh, you mean for the grids generate with `generate_grid` optimal? 

Have you by any chance tried grids coming from external tools like gmsh then?

test/test_grid_dofhandler_vtk.jl
```diff
@@ -321,36 +328,38 @@ end
...
+    @test topology.edge_edge_neighbor[6,1] == [EdgeIndex(4,3)]
+    @test topology.edge_edge_neighbor[6,2] == Ferrite.BoundaryIndex[]
+    @test topology.edge_edge_neighbor[6,3] == Ferrite.BoundaryIndex[]
```
KnutAM, line 359:
```suggestion
    @test isempty(topology.edge_edge_neighbor[6,2])
    @test isempty(topology.edge_edge_neighbor[6,3])
```

src/Grid/topology.jl
KnutAM: No, I found that to be out of scope for this PR since this is just an optimization. Of course for non-structured grids, there will be cases when the numbers are higher for some entities, but we probably don't won't these sizehints for the worst case as that will over-allocate. From what I've seen during development, the key is to have most connections below the sizehint. 

From my perspective, if someone wants to do a proper investigation into this, that can be done at any point, since changing these sizehints only changes the performance.

test/test_grid_dofhandler_vtk.jl
termi-official: Right

src/Grid/topology.jl
```diff
@@ -86,12 +110,31 @@ function _get_facet_facet_neighborhood(::ExclusiveTopology, #=rdim=#::Val{:mixed
     Access the `vertex_vertex_neighbor`, `edge_edge_neighbor`, or `face_face_neighbor` fields explicitly instead."))
 end
 
-function Base.show(io::IO, ::MIME"text/plain", topology::ExclusiveTopology)
-    println(io, "ExclusiveTopology\n")
-    print(io, "  Vertex neighbors: $(size(topology.vertex_vertex_neighbor))\n")
-    print(io, "  Face neighbors: $(size(topology.face_face_neighbor))\n")
-    println(io, "  Edge neighbors: $(size(topology.edge_edge_neighbor))")
+# Guess of how many neighbors depending on grid dimension and index type.
+# Better to guess a bit too high than a bit too low.
```
KnutAM, line 114:
```suggestion
# Guess of how many neighbors depending on grid dimension and index type. 
# This could be possible to optimize further by studying connectivities of non-uniform
# grids, see https://github.com/Ferrite-FEM/Ferrite.jl/pull/974#discussion_r1660838649
```

KnutAM:
> I left a few comments.

Thanks!

src/Grid/topology.jl
```diff
@@ -14,65 +13,90 @@ the given entity is included in the returned list as well.
...
+    # vertex_vertex_neighbor[cellid,local_vertex_id] -> exclusive connected entities to the given vertex
+    vertex_vertex_neighbor::ArrayOfVectorViews{VertexIndex, 2}
+
     face_skeleton::Union{Vector{FaceIndex}, Nothing}
     edge_skeleton::Union{Vector{EdgeIndex}, Nothing}
     vertex_skeleton::Union{Vector{VertexIndex}, Nothing}
-    # TODO reintroduce the codimensional connectivity, e.g. 3D edge to 2D face
+end
+
+function ExclusiveTopology(grid::AbstractGrid{sdim}) where sdim
```
koehlerson, line 57: Just want to highlight that this is more restrictive since before the argument was `Vector{C} where C <: AbstractCell`

src/Grid/topology.jl
```diff
@@ -106,164 +150,125 @@ function _num_shared_vertices(cell_a::C1, cell_b::C2) where {C1, C2}
...
+        end
+    return vertex_to_cell
+end
 
-getcells(neighbor::EntityNeighborhood{T}) where T <: BoundaryIndex = first.(neighbor.neighbor_info)
-getcells(neighbor::EntityNeighborhood{CellIndex}) = getproperty.(neighbor.neighbor_info, :idx)
-getcells(neighbors::Vector{T}) where T <: EntityNeighborhood = reduce(vcat, getcells.(neighbors))
-getcells(neighbors::Vector{T}) where T <: BoundaryIndex = getindex.(neighbors,1)
+function build_cell_neighbor(grid, cells, vertex_to_cell; ncells)
+    # In this case, we loop over the cells in order and all all neighbors at once.
```
koehlerson, line 222:
```suggestion
    # In this case, we loop over the cells in order and all neighbors at once.
```

koehlerson (approved):

src/Grid/topology.jl
KnutAM: Good point - I've added this change (and motivation) in the changelog.

termi-official (approved):
LGTM. Can you dump the benchmark code in the benchmark folder (here or separate PR)?

KnutAM:
> LGTM. Can you dump the benchmark code in the benchmark folder (here or separate PR)?

Thanks! I haven't looked so much at the benchmark setup, will just open an issue for now to not forget to do it later unless someone solves it before I have time.

PR #981, opened 2024-06-08
Allow geometric_interpolation(::AbstractCell)

### LINKED ISSUE #978: docstring of geometric_interpolation is wrong
Also the deprecation, see  #953

```julia
help?> geometric_interpolation
search: geometric_interpolation

  Each AbstractCell type has a unique geometric interpolation describing its geometry. This function returns that
  interpolation.
```

KnutAM:
Moves docstring to public API (where it should be since the function is exported)
Fixes wrong deprecation dispatch
Closes #978

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -132,12 +132,13 @@ reference_facets(::Type{<:AbstractRefShape})
 @inline reference_facets(refshape::Type{<:AbstractRefShape{3}}) = reference_faces(refshape)
 
 """
-    geometric_interpolation(::AbstractCell)::Interpolation
+    geometric_interpolation(::AbstractCell)
+    geometric_interpolation(::Type{<:AbstractCell})
 
 Each `AbstractCell` type has a unique geometric interpolation describing its geometry.
-This function returns that interpolation.
+This function returns that interpolation, which is always a scalar interpolation.
 """
-geometric_interpolation(::AbstractCell)
+geometric_interpolation(cell::AbstractCell) = geometric_interpolation(typeof(cell))
 
 """
     Ferrite.get_node_ids(c::AbstractCell)
--- a/test/test_deprecations.jl
+++ b/test/test_deprecations.jl
@@ -123,7 +123,7 @@ end
 end
 
 @testset "default_interpolation" begin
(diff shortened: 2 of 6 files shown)


src/Grid/grid.jl
```diff
@@ -132,12 +132,13 @@ reference_facets(::Type{<:AbstractRefShape})
 @inline reference_facets(refshape::Type{<:AbstractRefShape{3}}) = reference_faces(refshape)
 
 """
-    geometric_interpolation(::AbstractCell)::Interpolation
+    geometric_interpolation(::AbstractCell)
+    geometric_interpolation(::Type{<:AbstractCell})
 
 Each `AbstractCell` type has a unique geometric interpolation describing its geometry.
-This function returns that interpolation.
+This function returns that interpolation, which is always a scalar interpolation.
```
termi-official, line 139: Do we really want to enforce this behavior?

Also, since I just realize, we probably should double check the internal implementation where this is used to see if it might break when interacting with embedded elements.

test/test_deprecations.jl
```diff
@@ -122,4 +122,8 @@ end
     @test_throws ErrorException(msg) Ferrite.getfielddim(dh.subdofhandlers[1], :u)
 end
 
+@testset "default_interpolation" begin
+    @test Ferrite.default_interpolation(Triangle) == geometric_interpolation(Triangle)
```
termi-official, line 126:
```suggestion
    @test Ferrite.default_interpolation(Triangle) == geometric_interpolation(Triangle)
    @test Ferrite.default_interpolation(Triangle((1,2,3))) == geometric_interpolation(Triangle((1,2,3)))
```
To have a regression for this.

KnutAM: That never worked AFAIU (verified by checking out the commit before merging #953:  572c1068c7ed38720ef54303d97c8a829263be57)

src/Grid/grid.jl
KnutAM: Since we are giving only the cell, we don't know anything about the spatial dimension, and therefore we cannot return correct vectorized interpolation in order to support embedded elements. 

`default_geometric_interpolation(ip::Interpolation)` returns a `Vectorized` interpolation with `fielddim = refdim`, which is used to give a default value for `AbstractValues`

PR #982, opened 2024-06-08
Update to BlockArrays.jl v1

### LINKED ISSUE #980: Update to BlockArrays 1.0
Update to https://github.com/JuliaArrays/BlockArrays.jl/releases/tag/v1.0.0  should be rather straight forward.

KnutAM:
Only changes in docs required

Closes #980

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -24,7 +24,7 @@ FerriteBlockArrays = "BlockArrays"
 FerriteMetis = "Metis"
 
 [compat]
-BlockArrays = "0.16"
+BlockArrays = "0.16, 1"
 EnumX = "1"
 Metis = "1.3"
 NearestNeighbors = "0.4"
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.2"
+julia_version = "1.10.3"
 manifest_format = "2.0"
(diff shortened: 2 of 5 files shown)


Project.toml
```diff
@@ -24,7 +24,7 @@ FerriteBlockArrays = "BlockArrays"
 FerriteMetis = "Metis"
 
 [compat]
-BlockArrays = "0.16"
+BlockArrays = "0.16, 1"
```
termi-official, line 27: How do you handle backwards compat?

termi-official (commented):
Do we have test coverage for the FerriteBlockArrays extension?

Project.toml
fredrikekre: I believe there were no  breaking changes for the functionality used in the extension, just in the docs.

fredrikekre:
Silly thing to use a breaking release for...

Project.toml
KnutAM: No need, since the only changes required where in docs.
At least I couldn't find any use in ext/FerriteBlockArrays.jl relying on any breaking changes, the only thing was the rename of PseudoBlockArrays AFAIU which was only used in docs.

KnutAM:
> Do we have test coverage for the FerriteBlockArrays extension?

Decent coverage, https://app.codecov.io/gh/Ferrite-FEM/Ferrite.jl/pull/982/blob/ext/FerriteBlockArrays.jl

KnutAM:
> Silly thing to use a breaking release for...

A changelog would have been nice :)

PR #992, opened 2024-06-26
Allow more vector types for nodal evaluation of solution.

termi-official:
To allow e.g. operating on solutions which are views.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -912,7 +912,7 @@ getfieldinterpolation(sdh::SubDofHandler, field_idx::Int) = sdh.field_interpolat
 getfieldinterpolation(sdh::SubDofHandler, field_name::Symbol) = getfieldinterpolation(sdh, find_field(sdh, field_name))
 
 """
-    evaluate_at_grid_nodes(dh::AbstractDofHandler, u::Vector{T}, fieldname::Symbol) where T
+    evaluate_at_grid_nodes(dh::AbstractDofHandler, u::AbstractVector{T}, fieldname::Symbol) where T
 
 Evaluate the approximated solution for field `fieldname` at the node
 coordinates of the grid given the Dof handler `dh` and the solution vector `u`.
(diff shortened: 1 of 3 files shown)


fredrikekre:
Add a basic test?

src/Dofs/DofHandler.jl
```diff
@@ -912,7 +912,7 @@ getfieldinterpolation(sdh::SubDofHandler, field_idx::Int) = sdh.field_interpolat
 getfieldinterpolation(sdh::SubDofHandler, field_name::Symbol) = getfieldinterpolation(sdh, find_field(sdh, field_name))
 
 """
-    evaluate_at_grid_nodes(dh::AbstractDofHandler, u::Vector{T}, fieldname::Symbol) where T
+    evaluate_at_grid_nodes(dh::AbstractDofHandler, u::AbstractVector{T}, fieldname::Symbol) where T
 
 Evaluate the approximated solution for field `fieldname` at the node
 coordinates of the grid given the Dof handler `dh` and the solution vector `u`.
```
KnutAM, line 918: Would it make sense to add some note like this? 
```suggestion
coordinates of the grid given the Dof handler `dh` and the solution vector `u`.
Note that `u` is indexed by the DoF number from `dh`, passing only the solution 
pertinent to `fieldname` will therefore give the wrong result. 
```

KnutAM (approved):
LGTM, don't see any bad things that could happen from allowing an `AbstractVector`. 
I assume the use-case of the view would be if having more than one DofHandler, but a single solution vector? Or is this to rather support a GPU-array type?

src/Dofs/DofHandler.jl
termi-official: This is indeed one example. GPU should also work here with some trick, although probably a bit slower.

termi-official: Not sure, it is really difficult to pull the solution vector into subfields. We can just add some assertion in the function which checks if the length is okay and throw an error that the user probably passed the wrong aray.

KnutAM: True, `ndofs(dh)==length(u)` should always be fullfilled, so that is a nice check to add! That is probably better than my additional note!

termi-official: Can you add that in a separate PR?

PR #996, opened 2024-06-28
Change all deprecations to hard errors.

fredrikekre:


test/test_deprecations.jl
```diff
@@ -40,90 +40,88 @@ end
...
 
 @testset "facesets" begin
-grid = generate_grid(Quadrilateral, (2,2))
-addfaceset!(grid, "right_face", x -> x[1] ≈ 1)
-@test getfacetset(grid, "right_face") == getfacetset(grid, "right")
-addfaceset!(grid, "right_face_explicit", Set(Ferrite.FaceIndex(fi[1], fi[2]) for fi in getfacetset(grid, "right")))
-@test getfacetset(grid, "right_face_explicit") == getfacetset(grid, "right")
+    grid = generate_grid(Quadrilateral, (2,2))
+    @test_throws Ferrite.DeprecationError addfaceset!(grid, "right_face", x -> x[1] ≈ 1)
+    @test_throws Ferrite.DeprecationError addfaceset!(grid, "right_face_explicit", Set(Ferrite.FaceIndex(fi[1], fi[2]) for fi in getfacetset(grid, "right")))
```
fredrikekre, line 94: @KnutAM this is supposed to be deprecated? I guess we removed all methods of `addfaceset!`?

KnutAM: Yes, but done with `@warn ... maxlog=1` to actually show the warning for normal users not running tests. 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/900f6ed3547ce5d70e0ecdc998c9fb2f81aaeb94/src/deprecations.jl#L356-L366
But if we error, they will see it too so that is also fine by me (I think this way, it was possible to hack around to get e.g. FerriteGmsh working, but this is no longer required after updating FerriteGmsh)

PR #999, opened 2024-06-28
Flip the sign in the definition of jump for interface values

KnutAM:
As discussed on [slack](https://julialang.slack.com/archives/C01R8JU271D/p1718911945839659), this makes the definitions consistent for interface elements (FerriteInterfaceElements.jl).


docs/src/literate-tutorials/dg_heat_equation.jl
termi-official, line 1: I think the theory section above needs an update, too.

KnutAM: I couldn't find any place, since it uses the u^+ * n^+ + u^- * n^- definition? Or did I miss a place where the Ferrite jump definition is discussed?

fredrikekre: We could add a discussion in the theory section how the jump defined there relates to the jump in Ferrite, but I don't think it is inconsistent as it is with this PR.

PR #1000, opened 2024-06-28
Set version to 1.0.0.

fredrikekre:


ChrisRackauckas (changes requested):
I don't know if now is the time.

fredrikekre:
> I don't know if now is the time.

Issue #1001, opened 2024-06-30
Use separate functions for creating quadrature rule instead of the `QuadratureRule` constructor


KristofferC:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/991#issuecomment-2198520827

 I think the current way of creating quadrature rules is not ideal. For example we write:

```
qr = QuadratureRule{RefTetrahedron}(2)
```

but there are many quadrature rules and it isn't clear which one this creates etc. This was changed by me in https://github.com/Ferrite-FEM/Ferrite.jl/pull/58 and 8 years later I am suggesting we change it back to standard functions like:

```
create_legendre_quadrature(RefTetrahedron::RefShape, 1::Int)
create_lobotto_quadrature(RefCube::RefShape, 1::Int)
```

etc.  Should be decided before 1.0 I guess.

fredrikekre:
I quite like the current symmetry with interpolation constructors.

KristofferC:
Yes it looks nice but for the interpolation there are not really multiple "choices". To me, this is a bit similar to how there exists a `rand` function and a `zeros` function etc that all create an `Array` instead of shoe horning everything into `Array`.

fredrikekre:
There should imo at least be a function that works for any reference shape and returns a good default

KnutAM:
I think it makes sense to factor out the functions to create the quadrature points and weights (especially to allow easy reuse for custom quadrature rules) that the constructor calls. But I also think it is nice to have the current default behavior since it becomes very clear what you create. I would find it more logical to provide the rule as a kwarg, but perhaps not enough to motivate another breaking change.

fredrikekre:
Something like this?
```julia
create_quadrature(::RefShape, order) -> some default rule
create_quadrature(Legendre(), ::RefShape, order)
create_quadrature(Lobatto(), ::RefShape, order)
...
```
but if you extend this you probably just supply points and weights directly instead of building into this interface?

KnutAM:
Sorry, was on the phone writing yesterday...
I meant something like keeping the current user interface, but factoring out the point and weight generation,
```julia
create_quadrature(type::Legendre, ::RefShape, order::Int) -> (weights, points)
create_quadrature(type::Lobatto,  ::RefShape, order::Int) -> (weights, points)
function QuadratureRule{RefShape}(order::Int) where RefShape
    weights, points = create_quadrature(Legendre(), RefShape, order::Int)
[...]

PR #1007, opened 2024-07-01
Fix Quadrature rule docs and syms

termi-official:
Following the discussion in #1001 I just noticed that some references for the quadrature roles and the doc strings were not correct. This PR fixes both.


src/Quadrature/quadrature.jl
```diff
@@ -101,9 +108,11 @@ end
 for dim in 2:3
     @eval begin
         function QuadratureRule{RefSimplex{$dim}, T}(quad_type::Symbol, order::Int) where T
-            if $dim == 2 && quad_type === :legendre
+            if $dim == 2 && quad_type === :dunavant && order ≤ 8
```
KnutAM, line 112: Wouldn't this fail if I put `QuadratureRule{RefTriangle}(9)`?

src/Quadrature/quadrature.jl
```diff
@@ -57,13 +60,17 @@ function QuadratureRule{shape}(weights::Vector{T}, points::Vector{Vec{dim, T}})
     QuadratureRule{shape, T}(weights, points)
 end
 
+@inline _default_quadrature_rule(::RefHypercube) = :legendre
```
KnutAM, line 64: If we want the order-dependent defaults that we currently have, perhaps add `order` to these functions as well?

KnutAM (commented):
Nice that you found this! 

Since we probably don't want to change the defaults in the future, are these reasonable defaults? Seems strange to switch at order = 8? 

(Side-note: Also looks like dispatching on a type would be nicer as @fredrikekre suggested, but that would be a different PR)

src/Quadrature/quadrature.jl
termi-official: Fixed. I hope.

termi-official: Not sure. My argument is as follows. If we provide some default rule, then we should stick to it to not introduce behavior which could lead to potentially hard to track down bugs due to different properties in the quadrature rule.

src/Quadrature/quadrature.jl
termi-official: Note that (if I have not messed up anything here when reading through the references) the rule for order > 8 is actually a [Gauss-Jacobi](https://en.wikipedia.org/wiki/Gauss%E2%80%93Jacobi_quadrature) rule and the rule for order < 9 can properly integrate all polynomials on the triangle.

termi-official:
> Nice that you found this!
> 

Thanks, took me a while to realize too tbh.

> Since we probably don't want to change the defaults in the future, are these reasonable defaults? Seems strange to switch at order = 8?
> 

Actually I am really not sure about the rule or why it works in the in first place, because it comes with negative weights. I am pretty sure this can cause some trouble, but I cannot pin down why. See e.g. https://www.math.ntnu.no/emner/TMA4130/2021h/lectures/GaussQuadrature.pdf . But on the other hand rules with positive weights have more points in general.

> (Side-note: Also looks like dispatching on a type would be nicer as @fredrikekre suggested, but that would be a different PR)

I am not super sure about this. I think it would also be nice to provide extensions which add new quadrature rules.

fredrikekre:
The triangle ones are from https://github.com/Ferrite-FEM/Ferrite.jl/pull/514#issuecomment-1273850208 (unless you found it already).

termi-official:
> The triangle ones are from [#514 (comment)](https://github.com/Ferrite-FEM/Ferrite.jl/pull/514#issuecomment-1273850208) (unless you found it already).

Yes, and going into the basics source code tracking down the call graph manually I arrived at a call to Gauss-Jacobi generators.

termi-official:
Someone should verify this independently and then it should be good to go.

src/Quadrature/quadrature.jl
```diff
@@ -154,14 +172,14 @@ end
...
+    FacetQuadratureRule{shape}([::Type{T}, [quad_rule_type::Symbol|quad_rule_type::NTuple{Symbol},] order::Int)
     FacetQuadratureRule{shape}(face_rules::NTuple{<:Any, <:QuadratureRule{shape}})
     FacetQuadratureRule{shape}(face_rules::AbstractVector{<:QuadratureRule{shape}})
 
 Create a `FacetQuadratureRule` used for integration of the faces of the refshape `shape` (of
 type [`AbstractRefShape`](@ref)). `order` is the order of the quadrature rule.
-`quad_rule_type` is an optional argument determining the type of quadrature rule.
-Currently the :polyquad, `:legendre` and `:lobatto` rules are implemented, depending on the reference element.
+It defaults to the default rules for the individual face reference elements. Elements with mixed facet types
+requires passing a tuple of symbols for the rule, one for each facet type.
```
KnutAM, line 182: The ordering of the `quad_rule_type` is not clear/obvious here. I'm not sure if we want to introduce this as public API now? 

Would it be better (at least for now), to rather say that if you want non-default rules, you need to pass the `face_rules` manually?

termi-official: Indeed. I also struggle a bit with the wording in the docstring. It is essentially implemented as "first triangle, then quad" in the sense of "increasing size".

termi-official: bump

src/Quadrature/quadrature.jl
```diff
@@ -154,14 +172,14 @@ end
...
+    FacetQuadratureRule{shape}([::Type{T},] [quad_rule_type::Symbol,] order::Int)
     FacetQuadratureRule{shape}(face_rules::NTuple{<:Any, <:QuadratureRule{shape}})
     FacetQuadratureRule{shape}(face_rules::AbstractVector{<:QuadratureRule{shape}})
 
 Create a `FacetQuadratureRule` used for integration of the faces of the refshape `shape` (of
 type [`AbstractRefShape`](@ref)). `order` is the order of the quadrature rule.
-`quad_rule_type` is an optional argument determining the type of quadrature rule.
-Currently the :polyquad, `:legendre` and `:lobatto` rules are implemented, depending on the reference element.
+It defaults to the default rules for the individual face reference elements. Elements with mixed facet types
+do not have a way of specifying the quadrature rule via symbol for now.
```
KnutAM, line 182:
```suggestion
```suggestion
If not provided, the default `quad_rule_type` for each facet's reference shape is used.
For non-default `quad_rule_type`s on cells with mixed facet types (e.g. `Wedge`), the `face_rules` must be provided explicitly. 
```
```

File: src/Quadrature/quadrature.jl [hunk 5]
```diff
@@ -96,10 +108,16 @@ end
...
-            if $dim == 2 && quad_type === :legendre
-                data = _get_gauss_tridata(order)
-            elseif $dim == 3 && quad_type === :legendre
-                data = _get_gauss_tetdata(order)
+            if $dim == 2 && quad_type === :dunavant
+                data = _get_dunavant_gauss_tridata(order)
+            elseif $dim == 2 && quad_type === :gaussjacobi
+                data = _get_gaussjacobi_tridata(order)
+            elseif $dim == 3 && quad_type === :jinyun
+                data = _get_jinyun_tet_quadrature_data(order)
```
KnutAM, line 116: I think we should delete this option, since the current implementation is "keast" and we don't need to implement all possible quadrature rules. Of course, in the docs, we should acknowledge that `keast` orders 2 and 3 comes from jinyun.

src/Quadrature/quadrature.jl
```diff
@@ -16,8 +16,15 @@ using Base.Cartesian: @nloops, @ntuple, @nexprs
...
-Currently the :polyquad, `:legendre` and `:lobatto` rules are implemented, depending on the reference element.
+`quad_rule_type` is an optional argument determining the type of quadrature rule,
+currently the `:legendre` and `:lobatto` rules are implemented for hypercubes.
+For triangles up to order 8 the default rule is the one by `:dunavant` (see [Dun:1985:hde](@cite)) and for
+tetrahedra the default rule is `keast_positive` (see [Keast:1986:mtq](@cite)). Wedges and pyramids default
+to `:polyquad` (see [WitVin:2015:isq](@cite)).
+Furthermore we have implemented
+* `:gaussjacobi` for triangles (order 9-15)
+* `:jinyun` (see [Jin:1984:sgq](@cite)) for tetrahedra (order 1-4)
+* `:keast_minimal` (see [Keast:1986:mtq](@cite)) for tetrahedra (order 1-5)
```
KnutAM, line 27: The references don't work well in the docstrings, so I'd suggest only mentioning the defaults here, and refer the user to the online docs for further details. 
There we could have a table or similar showing the defaults and options for each reference shape, as well as the literature references and possibly some key properties of the rules.

src/Quadrature/quadrature.jl
```diff
@@ -109,7 +127,7 @@ for dim in 2:3
             for p in 1:size(data, 1)
                 points[p] = Vec{$dim,T}(@ntuple $dim i -> data[p, i])
             end
-            weights = data[:, $dim + 1]
+            weights = T.(data[:, $dim + 1])
```
KnutAM, line 130: I suppose this is to support Float32. 
This doesn't seem to be tested?

KnutAM (commented):
OK, started looking through this, and as I understand, we can skip the `:jinyun` type, as this is covered by `:keast`. 

I've checked that `:jinyn` matches the reference so far (except for n=4 where it matches :keast), but don't have more time now.

KnutAM (commented):
Checked keast 5 and keast positive 4 as well. They look ok (didn't check all points of course). 

I think adding a test that checks that the defaults are actually used would be good.

src/Quadrature/gaussquad_tri_table.jl
```diff
@@ -5,25 +5,25 @@
 #
 # Quadrature rules for orders 9 to 20 have been obtained using the
 # basix.make_quadrature(basix.CellType.triangle, n) calls of the
-# FEniCS / basix python package
+# FEniCS / basix python package, which corresponds to Gauss-Jacobi rules.
 #
 # see
 # https://docs.fenicsproject.org/basix/main/python/_autosummary/basix.html?highlight=quadraturetype#basix.make_quadrature
```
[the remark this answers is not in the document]
termi-official: I really struggle with finding the reference here. From what I understand it should be https://www.degruyter.com/document/doi/10.1515/crll.1826.1.301/html but the formula is slightly different ($w(x)=(x-1)^\alpha$?) from the formula I find in modern literature on ($w(x)=(x-1)^\alpha (x+1)^\beta$, e.g. https://epubs.siam.org/doi/pdf/10.1137/120889873).

docs/src/assets/references.bib
```diff
@@ -123,6 +123,49 @@ @article{Mu:2014:IP
...
+}
+@article{Keast:1986:mtq,
+  title={Moderate-degree tetrahedral quadrature formulas},
+  author={Keast, Patrick},
+  journal={Computer methods in applied mechanics and engineering},
+  volume={55},
+  number={3},
+  pages={339--348},
+  year={1986},
+  publisher={Elsevier}
```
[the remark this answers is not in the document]
termi-official: Oh, nice catch. Not sure what happened here.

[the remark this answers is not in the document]
termi-official: Okay sorry I thought this was some missing comma. What exactly is the problem here?

src/Quadrature/quadrature.jl
termi-official: Good catch. I thought this paths are already tested, because we had the option to set the type already before this PR. Will add it.

src/Quadrature/quadrature.jl
termi-official: I would leave it for reference, because for me it was not that straight forward to figure out how to get the rules for order 1..3 from the Keast paper. This way it is directly obvious.

docs/src/assets/references.bib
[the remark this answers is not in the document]
KnutAM: the doi / url was missing from the refs, thought that at least having one of them would be nice for getting the paper?

src/Quadrature/quadrature.jl
termi-official: Let us leave this as a separate PR because this is a bigger change. Especially working out the properties.

termi-official:
> I think adding a test that checks that the defaults are actually used would be good.

Should we do this separately, as I think we have not yet closed the discussion on the defaults?

src/Quadrature/quadrature.jl
KnutAM: IMO, we should make this clear from the docs rather than adding more complexity/options to the code.

KnutAM:
> Should we do this separately, as I think we have not yet closed the discussion on the defaults?

Can be done in a separate PR, but no need to wait until the defaults discussion is closed I'm thinking that such a test is just to ensure that we don't change by mistake, and changing the test makes it clear that it was intended.

src/Quadrature/gaussquad_tri_table.jl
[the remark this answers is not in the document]
KnutAM: But these are references for line intervals? Or do I miss something? 

Seems like dunavant provides rules up to `n=20`. (https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.1620210612) (Although some are based on previous works, as noted on the bottom of page 1131). 

Since higher than 7 were added in #514 and probably no one are saving quadrature data for such high orders, I think it could make sense to only use `dunavant`. But due to the different coordinate systems used in the papers and in the code, I find it hard to check where the points come from. 
How did you check this to figure out that starting at n = 9 this is gauss_jacobi?

PR #1010, opened 2024-07-02
Throw when asking for dof indices of vectorized interpolations

KnutAM:
Master
```julia
Ferrite.facedof_indices(ip)   # ((1, 2, 3, 4, 5, 6, 7, 8, 9),)
Ferrite.facedof_indices(ip^2) # ((),)
```
The last call errors with this PR. 
This works for `<entity>dof_indices`, `dirichlet_<entity>dof_indices`, and `<entity>dof_interior_indices`.

Example for when this behavior gives a problem,
```julia
julia> Ferrite.BCValues(ip, ip, FacetIndex)
Ferrite.BCValues{Float64}([1.0 0.0 0.0; 0.0 1.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 0.0 0.0; 1.0 0.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 0.0 0.0; 0.0 0.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 1.0 0.0; 0.0 0.0 0.0; … ; 0.0 0.0 1.0; 0.0 0.0 0.0], [3, 3, 3, 3], 0)

julia> Ferrite.BCValues(ip^2, ip^2, FacetIndex)
Ferrite.BCValues{Float64}(Array{Float64, 3}(undef, 18, 0, 4), [0, 0, 0, 0], 0)
```
The last call errors with this PR.

Of course, it could be discussed if these calls should work and return the correct indices as a "true" vector interpolation should (that does make sense to me), but there are some performance pitfalls that could occur if that is passed around, so for now I suggest erroring, and then we can add support for this in the future if desired.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -404,28 +404,29 @@ function _close_subdofhandler!(dh::DofHandler{sdim}, sdh::SubDofHandler, sdh_ind
     ip_infos = InterpolationInfo[]
     for interpolation in sdh.field_interpolations
         ip_info = InterpolationInfo(interpolation)
+        base_ip = get_base_interpolation(interpolation)
         begin
             next_dof_index = 1
-            for vdofs ∈ vertexdof_indices(interpolation)
+            for vdofs ∈ vertexdof_indices(base_ip)
                 for dof_index ∈ vdofs
                     @assert dof_index == next_dof_index "Vertex dof ordering not supported. Please consult the dev docs."
                     next_dof_index += 1
                 end
             end
-            for vdofs ∈ edgedof_interior_indices(interpolation)
+            for vdofs ∈ edgedof_interior_indices(base_ip)
                 for dof_index ∈ vdofs
                     @assert dof_index == next_dof_index "Edge dof ordering not supported. Please consult the dev docs."
                     next_dof_index += 1
                 end
             end
-            for vdofs ∈ facedof_interior_indices(interpolation)
+            for vdofs ∈ facedof_interior_indices(base_ip)
                 for dof_index ∈ vdofs
                     @assert dof_index == next_dof_index "Face dof ordering not supported. Please consult the dev docs."
                     next_dof_index += 1
                 end
             end
-            for dof_index ∈ volumedof_interior_indices(interpolation)
-                @assert next_dof_index <= dof_index <= getnbasefunctions(interpolation) "Cell dof ordering not supported. Please consult the dev docs."
+            for dof_index ∈ volumedof_interior_indices(base_ip)
+                @assert next_dof_index <= dof_index <= getnbasefunctions(base_ip) "Cell dof ordering not supported. Please consult the dev docs."
             end
         end
         push!(ip_infos, ip_info)
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -1505,6 +1505,18 @@ end
 get_n_copies(::VectorizedInterpolation{vdim}) where vdim = vdim
 InterpolationInfo(ip::VectorizedInterpolation) = InterpolationInfo(ip.ip, get_n_copies(ip))
 
+# Error when trying to get dof indicies from vectorized interpolations.
+# Currently, this should only be done for the scalar interpolation.
+function _entitydof_indices_vectorized_ip_error(f::Symbol)
+    throw(ArgumentError(string(f, " is not implemented for VectorizedInterpolations and should be called on the scalar base interpolation")))
+end
+vertexdof_indices(::VectorizedInterpolation) = _entitydof_indices_vectorized_ip_error(:vertexdof_indices)
+edgedof_indices(::VectorizedInterpolation) = _entitydof_indices_vectorized_ip_error(:edgedof_indices)
+facedof_indices(::VectorizedInterpolation) = _entitydof_indices_vectorized_ip_error(:facedof_indices)
+edgedof_interior_indices(::VectorizedInterpolation) = _entitydof_indices_vectorized_ip_error(:edgedof_interior_indices)
+facedof_interior_indices(::VectorizedInterpolation) = _entitydof_indices_vectorized_ip_error(:facedof_interior_indices)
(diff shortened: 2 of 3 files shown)


termi-official:
I think this goes into the wrong direction. Shoudn't we rather fix the dof indices for vectorized interpolations than making them error?

KnutAM:
Currently, we special-case a lot of code by using the base interpolation. So I agree that we should support this somehow. But since we currently give the wrong result, I think this solution improves the situation. When including #798 and supporting more features for this (such as L2 projection and vtk export) it will be easier to see how, where and if this makes sense.

Right now, this actually caught the error that the dof-distribution check was skipped for vectorized interpolations [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/8c897daa10d74ebff0d38b276c92e07d42690605/src/Dofs/DofHandler.jl#L408-L430).

While mathematically nice to use `VectorizedInterpolations` just as a standard vector interpolation, my guess is that there could be several places where this is `vdim` times more costly (at least, cf. e.g. L2Projection). 

For future reference though, here is some code I had to do rather generalize the results. 
But I don't think it makes sense to change to code-base this much now right before 1.0 (since this *should* be possible to do in a no-breaking way later). 

```julia
# Don't know any function for `jointuple`, ought to exist. 
jointuple(t1::Tuple, t2::Tuple, args::Tuple...) = jointuple((t1..., t2...), args...)
jointuple(t::Tuple) = t

function dofinds(dof_ind_fun::F, ip::VectorizedInterpolation{dim}) where {F, dim}
    return map(inds -> jointuple(map(i -> ntuple(j -> (i-1)*dim + j, dim), inds)...), dof_ind_fun(ip.ip))
end
Ferrite.edgedof_indices(ip::VectorizedInterpolation) = dofinds(Ferrite.edgedof_indices, ip)
```

termi-official:
> Currently, we special-case a lot of code by using the base interpolation. So I agree that we should support this somehow. But since we currently give the wrong result, I think this solution improves the situation. When including #798 and supporting more features for this (such as L2 projection and vtk export) it will be easier to see how, where and if this makes sense.
> 
> [...]
> While mathematically nice to use `VectorizedInterpolations` just as a standard vector interpolation, my guess is that there could be several places where this is `vdim` times more costly (at least, cf. e.g. L2Projection).

We can still support specializations for these cases. I am more concerned about the API doing the right thing if you want to check e.g. consistency or you want to try out something without worring about the correct way to vectorization (/tensorization) manually.

> 
> For future reference though, here is some code I had to do rather generalize the results. But I don't think it makes sense to change to code-base this much now right before 1.0 (since this _should_ be possible to do in a no-breaking way later).
> 
> ```julia
> # Don't know any function for `jointuple`, ought to exist. 
> jointuple(t1::Tuple, t2::Tuple, args::Tuple...) = jointuple((t1..., t2...), args...)
> jointuple(t::Tuple) = t
> 
> function dofinds(dof_ind_fun::F, ip::VectorizedInterpolation{dim}) where {F, dim}
>     return map(inds -> jointuple(map(i -> ntuple(j -> (i-1)*dim + j, dim), inds)...), dof_ind_fun(ip.ip))
> end
> Ferrite.edgedof_indices(ip::VectorizedInterpolation) = dofinds(Ferrite.edgedof_indices, ip)
> ```

Yea I also ran into the issue on how to represent the vdim properly when designing this, but I stopped when I realized that access to the interpolation is not type stable in the dof handler. Will try to approach this in the future again (or Abd will pick up https://github.com/Ferrite-FEM/Ferrite.jl/pull/829 again).

Should be non-breaking for most users tho.

KnutAM:
I guess we we both agree that in the future we would like to allow this! 
Are you ok with including this fix for now to prevent silent errors? 

We could open a separate issue "treat VectorizedInterpolation as true vector interpolation" which would be cool if that doesn't degrade performance and we could get rid of some of the special-casing.

termi-official:
> I guess we we both agree that in the future we would like to allow this! Are you ok with including this fix for now to prevent silent errors?
> 
Yes, good. :)

> We could open a separate issue "treat VectorizedInterpolation as true vector interpolation" which would be cool if that doesn't degrade performance and we could get rid of some of the special-casing.

Agreed that we should add an issue (independent of this PR). We will see a performance hit for some operations if we do not use the property, because we increase the number of arithmetic operations. However, I think this should be fine in cases which are not performance-critical to keep the code easier.

termi-official:
> Example for when this behavior gives a problem,
> 
> ```julia
> julia> Ferrite.BCValues(ip, ip, FacetIndex)
> Ferrite.BCValues{Float64}([1.0 0.0 0.0; 0.0 1.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 0.0 0.0; 1.0 0.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 0.0 0.0; 0.0 0.0 0.0; … ; 0.0 0.0 0.0; 0.0 0.0 0.0;;; 0.0 1.0 0.0; 0.0 0.0 0.0; … ; 0.0 0.0 1.0; 0.0 0.0 0.0], [3, 3, 3, 3], 0)
> 
> julia> Ferrite.BCValues(ip^2, ip^2, FacetIndex)
> Ferrite.BCValues{Float64}(Array{Float64, 3}(undef, 18, 0, 4), [0, 0, 0, 0], 0)
> ```
> 
> The last call errors with this PR.
> 
Should the last one error? If not, could you add the fix in this PR? If I see it correctly then it should be a straightforward, as there seems to be some dispatch missing in BCValues to de-vectorize the interpolation correctly.

KnutAM:
> Should the last one error?

BCValues are internal, and we de-vectorize [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/8c897daa10d74ebff0d38b276c92e07d42690605/src/Dofs/ConstraintHandler.jl#L828-L830), and the devectorized ip is used for multiple places.

And in the future, if we want to go down the path of treating a vectorized ip as a true vector ip, then BCValues should probably duplicate the reference coordinates etc. `vdim` times since that is what a vector interpolation would have, so I'm not sure if it is correct to simply devectorize.

Issue #1009, opened 2024-07-02
Constraining dof > ndofs(dh) fails


KnutAM:
With the nice new SparsityPattern (#888), we can now extend the equation system beyond ndofs(dh). (I'm seeing how #596 could work now without changing the dofhandler). 

I realized now that the `ConstraintHandler` fails if adding constraints to these, since it assumes the maximum size is ndofs(dh). (This now happens in `_sorted_setdiff!` from #860 , before I think `ch.free_dofs` would simply be wrong but not error). In other places, we use `ndofs(ch.dh)` to check dimensions of the "to-be-constrained" vector, but this should really check for `last(ch.prescribed_dofs)` to be safe from out of bounds access. 

An easy solution would be to add an optional keyword arg to `ConstraintHandler(dh::DofHandler; n_dofs=ndofs(dh))`, but I'm also not sure what we should call this since `ndofs` is taken for the function name. Anyone have some thoughts on this? 

MWE
```julia
grid = generate_grid(Quadrilateral, (2,2));
dh = close!(add!(DofHandler(grid), :u, Lagrange{RefQuadrilateral,1}()));
ch = ConstraintHandler(dh);
Ferrite.add_prescribed_dof!(ch, ndofs(dh) + 1, NaN, nothing)
close!(ch)
```
Gives `ERROR: BoundsError: attempt to access 8-element Vector{Int64} at index [9]`

lijas:
If a user wants to add global Lagrange multipliers/dofs, I think it should be possible to add these to the DofHandler directly, e.g. `add_global_dofs!(dh, :lambda, 3)`. The sparsity-pattern can then be handled by the user. It seems reasonable that DofHandler should be aware of all the dofs of the problem.

KnutAM:
Yes, that would be the solution here I guess: https://github.com/Ferrite-FEM/Ferrite.jl/pull/596/files (but with better names, global dofs is probably a good one)
It is nice to be able to store the dofs associated with a "field"

An alternative is adding a "GlobalInterpolation" that we discussed earlier on slack. These dofs are part of the celldof for each cell in the subdofhandler (https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...kam/GlobalDofs). It would also be possible to have two variants, one connected to cells and one not connected?

termi-official:
[...]

PR #1012, opened 2024-07-03
Add global dofs

KnutAM:
Solves #1009 following @lijas's suggestion. 

Currently the new functionality is not exported, and I'm considering if it makes sense to add this to devdocs until we battle-test it a bit before committing to this interface.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -96,6 +96,7 @@ end
 mutable struct DofHandler{dim,G<:AbstractGrid{dim}} <: AbstractDofHandler
     const subdofhandlers::Vector{SubDofHandler{DofHandler{dim, G}}}
     const field_names::Vector{Symbol}
+    const global_dofs::OrderedDict{Symbol, UnitRange{Int}}
     # Dofs for cell i are stored in cell_dofs at the range:
     #     cell_dofs_offset[i]:(cell_dofs_offset[i]+ndofs_per_cell(dh, i)-1)
     const cell_dofs::Vector{Int}
@@ -131,7 +132,8 @@ close!(dh)
 function DofHandler(grid::G) where {dim, G <: AbstractGrid{dim}}
     ncells = getncells(grid)
     sdhs = SubDofHandler{DofHandler{dim, G}}[]
-    DofHandler{dim, G}(sdhs, Symbol[], Int[], zeros(Int, ncells), zeros(Int, ncells), false, grid, -1)
+    global_dofs = OrderedDict{Symbol, UnitRange{Int}}()
+    DofHandler{dim, G}(sdhs, Symbol[], global_dofs, Int[], zeros(Int, ncells), zeros(Int, ncells), false, grid, -1)
 end
 
 function Base.show(io::IO, mime::MIME"text/plain", dh::DofHandler)
@@ -150,6 +152,10 @@ function Base.show(io::IO, mime::MIME"text/plain", dh::DofHandler)
             end
             println(io, "    ", repr(fieldname), ", ", field_type)
         end
+    end
+    if length(dh.global_dofs) > 0
+        gdof_str = join([string(key, " (", length(r), ")") for (key, r) in dh.global_dofs], ", ")
+        println(io, "  Global dofs: ", gdof_str)
     end
     if !isclosed(dh)
         print(io, "  Not closed!")
@@ -323,6 +329,29 @@ function add!(dh::DofHandler, name::Symbol, ip::Interpolation)
 end
 
 """
+    add_global_dofs!(dh::DofHandler, name::Symbol, n::Int)
+
+Add `n` global degrees of freedom to `dh`. The degree of freedom numbers can be obtained
+after closing `dh` with [`get_global_dofs`](@ref).
+"""
+function add_global_dofs!(dh::DofHandler, name::Symbol, n::Int)
+    @assert !isclosed(dh)
+    haskey(dh.global_dofs, name) && throw(ArgumentError(string(":", name, " is already a global dof")))
+    dh.global_dofs[name] = (-n):(-1) # Range shifted in close
+end
+
+"""
+    get_global_dofs(dh::DofHandler, name::Symbol)
+
+Get a `UnitRange{Int}` describing the global dofs associated with `name`.
+Such dofs can be added with [`add_global_dofs!`](@ref).
+"""
+function get_global_dofs(dh::DofHandler, name::Symbol)
+    @assert isclosed(dh)
+    return dh.global_dofs[name]
+end
+
+"""
     close!(dh::AbstractDofHandler)
 
 Closes `dh` and creates degrees of freedom for each cell.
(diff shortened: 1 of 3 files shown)


src/Dofs/DofHandler.jl
termi-official, line 1: Shouldn't `celldofs[!]` return these additional dofs, too?

src/Dofs/DofHandler.jl
```diff
@@ -322,6 +328,29 @@ function add!(dh::DofHandler, name::Symbol, ip::Interpolation)
...
+    dh.global_dofs[name] = (-n):(-1) # Range shifted in close
+end
+
+"""
+    get_global_dofs(dh::DofHandler, name::Symbol)
+
+Get a `UnitRange{Int}` describing the global dofs associated with `name`.
+Such dofs can be added with [`add_global_dofs!`](@ref).
+"""
+function get_global_dofs(dh::DofHandler, name::Symbol)
```
fredrikekre, line 349: Perhaps
```suggestion
function global_dof_range(dh::DofHandler, name::Symbol)
```
(https://github.com/Ferrite-FEM/Ferrite.jl/blob/17c9ef20a77ebd6315c61ce9cc561c1e8504ef54/src/Dofs/DofTools.jl#L86)

src/Dofs/DofHandler.jl
KnutAM: No, the global dofs in this PR have no coupling with the cells, and this can be added for example via AffineConstraints. 

For "the other type of global dofs", I think we should go with something like https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...kam/GlobalDofs, which allows you to do 
```julia
dh = DofHandler(grid)
sdh1 = SubDofHandler(dh, domain1)
add!(sdh1, :u, Lagrange{RefTriangle,1}())
add!(sdh1, :g, GlobalInterpolation(n)) # Add n extra dofs that are connected to all cells in sdh1
sdh2 = SubDofHandler(dh, domain2)
add!(sdh2, :u, Lagrange{RefTriangle,1}())
close!(dh)
```
If you only want the global dof on some subdofhandlers.

src/Dofs/DofHandler.jl
KnutAM: Thought about it, but not sure how this plays with `dof_range` which gives you the local dof range in the cell. Global in that respect here refers both to that the dofs are global in nature, and that it is the global (and not local) numbering that is included in the range. 

But if you think that is ok, then I think your naming suggestion is better.

fredrikekre: I think it has the same meaning: "What range in the global dof numbering is field x" is what both of them correspond to, no?

src/Dofs/DofHandler.jl
termi-official: > [...]  the global dofs in this PR have no coupling with the cells [...]

Okay, then I do not understand what the purpose of feature. If you have an independent dof, then you have effectively two independent systems. Hence we can just have a separate dof handler with the additional dofs and assemble two separate systems. Can you point me to the detail which I am missing out?

> [...]  this can be added for example via AffineConstraints.

Not sure how this fits in the statement above. From what I originally understood in the related discussion in the linked issue is that you just want to have some mechanism to add "simple constraints". Why should the dof handler manage this and why can't we add have some `LagrangeMultiplierConstraint` which is managed by the constraint handler instead? Or what exactly is the advantage of this design, having the dof handler managing these constraints (partially)?

src/Dofs/DofHandler.jl
termi-official: Related to this: How will this interact with the renumbering logic?

src/Dofs/DofHandler.jl
KnutAM: > Okay, then I do not understand what the purpose of feature. If you have an independent dof, then you have effectively two independent systems.

By having the DofHandler aware of the total number of dofs, the constraint handler can rely on `ndofs(dh)` to return the number of rows and columns in the system matrix to which it should add constraints. In #1009 I first proposed passing this info directly to the ConstraintHandler, but this PR introduces the alternative solution that passes this information via the DofHandler. With this design, one can do
```julia
dh = create_dh(...)
ch = ConstraintHandler(dh) ...
K = allocate_matrix(dh, ch)
```
whereas with my original suggestion you need 
```julia
dh = create_dh(...)
n = setup_extra_dofs(...)
ndofs_total = ndofs(dh) + n
ch = ConstraintHandler(dh; ndofs_total) # New feature that we would need to add. 
sp = SparsityPattern(ndofs_total, ndofs_total)
add_sparsity_entries!(sp, dh, ch)
K = allocate_matrix(sp)
```

> Or what exactly is the advantage of this design, having the dof handler managing these constraints (partially)?

The DofHandler doesn't really manage these constraints, it is just aware of the extra dofs. Currently, my workaround is doing `dh.ndofs += n` after `close!(dh)`, which tricks the setup to work, but provides no info about the additional dofs, this I have to manage myself. The main advantage with adding fields to the dofhandler is IMO that you can directly request these from the `DofHandler`, and that `ndofs(dh)` gives the correct number of the degrees of freedom in the system.

KnutAM: But it might also be possible to do
```julia
dh = DofHandler(grid)
add!(dh, ...)
add_global_dofs!(dh, ...)
close!(dh)
ch = ConstraintHandler(dh) ...
add!(ch, ...)
add!(ch, LagrangeMultiplierConstraint(:lambda, n, AffineConstraintInfos)) # Extends with n number of dofs
close!(ch)
K = allocate_matrix(dh, ch)
```
One advantage with this setup is that it potentially could make it easier to "deactivate" some multipliers, keeping the sparsity pattern the same (nice for e.g. contact simulations). The disadvantage is that it would be less flexible then the suggestions above.

termi-official: So we are really talking about just adding some constraint $\sum_i u_i = 0$? Or why can you skip the `setup_global_dofs` to build the connectivity in the the examples 1. and 3.?

> A disadvantage I see is that it is less flexible then the suggestions above.

I see this as an advantage, because is is clear what to do in all scenarios. Whereas the in proposed design it seems heavily underdesigned, as it is absolutely unclear to my how the interaction between sparsity pattern, constraints, assembly and distributed parallelization should be (and it is also not clear from going over the code). So maybe it is easier to have a special dof handler for such specific corner cases?

KnutAM: > So we are really talking about just adding some constraint $\sum_i u_i = 0$?

Yes, the case I hit was adding an extra dof. Then I add many AffineConstraints which constrain other dofs to this extra dof. 
(The reason for having it in the system, as opposed to fixing it by hard-coding it in the affine constraints is that I need this for computing correct residuals and sensitivities later, and there are cases when I don't constrain it, but still want the AffineConstraints to the extra dof).

> Whereas the in proposed design it seems heavily underdesigned, as it is absolutely unclear to my how the interaction between sparsity pattern, constraints, assembly and distributed parallelization should be (and it is also not clear from going over the code).

The only difference apart from the possibility to get information from the DofHandler, is that it increases the value of `ndofs(dh)`, implying that the total equation system is larger. This *allows* you to add constraints to these extra dofs, which then will affect e.g. the sparsity system as always. It does not change anything for the assembly. For distributed parallelization it should have no effect either, the effect comes if you tie two dofs together via Affine constraints, which is not different with this PR from before. 

So to summarize, an absolute minimal change with this pr would be to add a field `n_global_dofs::Int` to the dofhandler (allowing you to add a given number of extra dofs, but without any naming etc. of these), and redefine `ndofs(dh) = dh.ndofs + dh.n_global_dofs` (or update `dh.ndofs` to this value during `close!`.

termi-official:
I guess this is also resolved in #1482 ?

PR #1013, opened 2024-07-03
remove ApplyStrategy enums

KnutAM:


src/Ferrite.jl
```diff
@@ -8,7 +8,7 @@ using Base:
 using EnumX:
     EnumX, @enumx
 using LinearAlgebra:
-    LinearAlgebra, Symmetric, Transpose, cholesky, det, norm, pinv, tr
```
fredrikekre, line 11: I guess this was a false positive in ExplicitImports.jl then?

fredrikekre: @ericphanson ^

KnutAM: Yes, looks like [this](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1013/commits/63f62ed9e4f8a74f0856abf40b3a672a091a464f#diff-82ed2dacb8787323aa0010af6c9938016eff02dd4154c4afe0e9d492bb84ef94L599) was the offending line

KristofferC (approved):

Issue #1004, opened 2024-07-01
The WriteVTK wrapper: Multiblockfiles and VTKFileCollection


lijas:
Before the introduction of the WriteVTK wrapper in #692 , it was easier to work with multiblock files and paraview collection files (to save multiple time steps). From one of my packages:

```
pvd = paraview_collection(filename)
...
vtmfile = WriteVTK.vtk_multiblock(filename)
for part in parts
    vtkfile = vtk_grid(partname, part) #e.g special output for IGA or Immersed grids
    WriteVTK.multiblock_add_block(vtmfile, vtkfile)
    ...
end
vtk_save(vtmfile)
pvd[state.t] = vtmfile
```

This has changed now:

1. The creation of paraview collection files has changed. Now they require a grid/dofhadler as an argument for the [constructor](https://github.com/Ferrite-FEM/Ferrite.jl/blob/2afbd6a2f0cba65070ff6c8453bde4c7a4f89995/src/Export/VTK.jl#L74) (`VTKFileCollection(name, grid_or_dh)`), which is not always known at the construction of the paraview collection. We should either make it possible to have `grid_or_dh===nothing` or remove it completely (depending on if we only want to have a thin wrapper or extend it with our own functionality).
2. We dont have any wrappers for multiblockfiles atm.

KnutAM:
Adding 1.0 since removing `grid_or_dh` from `VTKFileCollection` will be breaking, since this will make it required to include when calling `addstep!`. 

Also pointing the comments where the multiblock was discussed in #692 since the discussion is so long there. 
* https://github.com/Ferrite-FEM/Ferrite.jl/pull/692#issuecomment-1733075533
* https://github.com/Ferrite-FEM/Ferrite.jl/pull/692#issuecomment-1733338201
* https://github.com/Ferrite-FEM/Ferrite.jl/pull/692#issuecomment-1733382029
* https://github.com/Ferrite-FEM/Ferrite.jl/pull/692#issuecomment-1735017761

KnutAM:
Just thinking out loud for some options

## 1) Introduce wrappers for the collection types
```julia
pvd = VTMFileCollection(filename)
...
addstep!(pvd, t) do vtm # ::VTMFile
    for part in parts
        VTKFile(vtm, partname, part.grid) do vtk 
            write_solution(vtk, dh, u)
            write_...
        end
    end
end

# Without `addstep!`
pvd = VTMFileCollection(filename)
...
[...]

PR #1014, opened 2024-07-03
Rename VTKFile to VTKGridFile

KnutAM:
xref #1004 

Does not solve #1004, but as pointed out by @termi-official, it is clear that having `Ferrite.VTKFile` (concrete type, wrapper of `WriteVTK.DatasetFile`) being something different than `WriteVTK.VTKFile` (abstract supertype of most `WriteVTK` filetypes) can be very confusing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -219,7 +219,7 @@ more discussion).
   -     vtk_point_data(vtk, nodal_data, "my node data")
   -     vtk_point_data(vtk, proj, projected_data, "my projected data")
   -     vtk_cell_data(vtk, proj, projected_data, "my projected data")
-  + VTKFile(name, dh) do vtk
+  + VTKGridFile(name, dh) do vtk
   +     write_solution(vtk, dh, a)
   +     write_node_data(vtk, nodal_data, "my node data")
   +     write_projection(vtk, proj, projected_data, "my projected data")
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -427,8 +427,8 @@ more discussion).
   omitting type piracy, ([#835][github-835]) implemented `isequal` and `hash` for `BoundaryIndex` datatypes.
 
 - **VTK export**: Ferrite no longer extends methods from `WriteVTK.jl`, instead the new types
-  `VTKFile` and `VTKFileCollection` should be used instead. New methods exists for writing to
-  a `VTKFile`, e.g. `write_solution`, `write_cell_data`, `write_node_data`, and `write_projection`.
+  `VTKGridFile` and `VTKFileCollection` should be used instead. New methods exists for writing to
```
fredrikekre, line 430:
```suggestion
  `VTKGridFile` and `VTKCollectionFile` should be used instead. New methods exists for writing to
```
for symmetry?

KnutAM: See suggestion to remove completely here: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1015

KnutAM:
An alternative name is perhaps simply `VTKGrid` (motivation that it combines `[Write]VTK` and `Ferrite.Grid`). `File` doesn't perhaps add much information to the name I think as VTK is already a file type? What do you think?

KnutAM:
> What do you think? 

After discussions with @termi-official on Slack, realized that isn't specific enough. So let's stay with VTKGridFile then.

PR #1020, opened 2024-07-09
Update degrees of freedom topic to reflect updates for 1.0

KnutAM:
Noticed that this was out of date, and also errored in the docs due to deprecated syntax. 

Perhaps we should have a bit stricter docs build to not let these errors pass? 

Edit: 
* Doc error fixed in #889
* Issue regarding stricter docs build: #1021 
* Fix here: In v1.0 we only allow adding fields by providing the interpolation, the docs are updated in this PR to reflect that.


termi-official:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/889

KristofferC (approved):

Issue #678, opened 2023-04-14
Binary export extension


koehlerson:
I use JLD2 for saving my simulation results. The two most important pieces that I save are solution vectors and a `DofHandler`.  The only annoying part is if structs change they can't be reloaded that easily. However, JLD2 offers a lot of API to cover that by e.g. [custom serialization](https://juliaio.github.io/JLD2.jl/dev/customserialization/) and [type remapping](https://juliaio.github.io/JLD2.jl/dev/advanced/#Explicit-Type-Remapping)

What do you think about providing a binary export extension for e.g. JLD2 where we specify a custom serialization for the `DofHandler`?

KristofferC:
In my opinion, if you want to reliably serialize/deserialize arbitrary Julia object then the best is to

- record Julia version
- record package versions (with a manifest)
- use the `Serialization` stdlib.

Trying to keep backwards compatibility by writing these cusom serialization functions sounds kind of annoying to me.

koehlerson:
I'd imagine to write as a sane default the `.toml` files of the current env to the binary output in the same way JLSO does. I'm not sure if I agree fully on the custom serialization part. There are building blocks of the `DofHandler` which should be easily abstracted away by e.g. scalars that doesn't change over versions. One particular example I can think of are the interpolations.

It is very likely that I can describe a Lagrange interpolation by a triple of `refshape`, `order` and `dim` which are invariant to changes of `Interpolation`. I wouldn't aim for a full backwards compatibility, just deconstructing the structs to julia native/primitive types as much as possible. I thought that the type remapping part could potentially be used after big breaking changes

KnutAM:
I think this would be a nice feature! 
But I do share Kristoffer's concerns that it might add quite a lot to Ferrite's maintenance load...

From my point of view, (or perhaps as a starting point), the current issue is to save the constraint handler when used with anonymous functions. Are there any nice solutions for this? (xref my question/attempt https://github.com/JuliaIO/JLD2.jl/issues/288#issuecomment-1519076407). 
For now I'm creating custom structs/using `Returns` ([example](https://knutam.github.io/FerriteProblems.jl/dev/examples/plasticity/#Problem-definition)). But it would be nice to be able to reload without erroring (even if one cannot recover the function)

koehlerson:
> I think this would be a nice feature! But I do share Kristoffer's concerns that it might add quite a lot to Ferrite's maintenance load...

[...]

PR #1024, opened 2024-07-19
Save the grids

Firimo:
Is nobody thinking of the grids?
TOML can save the grids!


src/Grid/parse_the_grids.jl
```diff
@@ -0,0 +1,120 @@
+function parse_the_grids(path::String="/tmp/grid.toml")
```
termi-official, line 1: Since this function is exported users will see it. Hence a doc string should be added for users.

src/Grid/save_the_grids.jl
```diff
@@ -0,0 +1,213 @@
...
+
+	t_string = t_string[1:end-1]
+
+	t_string *= "]"
+
+	return t_string
+end
+
+# find a small number of digits for precise storage
+function toml_saves_the_grids(x::AbstractFloat)
```
termi-official, line 138: I am not confident that this works properly for all float types. Seems like this implementation is specific to Float64 and (more narrow types). But this function essentially generates a string of the truncated real number represenation?

src/Grid/save_the_grids.jl
```diff
@@ -0,0 +1,213 @@
...
+		t_string = @sprintf("%.12E",x)
+	elseif N == 14
+		t_string = @sprintf("%.13E",x)
+	elseif N == 15
+		t_string = @sprintf("%.14E",x)
+	elseif N == 16
+		t_string = @sprintf("%.15E",x)
+	elseif N == 17
+		t_string = @sprintf("%.16E",x)
+	end
```
termi-official, line 183: This should just be something like
```suggestion
	t_string = @sprintf("%.$(N-1)E",x)
```

termi-official (commented):
Thanks for the PR! I think having something along these lines can be helpful to users. I left some comments. Furthermore, the PR is missing test coverage and docs for the new functionality.

KnutAM:
> Is nobody thinking of the grids?

https://github.com/Ferrite-FEM/Ferrite.jl/issues/678#issuecomment-1810739315

Instead of parsing to TOML, then I think @koehlerson's idea of supporting saving to HDF5 should be easier and more efficient.

KristofferC:
I personally rather not add a bespoke textual format for this. I don't really see the use case. I find it best to go all out Julia which means using Serializatin stdlib or something standard like HDF5, Arrow, etc.

Firimo:
Yes. All you make good arguments. I could counter them, but that annoys me.
Let me just ask you this: Should there not be the option of  plain-text?

fredrikekre:
Perhaps it could/should live in https://github.com/Ferrite-FEM/FerriteMeshParser.jl?

KnutAM:
> Let me just ask you this: Should there not be the option of plain-text?

From my point of view, I'm not sure what the advantage is of having a plain text format? 
Especially since you normally don't create the grid using Ferrite, and rather import it from some other format? But I'd be interested to hear your thoughts, maybe I'm missing an important case here? 

> Perhaps it could/should live in https://github.com/Ferrite-FEM/FerriteMeshParser.jl?

There, I would suggest to follow some existing mesh format in that case. 
* The vtk can be written in ASCII format, so perhaps that would be an easy way to allow reading in VTK by using ReadVTK? 
* I don't think using the Abaqus input file format is nice, since the element codes are not bijective (e.g. C3PS and C3PE are both Triangle). 
* Perhaps there are other existing nice plain-text formats to which it could make sense to have a writer and parser?

Issue #993, opened 2024-06-27
Sparsity pattern entries for interface coupling is slow


fredrikekre:
For example:
```julia
using Ferrite

const n = 20

const grid = generate_grid(Hexahedron, (n, n, n))
const top = ExclusiveTopology(grid)

const dh = DofHandler(grid)
add!(dh, :u, Lagrange{RefHexahedron, 2}()^3)
add!(dh, :p, Lagrange{RefHexahedron, 1}())
add!(dh, :q, DiscontinuousLagrange{RefHexahedron, 1}())
close!(dh)

# Just couple q-q over the interface
const interface_coupling = [
    false false false
    false false false
    false false true
]

@time create_sparsity_pattern(dh)
@time create_sparsity_pattern(dh; topology=top, cross_coupling = interface_coupling)
```
Gives 
```
  0.708471 seconds (3.04 k allocations: 3.653 GiB, 13.00% gc time, 2.40% compilation time)
 16.754851 seconds (217.19 k allocations: 4.485 GiB, 1.05% gc time, 0.03% compilation time)
```

(Also happens with #888).

KristofferC:
Unclear if https://github.com/Ferrite-FEM/Ferrite.jl/pull/1028 is enough to close this.

PR #1028, opened 2024-07-29
speed up adding sparsity pattern entries with coupling

KristofferC:
Ref https://github.com/Ferrite-FEM/Ferrite.jl/issues/993

As written then `in` check for `j` was performed on every iteration over `i`.

```
Before:
julia> @time allocate_matrix(dh; topology=top, interface_coupling = interface_coupling)
 25.123126 seconds (217.41 k allocations: 1.820 GiB, 0.87% gc time)


After:
julia> @time allocate_matrix(dh; topology=top, interface_coupling = interface_coupling)
 12.786278 seconds (217.84 k allocations: 1.821 GiB, 1.79% gc time)
```

I think the approach here of doing this `in` check like this has to be rethought do.

### DIFF
--- a/src/Dofs/sparsity_pattern.jl
+++ b/src/Dofs/sparsity_pattern.jl
@@ -587,11 +587,9 @@ function _add_constraint_entries!(
     return sp
 end
 
-function _add_interface_entry(sp::SparsityPattern, coupling_sdh::Matrix{Bool}, dof_i::Int, dof_j::Int,
+function _add_interface_entry(sp::SparsityPattern,
         cell_field_dofs::Union{Vector{Int}, SubArray}, neighbor_field_dofs::Union{Vector{Int}, SubArray},
         i::Int, j::Int, keep_constrained::Bool, ch::Union{ConstraintHandler, Nothing})
-
-    coupling_sdh[dof_i, dof_j] || return
     dofi = cell_field_dofs[i]
     dofj = neighbor_field_dofs[j]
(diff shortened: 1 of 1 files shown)


KristofferC:
I see that the check for the coupling being non-zero is done inside `_add_interface_entry`. Lifting it out to avoid all these `in` for the zero case is probably worth while.

fredrikekre:
There is already a duplicate-check here https://github.com/Ferrite-FEM/Ferrite.jl/blob/98e2b674a8c6d740a779c022de31f73051cb353c/src/Dofs/sparsity_pattern.jl#L170 so possibly there shouldn't be one in the function modified here at all?

KristofferC:
I now have

```
julia> @time allocate_matrix(dh);
  2.430063 seconds (1.21 k allocations: 1.767 GiB, 14.28% gc time)

julia> @time allocate_matrix(dh; topology, interface_coupling)
  3.207378 seconds (1.04 M allocations: 1.880 GiB, 13.06% gc time)
```

In #993, the first only took 0.7 seconds. I'm on a pretty fast CPU so did something regress @fredrikekre?

KristofferC:
> so possibly there shouldn't be one in the function modified here at all?

If I just flat out comment out the `in` checks the tests fail at least.

KristofferC:
@AbdAlazezAhmed maybe you could look over this.

AbdAlazezAhmed:
LGTM :D

PR #1035, opened 2024-08-01
Update topology example

### LINKED ISSUE #1031: Topology optimization example in code gallery should use better use of topology
https://github.com/Ferrite-FEM/Ferrite.jl/blob/f7f2592bd26e8005a739cfe19169097c0146e269/docs/src/literate-gallery/topology_optimization.jl#L194-L216

Some problems with this code:

- `_nfacets = nfacets(dh.grid.cells[1]) `
- `dh.grid` is an implementation detail?
- `opp` could be a tuple
- It should use `get_facet_facet_neighborhood` instead (but this is undocumented?)

KristofferC:
Fixes #1031

### DIFF
--- a/docs/src/literate-gallery/topology_optimization.jl
+++ b/docs/src/literate-gallery/topology_optimization.jl
@@ -110,9 +110,9 @@ function create_dofhandler(grid)
     return dh
 end
 
-function create_bc(dh)
+function create_bc(dh, grid)
     dbc = ConstraintHandler(dh)
-    add!(dbc, Dirichlet(:u, getnodeset(dh.grid, "clamped"), (x, t) -> zero(Vec{2}), [1, 2]))
+    add!(dbc, Dirichlet(:u, getnodeset(grid, "clamped"), (x, t) -> zero(Vec{2}), [1, 2]))
     close!(dbc)
     t = 0.0
     update!(dbc, t)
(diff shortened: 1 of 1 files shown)


KristofferC:
Not sure what is going on with the doc build...


```
Run julia --project=docs -e 'using Pkg; Pkg.instantiate(); Pkg.precompile()'
   Installed ArraysOfArrays ─ v0.6.5
Precompiling project...
  ✓ ArraysOfArrays
  ✓ ArraysOfArrays → ArraysOfArraysStaticArraysCoreExt
  ✗ LazyArrays → LazyArraysBlockArraysExt
  ✓ ArraysOfArrays → ArraysOfArraysAdaptExt
  ✓ ArraysOfArrays → ArraysOfArraysChainRulesCoreExt
  ✓ Ferrite
  ✗ Ferrite → FerriteBlockArrays
  ✓ FerriteMeshParser
  ✓ FerriteGmsh
  7 dependencies successfully precompiled in 8 seconds. 346 already precompiled.

The following 1 direct dependency failed to precompile:

BlockArrays [8e7c35d0-a365-5155-bbbb-fb81a777f[2](https://github.com/Ferrite-FEM/Ferrite.jl/actions/runs/10195746128/job/28205064552?pr=1035#step:5:2)4e]

Error: Missing source file for BlockArrays [8e7c35d0-a365-5155-bbbb-fb81a777f24ePrecompiling project...
```

termi-official:
I had the exact same issue a while ago with the docs with Julia 10.4. It turned out that the dependency (FerriteMeshParser) was only partially downloaded. Found this by manually inspecting the depot folder (.julia/package/FerriteMeshParser). I could come up with a reproducer and eventually forgot about it.

KnutAM:
> Not sure what is going on with the doc build...

Perhaps? 
```diff
- git-tree-sha1 = "5c0ffe1dff8cb7112de075f1b1cb32191675fcba"
+ path = "dev/BlockArrays"
```

https://github.com/Ferrite-FEM/Ferrite.jl/pull/1035/files#diff-54a801f523c8f767061bbcd8c8da625a8d8ce3354a1513a7d674badb77fa14e9R154

KristofferC:
Ah, thanks, that was not intended to commit. Weird that Pkg didn't complain about that indeed.

docs/src/literate-gallery/topology_optimization.jl
```diff
@@ -191,26 +191,25 @@ end
...
+        _nfacets = nfacets(element)
+        resize!(nbg, _nfacets)
         i = cellid(element)
         for j in 1:_nfacets
-            nbg_cellid = getneighborhood(topology, dh.grid, FacetIndex(i,j))
-            if(!isempty(nbg_cellid))
-                nbg[j] = first(nbg_cellid)[1] # assuming only one face neighbor per cell
+            neighbor_facet = neighborhood[i, j]
+            if(!isempty(neighbor_facet))
+                nbg[j] = neighbor_facet[1][1] # assuming only one face neighbor per cell
```
KnutAM, line 206:
```suggestion
            neighbor_facets = neighborhood[i, j]
            if(!isempty(neighbor_facets))
                nbg[j] = neighbor_facets[1][1] # assuming only one facet neighbor per cell
```

docs/src/literate-gallery/topology_optimization.jl
```diff
@@ -191,26 +191,25 @@ end
 # the function call will return an empty object. In that case we use the dictionary to instead find the opposite
 # face, as discussed in the introduction.
```
KnutAM, line 192: The above needs to reflect that a tuple is used I think?

KnutAM (approved):
LGTM, just some minor suggestions. 
 
Are there any measureable improvements when not creating many `Vectors{Vectors}` here as well? In that case, perhaps a small comment would be nice for users to understand why we use `VectorOfArrays`?

Issue #1033, opened 2024-07-31
Having mesh files in the example be downloaded as a part of building the docs is surprising


KristofferC:
I was wondering how the heck 

https://github.com/Ferrite-FEM/Ferrite.jl/blob/f7f2592bd26e8005a739cfe19169097c0146e269/docs/src/literate-tutorials/porous_media.jl#L253

was working when I couldn't find a reference to it in the repo or anywhere.

In my opinion it is better to download it as a part of the example so that you can just include the file. As it is right now, how are you supposed to know where the file is even coming from?

PR #1037, opened 2024-08-02
Download meshes for tutorials

KnutAM:
Solve #1033

### DIFF
--- a/docs/src/literate-tutorials/computational_homogenization.jl
+++ b/docs/src/literate-tutorials/computational_homogenization.jl
@@ -197,17 +197,21 @@ using FerriteGmsh
 #src notebook: use coarse mesh to decrease build time
 #src   script: use the fine mesh
 #src markdown: use the coarse mesh to decrease build time, but make it look like the fine
-#nb ## grid = togrid("periodic-rve.msh")
-#nb grid = togrid("periodic-rve-coarse.msh")
-#jl ## grid = togrid("periodic-rve-coarse.msh")
-#jl grid = togrid("periodic-rve.msh")
-#md grid = togrid("periodic-rve.msh")
 #-
+import Downloads #hide
+meshfile = "periodic-rve.msh"            #src
+#jl meshfile = "periodic-rve.msh"
+#md meshfile = "periodic-rve-coarse.msh" #hide
+#nb meshfile = "periodic-rve-coarse.msh"
+isfile(meshfile) || Downloads.download(  #hide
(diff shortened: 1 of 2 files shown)


docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -197,17 +197,21 @@ using FerriteGmsh
...
+import Downloads #hide
+meshfile = "periodic-rve.msh"            #src
+#jl meshfile = "periodic-rve.msh"
+#md meshfile = "periodic-rve-coarse.msh" #hide
+#nb meshfile = "periodic-rve-coarse.msh"
+isfile(meshfile) || Downloads.download(  #hide
+    string("https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/", meshfile), #hide
+    meshfile) #hide
+grid = togrid(meshfile) #src
+#nb grid = togrid(meshfile)
```
termi-official, line 210: Wouldn't it be easier to add docs for the download file (i.e. docs/download_resources) instead of moving the download logic into the tutorial files? I think having a central point where the external resources are defined for the download is easier to maintain in the long run than splitting it apart like this,.

KnutAM: I was thinking about it, but for a user who copies the tutorial, it is nice to be able to copy the code and run directly I suppose? 

Or did you mean to include via literate, so that the shown case (in cases we actually show the downloads in the docs) will include (part of) the code in `docs/download_resources`?

KristofferC: > I think having a central point where the external resources are defined for the download is easier to maintain in the long run than splitting it apart like this,.

These are tutorials that should be able to be runnable by someone reading them so whatever is done it should be visible from the source code itself.

KnutAM: One solution would be to add a non-exported function
```julia
asset_url(file) = string("https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/", file)
```
to Ferrite itself, which would at least allow using 
```julia
isfile(meshfile) || Downloads.download(Ferrite.asset_url(meshfile), meshfile)
```
above, which looks prettier at least, and would allow changing the location of the assets easily in the future. But feels a bit uggly to add such things to the base package?

KnutAM: Currently, this is only used in two places, so I suggest to do it this "distributed" way for now, to solve #1033. Would be great with a nice solution including a "central point" in the future though.

PR #1038, opened 2024-08-02
Fix notebook links in tutorials

KnutAM:
Fix according to changes introduced in #737


fredrikekre:
Thanks, noticed this the other day too but forgot to open issue/PR about it.

PR #1044, opened 2024-08-07
Fix return type of evaluate_at_grid_nodes

termi-official:


src/Dofs/DofHandler.jl
```diff
@@ -938,10 +938,10 @@ function _evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector{T}, fieldname
         # VTK output of solution field (or L2 projected scalar data)
         n_c = n_components(ip)
         vtk_dim = n_c == 2 ? 3 : n_c # VTK wants vectors padded to 3D
-        data = fill(NaN * zero(T), vtk_dim, getnnodes(get_grid(dh)))
+        data = fill(T(NaN) * zero(T), vtk_dim, getnnodes(get_grid(dh)))
```
KnutAM, line 941:
```suggestion
        data = fill(T(NaN), vtk_dim, getnnodes(get_grid(dh)))
```

KnutAM (approved):
LGTM

test/test_mixeddofhandler.jl
```diff
@@ -649,7 +649,7 @@ function test_vtk_export()
     sdh_tri = SubDofHandler(dh, Set(2))
     add!(sdh_tri, :u, ip_tri)
     close!(dh)
-    u = collect(1:ndofs(dh))
+    u = collect(1.:ndofs(dh))
```
fredrikekre, line 652: So Int types won't work after this? Perhaps you can use `float(T)` when computing the type.

termi-official: I would argue that integer types should not work here, because we assume that the solutions are in some continuous space.

src/Dofs/DofHandler.jl
termi-official: Should I fix this consistently in this PR? Because other places use this.

KnutAM: Could do if you want to :)
Just noticed that this was not required after the changes here. 
But it could depend on the resolution of the `Int` discussion below...

test/test_mixeddofhandler.jl
termi-official: How to proceed?

PR #1053, opened 2024-09-06
Implement hermitian polynomials

lijas:
Me and @fredrikekre had som fun and implemented hermitian polynomials :grimacing: 

Hermitian polynomials cant use identity-mapping for every second shape functions; instead we have that `dNdx = dNd\xi`. This means that it is actually the shape functions N that are modified/mapped.

Not sure how to extend to 2d-element :P 
Nice to see that the interpolation and cellvalues interface is quite generic :) 

Edit: this is only draft PR, will not finish anytime soon since there is currently no need for it.


termi-official:
> Not sure how to extend to 2d-element :P

$H(x,y) = H(x) \otimes H(y)$ 🥴

I would suggest that we also should check analytical convergence for harmonic and biharmonic problems in L2 and H1.

PR #1055, opened 2024-09-18
Fix wrong function to save pvd files in examples and docs

KnutAM:
This was a mistake in #1015, `close` is not defined for `pvd` files by `WriteVTK`, so users have to use `vtk_save` instead for saving `pvd` files.


lijas (approved):
Why did this not error? Is `close(pvd)` defined somewhere?

KnutAM:
After https://github.com/JuliaVTK/WriteVTK.jl/pull/149 this would actually not have been required. 
Should we revert now @fredrikekre? 
Or maybe reasonable to use `vtk_save` for some time before updating to the latest version of WriteVTK (we should also call `close` internally in Ferrite then probably)?

PR #1058, opened 2024-09-19
Extend COOAssembler to also handle vector assembly

fredrikekre:
Follow up to #916.

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -2,58 +2,98 @@ abstract type AbstractAssembler end
 abstract type AbstractCSCAssembler <: AbstractAssembler end
 
 """
-This assembler creates a COO (**coo**rdinate format) representation of a sparse matrix during
-assembly and converts it into a `SparseMatrixCSC{T}` on finalization.
+    struct COOAssembler{Tv, Ti}
+
+This assembler creates a COO (**coo**rdinate format) representation of a sparse matrix
+during assembly and converts it into a `SparseMatrixCSC{Tv, Ti}` on finalization.
 """
-struct COOAssembler{T} # <: AbstractAssembler
-    I::Vector{Int}
-    J::Vector{Int}
-    V::Vector{T}
+struct COOAssembler{Tv, Ti} # <: AbstractAssembler
+    nrows::Int
+    ncols::Int
+    f::Vector{Tv}
+    I::Vector{Ti}
+    J::Vector{Ti}
+    V::Vector{Tv}
 end
 
-COOAssembler(N) = COOAssembler(Float64, N)
-
-function COOAssembler(::Type{T}, N) where T
+function COOAssembler{Tv, Ti}(nrows::Int, ncols::Int; sizehint::Int = 0) where {Tv, Ti}
     I = Int[]
     J = Int[]
-    V = T[]
-    sizehint!(I, N)
-    sizehint!(J, N)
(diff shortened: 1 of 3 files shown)


src/assembler.jl
```diff
@@ -2,58 +2,82 @@ abstract type AbstractAssembler end
...
+and the [manual section on assembly](@ref man-assembly).
 
 !!! note
     When the same matrix pattern is used multiple times (for e.g. multiple time steps or
     Newton iterations) it is more efficient to create the sparse matrix **once** and reuse
     the same pattern. See the [manual section](@ref man-assembly) on assembly.
 """
-function start_assemble(N::Int=0)
-    return start_assemble(Float64, N)
+function start_assemble(nrows::Int, ncols::Int; sizehint::Int = 0)
```
termi-official, line 52: Should we include some dummy data type, e.g. `COOMatrixInfo`, to make the interface more symmetric (cc @KnutAM )? I left the interface intact as it was in the last PR, because the last PR was about refcatoring internals only.

fredrikekre: To encode the size or what?

KnutAM: I don't think that makes it more symmetric, since we do pass the actual matrix [and vector] in the other cases, which you can refer to directly after assembly. That would not be the case here AFAIU. 

So I guess its ok as it is, but I think the best would be to have `start_assemble(data_to_be_filled...)::AbstractAssembler`, and that users who wants the `COOAssembler` should simply call `assembler = COOAssembler(args...; kwargs...)`

src/assembler.jl
```diff
@@ -2,58 +2,82 @@ abstract type AbstractAssembler end
...
+    if fe !== nothing
+        # If the user passes `fe` we can be nice and extend the vector here.
+        m = maximum(dofs; init = 0)
+        lf = length(a.f)
+        if lf < m
+            resize!(a.f, m)
+            for i in (lf + 1):m
+                a.f[i] = 0
+            end
+        end
```
KnutAM, line 77: Avoid frequent resizes? 
```suggestion
        if a.nrows == -1
            m = maximum(dofs; init = 0)
            lf = length(a.f
            if lf < m
                resize!(a.f, m)
                for i in (lf + 1):m
                    a.f[i] = 0
                end
            end
        elseif isempty(a.f)
            resize!(a.f, a.rows)
            fill!(a.f, 0)
        end
```

src/assembler.jl
```diff
@@ -78,13 +102,31 @@ function assemble!(a::COOAssembler{T}, rowdofs::AbstractVector{Int}, coldofs::Ab
 end
 
 """
-    finish_assemble(a::Assembler) -> K
+    finish_assemble(a::COOAssembler) -> K, f
```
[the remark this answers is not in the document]
fredrikekre: So you would store the matrix also in the assembler here?

fredrikekre:
In order to make progress with 1.0 I did this:

> users who wants the COOAssembler should simply call assembler = COOAssembler(args...; kwargs...)

for now and we can decide later if we are introducing a COOMatrix or something to call `start_assemble` on.

src/assembler.jl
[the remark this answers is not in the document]
fredrikekre: Can be introduced in a non-breaking way.

PR #1059, opened 2024-09-19
Deprecate assemble! with local vector before local matrix

fredrikekre:
This patch deprecates the signature
`assemble!(assembler, dofs, local_vector, local_matrix)` in favor of `assemble!(assembler, dofs, local_matrix, local_vector)`. The former signature was not used in many places, not supported by e.g. the block matrix assembler, and it is confusing that there are two options. See also #707 for a similar change to `start_assemble`.

In addition, if the assembler is used just for matrix assembly, this patch removes an unnecessary allocation of an empty vector for every call to `assemble!`.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -438,3 +438,8 @@ export VTKFile
 function VTKFile(args...)
     throw(DeprecationError("VTKFile(args...)" => "VTKGridFile(args...)"))
 end
+
+# assemble with vector first
+function assemble!(::AbstractAssembler, ::AbstractVector{<:Integer}, ::AbstractVector, ::AbstractMatrix)
+    throw(DeprecationError("assemble!(assembler, dofs, fe, Ke)" => "assemble!(assembler, dofs, Ke, fe)"))
+end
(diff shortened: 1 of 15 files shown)


src/deprecations.jl
```diff
@@ -438,3 +438,8 @@ export VTKFile
 function VTKFile(args...)
     throw(DeprecationError("VTKFile(args...)" => "VTKGridFile(args...)"))
 end
+
+# assemble with vector first
+function assemble!(::AbstractAssembler, ::AbstractVector{<:Integer}, ::AbstractVector, ::AbstractMatrix)
```
KnutAM, line 443:
```suggestion
function assemble!(::Union{AbstractAssembler, COOAssembler}, ::AbstractVector{<:Integer}, ::AbstractVector, ::AbstractMatrix)
```
And add test since this is important that users see if they have this error?

KnutAM (commented):
Good change! 

Needs changelog I guess, and I think my suggestion is required?

src/deprecations.jl
fredrikekre: COOAssembler doesn't support vector assembly.

KnutAM: Got confused since this is being added now :)

Issue #1049, opened 2024-08-20
Remove dofhandler from CellCache


lijas:
I find it a bit awkward that the `CellCache` holds the dofhandler and/or grid. Sometimes I have wanted to initialize and store a `CellCache` at the initialization phase of my code, when the dofhandler has not been constructed yet. I think it makes more sense to move the dh/grid to the `CellIterator`, and only store the current dofs,coords and nodes in the `CellCache`:

```
struct CellCache{X}
    nodes::Vector{Int}
    coords::Vector{X}
    dofs::Vector{Int}
end

mutable struct CellIterator{CC<:CellCache, IC<:IntegerCollection}
    set::IC
    cc::CC
    const flags::UpdateFlags
    const grid::Grid
    const dh::DofHandler
    cellid::Int
end
```

termi-official:
I heavily rely on having access to the cellid (and grid) in the cell cache. However, I wanted to investigate the cell cache in more detail with my student soon, because we want to be able to use it on the GPU. My current suggestion is that the cell iterator has some prototype to derive the cache information from. My first sketch was essentially this

```
struct CellCachePrototype{...}
    flags::UpdateFlags
    grid::G
    nodes_buffer::VectorType{Int}
    coords_buffer::VectorType{X}
    dh::DH
    dofs_buffer::VectorType{Int}
end

struct CellIterator{CC<:CellCachePrototype, IC<:IntegerCollection}
    cc::CC
    set::IC
end

function Base.iterate(iterator::GridIterators, state_in...)
    it = iterate(_getset(iterator), state_in...)
    it === nothing && return nothing
    item, state_out = it
    cache = _makecache(iterator, item)
    return (cache, state_out)
end

struct CellCache{...}
    grid::G
    nodes::VectorType{Int}
    coords::VectorType{X}
    dh::DH
    dofs::VectorType{Int}
    cellid::Int
end

function _makecache(iterator, item)
   cache = CellCache(iterator.cc.grid, ..., item)
   reinit!(cache, item)
   return cache
end
```

Abdelrahman912:
> ```
> function Base.iterate(iterator::GridIterators, state_in...)
>     it = iterate(_getset(iterator), state_in...)
>     it === nothing && return nothing
>     item, state_out = it
>     cache = _makecache(iterator, item)
>     return (cache, state_out)
> end
> 
> struct CellCache{...}
>     grid::G
>     nodes::VectorType{Int}
[...]

PR #1060, opened 2024-09-19
WIP: Add cell to CellCache

KnutAM:
Just to ensure that this can be done in a non-breaking way after #798 is merged. 

Did not see any performance degradation on the porous media example where the cell type is `Union{Quadrilateral, Triangle}`, so the approach in this PR does look ok AFAIU.

No need to merge now - just wanted to prevent any issues after 1.0 release..

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -4,10 +4,11 @@ struct UpdateFlags
     nodes::Bool
     coords::Bool
     dofs::Bool
+    cell::Bool
 end
 
-UpdateFlags(; nodes::Bool=true, coords::Bool=true, dofs::Bool=true) =
-    UpdateFlags(nodes, coords, dofs)
+UpdateFlags(; nodes::Bool=true, coords::Bool=true, dofs::Bool=true, cell::Bool=true) =
+    UpdateFlags(nodes, coords, dofs, cell)
 
 
 ###############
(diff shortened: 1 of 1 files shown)


termi-official:
Can you ELI5 why the celltype should be in the CellCache (and why it should be a union)? 

Having a union might be an issue for the GPU stuff. We are currently working towards making CellCache type stable for the GPU (xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/1049#issuecomment-2298441315).

KnutAM:
Needed to support reinit! for vector interpolations using the CellIterator. 

Union: the grid has a Union, the cellCache not, but the constructor should be type unstable. Iteration is not.

termi-official:
Thanks!

> Needed to support reinit! for vector interpolations using the CellIterator.

`reinit!` acts on  some FEValues and you can obtain the information from the geometric interpolation, right?

KnutAM:
> `reinit!` acts on some FEValues and you can obtain the information from the geometric interpolation, right?

The actual cell is required to determine the global direction of faces or edges based on its node ids during reinit of FEValues with non-identity mappings.

fredrikekre:
Couldn't the cell be looked up in `reinit!` instead? We have the cellid

KnutAM:
> Couldn't the cell be looked up in reinit! instead? We have the cellid

Yes, would be possible. But that felt like a split design, to partly reinit the cell cache, and then do more lookup based on the grid in the cache during reinit of the fe values. 

And also not sure if there are performance pitfalls regarding type stability here, especially for GPU if a Union isn't allowed during `reinit!`, but haven't checked/thought carefully about this yet.

fredrikekre:
We should not consider type params as part of the API so this is not breaking.

PR #1063, opened 2024-09-20
Add a doc section comparing different assembly strategies

fredrikekre:


docs/src/topics/assembly.md
```diff
@@ -83,3 +83,310 @@ for t in 1:timesteps
...
+A variant of the first strategy is to explicitly loop over the indices and add add the
+elements individually as scalars:
+
+```@example assembly-perf
+function assemble_v2(_, K, dofs, Ke)
+    for (i, I) in pairs(dofs)
+        for (j, J) in pairs(dofs)
+            K[I, J] += Ke[i, j]
+        end
+    end
```
fredrikekre, line 181: Based on the timings this is a perfectly good way to assemble so it could be used in the very first tutorial if one doesn't want to introduce the assembler already there (as suggested by Chris during FerriteCon).

termi-official: Maybe we should also link to this topic section in the first tutorial?

KnutAM:
Quick comments (didn't read super carefully).

I think this is very nice to put numbers on! Thanks for putting this together!

Wouldn't using broadcasting be the more "julian" way instead of manually looping or directly doing non-dot `+=`? Using a view seems to give same perf as loop - not sure if this is an issue with `SparseArrays.jl` that its not equivalent to broadcasting. 
```julia
function assemble_v1d(_, K, dofs, Ke)
    @. K[dofs, dofs] += Ke # Or K[dofs, dofs] .+= Ke
    return
end;

function assemble_v1v(_, K, dofs, Ke)
    Kv = view(K, dofs, dofs);
    Kv .+= Ke
    return
end;

julia> @btime assemble_v1d($assembler, $K, $dofs, $Ke);
  22.300 μs (11 allocations: 716.23 KiB)

julia> @btime assemble_v1v($assembler, $K, $dofs, $Ke);
  2.511 μs (0 allocations: 0 bytes)
```

Regarding the content, personally I find it a bit too verbose with both the single and full assembly for what its showing. Perhaps a golden midway would be to include the timings for the full assembly, but collapse the actual element routine and assembly routine, and simply report the results?

Finally, why not use the heat equation assembly which is simpler for new users to follow?

fredrikekre:
> Wouldn't using broadcasting be the more "julian" way instead of manually looping or directly doing non-dot +=?

Yea, maybe, I can add a note about this.

> Regarding the content, personally I find it a bit too verbose

I can hide the global assembly routine perhaps? I don't think verbosity is really a problem though. This is in a "topic" section where we can dig a bit deeper.

> Finally, why not use the heat equation assembly which is simpler for new users to follow?

I just wanted a bit of a bigger local matrix but probably the point can be made also with the smaller heat eq one.

docs/src/topics/assembly.md
```diff
@@ -83,3 +84,302 @@ for t in 1:timesteps
...
+scenario we compare the full matrix assembly including the integration of all the elements.
+
+!!! note "Pre-allocated global matrix"
+    All strategies that we compare below uses a pre-allocated global matrix `K` with the
+    correct sparsity pattern. Starting with something like
+    `K = spzeros(ndofs(dh), ndofs(dh))` and then inserting entries is excruciatingly slow
+    due to the sparse data structure so this method is not even considered.
+
+For the comparison we need a representative global matrix to assemble into. In the following
+setup code we create grid with triangles and a DofHandler with a quadratic scalar field.
```
KnutAM, line 104:
```suggestion
setup code we create a grid with triangles and a DofHandler with a quadratic scalar field.
```

docs/src/topics/assembly.md
```diff
@@ -83,3 +84,302 @@ for t in 1:timesteps
...
+
+```@example assembly-perf
+function assemble_v1(_, K, dofs, Ke)
+    K[dofs, dofs] += Ke
+    return
+end
+nothing # hide
+```
+
+This looks very simple but it is very inefficient (as the numbers will show later). To
```
KnutAM, line 138:
```suggestion
This looks very simple, but it is very inefficient (as the numbers will show later). To
```

docs/src/topics/assembly.md
```diff
@@ -83,3 +84,302 @@ for t in 1:timesteps
...
+tmp2 = tmp1 + Ke       # 2
+K[dofs, dofs] = tmp2   # 3
+```
+
+Now the problem with this strategy becomes a bit more obvious:
+ - In line 1 there is first an allocation of a new matrix (`tmp1`) followed by indexing into
+   `K` to copy elements from `K` to `tmp1`. Both of these operations are rather costly:
+   allocations should always be minimized in tight loops, and indexing into a sparse matrix
+   is non-trivial due to the data structure. In addition, since the `dofs` vector contains
+   the global indices (which are not sorted nor consecutive) we have a random access
```
KnutAM, line 162:
```suggestion
   the global indices (which are neither sorted nor consecutive) we have a random access
```

docs/src/topics/assembly.md
```diff
@@ -83,3 +84,302 @@ for t in 1:timesteps
...
+The key difference here is that we index using integers (`I`, `J`, `i`, and `j`) which means
+that `tmp1` and `tmp2` are scalars which don't need to be allocated on the heap. This
+stragety thus eliminates all allocations that were present in the first strategy. However,
+we still lookup the same location in `K` twice, and we still have a random access pattern.
+
+#### Strategy 3: scalar indexing with single lookup
+
+To improve on the second strategy we will get rid of the double lookup into the sparse
+matrix `K`. While Julia doesn't have a "`+=`"-operation there is a `addindex!`-function in
+Ferrite which does exactly what we want: it adds a value to a specific location in a sparse
```
KnutAM, line 211: Just to be clear that this is internal? 
```suggestion
matrix `K`. While Julia doesn't have a "`+=`"-operation, Ferrite has an internal `addindex!`-function which does exactly what we want: it adds a value to a specific location in a sparse
```

docs/src/topics/assembly.md
```diff
@@ -83,3 +84,302 @@ for t in 1:timesteps
...
+assemble_v2(assembler, K, dofs, Ke) # hide
+assemble_v3(assembler, K, dofs, Ke) # hide
+assemble_v4(assembler, K, dofs, Ke) # hide
+nothing                             # hide
+```
+
+```julia
+using BenchmarkTools
+
+@btime assemble_v1(assembler, K, dofs, Ke) evals = 10 setup = Ferrite.fillzero!(K)
```
KnutAM, line 276: Any reason why the regular `fill!` cannot be used? If so, a small remark should probably be made? 
```suggestion
@btime assemble_v1(assembler, K, dofs, Ke) evals = 10 setup = fill!(K, 0)
```

KnutAM (approved):
Nice!
Just a few minor suggestions.
Of course, would be relevant to compare the time with `K\f`, but that also destroys the point a bit since in this case it will probably cleary dominate.

fredrikekre:
> Of course, would be relevant to compare the time with K\f, but that also destroys the point a bit since in this case it will probably cleary dominate.

Yea I don't know, that doesn't scale linearly though so then you can tweak it any way you want. Perhaps there is no need to show the global assembly at all though, but I thought it was fun to relate it to something.

PR #1064, opened 2024-09-21
Replace `sparse` with `spzeros!` in grid coloring

fredrikekre:
This patch uses the more optimal `spzeros!` instead of `sparse` when creating the incidence matrix for the grid coloring. This doesn't improve the performance by much, but no reason not to use it.

For a grid with 1 million elements:
  3.181 s (4537096 allocations: 1.99 GiB) # PR
  3.534 s (4537126 allocations: 2.29 GiB) # master


fredrikekre:
Looks like latest NearestNeighbors broke tests here @KristofferC

PR #1069, opened 2024-09-25
Consistently return the vtk file in export methods

fredrikekre:
This patch make sure that the vtk file is always returned in export related methods. In particular, this make sure that return values of WriteVTK functions don't leak out to Ferrite users. For example, `write_cell_data` returned an `LightXML.XMLElement`.

### DIFF
--- a/src/Export/VTK.jl
+++ b/src/Export/VTK.jl
@@ -44,9 +44,13 @@ function VTKGridFile(f::Function, args...; kwargs...)
     finally
         close(vtk)
     end
+    return vtk
 end
 
-Base.close(vtk::VTKGridFile) = WriteVTK.vtk_save(vtk.vtk)
+function Base.close(vtk::VTKGridFile)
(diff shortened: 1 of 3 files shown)


src/Export/VTK.jl
```diff
@@ -56,9 +60,10 @@ end
 
 function WriteVTK.collection_add_timestep(pvd::WriteVTK.CollectionFile, datfile::VTKGridFile, time::Real)
     WriteVTK.collection_add_timestep(pvd, datfile.vtk, time)
+    return
```
KnutAM, line 63: Since this extends WriteVTK, would it make sense to return the same as WriteVTK does instead of nothing?

KnutAM (commented):
Thanks for cleaning this up!
Would perhaps be nice to add some tests for this?

fredrikekre:
Added tests

KnutAM (approved):

PR #1070, opened 2024-09-26
WIP add task_local for duplication things for multithreading

fredrikekre:


src/FEValues/FunctionValues.jl
```diff
@@ -107,6 +107,14 @@ function Base.copy(v::FunctionValues)
     d2Ndξ2_copy = _copy_or_nothing(v.d2Ndξ2)
     return FunctionValues(copy(v.ip), Nx_copy, Nξ_copy, dNdx_copy, dNdξ_copy, d2Ndx2_copy, d2Ndξ2_copy)
 end
+function task_local(v::FunctionValues)
+    Nξ = task_local(v.Nξ)
+    Nx = v.Nξ === v.Nx ? Nξ : task_local(v.Nx) # Preserve aliasing
```
termi-official, line 112: This should never happen

KnutAM: What shouldn't happen?

KnutAM:
I don't have time to look in detail, but I really like to have this interface!

src/FEValues/FunctionValues.jl
termi-official: Nx should essentially never alias Nξ

KnutAM: Yes, for identity mapping it does to save memory footprint

termi-official: But for identity mapped elements (like for example Lagrange) we still have Nx = J Nξ . For these Nx = Nξ iff J = I .

KnutAM: Not according to https://ferrite-fem.github.io/Ferrite.jl/dev/topics/FEValues/#Identity-mapping.

termi-official: Oh, sorry, I mixed up the derivatives with the function values. Mb.

termi-official (commented):
Is there a sane way to unit test this?

fredrikekre:
Thoughts on the name? `task_local` is OK I guess, but then it sounds like it would return something like https://github.com/vchuravy/TaskLocalValues.jl, which it doesn't (but perhaps it should?). `scratch_copy`?

KnutAM:
I think having `task` in the name makes sense. Some suggestions are
`task_copy`, `task_scratch`, or `task_local_copy`

src/multithreading.jl
```diff
@@ -0,0 +1,40 @@
...
+
+"""
+    task_local(x::T) -> T
+
+Duplicate `x` for a new task such that it can be used concurrently with the original `x`.
+"""
+task_local(::Any)
+
+# Vector/Matrix (e.g. local matrix and vector)
+# task_local(A::Array) = copy(A)
```
KnutAM, line 11:
```suggestion
```
I just looked at this yesterday and wanted to suggest adding a definition like below 👍

test/test_multithreading.jl
```diff
@@ -0,0 +1,75 @@
...
+    end
+    # CellValues
+    for cv in (CellValues(qr, ip), CellValues(qr, ip; update_hessians = true))
+        equivalent_but_distinct(cv, task_local(cv))
+    end
+    # FacetValues
+    fqr = FacetQuadratureRule{RefTriangle}(2)
+    for fv in (FacetValues(fqr, ip), FacetValues(fqr, ip))
+        equivalent_but_distinct(fv, task_local(fv))
+    end
```
KnutAM, line 63: Remove the tests of `copy` for `CellValues`, `FacetValues`.

KnutAM (commented):
Nice! 

Took a bit deeper look, looks good. Docs/changelog etc of course needed as well, but still just WIP AFAIU. 

Should it also be added to `InterfaceValues`?

src/FEValues/CellValues.jl
```diff
@@ -68,6 +68,12 @@ end
 function Base.copy(cv::CellValues)
     return CellValues(copy(cv.fun_values), copy(cv.geo_mapping), copy(cv.qr), _copy_or_nothing(cv.detJdV))
 end
+function task_local(cv::CellValues)
+    return CellValues(
+        task_local(cv.fun_values), task_local(cv.geo_mapping), task_local(cv.qr),
```
termi-official, line 73: It is an invariant that CellValues' qr does not change, so we do not need to task_local qr here right?

src/Quadrature/quadrature.jl
```diff
@@ -319,3 +319,12 @@ getrefshape(::QuadratureRule{RefShape}) where {RefShape} = RefShape
 
 # TODO: This is used in copy(::(Cell|Face)Values), but it it useful to get an actual copy?
 Base.copy(qr::Union{QuadratureRule, FacetQuadratureRule}) = qr
+
+# TODO: For typical use the quadrature rule is read-only, but seems safer to copy anyway?
+#       And might even be beneficial with e.g. NUMA?
```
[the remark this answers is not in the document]
termi-official: I would argue that `task_local` should always be safe and that the caller is responsible to not call `task_local` on immutable stuff, if he does not want the duplication.

src/multithreading.jl
```diff
@@ -0,0 +1,40 @@
+function task_local end
+
+"""
+    task_local(x::T) -> T
+
+Duplicate `x` for a new task such that it can be used concurrently with the original `x`.
```
termi-official, line 6: Should we add to the docs that users with custom types might need to dispatch this function on some of their types for threaded assembly?

src/multithreading.jl
```diff
@@ -0,0 +1,40 @@
...
+"""
+task_local(::Any)
+
+# Vector/Matrix (e.g. local matrix and vector)
+# task_local(A::Array) = copy(A)
+function task_local(x::T) where {S, T <: DenseArray{S}}
+    if isbitstype(T)
```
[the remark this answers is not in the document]
termi-official: + test coverage

test/test_multithreading.jl
termi-official, line 1: Some code paths are not covered yet (see CodeCov result).

src/multithreading.jl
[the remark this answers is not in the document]
fredrikekre: Yea I thought StaticArrays would hit this but I guess not. The reason for limiting at all (and not using AbstractArray) is that I didn't want people to accidentally call this on the global arrays which don't need to be duplicate (unless you are not doing coloring).

src/Quadrature/quadrature.jl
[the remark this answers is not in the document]
fredrikekre: IIRC, at least in the past, it was beneficial that the thread that used the data was the one allocating it so thats why I thought maybe it is better to duplicate this too.

src/FEValues/CellValues.jl
termi-official: > It would allow task_local(pv::PointValues) = PointValues(task_local(pv.cv)) which would be nice, and I think there wouldn't be any significant cost of doing this?

Good point.

src/Quadrature/quadrature.jl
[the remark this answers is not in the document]
termi-official: I am not sure if that is also true for constant shared memory data. I suggest that we keep the rationale in a doc string here tho, so if we revisit this we have the exact arguments at hand (and also for users).

test/test_multithreading.jl
termi-official: Still.

KristofferC:
FWIW, I am not 100% sold on this. What was the issue with the previous way?

Also, calling it `task_local` when it basically seems to create a (somewhat selective) `copy` feels strange to me? What's task local about that?

src/Quadrature/quadrature.jl
```diff
@@ -317,5 +317,11 @@ getpoints(qr::FacetQuadratureRule, face::Int) = getpoints(qr.face_rules[face])
...
 
-# TODO: This is used in copy(::(Cell|Face)Values), but it it useful to get an actual copy?
-Base.copy(qr::Union{QuadratureRule, FacetQuadratureRule}) = qr
+# TODO: For typical use the quadrature rule is read-only, but seems safer to copy anyway?
+#       And might even be beneficial with e.g. NUMA?
+function task_local(qr::QR) where {refshape, QR <: QuadratureRule{refshape}}
+    return QuadratureRule{refshape}(task_local(qr.weights), task_local(qr.points))::QR
+end
+function task_local(qr::QR) where {refshape, QR <: FacetQuadratureRule{refshape}}
+    return FacetQuadratureRule{refshape}(map(task_local, qr.face_rules))::QR
```
fredrikekre, line 326:
```suggestion
    return FacetQuadratureRule{refshape}(task_local(qr.face_rules))::QR
```

src/FEValues/FacetValues.jl
```diff
@@ -68,10 +68,12 @@ function FacetValues(::Type{T}, qr::FacetQuadratureRule, ip::Interpolation, ip_g
     return FacetValues(T, qr, ip, ip_geo, ValuesUpdateFlags(ip; kwargs...))
 end
 
-function Base.copy(fv::FacetValues)
-    fun_values = map(copy, fv.fun_values)
-    geo_mapping = map(copy, fv.geo_mapping)
-    return FacetValues(fun_values, geo_mapping, copy(fv.fqr), copy(fv.detJdV), copy(fv.normals), fv.current_facet)
+function task_local(fv::FacetValues)
+    return FacetValues(
+        map(task_local, fv.fun_values), map(task_local, fv.geo_mapping),
```
fredrikekre, line 73:
```suggestion
        task_local(fv.fun_values), task_local(fv.geo_mapping),
```

fredrikekre:
> What was the issue with the previous way?

The issue is basically that there isn't a previous way:
 - CellValues you can `copy` (so you don't have to thread the quadrature and interpolation all the way into the global assembly in order to call the `CellValue` constructor to obtain a copy)
 -  `CellCache` is not a problem to call the constructor for each task because the input `dh` should be available
 - local matrix/vector is OK to recreate with `zeros` or whatever
 - To get a task local assembler you have to call `start_assemble(K, f; fillzero = false)`... what?

The point of this PR is to get something consistent I guess. I am not really fond of the name either, but since it is typically used when you set up a new task the name "looks good", i.e.
```julia
cv = task_local(cellvalues)
Threads.@spawn assemble!( cv...)
```

Would a name like `scratch_copy` or something be better?

termi-official:
> Would a name like `scratch_copy` or something be better?

To get the ball rolling on this, in Thunderbolt I have implemented this exact concept some time ago decided for the name `duplicate_for_parallel`.

KnutAM:
`copy_for_task`? 

IMO `scratch` isn't so nice since we also pass information (e.g. `DofHandler` in `CellCache`)

fredrikekre:
I still think something like this would be nice to have. Can we come up with a better name? I don't think `copy` is the right verb exactly...

KnutAM:
Would be good to get this settled I think, and also related to the discussion in #1291. 

I suggest we make a vote - add 👍 or 👎 for the suggestions you like or don't like below, and ❤️ for your favourite. If not wanting to have a function like this at all, add a 💀 on this comment :) 

(Hope I didn't miss any suggestions, if so please add)

KnutAM:
`task_local`

KnutAM:
`scratch_copy`

KnutAM:
`task_copy`

KnutAM:
`task_scratch`

KnutAM:
`task_local_copy`

KnutAM:
`duplicate_for_parallel`

KnutAM:
`copy_for_task`

KnutAM:
`task_local_instance`

fredrikekre:
`task_local_copy` perhaps (if not `task_local`). Alternatives could be something with `localize` or `deduplicate`? `localize_for_task` maybe idk.

KnutAM:
> task_local_copy perhaps (if not task_local). Alternatives could be something with localize or deduplicate? localize_for_task maybe idk.

yes - (summarizing so not everyone has to count) `task_local_copy` has (+5, -0) votes and `task_local` (+4, -1). I guess @termi-official don't like `task_local` due to association with `TaskLocalValues.jl`? 

Personally, `localize` doesn't imply no shared memory reference to me, and neither `deduplicate`. 

From my PoV we can go with `task_local_copy` - but that was also my 1st choice :D

termi-official:
> task_local

Sounds like a variable name, not like a function which does some operation.

fredrikekre:
🤖 suggests `duplicate` as the name.

fredrikekre:
For the GPU stuff we now have `distribute_to_workers` if we want to reuse some terminology from that?

termi-official:
Happy about either way, we should just make sure that users understand the difference here (vector of objects vs object that behaves like a vector).

PR #1071, opened 2024-09-26
Add documentation preview cleanup

fredrikekre:
Github refuses to push the gh-pages branch because it has become too big.


fredrikekre:
Cleaned it up manually, should be fine for a while again and it can be occasionaly useful to have the commit history.

PR #1074, opened 2024-09-27
Correcting weak form + spelling mistake

mipals:
* Changing `b` to `f` in the weakform in order to be consistent with the notation in the introduced heat equation.
* base function -> basis function.


fredrikekre:
Thanks!

PR #1081, opened 2024-09-30
WIP HYPRE extension

fredrikekre:


test/hypre_mpi.jl
termi-official, line 1: I think this should be a "how-to" with the hint that one has to be very careful on big meshes, because we generate N (=number of ranks) instances of the full grid and dof handler.

PR #1095, opened 2024-10-23
Update Downstream.yml to include testing FerriteInterfaceElements.jl

DRollin:
I just updated FerriteInterfaceElements.jl to Ferrite.jl 1.0.
This might be a good time to also test FerriteInterfaceElements.jl when testing Ferrite.jl.

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -26,6 +26,7 @@ jobs:
           # - {user: 'Ferrite-FEM', repo: 'FerriteDistributed.jl'} # Requires more efforts to be updated to 1.0
           - {user: 'Ferrite-FEM', repo: 'FerriteGmsh.jl'}
           - {user: 'Ferrite-FEM', repo: 'FerriteMeshParser.jl'}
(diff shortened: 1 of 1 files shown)


KnutAM:
Perhaps also add in the readme of Ferrite to increase visibility? 
Or wait until registered?

DRollin:
I will add a note somewhere :+1: 
First, I want to have a first release of FerriteInterfaces.jl and then update to Ferrite#master.
(Currently the test still use `Lagrange` with three arguments)
Once that is working, I will add a note :smile:

fredrikekre:
> update to Ferrite#master.

Looks like you need to do this first to make tests pass here.

DRollin:
Yes, but first I wanted to fix version 1.0.0 of FerriteInterfaceElements.jl, so this version remains compatible wit Ferrite.jl 1.0.0.
I already started a PR, but I am not sure what needs to be done to release a version.

DRollin:
I fixed the issue in FerriteInterfaceElements.jl

PR #1097, opened 2024-10-29
Simplify evaluate_at_grid_nodes

lijas:
There were some hacks in `evaluate_at_grid_nodes` for embedded element which I don't think is needed anymore.
Also added `shape_value_type(::Interpolation, T)` which I think will be required for #958  also.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -930,7 +930,7 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
 end
 
 # Internal method that have the vtk option to allocate the output differently
-function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk}=Val(false)) where {T, vtk, sdim}
+function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk} = Val(false)) where {T, vtk, sdim}
     # Make sure the field exists
     fieldname ∈ getfieldnames(dh) || error("Field $fieldname not found.")
     # Figure out the return type (scalar or vector)
@@ -966,8 +966,10 @@ function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fie
 end
 
 # Loop over the cells and use shape functions to compute the value
-function _evaluate_at_grid_nodes!(data::Union{Vector,Matrix}, sdh::SubDofHandler,
-        u::AbstractVector{T}, cv::CellValues, drange::UnitRange) where {T}
+function _evaluate_at_grid_nodes!(
+        data::Union{Vector, Matrix}, sdh::SubDofHandler,
+        u::AbstractVector{T}, cv::CellValues, drange::UnitRange
+    ) where {T}
     ue = zeros(T, length(drange))
     for cell in CellIterator(sdh)
         # Note: We are only using the shape functions: no reinit!(cv, cell) necessary
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -930,13 +930,13 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
 end
 
 # Internal method that have the vtk option to allocate the output differently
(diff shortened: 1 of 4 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -954,37 +954,26 @@ function _evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector{T}, fieldname
         ip_geo = geometric_interpolation(CT)
         local_node_coords = reference_coordinates(ip_geo)
         qr = QuadratureRule{getrefshape(ip)}(zeros(length(local_node_coords)), local_node_coords)
-        if ip isa VectorizedInterpolation
-            # TODO: Remove this hack when embedding works...
-            cv = CellValues(qr, ip.ip, ip_geo)
-        else
-            cv = CellValues(qr, ip, ip_geo)
-        end
+        cv = CellValues(qr, ip, ip_geo^sdim; update_gradients = Val(false), update_hessians = Val(false), update_detJdV = Val(false))
```
KnutAM, line 957: Since it is type unstable anyways, I think we could avoid the undocumented option of sending a `Val` here? 
```suggestion
        cv = CellValues(qr, ip, ip_geo^sdim; update_gradients = false, update_hessians = false, update_detJdV = false)
```

src/interpolations.jl
```diff
@@ -57,6 +57,10 @@ n_components(::VectorInterpolation{vdim}) where {vdim} = vdim
 # Number of components that are allowed to prescribe in e.g. Dirichlet BC
 n_dbc_components(ip::Interpolation) = n_components(ip)
 
+shape_value_type(::ScalarInterpolation, T::Type) = T
+shape_value_type(::VectorInterpolation{vdim}, T::Type) where {vdim} = Vec{vdim,T}
```
KnutAM, line 61: Ensure specialization, and require numeric data type? 
```suggestion
shape_value_type(::ScalarInterpolation, ::Type{T}) where {T <: Number} = T
shape_value_type(::VectorInterpolation{vdim}, ::Type{T}) where {vdim, T <: Number} = Vec{vdim, T}
```

KnutAM (commented):
Looks like good changes to me, but GitHub/CodeCov complains that it isn't covered by tests, that looks odd to me?

src/interpolations.jl
```diff
@@ -57,6 +57,10 @@ n_components(::VectorInterpolation{vdim}) where {vdim} = vdim
 # Number of components that are allowed to prescribe in e.g. Dirichlet BC
 n_dbc_components(ip::Interpolation) = n_components(ip)
 
+shape_value_type(::ScalarInterpolation, T::Type) = T
```
KnutAM, line 60: Add a docstring to this one in devdocs for interpolations, and filter the one in [docs/src/devdocs/FEValues.md](https://github.com/Ferrite-FEM/Ferrite.jl/blob/212115a3e041ac19be7c78fe50b2e8a251aa03d3/docs/src/devdocs/FEValues.md?plain=1#L25) to show for `AbstractValues` and `FunctionValues` inputs?

src/Dofs/DofHandler.jl
```diff
@@ -930,13 +930,13 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
 end
 
 # Internal method that have the vtk option to allocate the output differently
-function _evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk} = Val(false)) where {T, vtk}
+function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk}=Val(false)) where {T, vtk, sdim}
```
KnutAM, line 933:
```suggestion
function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk} = Val(false)) where {T, vtk, sdim}
```

KnutAM (approved):
LGTM if formatting is fixed such that runic / pre-commit workflows pass. 

I guess tests should test the vtk path, but did you also check manually that export of the solution gives correct vtk output?

lijas:
Yes, the vtk output looks correct (checked manually). And we have tests that check if the hash of vtk-files changes, right?

KnutAM:
> And we have tests that check if the hash of vtk-files changes, right?

True, I wasn't sure if the writing of solution was covered by those, but seems so!

Issue #1022, opened 2024-07-12
Default quadrature rule for Tets contains negative weight


termi-official:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/98e2b674a8c6d740a779c022de31f73051cb353c/src/Quadrature/gaussquad_tet_table.jl#L29

Negative quadrature weights can cause all sorts of issues when integrating functions, which are outside the basis space they are designed for. While such rules can be more efficient on many problems, I do not think that we should provide a problematic rule as the default rule to users, such that users must explicitly make the decision to use a rule with negative weights if they want the additional performance. After a bit googling I think https://portwooddigital.com/2020/07/03/integration-points-with-negative-weight/ explains the issue, which can be really hard to track down.

Hence I would propose that we should provide, as the default, only rules with positive weights to not surprise new users. Furthermore I think it makes sense to spend a few lines on this in the docs.

KnutAM:
> Negative quadrature weights can cause all sorts of issues when integrating functions, which are outside the basis space they are designed for.

Do you here mean the basis space that the quadrature rule is designed for? 

While I see that the why is hard to explain. Do you know of any references that report that his is an issue for the tet cases where we have negative weights currently? 

The blogpost seems to be more focused on structural elements, and in particular says 

*These approaches lead to varying degrees of strange results, but are generally harmless because the negative integration weights are not at the ends of the element. However, there could be unexpected results with member loads.*

I interpret this as saying that negative weights are only an issue at the end points. 

The example deal with plastic hinges and shows what happens when such a hinge is outside the element, leading to a negative integration weight. But I don't see how the fact that an input error in defining the plastic hinge which gives a negative weight relates to the properties of having negative weights in general.

termi-official:
> > Negative quadrature weights can cause all sorts of issues when integrating functions, which are outside the basis space they are designed for.
> 
> Do you here mean the basis space that the quadrature rule is designed for?
> 
Yes. E.g. for some of the rules the space is polynomials of order p on the reference geometry.

> While I see that the why is hard to explain. Do you know of any references that report that his is an issue for the tet cases where we have negative weights currently?
> 

No reference here at hand, but you can just take any problem with localized nonlinearity and positivity. For example you can center the nonlinearity (e.g. something like exp(-100(x-c)^2)) on the point with the negative weight and obtain entries with the wrong sign in your stiffness matrix.

termi-official:
[...]

PR #1098, opened 2024-10-30
Making the meaning of order more clear in QuadratureRule

kreutzheidephil:
- docstring for evaluate_at_grid_nodes() and 
- suggestion for more detailed explanation on quadrature order


src/Quadrature/quadrature.jl
```diff
@@ -15,8 +15,8 @@ using Base.Cartesian: @nloops, @ntuple, @nexprs
     QuadratureRule{shape}(weights::AbstractVector{T}, points::AbstractVector{Vec{rdim, T}})
 
 Create a `QuadratureRule` used for integration on the refshape `shape` (of type [`AbstractRefShape`](@ref)).
-`order` is the order of the quadrature rule.
-`quad_rule_type` is an optional argument determining the type of quadrature rule,
+`order` is related to the number of quadrature points per direction (i.e. a `QuadratureRule{RefQuadrilateral}(2)` 
+uses 2 quadrature points per direction). `quad_rule_type` is an optional argument determining the type of quadrature rule,
```
KnutAM, line 19: This is only true for hypercubes. For simplexes the concept of "each direction" is a bit harder to describe. I don't think we are fully consistent with what is meant with order, see also #1022 for some discussions about the options. For now, would something like this explain good enough? 
```suggestion
`order` is related to the number of quadrature points, but is dependent on the reference shape. 
`quad_rule_type` is an optional argument determining the type of quadrature rule,
```

src/L2_projection.jl
```diff
@@ -326,6 +326,13 @@ function assemble_proj_rhs!(f::Matrix, cellvalues::CellValues, sdh::SubDofHandle
     end
 end
 
+"""
+    evaluate_at_grid_nodes(proj::L2Projector, vals::AbstractVector)
+Return a vector of length `getnnodes(grid)` where the order corresponds with the node order 
+in the `grid` used to create `proj`.
+
+`vals` must have length `getnnodes(grid)`.
```
KnutAM, line 334: The length can be different from `getnnodes`. 

```suggestion
`vals` should be the output from `project` using `proj`.
```

KnutAM (commented):
Welcome @kreutzheidephil 🚀 

Some comments below. It will also be required to merge with the master branch. 
After that, need to fix formatting to remove errors reported by the "pre-commit" workflow (and later "runic" after merge with master)

src/Quadrature/quadrature.jl
DRollin: I discussed that with Phil: We thought the general "is related to" would be enough to make clear that this is not always exactly true
Now, I see that this is still not perfectly clear.
What you suggest could be misread as "The order depends on the reference shape." But this is not what you mean, right?

termi-official: Ohh, good catch here! Not sure how we missed that inconsistency earlier. We should definitely clarify this.

When writing the docs I found it difficult to make a balance between 1) being precise and 2) being still useful for users which are not fully aware (and interested) of all details. So for everything but hypercubes we refer with order to the actual integration order of the rule w.r.t. to the function system it integrates. Note that the function system varies between rules, see for example the [Gauss-Laguerre rule](https://en.wikipedia.org/wiki/Gauss%E2%80%93Laguerre_quadrature) for an example of a fancy system. So, if you have any idea how to communicate what "order" means to users, I am open for any suggestion.

KnutAM: Good point! 
Something like this, and if added in topics or somewhere else, we should document the rules somewhere and point to that location. Probably nice not having it in the docstring directly, as this would get too verbose. 

```suggestion
`order` affects the number of quadrature points, and thus the precision of the integration rule. The exact relationship between `order` and the number of quadrature points depends on the `quad_rule_type` and `shape`.
`quad_rule_type` is an optional argument determining the type of quadrature rule,
```

DRollin: I think this is nice for the docstring :+1: 
Maybe we could add a more detailed description (hypercubes vs. other shapes) somewhere in the documentation?
Suggestions where to add it:

1) [at the beginning of "Topic guides/Reference shapes"](https://github.com/Ferrite-FEM/Ferrite.jl/blob/3b081bea81cb718cf33b004ea16f90fdbb8cc5f9/docs/src/topics/reference_shapes.md?plain=1#L14)

2) [at the beginning of "Topic guides/Reference shapes"](https://github.com/Ferrite-FEM/Ferrite.jl/blob/3b081bea81cb718cf33b004ea16f90fdbb8cc5f9/docs/src/topics/reference_shapes.md?plain=1#L84)

3) [in the tutorial "Heat equation"](https://github.com/Ferrite-FEM/Ferrite.jl/blob/3b081bea81cb718cf33b004ea16f90fdbb8cc5f9/docs/src/literate-tutorials/heat_equation.jl#L56)

src/Quadrature/quadrature.jl
```diff
@@ -15,7 +15,7 @@ using Base.Cartesian: @nloops, @ntuple, @nexprs
     QuadratureRule{shape}(weights::AbstractVector{T}, points::AbstractVector{Vec{rdim, T}})
 
 Create a `QuadratureRule` used for integration on the refshape `shape` (of type [`AbstractRefShape`](@ref)).
-`order` is the order of the quadrature rule.
+`order` affects the number of quadrature points, and thus the precision of the integration rule. The exact relationship between `order` and the number of quadrature points depends on the `quad_rule_type` and `shape`.
```
KnutAM, line 18: As discussed, this can be further refined with #1220, but now I know more than what I commented last time :)

```suggestion
`order` affects the number of quadrature points, and thus the precision of the integration rule. For `RefHyperCube`s, it denotes the number of quadrature points along each axis, and for `RefSimplex`, the polynomial order which is integrated exactly.
```

src/L2_projection.jl
```diff
@@ -332,6 +332,13 @@ function assemble_proj_rhs!(f::Matrix, cellvalues::CellValues, sdh::SubDofHandle
     return
 end
 
+"""
+    evaluate_at_grid_nodes(proj::L2Projector, vals::AbstractVector)
+Return a vector of length `getnnodes(grid)` where the order corresponds with the node order
+in the `grid` used to create `proj`.
+
+`vals` should be the output from `project` using `proj`.
```
KnutAM, line 340: To align more with the docstring of `evaluate_at_grid_nodes(::DofHandler, ...)`

```suggestion
Return a vector of length `getnnodes(grid)` where the `i`th entry is the projected value evaluated at node `i`. `vals` should be the output from `project` using `proj`.
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Next] - xxxx-xx-xx
  - Make `shape_symmetric_gradient` work for `PointValues` ([#1325])
+ - Update docs for `QuadratureRule` to better reflect the meaning of order. Add doc string for `evaluate_at_grid_nodes` ([#1098])
```
KnutAM, line 10:
```suggestion
 - Update docs for `QuadratureRule` to better reflect the meaning of order. ([#1098])
 - Add docstring for `evaluate_at_grid_nodes` from `L2Projector`. ([#1098])
```

kreutzheidephil:
Closing this as I split this into two PRs.

- `evaluate_at_grid_nodes` docstring #1513

- updating docs for `order` in `QuadatureRule`  #1514

PR #1100, opened 2024-11-05
Use `Tensor` for interface transformation matrices

AbdAlazezAhmed:
The return type kept changing to `Matrix` while testing my amr stuff. Turns out mixing `SMatrix` and `Tensor{2,3}` was not the wisest thing to do :D.

Also `InterfaceIterator` needed these changes so it can be used with other grid/topology types.

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -341,10 +341,10 @@ end
     `InterfaceIterator` is stateful and should not be used for things other than `for`-looping
     (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
 """
-struct InterfaceIterator{IC <: InterfaceCache, G <: AbstractGrid, Topology <: AbstractTopology}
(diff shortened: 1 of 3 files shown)


termi-official:
Regression tests pls. :smile:

src/iterators.jl
```diff
@@ -341,10 +341,10 @@ end
     `InterfaceIterator` is stateful and should not be used for things other than `for`-looping
     (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
 """
-struct InterfaceIterator{IC <: InterfaceCache, G <: Grid}
+struct InterfaceIterator{IC <: InterfaceCache, G <: AbstractGrid, Topology <: AbstractTopology}
     cache::IC
     grid::G
-    topology::ExclusiveTopology
+    topology::Topology
```
termi-official, line 347:
```suggestion
    topology::TopologyType
```

Or smth in this direction. I think using a name which could be some actual type name can bite us in the future.

termi-official (approved):
LGTM

PR #1101, opened 2024-11-06
Graphical package overview

### LINKED ISSUE #696: Package flowchart
image

I think something like the above would be pretty useful to have in the docs to see how things are connected. This is done using https://mermaid.js.org/ which can now integrate with Documenter using https://github.com/JuliaDocs/DocumenterMermaid.jl

termi-official:
Includes two proposals to resolve #696 .

## A (f32cd96beaa64e88aad36238506fd4d9f5d1c3df)

image

## B (a926a137584d5a0bbe38ca6262150b161a815733)
image

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -78,7 +78,6 @@ flowchart TD
     sptype(SparsityPattern)
     mattype[[System matrix]]
     vectype[[System vector]]
-    assemblertype(Assembler)
     assemblycode{{User assembly code}}
     gridtype-->dhtype
     iptype-->dhtype
@@ -94,11 +93,10 @@ flowchart TD
     sptype-->mattype
     chtype-->vectype
     dhtype-->vectype
-    vectype-->assemblertype
-    mattype-->assemblertype
     cvtype-->assemblycode
     fvtype-->assemblycode
-    assemblertype-->assemblycode
+    assemblycode-->vectype
+    assemblycode-->mattype
 ```
 
 ### Getting help
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -61,6 +61,46 @@ understand how Ferrite works. Already this rather simple program discusses many 
 important concepts. See the [tutorials overview](@ref Tutorials) for suggestion on how to
 progress to more advanced usage.
 
+A high-level graphical overview of Ferrite is given in the following diagram
+
+```mermaid
+flowchart TD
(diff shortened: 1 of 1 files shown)


termi-official:
Thanks for taking a look Knut!

> 
>     * ConstraintHandler -> SparsityPattern missing?
> 
I actually removed this one again, because it really messed around with the layouting again and most users do not need this anyway. I can add it back if you think it is important to show in the overview.

>     * Can Ferrite types have a different color coding or similar to other things (user code / system matrices / vectors)
> 
Probably, but I gave up with the styling at some point, as I could not get meaningful error message out of mermaid about what I am doing wrong. So I decided to give the different parts different node geometries instead, as I could get this one working.

> 
> An overall comment, What is the main purpose? Showing what components Ferrite has and the connections between, or a typical workflow? To me, it looks a bit mixed at the moment.

The main purpose is the former, to show users on a high level how the commonly used main components in Ferrite are connected to each other. I hope that new users can grasp the overall architecture with more ease using this diagram. If there is anything else which can be imporved or if anything is missing just dump in in this PR and let us try to figure out how to add it then.

KnutAM:
> I hope that new users can grasp the overall architecture with more ease using this diagram. 

TBH for me it's hard to follow both diagrams (no criticism, because it's hard to make them easy to follow, and I wouldn't know how to do it better). I would think that a diagram showing the typical workflow could be easier (e.g. for the heat eq. tutorial), but its hard to judge when already being so used to the way Ferrite works... 

And I'm more a fan of structured/aligned flowcharts, and would probably have made an svg instead of mermaid, but that is also just a matter of taste. And is again because I tend to think in workflow.

fredrikekre:
Perhaps there should be two diagrams, one which is "devdocs"-like describing dependencies between the types and another one which is "workflow"-like?

termi-official:
> Perhaps there should be two diagrams, one which is "devdocs"-like describing dependencies between the types and another one which is "workflow"-like?

If we want to add more diagrams, I think that for the devdocs it is beneficial to add (or generate) one for each component with more details, such that we see more how the internal types relate to each other. See e.g. CellValues internals.

Regarding the workflow visualization, I am not very sure what exactly should be shown. Sequence diagrams? User journeys?

termi-official:
Let's then opt for Knuts variant from the paper draft.

PR #1106, opened 2024-11-07
Update degrees_of_freedom.md -> make title consistent with other titles

DRollin:
I just changed "Freedom" to "freedom" to be consistent with the other titles.
(Was there a reason for the capital F?)


DRollin:
Good point.
Should it be "dofs" or "DOFs" since it is an abbreviation?
I have teaching soon, but after that, I could go through all titles (and occurrances of "dof") and make everything consistent.

PR #1113, opened 2024-11-29
Recover export for non-Grid grid types

termi-official:


KnutAM (commented):
Nice one!

I assume you already checked that it works, but if you could add a test using `SmallGrid` in [`test_abstractgrid.jl`](https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_abstractgrid.jl) to export some cell data and compare hash with `Grid`, then that would be great IMO.

KnutAM (approved):
The changes look good to me! 

The only possible reason I can see for the codecov fail would be the updated CI on master? Not sure exactly why though, but maybe try merging the master changes into the PR branch before merging the PR? 

Edit: I guess this is fixed by #1115, merging that reported getting full coverage again.

PR #1119, opened 2024-12-08
Rigid body connector

lijas:
Rigid body connectors

It would be nice to support [Rigid body connectors (RBE2 and RBE3)](https://www.fidelisfea.com/post/what-is-the-difference-between-a-kinematic-rbe2-and-a-distributing-rbe3-coupling-in-fea) in Ferrite. They are quite common to use in industry, and there has been questions about them in the slack channel before. They are usually modeled with rigid bodies which are connected to some part of the mesh (for example the nodes/facets around a hole).

This PR adds:

1. `Point<:AbstractCell` and an interpolation which can be used to add rigid bodies dofs to the system.
2. A `RigidBodyConnector<:Constraint` (RBE2) which sets up the affine constraints  between the rigid body and connecting mesh (This constraint is pretty specific for solid mechanics problems, so it should probably not be included in core Ferrite).
3. A simple example illustrating the constraint (a plate with a hole, and a rigid body at the center that is connected to the nodes around the hole. The rigid body is rotated pi/10).

Screenshot 2024-12-08 at 10 28 47

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -193,6 +193,11 @@ function faces(c::AbstractCell{RefShape}) where {RefShape}
     end
 end
 
+# RefPoint (refdim = 0)
+reference_vertices(::Type{RefPoint}) = (1,)
+reference_edges(::Type{RefPoint}) = () # -
+reference_faces(::Type{RefPoint}) = () # -
+
 # RefLine (refdim = 1)
 reference_vertices(::Type{RefLine}) = (1, 2)
 reference_edges(::Type{RefLine}) = ((1, 2),) # e1
@@ -263,6 +268,9 @@ end
 ######################################################
 
 # Lagrange interpolation based cells
+struct Point <: AbstractCell{RefPoint}
+    nodes::NTuple{1, Int}
+end
 struct Line <: AbstractCell{RefLine}
     nodes::NTuple{2, Int}
 end
@@ -300,6 +308,7 @@ struct Pyramid <: AbstractCell{RefPyramid}
     nodes::NTuple{5, Int}
 end
 
+geometric_interpolation(::Type{Point}) = Lagrange{RefPoint, 1}()
 geometric_interpolation(::Type{Line}) = Lagrange{RefLine, 1}()
 geometric_interpolation(::Type{QuadraticLine}) = Lagrange{RefLine, 2}()
 geometric_interpolation(::Type{Triangle}) = Lagrange{RefTriangle, 1}()
--- a/src/exports.jl
+++ b/src/exports.jl
@@ -4,6 +4,7 @@ export
     VectorInterpolation,
     ScalarInterpolation,
     VectorizedInterpolation,
+    #RefPoint,
     RefLine,
     RefQuadrilateral,
     RefHexahedron,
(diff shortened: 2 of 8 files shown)


KristofferC:
> This constraint is pretty specific for solid mechanics problems, so it should probably not be included in core Ferrite).

Is there a way to formulate this constraint in a somewhat "general" way? I think it is fine to have in Ferrite even if the most common use case is for a certain class of problems.

lijas:
I tried to think of way to generalize it, but i dont think there is an analog constraint for e.g. heat problems.

termi-official:
I think rather than having this directly in Ferrite we should think about providing the building blocks (and howtos) to build such special constraints in an efficient way. In the end this just another affine constraint, since I think building most affine constraints is right now difficult for most users.

lijas:
Yes, it could work as a how-to. It would also be fun to replicate [this phase field](https://www.comsol.com/model/brittle-fracture-of-a-holed-plate-89321) problem, where they use these rigid body connectors to apply the boundary conditions.

src/Grid/grid.jl
```diff
@@ -300,6 +308,7 @@ struct Pyramid <: AbstractCell{RefPyramid}
     nodes::NTuple{5, Int}
 end
 
+geometric_interpolation(::Type{Point}) = Lagrange{RefPoint, 1}()
```
KnutAM, line 311: Since it is a constant value, shouldn't this be 0th order?
```suggestion
geometric_interpolation(::Type{Point}) = Lagrange{RefPoint, 0}()
```

src/exports.jl
```diff
@@ -4,6 +4,7 @@ export
     VectorInterpolation,
     ScalarInterpolation,
     VectorizedInterpolation,
+    #RefPoint,
```
KnutAM, line 7:
```suggestion
    RefPoint,
```

src/exports.jl
```diff
@@ -58,6 +59,7 @@ export
     # Grid
     Grid,
     Node,
+    #Point,
```
KnutAM, line 62:
```suggestion
    Point,
```

src/interpolations.jl
```diff
@@ -1308,6 +1308,19 @@ function reference_shape_value(ip::Lagrange{RefPyramid, 2}, ξ::Vec{3, T}, i::In
...
+#####################################
+# Lagrange dim 0 RefPoint order 0   #
+#####################################
+getnbasefunctions(::Lagrange{RefPoint}) = 1
+vertexdof_indices(::Lagrange{RefPoint}) = ((1,),)
+function reference_coordinates(::Lagrange{RefPoint})
+    return [Tensor{1, 0, Float64, 0}(())] # zero dim Vec{0}
+end
+function reference_shape_value(ip::Lagrange{RefPoint}, ξ::Vec{0}, i::Int)
+    return i == 1 && 1.0
```
KnutAM, line 1320:
```suggestion
    i == 1 && return 1.0
```

KnutAM (commented):
I think the `RefPoint` and `Point` are really cool features to get in 🚀 

I think that `RigidBodyConnector` potentially could be generalized by accepting some function as argument, I think it could be nice to think more about how we could have an interface and separate out that part in a standalone PR?

KnutAM:
>  tried to think of way to generalize it, but i dont think there is an analog constraint for e.g. heat problems.

The equivalent there would be to prescribe a constant, but unknown, temperature on a boundary, AFAIU.

PR #1125, opened 2025-01-06
Update value and derivative deprecation message to make users aware of non-allocating batch evaluation

termi-official:


src/deprecations.jl
```diff
@@ -341,10 +341,10 @@ function FacetValues(
 end
 
 function value(ip::Interpolation, ξ::Vec)
-    throw(DeprecationError("value(ip::Interpolation, ξ::Vec)" => "[reference_shape_value(ip, ξ, i) for i in 1:getnbasefunctions(ip)]"))
+    throw(DeprecationError("value(ip::Interpolation, ξ::Vec)" => "[reference_shape_value(ip, ξ, i) for i in 1:getnbasefunctions(ip)] or reference_shape_values!(N, ip, ξ) with preallocated N"))
```
KnutAM, line 344: (Same for derivative below)

The custom highlighting doesn't really match here, so the simplest option would be
```suggestion
    throw(DeprecationError("value(ip::Interpolation, ξ::Vec) is deprecated, use `[reference_shape_value(ip, ξ, i) for i in 1:getnbasefunctions(ip)]` or `reference_shape_values!(N, ip, ξ)` (with preallocated N) instead"))
```

termi-official: Okay was a bit more tricky to recover styling.

KnutAM: Yes, I didn't suggest having the styling (thinking no styling is better than wrong styling) since that seemed to be too much work... So sorry for creating the extra work!

src/deprecations.jl
```diff
@@ -341,10 +341,30 @@ function FacetValues(
...
-    throw(DeprecationError("derivative(ip::Interpolation, ξ::Vec)" => "[reference_shape_gradient(ip, ξ, i) for i in 1:getnbasefunctions(ip)]"))
+    io = _iobuffer()
+    printstyled(io, "`derivative(ip::Interpolation, ξ::Vec)`", color = :red)
+    print(io, " is deprecated, use ")
+    printstyled(io, "`[reference_shape_gradient(ip, ξ, i) for i in 1:getnbasefunctions(ip)]`", color = :green)
+    print(io, " or ")
+    printstyled(io, "`reference_shape_gradients!(dNdξ, ip, ξ)`", color = :green)
+    print(io, " (with preallocated ")
+    printstyled(io, "`dNdξ`", color = :green)
+    print(io, ") instead.")
```
KnutAM, line 365:
```suggestion
    print(io, " (with preallocated `dNdξ`) instead.")
```

src/deprecations.jl
```diff
@@ -341,10 +341,30 @@ function FacetValues(
...
-    throw(DeprecationError("value(ip::Interpolation, ξ::Vec)" => "[reference_shape_value(ip, ξ, i) for i in 1:getnbasefunctions(ip)]"))
+    io = _iobuffer()
+    printstyled(io, "`value(ip::Interpolation, ξ::Vec)`", color = :red)
+    print(io, " is deprecated, use ")
+    printstyled(io, "`[reference_shape_value(ip, ξ, i) for i in 1:getnbasefunctions(ip)]`", color = :green)
+    print(io, " or ")
+    printstyled(io, "`reference_shape_values!(N, ip, ξ)`", color = :green)
+    print(io, " (with preallocated ")
+    printstyled(io, "`N`", color = :green)
+    print(io, ") instead.")
```
KnutAM, line 352:
```suggestion
    print(io, " (with preallocated `N`) instead.")
```

src/deprecations.jl
termi-official: No worries! :)

KnutAM (approved):
LGTM
I don't think it is necessary to have the N and dNd\xi in green which simplifies a bit, but this will be removed after some time anyway (after a few releases I guess), so doesn't really matter much.

PR #1134, opened 2025-01-23
Add example for plotting with plots.jl

TheFibonacciEffect:
Show how the `PointEvalHandler` can be used to evaluate the solution for plotting using Plots.jl in the case of the heat uquation.

### DIFF
--- a/docs/src/topics/export.md
+++ b/docs/src/topics/export.md
@@ -61,7 +61,7 @@ vtk_save(pvd);
 See [Transient heat equation](@ref tutorial-transient-heat-equation) for an example
 
 # Evaluating the Solution at a list of Points for Plotting using Plots.jl
-The solution can also be evaluated at a list of points and plotted using `Plots.jl`. For the heat equation example it can be done like this:
+The solution can also be evaluated at a list of points and plotted using `Plots.jl`. For the [heat equation](@ref tutorial-heat-equation) example it can be done like this:
 ```
 using Plots
 # The domain extends from -1 to 1 by default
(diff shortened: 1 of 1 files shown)


KnutAM:
Thank you for the suggestion. However, in my opinion, this is not a recommended way to plot an FE solution in Ferrite, so we shouldn't recommend that in the docs. 

FerriteViz.jl aims to make such plots easy (but currently unfortunately requires using a [development](https://github.com/Ferrite-FEM/FerriteViz.jl/pull/103) branch).

termi-official:
I think maybe we can still add this to the how-to section and elaborate in detail why this should not be used in general.

I will try to update FerriteViz in the meantime.

TheFibonacciEffect:
I think it shouldnt be used generally, because you would have to find a way to sample from the more complex geometries and display it with plots.jl, which is what FerriteViz should be used for instead.

However I think it would be nice to add it, because it shows how to _sample from a solution_, which is important for solving optimization problems involving FEM. I couldnt find an example for that in the docs yet.

termi-official:
Thanks for elaborating. It was not ment to offload writing up this section to you, no worries.

> However I think it would be nice to add it, because it shows how to sample from a solution, which is important for solving optimization problems involving FEM. I couldnt find an example for that in the docs yet.

Sorry, I think I do not understand. Is there something missing in https://ferrite-fem.github.io/Ferrite.jl/stable/howto/postprocessing/ (which should be visible from https://ferrite-fem.github.io/Ferrite.jl/stable/howto/) or do you have something different in mind? If you want to solve optimization stuff with FEM (or PDEs in general), then I would highly recommend you to take a look into adjoint methods, as this might, without further knowledge about your specific form, make you optimization problem easier to handle.

TheFibonacciEffect:
Yes, true, youre right its definetly there.

Issue #1135, opened 2025-01-27
Undocumented and unused: getlowerorder


KnutAM:
We have the internal function `getlowerorder` which seems to be legacy (at least I don't see that we use it internally, and it is not documented in devdocs). 

Only a few interpolations are supported. Would be good to either remove (since we don't use it), or (dev)document it at least. Keeping it leads to extra code for implementing new interpolations...

termi-official:
I think `getorder` is also unsued and the order parameter in the interpolation is effectively deprecated.

KnutAM:
`getorder` is documented though - 
https://ferrite-fem.github.io/Ferrite.jl/stable/reference/interpolations/#Ferrite.getorder

PR #1136, opened 2025-01-27
Modularize interpolation testing

KnutAM:
This PR includes some modularization of the testing of interpolations started in #798, and is the first breakout from that PR. 

Specifically all tests that every base interpolations should pass should be put into `test_interpolation_properties(ip::Interpolation)`. Having such a function was very valuable when developing the interpolations in #798, and it also makes it easier to include it in testing of such exotic interpolations when all assumptions currently tested for scalar and nodal interpolations don't hold. 

It also separates testing of the reference shapes from the testing of the interpolations, and fixes one discovered issue with the thrown error from `reference_shape_value(ip::DiscontinuousLagrange{shape, 0}, ...` (it silently accepted `i > n_basefunctions`)

### DIFF
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -1,89 +1,184 @@
 using Ferrite: reference_shape_value, reference_shape_gradient
 
-@testset "interpolations" begin
+"""
+    test_interpolation_properties(ip::Interpolation)
 
-    @testset "Value Type $value_type" for value_type in (Float32, Float64)
-        @testset "Correctness of $interpolation" for interpolation in (
-                Lagrange{RefLine, 1}(),
-                Lagrange{RefLine, 2}(),
-                Lagrange{RefQuadrilateral, 1}(),
-                Lagrange{RefQuadrilateral, 2}(),
-                Lagrange{RefQuadrilateral, 3}(),
-                Lagrange{RefTriangle, 1}(),
-                Lagrange{RefTriangle, 2}(),
-                Lagrange{RefTriangle, 3}(),
-                Lagrange{RefTriangle, 4}(),
-                Lagrange{RefTriangle, 5}(),
-                Lagrange{RefHexahedron, 1}(),
-                Lagrange{RefHexahedron, 2}(),
-                Serendipity{RefQuadrilateral, 2}(),
-                Serendipity{RefHexahedron, 2}(),
-                Lagrange{RefTetrahedron, 1}(),
-                Lagrange{RefTetrahedron, 2}(),
-                Lagrange{RefPrism, 1}(),
-                Lagrange{RefPrism, 2}(),
-                Lagrange{RefPyramid, 1}(),
-                Lagrange{RefPyramid, 2}(),
-                #
-                DiscontinuousLagrange{RefLine, 0}(),
-                DiscontinuousLagrange{RefQuadrilateral, 0}(),
-                DiscontinuousLagrange{RefHexahedron, 0}(),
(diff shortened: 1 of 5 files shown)


test/test_interpolations.jl
```diff
@@ -1,89 +1,200 @@
...
-                RannacherTurek{RefQuadrilateral, 1}(),
-                RannacherTurek{RefHexahedron, 1}(),
-            )
-            # Test of utility functions
-            ref_dim = Ferrite.getrefdim(interpolation)
-            ref_shape = Ferrite.getrefshape(interpolation)
-            func_order = Ferrite.getorder(interpolation)
-            @test typeof(interpolation) <: Interpolation{ref_shape, func_order}
+This function tests the following implementation details for an
+interpolation. All base interpolations should pass this test, but
```
termi-official, line 7: Technically we are not checking for "implementation details" but we check for invariants of the interpolation.

test/test_utils.jl
termi-official, line 1: Stuff for a separate PR, but wouldn't it make sense to provide these via Ferrite directly instead of just in the tests? I think they can also be helpful in FerriteViz later to visualize H(div) vector-valued basis functions directly.

termi-official (approved):
Other than some minor detail this is fine for me. This should also make it easier in the future to extend the tests with interpolations having more fancy properties to check for.

test/test_interpolations.jl
KnutAM: What do you mean with invariants in this setting? 

I agree that implementation details is perhaps not the best word, but I couldn't come up with something covering all the test cases either... Since we don't check for some properties here, e.g. the Dirac-delta property, interpolation properties seems a bit off, especially since we also check the ArgumentError checks for `reference_shape_value`...

termi-official: We can also just use properties. Invariants are essentially "correctness properties" of functions which hold true under a given program state.

test/test_utils.jl
KnutAM: Not sure why you would need these for visualization? In principle I agree that some of these could belong in the code base, if nothing else to allow automatically plotting the reference shapes for documentation. 

If it is only for testing purposes, but also useful for downstream testing, I think it at least should be put in a submodule to separate the namespace and make it clear that it isn't used by Ferrite itself. For example #1135 is an example of code I find quite confusing that we have (but that's not even used in testing, so can probably be removed). I guess if we want to move test util code into src is a question for @fredrikekre? 

Regarding the implementations herein, I think we should also attribute more implementation to the reference shapes themselves (as I did here for the normals), and not to an interpolation on the reference shape. I always find it a bit strange that we don't have `reference_coordinates(::Type{AbstractRefShape})`, but of course, these coordinates are rarely used. 

I also noticed now that `reference_face_area` here should be updated to use facet instead of face, but that's also for a separate PR.

test/test_interpolations.jl
KnutAM: OK, let's go with that then. After all, this is just a docstring for a test function (besides, it was already called `test_interpolation_properties` :))

```suggestion
This function tests the following
interpolation properties. All base interpolations should pass this test, but
```

test/test_utils.jl
termi-official: > Not sure why you would need these for visualization?

With this stuff it is easier to show (and visually debugging) normal mappings.

> I think we should also attribute more implementation to the reference shapes [...]

Totally agree. For me Ferrite is a toolbox and should provide all basic tools to build more complex finite element codes on top.

PR #1138, opened 2025-02-02
Clarify construction of Jacobian in Navier-Stokes tutorial

jfdev001:
In the incompressible Navier-Stokes tutorial, there are two things I noticed that might be improved:

P1. The construction of the Jacobian is a bit hand-wavy in the sense that there is a claim that the Jacobian and stiffness matrix share the same sparsity pattern.

P2. It is not necessarily clear what motivates the `navierstokes_jac_element!` function.

To address these points, I have added a small derivation for the analytical form of the Jacobian. In particular, I note that the nonlinear advection term must be handled explicitly when constructing the Jacobian. The functions `navierstokes_jac!` and
`navierstokes_jac_element!` already implement what I describe in the derivation, but to make things explicit, I believe the derivation is worth including.

The derivation is currently in the section
`Commented Implementation > Solution of the semi-discretized system via DifferentialEquations.j` and may be more appropriate in the theory section.

### DIFF
--- a/docs/src/literate-tutorials/ns_vs_diffeq.jl
+++ b/docs/src/literate-tutorials/ns_vs_diffeq.jl
@@ -364,11 +364,61 @@ u₀ = zeros(ndofs(dh))
 apply!(u₀, ch);
 
 # DifferentialEquations assumes dense matrices by default, which is not
-# feasible for semi-discretization of finite element models. We communicate
-# that a sparse matrix with specified pattern should be utilized through the
-# `jac_prototyp` argument. It is simple to see that the Jacobian and the
-# stiffness matrix share the same sparsity pattern, since they share the
-# same relation between trial and test functions.
+# feasible for the semi-discretization of finite element models. Since we use 
+# an implicit method to solve the ODEs, it is more efficient to provide 
+# information to the solver about the sparsity pattern and values of the Jacobian.
+# We communicate that a sparse matrix with specified pattern should be utilized 
+# through the `jac_prototype` argument. The sparsity pattern and values for the
+# Jacobian can be shown by taking the directional gradient (see also [deal.ii: step 57](https://www.dealii.org/current/doxygen/deal.II/step_57.html)) of
+# $F(\bold{x})$ along $\delta \bold{x}$ where $F(\bold{x}) = F(v, p)$ is the 
+# semi-discrete weak form of the incompressible Navier-Stokes equations such that
+# ```math
+# F(v, p) = \begin{pmatrix}
+# F_1 \\
+# F_2
+# \end{pmatrix} = \begin{pmatrix}
+# - \int_\Omega \nu \nabla v : \nabla \varphi - \int_\Omega (v \cdot \nabla) v \cdot \varphi + \int_\Omega p (\nabla \cdot \varphi) \\
+# \int_{\Omega} (\nabla \cdot v) \psi
+# \end{pmatrix}.
+# ```
+# The variable $t$ is omitted as it is not relevant to the directional gradient. 
+# The directional gradient is thus given by
+# ```math
+# \begin{aligned}
+#   \nabla F(v, p) \cdot (\delta v, \delta p) &= \lim_{\epsilon \to 0 } \frac{1}{\epsilon} (F(v + \delta v, p + \delta p ) - F(v, p)) \\
+#   &= \lim_{\epsilon \to 0}  \frac{1}{\epsilon} \begin{pmatrix}
+#   F_1(v + \epsilon \delta v, p + \epsilon \delta p) - F_1(v,p) \\
+#   F_2(v + \epsilon \delta v) - F_2(v)
+#   \end{pmatrix} \\
+#   &= \lim_{\epsilon \to 0}  \frac{1}{\epsilon} \begin{pmatrix}
+# - \int_\Omega \nu \nabla (v + \epsilon \delta v) : \nabla \varphi - \int_\Omega ((v + \epsilon \delta v) \cdot \nabla) (v + \epsilon \delta v) \cdot \varphi + \int_\Omega (p + \epsilon \delta p)(\nabla \cdot \varphi)  - F_1 \\
+# \int_{\Omega} (\nabla \cdot (v + \epsilon \delta v)) \psi - F_2 
+#   \end{pmatrix} \\
+#   &= \lim_{\epsilon \to 0} \frac{1}{\epsilon} \begin{pmatrix}
+#   - \int_{\Omega} \nu \epsilon \nabla \delta v : \nabla \varphi - \int_{\Omega} (\epsilon \delta v \cdot \nabla v + v \cdot \epsilon \nabla \delta v + \epsilon^2 \delta v \cdot \nabla \delta v) \cdot \varphi + \int_{\Omega} \epsilon \delta p (\nabla \cdot \varphi) \\ 
+#   - \int_{\Omega} (\nabla \cdot \epsilon \delta v) \psi
+#   \end{pmatrix} \\
+#   &=  \begin{pmatrix}
+#   - \int_{\Omega} \nu \nabla \delta v : \nabla \varphi - \int_{\Omega} (\delta v \cdot \nabla v + v \cdot \nabla \delta v) \cdot \varphi + \int_{\Omega} \delta p (\nabla \cdot \varphi) \\ 
+#   - \int_{\Omega} (\nabla \cdot \delta v) \psi
+#   \end{pmatrix}.
+# \end{aligned}
+# ```
+# The directional gradient of the nonlinear advection term is the only term that 
+# does not match the pattern described by the stiffness matrix. This implies 
+# that local contributions from the nonlinear elements 
+# ```math
+#  - \int_{\Omega} (\delta v \cdot \nabla v + v \cdot \nabla \delta v) \cdot \varphi 
+# ```
+# must be added to the Jacobian explicitly, and $v$ must be a function value
+# while $\nabla v$ must be a function gradient.
+# Conversely,
+# $\delta v$ and $\delta p$ are approximated by $\varphi$ and $\psi$, 
+# respectively, so the nonzero values of $K$ *are* the local contributions of the 
+# linear terms and can be used to populate the Jacobian directly.
+# It is therefore simple to see that the Jacobian and the stiffness matrix share 
+# the same sparsity pattern, since they share the same relation between trial 
+# and test functions.
 jac_sparsity = sparse(K);
 
 # To apply the nonlinear portion of the Navier-Stokes problem we simply hand


docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -364,11 +364,61 @@ u₀ = zeros(ndofs(dh))
...
+# ```
+# must be added to the Jacobian explicitly, and $v$ must be a function value
+# while $\nabla v$ must be a function gradient.
+# Conversely,
+# $\delta v$ and $\delta p$ are approximated by $\varphi$ and $\psi$, 
+# respectively, so the nonzero values of $K$ *are* the local contributions of the 
+# linear terms and can be used to populate the Jacobian directly.
+# It is therefore simple to see that the Jacobian and the stiffness matrix share 
+# the same sparsity pattern, since they share the same relation between trial 
+# and test functions.
```
termi-official, line 421: If it is still necessary after fixing the comment above, can you rephrase this? The continuous and discrete notation/theory is mixed up (plus notation).

termi-official (commented):
Thanks for this proposal.

> The construction of the Jacobian is a bit hand-wavy in the sense that there is a claim that the Jacobian and stiffness matrix share the same sparsity pattern.

From reading over your suggestion, your actual question here is: How do we compute the Jacobian for the right hand side? (See comments in the diff)

> It is not necessarily clear what motivates the navierstokes_jac_element! function.

Good point. We can probably clarify be renaming this function.

docs/src/literate-tutorials/ns_vs_diffeq.jl
jfdev001: I will address both of your comments (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1138#discussion_r1945190356 and https://github.com/Ferrite-FEM/Ferrite.jl/pull/1138#discussion_r1945162665) this week/weekend (2025-02-10). Just confirming that I am actively working on it.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -113,6 +113,14 @@ nothing                    #hide
 # in the implementation and only stated for clarity in this section.
 #
 #
+# ### Derivation of the Jacobian
+# * The solution of nonlinear problems with newton-like methods (deal.ii, bueller)
+# * The Jacobian's relevance, provide the newton update equation (deal.ii, bueller, heath)
```
termi-official, line 118: Please take note that no Newton is involved in this example.

jfdev001: Good point. I see now that the solver is `Rodas5P`, which relies on the jacobian (see 1.3 in Steinebach G. Construction of Rosenbrock–Wanner method Rodas5P and numerical benchmarks within the Julia Differential Equations package. In: BIT Numerical Mathematics, 63(2), 2023 from [OrdinaryDiffEqRosenbrock.Rodas5P](https://docs.sciml.ai/OrdinaryDiffEq/stable/semiimplicit/Rosenbrock/#OrdinaryDiffEqRosenbrock.Rodas5P)) but is *not* a newton method. I will correct this.

termi-official: We do not need to go over board with details here, as we really just want to showcase here that there is not too much to do to use a wide range of solvers from DiffEq.

termi-official:
> Regarding the header/subheader captilization, this is not internally
consistent in the Ferrite docs anyways (maybe I open a new issue for
this? 

Please go ahead. :) 


I will need a few days to come back to this unfortunately, sorry for the delay.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -112,6 +112,87 @@ nothing                    #hide
...
+# pattern of the Jacobian. By definition, and omitting $t$, the directional 
+# gradient is given by 
+# 
+# ```math
+# \begin{aligned}
+#   \nabla f(v, p) \cdot (\delta v, \delta p) &= \lim_{\epsilon \to 0 } \frac{1}{\epsilon} (f(v + \epsilon \delta v, p + \epsilon \delta p ) - f(v, p)) \\
+#   &= \lim_{\epsilon \to 0}  \frac{1}{\epsilon} \begin{bmatrix}
+# - \int_\Omega \nu \nabla (v + \epsilon \delta v) : \nabla \varphi - \int_\Omega ((v + \epsilon \delta v) \cdot \nabla) (v + \epsilon \delta v) \cdot \varphi + \int_\Omega (p + \epsilon \delta p)(\nabla \cdot \varphi) \\
+# \int_{\Omega} (\nabla \cdot (v + \epsilon \delta v)) \psi
+#   \end{bmatrix} - f(v, p) \\
```
termi-official, line 133:
```suggestion
#   \end{bmatrix} - f(v, p) \\
```
is missing the eps or braces

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -112,6 +112,87 @@ nothing                    #hide
 # of $v$ and $p$ respectively, while $\hat{\varphi}$ is the choice for the test function in the discretization. The hats are dropped
 # in the implementation and only stated for clarity in this section.
 #
+# ### [Derivation of the Jacobian](@id jacobian-derivation)
+# To enable the use of a wide range of solvers, it is more efficient to provide 
+# information to the solver about the sparsity pattern and values of the Jacobian
+# of $f(u, t)$. The sparsity pattern and values for the Jacobian can be shown by 
+# taking the [directional gradient](https://en.wikipedia.org/wiki/Directional_derivative) of
```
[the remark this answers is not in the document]
jfdev001: It should be directional derivative, I have pushed a change

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -112,6 +112,87 @@ nothing                    #hide
...
+# To determine the values and sparsity pattern of the Jacobian given 
+# $\nabla f(v, p) \cdot (\delta v, \delta p)$, we now consider the 
+# finite element approximation of $\delta v$ and $\delta p$
+# ```math
+# \begin{aligned}
+# \delta v_h(\mathbf{x}) &= \sum_{i}^{N} \varphi_i(\mathbf{x}) \delta \hat{v}_i \\
+# \delta p_h(\mathbf{x}) &= \sum_{i}^{N} \psi_i(\mathbf{x}) \delta \hat{p}_i.
+# \end{aligned}
+# ```
+# on a mesh with element size $h$ and $N$ trial (aka nodal shape) functions
```
termi-official, line 154:
```suggestion
# on a mesh with characteristic element size $h$ and $N$ trial (aka nodal shape) functions 
```

termi-official (commented):
Okay, maybe we need to go one step back here. What exactly is the problem you are trying to solve here? Maybe I am missing something, but if you want to show how the Jacobian and Stokes operator relate to each other, then we can compute the Jacobian just directly without iterating over theory on the function space again.

The Jacobian here is *per definition* really just

$$J(u) := \partial_u (K u + N(u)) = K + \partial_u N(u)$$

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -112,6 +112,87 @@ nothing                    #hide
...
+# with the analagous term in the directional gradient,
+# ```math
+# \begin{aligned}
+# - \int_\Omega \nu \nabla \delta v : \nabla \varphi  & \xrightarrow{\delta v_h} - \sum_{i}^{N} (\int_{\Omega} \nu \nabla \varphi_i : \nabla \varphi_j) \delta \hat{v}_i \\ 
+# - \int_\Omega \nu \nabla v : \nabla \varphi  & \xrightarrow{v_h}  - \sum_{i}^{N} (\int_{\Omega} \nu \nabla \varphi_i : \nabla \varphi_j) \hat{v}_i \\
+# \end{aligned} 
+# ```
+# we see that the coefficients (i.e., the terms in the integrand on the right 
+# hand side of the arrow) for the nodal unknowns $\delta \hat{v}_i$ and $\hat{v}_i$
+# are the same. The same can be shown for the pressure term and the
```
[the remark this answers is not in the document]
jfdev001: This is just a notational error. I am not taking a limit. I wanted to indicate that when one substitutes the expression over the arrow (e.g., $\delta v_h$), the left side of the arrow becomes the right side. It would likely be clearer to write, "Substituting the finite element approximation $v_h$ for the viscoscity term and substituting the finite element approximation $\delta v_h$ for the analogous term in the directional derivative, ..."

jfdev001:
> Okay, maybe we need to go one step back here. What exactly is the problem you are trying to solve here? Maybe I am missing something, but if you want to show how the Jacobian and Stokes operator relate to each other, then we can compute the Jacobian just directly without iterating over theory on the function space again.
> 
> The Jacobian here is _per definition_ really just
> 
> J ( u ) := ∂ u ( K u + N ( u ) ) = K + ∂ u N ( u )

This is definitely the most succinct way of communicating what the Jacobian is, but given that it's a tutorial, I wanted to provide lower level details that correspond more directly to the function `navierstokes_jac!` and `navierstokes_jac_element!`. The problem I am therefore attempting to solve is to more clearly motivate how one would write a function(s) with the Ferrite API defining a Jacobian that could be used with DifferentialEquations.jl solvers. Perhaps the most clear explanation would be to do the following:

1. State J ( u ) := ∂ u ( K u + N ( u ) ) = K + ∂ u N ( u ) at the beginning
2. From this, one can very clearly see that the sparsity pattern and values of the Jacobian correspond to element contributions of K. Hence, the `nonzeros(J) .= nonzeros(K)` in `navierstokes_jac!`.  I could then remove the large part of the discussion of $K$ in my derivation.
3. Show the directional derivative derivation (see [the directional derivative is equivalent to the Jacobian](https://math.stackexchange.com/questions/3191003/directional-derivative-and-jacobian-matrix)) since I think showing this is particularly didactic and is not shown in any other Ferrite tutorials involving a Jacobian. 
4. From the derivation, highlight how ∂ u N ( u ) corresponds to $- \int_{\Omega} (\delta v \cdot \nabla v + v \cdot \nabla \delta v) \cdot \varphi$.

termi-official:
Thanks for elaborating. So, we can probably do it similar as in the hyperelasticity tutorial here: https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/hyperelasticity/ , i.e. just having the main points in the text and derivation details on the underlying function space in spoilers. What do you think?

jfdev001:
Thank you for working with me on this. It is my first proper open-source contribution beyond trivial documentation PRs, so I appreciate your patience!  I think the style in https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/hyperelasticity/ is perfect. I will rewrite my PR in that spoilers style. I am traveling for business currently, so may not have time until next week (20250-02-24) to make changes.

termi-official:
No problem, we have to thank for the help to clarify the tutorial for new users!

jfdev001:
Two comments that are unrelated to this direct PR but which might warrant opening separate issues related to the Navier-Stokes tutorial:

1. Maybe I open a separate issue based on the below [comment](https://github.com/Ferrite-FEM/Ferrite.jl/blob/41b237548b58d14f265f1b237c2635f80b53cf8a/docs/src/literate-tutorials/ns_vs_diffeq.jl#L541)? I did not see one (maybe i missed it?) in the list of issues that mentioned this one in particular.
> All norms must not depend on constrained dofs. A problem with the presented implementation is that we are currently unable to strictly enforce constraint everywhere in the internal time integration process of [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl), hence the values might differ, resulting in worse error estimates. We try to resolve this issue in the future. Volunteers are also welcome to take a look into this!
2. Maybe an extra details section on the NS weak form derivation would be useful? In https://github.com/Ferrite-FEM/Ferrite.jl/issues/1139, I mention that no books or tutorials I've seen actually show the integration by parts step involving vector fields, so perhaps showing it in this tutorial could be a nice addition.

jfdev001:
TODO: need to update parenthesis in derivation for `dot(v, gradv)`, should be `dot(v, grad)v`

jfdev001:
@termi-official  Any thoughts on my comments here https://github.com/Ferrite-FEM/Ferrite.jl/pull/1138#issuecomment-2763286050 ? Regarding opening another issue/PR for

> Maybe an extra details section on the NS weak form derivation would be useful?

if we include this in the current Navier-Stokes tutorial, it would eliminate the "need" to include some steps in the "how-to" mentioned in https://github.com/Ferrite-FEM/Ferrite.jl/issues/1139

Incidentally, the current PR is ready for review. Is there a better way to indicate that a PR is ready for review other than just converting the draft PR to a normal PR? I see that, for example, https://github.com/Ferrite-FEM/Ferrite.jl/pull/1183 has the awaiting review label, but it doesn't look like I can assign such a label.

termi-official:
> Incidentally, the current PR is ready for review. Is there a better way to indicate that a PR is ready for review other than just converting the draft PR to a normal PR? I see that, for example, #1183 has the awaiting review label, but it doesn't look like I can assign such a label.

This is correct. The label is for us to notify Fredrik that we need him.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# sparsity values and pattern can be used to initialize the Jacobian.
+#
+# To define $\frac{\partial N}{\partial u}$, we take the directional derivative
+# of $f$ along the vector $(\delta v, \delta p)$ and
+# define finite element approximations of $\delta v$ and $\delta p$ with trial
+# functions $\varphi$ and $\psi$, respectively. The discrete form of
+# $\frac{\delta N}{\delta u}$ is therefore given by
+#
+# ```math
+#     - \sum_{i}^{N} (\int_{\Omega} ((\varphi_i \cdot \nabla) v + (v \cdot \nabla) \varphi_i) \cdot \varphi_j) \delta \hat{v}_i,
```
termi-official, line 135: I think we should specify this further, as this expression is not correct in its dimension. The left side is dim(u) and the right side is dim(v). Also the continuous and discrete variables are mixed.

```suggestion
#     \frac{\partial N}{\partial u} = - \sum_{i}^{N} (\int_{\Omega} ((\varphi_i \cdot \nabla) v + (v \cdot \nabla) \varphi_i) \cdot \varphi_j) \cdot \delta \hat{v}_i,
```

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# mass matrix form given by $f(u, t) = Ku + N(u)$. By [definition](https://nl.mathworks.com/help/matlab/ref/odeset.html#d126e1203285),
+# ```math
+# J = \frac{\partial f}{\partial u} = \frac{\partial}{\partial u}(K u + N(u)) = K + \frac{\partial N}{\partial u}.
+# ```
+# It is simple to see that the Jacobian and the discretized Stokes operator $K$
+# share the same sparsity pattern since they share the same relation between
+# trial and test functions. This implies that once $K$ is assembled, its
+# sparsity values and pattern can be used to initialize the Jacobian.
+#
+# To define $\frac{\partial N}{\partial u}$, we take the directional derivative
```
termi-official, line 128: The expression is already defined.
```suggestion
# To derive $\frac{\partial N}{\partial u}$, we take the directional derivative
```

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# To enable the use of a wide range of solvers, it is more efficient to provide
+# information to the solver about the [sparsity pattern and values of the Jacobian](https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/#stiff)
+# $J$ of $f(u, t) \equiv f(v, p, t)$ where $f(u, t)$ is just the RHS of the
+# mass matrix form given by $f(u, t) = Ku + N(u)$. By [definition](https://nl.mathworks.com/help/matlab/ref/odeset.html#d126e1203285),
+# ```math
+# J = \frac{\partial f}{\partial u} = \frac{\partial}{\partial u}(K u + N(u)) = K + \frac{\partial N}{\partial u}.
+# ```
+# It is simple to see that the Jacobian and the discretized Stokes operator $K$
+# share the same sparsity pattern since they share the same relation between
+# trial and test functions. This implies that once $K$ is assembled, its
```
termi-official, line 125: Since we already clarify this, lets try to be specific.
```suggestion
# trial and test functions, ignoring the zero values in the form. This implies that once $K$ is assembled, its
```

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# sparsity values and pattern can be used to initialize the Jacobian.
+#
+# To define $\frac{\partial N}{\partial u}$, we take the directional derivative
+# of $f$ along the vector $(\delta v, \delta p)$ and
```
termi-official, line 129: I think we are mixing the derivative at the continuous weak form with the derivative of the discrete problem.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# define finite element approximations of $\delta v$ and $\delta p$ with trial
+# functions $\varphi$ and $\psi$, respectively. The discrete form of
+# $\frac{\delta N}{\delta u}$ is therefore given by
+#
+# ```math
+#     - \sum_{i}^{N} (\int_{\Omega} ((\varphi_i \cdot \nabla) v + (v \cdot \nabla) \varphi_i) \cdot \varphi_j) \delta \hat{v}_i,
+# ```
+#
+# on a mesh with characteristic element size $h$ and $N$ trial functions.
```
termi-official, line 138: h is not used. Also, N is the nonlinear form.

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+# It is simple to see that the Jacobian and the discretized Stokes operator $K$
+# share the same sparsity pattern since they share the same relation between
+# trial and test functions. This implies that once $K$ is assembled, its
+# sparsity values and pattern can be used to initialize the Jacobian.
+#
+# To define $\frac{\partial N}{\partial u}$, we take the directional derivative
+# of $f$ along the vector $(\delta v, \delta p)$ and
+# define finite element approximations of $\delta v$ and $\delta p$ with trial
+# functions $\varphi$ and $\psi$, respectively. The discrete form of
+# $\frac{\delta N}{\delta u}$ is therefore given by
```
termi-official, line 132:
```suggestion
# $\frac{\partial N}{\partial u}$ is therefore given by
```

I also think I see where the confusion comes from. We usually use $u$ for the continuous problem and also for the discrete $u$ to show that they are structurally the same thing (and to not do a mess with the notation).  However, in this tutorial we actually use $\hat{u}$ to make sure we distinguish these unambiguously. Can you fix this and make sure that this section is just using the discrete stuff?

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -105,13 +105,128 @@ nothing                    #hide
...
+#     corresponding to the nonlinear advection term is of particular interest since
+#     this integrand contains terms that evolve over time and thus these terms
+#     cannot be formulated as constant coefficients.
+#     To make this explicit, we construct a finite element approximation of
+#     $\delta v$ and $\delta p$ on a mesh with characteristic element size $h$
+#     and $N$ trial functions $\varphi$. An important note here is
+#     that the trial functions are the same as in the previous [section](@ref weak-form-derivation). The finite element approximation for $\delta v$ is thus given by
+#
+#     ```math
+#     \delta v_h(\mathbf{x}) = \sum_{i}^{N} \varphi_i(\mathbf{x}) \delta \hat{v}_i,
```
termi-official, line 200: I think we never used $v_h$ in this tutorial?

termi-official (commented):
There are multiple inconsistencies. Can you check the comments I left?

docs/src/literate-tutorials/ns_vs_diffeq.jl
jfdev001: Right! $v_h$ is never defined; however, $\delta v_h$ is the finite element approximation for the direction in which the derivative of $f$ is computed. If we do not explicitly define $\delta v_h$, then its substitution in [ns_vs_diffeq.jl:207](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1138/files#diff-86f6c62be033e4a97b5e334dd27ac6a9c3402556163dc345a26b5d49c1fa87d4R207) to get

 ```math
- \sum_{i}^{N} (\int_{\Omega} ((\varphi_i \cdot \nabla) v + (v \cdot \nabla) \varphi_i) \cdot \varphi_j) \delta \hat{v}_i,
```

may be less clear. It is therefore not necessary to define $v_h$, right?

File: docs/src/literate-tutorials/ns_vs_diffeq.jl (hunk 11 above)
Comment (jfdev001) [reply docs/src/literate-tutorials/ns_vs_diffeq.jl]: Removed $h$. $N$ is replaced with $n$, see c1d225ff5008735cc57730a5b3fbf5194f77d95e

File: docs/src/literate-tutorials/ns_vs_diffeq.jl (hunk 12 above)
Comment (jfdev001) [reply docs/src/literate-tutorials/ns_vs_diffeq.jl]: So to be clear with the notation, I am assuming the following:

1. The test functions $\varphi$ and $\psi$ for $v$ and $p$, respectively, are $\varphi \in H^{1}_{0}$ and $\psi \in L^2(\Omega)$ where 

```math
H^1_0 = \{ v \in H^1 (\Omega)^d | v = 0\ \text{on}\ \partial \Omega_D \},
```

and 

```math
H^{1} = \{ u_i : \Omega \rightarrow \mathbb{R}^d | u_i, \frac{\partial u_i}{\partial x_j} \in L^2(\Omega), i, j = 1, ..., d \},
```
where $d=2$ since we have a 2D domain. This notation is taken from [ch 2.2 in He2012](https://uu.diva-portal.org/smash/record.jsf?pid=diva2%3A544511&dswid=6449).
2. The finite element approximation of $v$ is $v_h = \sum_i^n \hat{\varphi}_i \hat{v}_i$ and the finite element approximation of $p$ is $p_h = \sum_i^n \hat{\psi}_i \hat{p}_i$. Then $\hat{v} = [ \hat{v}_1, \hat{v}_2, ..., \hat{v}_i, ..., \hat{v}_n]^{\text{T}}$ and $\hat{p} = [ \hat{p}_1, \hat{p}_2, ..., \hat{p}_i, ..., \hat{p}_n]^{\text{T}}$. Using notation from [Ferrite.jl: Introduction to FEM -- Finite Element Approximation](https://ferrite-fem.github.io/Ferrite.jl/dev/topics/fe_intro/#Finite-element-approximation). Similarly, the finite element approximations for $\delta v$ and $\delta p$ would be $\delta v_h =  \sum_i^n \hat{\varphi}_i \delta \hat{v}_i$ and $\delta p_h =  \sum_i^n \hat{\psi}_i \delta \hat{p}_i$ used in the directional derivative.
3. We restrict the test functions to the finite set of basis functions $\hat{\varphi}_j \in V^h$ and $\hat{\psi}_j \in Q^h$ where $V^h \subset H^1_0$ and $Q^h \subset L^2(\Omega)$. Standard notation, see also ch. 3.3.4 and ch. 2.7 in [Whiteley2017](https://link.springer.com/book/10.1007/978-3-319-49971-0).

This implies to me that the only place I need to use the hat notation (e.g., $\hat{v}$, $\hat{u}$, etc.) would be when I show the finite element approximation. So, from 2. and 3., the right hand side of 

```math
     \frac{\partial N}{\partial \hat{u}} = - \sum_{i}^{n} \left(\int_{\Omega} ((\hat{\varphi_i} \cdot \nabla) \hat{v} + (\hat{v} \cdot \nabla) \hat{\varphi_i}) \cdot \hat{\varphi_j} \cdot \delta \hat{v}_i\right),
```
would be correct, yes? I know I still need to resolve the dimensionality comment https://github.com/Ferrite-FEM/Ferrite.jl/pull/1138#discussion_r2042146389!

 The use of $v$, $p$, $\psi$, and $\varphi$ (i.e., the continuous forms) in the directional derivative derivation (in the long begin{aligned} end{aligned} section) should therefore be okay, right? I base my use of the continuous forms in the derivation on the [deal.ii step 57 tutorial](https://www.dealii.org/current/doxygen/deal.II/step_57.html)

termi-official: >  the right hand side of ... would be correct, yes?

No. That expression does not even make sense. You have matrix with (N+Mx(N+M) entries = vector with N entries. If you really want to learn how to do this stuff, then I can just highly recommend to take some time, a pen and some paper and do the derivation step by step yourself without peeking into the tutorial.

jfdev001: Here I justify using the continuous notation in order to compute the Jacobian since as far as I can tell, this is the prevailing approach:

For an implicit solver, we desire the Jacobian matrix $J$ of the discretized system of nonlinear equations resulting from the finite element approximation of the Navier-Stokes equations. The typical approach to find $J$ is to linearise the Navier-Stokes equations $f$ in their original continuous setting (see [8.3: Linearisation and Gateaux Derivatives in Imperial College London's "Finite Element Analysis and Implementation" Course Notes](https://finite-element.github.io/8_nonlinear_problems.html#linearisation-and-gateaux-derivatives) or [Ch. 5.2: Linearization of the Navier Stokes Equations in USING DEAL.II TO SOLVE PROBLEMS IN COMPUTATIONAL FLUID DYNAMICS (Bystricky 2016)](https://people.sc.fsu.edu/~lb13f/bystricky_thesis.pdf) or [Wolfgang Bangerth's Lecture 31.55 -- Nonlinear Problems Part 2: Newton's Method for PDEs](https://youtu.be/oVvIWMDctlE?si=3o2Iqie4PisQ0r-Y)). The function $f$ is in a continuous setting and is a function of functions, that is, $f$ is a function of a vector-valued velocity function $v$ and scalar-valued pressure function $p$. The derivative of $f$ in this setting corresponds to the Gateaux derivative. If we assume that the Gateaux derivative exists and that it is independent of the direction $\delta u = (\delta v, \delta p)$ and continuous in the neighborhood of the point of interest $u = (v, p)$, then $f$ is Frechet differentiable. Under these conditions, the Gateaux derivative is the continuous linear operator represented by the Jacobian matrix---this definition is on pages 57-58 from [*The Differentiation and Integration of Nonlinear Operators* in  Nonlinear Functional Analysis and Applications (Rall 1971)](https://www.sciencedirect.com/science/article/abs/pii/B9780125763509500059). Therefore, by deriving the Gateaux derivative in the continuous form (i.e., using symbols like $v$ and $\varphi$ without the hats) and then substituting the finite element approximation at the end, we recover a function that defines the action of the Jacobian. Since all the terms except for the advection term are linear, we can initialize the sparsity structure and values of $J$ with $K$, but we must still compute the values in $J$ corresponding to the nonlinear advection term separately.

Does this line of thinking seem reasonable? The updates to the PR that I would make would be the following:

* Summarize use of Gateaux derivative in this setting and state that it corresponds to the Jacobian.
* Use continuous notation for derivation of Jacobian (the derivation itself is already correct, just need to adjust notation) 
* Maybe mention that for complicated systems of equations, it is common to simply use autodiff functionality of differential solvers and/or symbolically solve for the Jacobian to avoid having to derive the Jacobian by hand? 
* Remove any nonsense like $\frac{\partial N}{\partial \hat{u}}$.
* I still think mentioning the finite element solution $\delta v_h$ of the direction $\delta v$ is worth it since this is actually what is implemented and can make clear the use of `function_value` vs. `shape_value` in `navierstokes_rhs_element!`.

I appreciate your patience as my response has been quite delayed, and I look forward to your feedback!

termi-official: Sorry, took me really forever to come back to this one.

Indeed. We typically use the continuous form in the analysis of the equations and you can find often that also the weak form is linearized directly (in contrast to linearizing the discretized form). IIRC both approaches should give you the same result for consistency reasons. I usually prefer the discretized form to be used directly, as it is easier to digest for people outside formal training in math, as we only have to deal with the "simple" finite dimensional case. I am fine also with the explanation you give tho. It comes down to a matter of "taste" anyway.

> Under these conditions, the Gateaux derivative is the continuous linear operator represented by the Jacobian matrix---this definition is on pages 57-58 from [The Differentiation and Integration of Nonlinear Operators in Nonlinear Functional Analysis and Applications (Rall 1971)](https://www.sciencedirect.com/science/article/abs/pii/B9780125763509500059).

Not quite. In your reference Tapia states this for the finite dimensional (i.e. discrete) case, not for the infinite-dimensional (i.e. continuous) case.

> [...] common to simply use autodiff [...]

We could probably just use autodiff here, this is correct. I do not know if the performance will be on par tho and if the resulting system will be correctly constrained (Dirichlet boundary conditions). Have you by any chance tried if it works right now?

jfdev001: No problem! Thanks for the feedback! I did not try with autodiff. I'll come back to this PR during the holidays (after 2025-12-23), had to step away for a while, thanks for your patience.

termi-official: If you have written something somewhere here, I have gotten some notificaiton on Github, but I think the comment got gobbled up by some Github outage.

jfdev001: Hi @termi-official, I've been quite caught up with work/application season right now, so will have to step away from Ferrite longer than I anticipated. I am not sure when I would be able to return to this PR. I can either leave it as a draft, or close it. Lmk what you think.

Unfortunately, same thing applies to the https://github.com/Ferrite-FEM/Ferrite.jl/issues/1139. I would recommend unassigning me from that one so that other contributors might be able to tackle it if they want.

I appreciate all your assistance with this PR up until this point! I will try to return to it at some point, but don't know a timeline yet.

termi-official: I can fully understand. And no worries, I won't run away in the meantime. :)

Issue #399, opened 2021-10-24
Tutorial on how to implement a new finite element


termi-official:
We should at some point provide at least a section where we describe how new elements can be implemented in Ferrite.

Technically we can also provide, to prottotype new elements, some kind of general Ciarlet element, because the construction of basis functions from a function space is a very mechanistic process (see e.g. https://finite-element.github.io/2_finite_elements.html for a reference). Basically we just have to provide the triple "reference geometry", "moments" and the "function space basis" for which we solve some systems of equations (e.g. vandermonde system for nodal basis functions on a polynomial function space), giving us the coefficients for our basis functions.

We can also show the alternative way via Lagrange polynomials for Lagrange interpolations.

PR #1140, opened 2025-02-02
Describe how to implement a new interpolation

KnutAM:
2nd PR based on experience from #798.

[Docs preview](https://ferrite-fem.github.io/Ferrite.jl/previews/PR1140/devdocs/interpolations/#devdocs-interpolations)

The goal here is to highlight how new interpolations should be implemented. While working on #798, I found that having a lot of the default fallbacks to no dof_indices could be confusing/easy to cause errors. While this PR doesn't remove those defaults, at least the devdocs would point to that they should be implemented. 

I've also re-organized the docstrings as to me at least, it was not so easy to follow what should, could, and didn't have to, be implemented. 

~Could make sense to add a ref to the test function given in #1136, can be done if that is merged first...~ (done)


termi-official:
Closes #399 .

termi-official:
This should address https://github.com/Ferrite-FEM/Ferrite.jl/issues/399 . I will go over the details later.

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,51 @@
...
+Draft = false
+```
+
 # [Interpolations](@id devdocs-interpolations)
 
-## Type definitions
+All `Interpolation`s should subtype `Interpolation{shape, order}`,
+where `shape <: AbstractRefShape` is the reference shape for which
+the interpolation is defined and `order` is the characteristic interpolation
+order. The [how-to at bottom of this page](@ref devdocs-howto_new-interpolation) describes how to implement a new interpolation.
```
termi-official, line 10: Should we add a comment what we mean by characteristic order? I think it can be surprising for some users that first order Lagrange on Quad and Hex is actually non-linear. :)

docs/src/devdocs/interpolations.md
```diff
@@ -22,30 +61,136 @@ Ferrite.reference_shape_hessians_gradients_and_values!
...
+## [How to implement a new interpolation, `QTI`](@id devdocs-howto_new-interpolation)
+As an example, we will implement a `Q`uadratic `T`riangle `I`nterpolation
+(`QTI`), corresponding to `Lagrange{RefTriangle, 2}()`. This is described
+in the devdocs as the API is not considered fully stable yet.
 
-Depending on the dimension of the reference element the following functions have to be implemented
+```@example InterpolationExample
+using Ferrite # hide
+using Test #hide
+struct QTI <: ScalarInterpolation{RefTriangle, 2}
```
termi-official, line 72: ScalarInterpolation is not mentioend before, so we might want to add a short comment on this one (and VectorInterpolation and VectorizedInterpolation to make sure users get what is going on here).

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,51 @@
...
-parametrized by the reference element and its characteristic order.
+### For special interpolations
+#### Discontinuous interpolations
+For discontinuous interpolations, the following methods should
+be implemented.
+```@docs
+Ferrite.is_discontinuous(::Interpolation)
+Ferrite.dirichlet_vertexdof_indices(::Interpolation)
+Ferrite.dirichlet_facedof_indices(::Interpolation)
+Ferrite.dirichlet_edgedof_indices(::Interpolation)
```
termi-official, line 35: Actually these are optional right now, as most "L2 continuous" elements have no dofs at their boundary.

docs/src/devdocs/interpolations.md
```diff
@@ -22,30 +61,136 @@ Ferrite.reference_shape_hessians_gradients_and_values!
...
+function Ferrite.reference_shape_value(ip::QTI, ξ::Vec{2}, shape_number::Int)
+    ξ₁ = ξ[1]
+    ξ₂ = ξ[2]
+    γ = 1 - ξ₁ - ξ₂ # Helper
+    shape_number == 1 && return ξ₁ * (2ξ₁ - 1) # v1: 1 at ξ = (1, 0)
+    shape_number == 2 && return ξ₂ * (2ξ₂ - 1) # v2: 1 at ξ = (0, 1)
+    shape_number == 3 && return γ * (2γ - 1)   # v3: 1 at ξ = (0, 0)
+    shape_number == 4 && return 4ξ₁ * ξ₂ # e1: 1 at ξ = (½, ½)
+    shape_number == 5 && return 4ξ₂ * γ  # e2: 1 at ξ = (0, ½)
+    shape_number == 6 && return 4ξ₁ * γ  # e3: 1 at ξ = (½, 0)
```
termi-official, line 99: Maybe I have missed this, but do you reference somewhere where you got this specific numbering and coordinate information from (for the readers)?

docs/src/devdocs/interpolations.md
```diff
@@ -22,30 +61,136 @@ Ferrite.reference_shape_hessians_gradients_and_values!
...
 ```
 
-for all entities which exist on that reference element. The dof functions default to having no
-dofs defined on a specific entity. Hence, not overloading of the dof functions will result in an
-element with zero dofs. Also, it should always be double checked that everything is consistent as
-specified in the docstring of the corresponding function, as inconsistent implementations can
-lead to bugs which are really difficult to track down.
+For this particular interpolation, we have one degree of freedom associated
+with each vertex, and one degree of freedom associated with each edge.
+Following the Ferrite numbering rules, we start by enumerating the
```
termi-official, line 87: See below if it is actually missing or if my day was just too long to find it :)
```suggestion
Following the [Ferrite numbering rules](<add ref to page here>), we start by enumerating the
```

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,51 @@
+```@meta
+Draft = false
+```
+
```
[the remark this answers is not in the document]
KnutAM: Good spot - was used such that it would evaluate during `liveserver.jl` execution to ensure that the hidden tests passed when developing.

```suggestion
```

docs/src/devdocs/interpolations.md
KnutAM: Not sure if I follow? The default is calling the non-dirichlet variant of these, but for L2 interpolations we do
https://github.com/Ferrite-FEM/Ferrite.jl/blob/1807f667086d12ec2b6a567264670980a309647c/src/interpolations.jl#L465-L467
to get dofs at the boundary?

docs/src/devdocs/interpolations.md
KnutAM: Not directly, could perhaps be more explicit. But the idea is that it comes from the reference shape that I show at the start (at
```Julia
```@docs; canonical=false
 RefTriangle
```

docs/src/devdocs/interpolations.md
KnutAM: True, perhaps `where order is the highest complete polynomial degree described by the interpolation on the reference shape.`? 

Potentially add in the examples that `(E.g. a triangle with linear shape functions have degree 1, and so does the quadrilateral with bilinear shape functions.)`?

KnutAM: Well, actually I think this should rather be described [here](https://ferrite-fem.github.io/Ferrite.jl/stable/reference/interpolations/) since that is public API. 

For the devdocs, I think it is reasonable to demand that users first are aquatinted with the public API / docs so that we don't have to repeat too much?

docs/src/devdocs/interpolations.md
KnutAM: I think we should document these [here](https://ferrite-fem.github.io/Ferrite.jl/stable/reference/interpolations/). Especially it looks like `ip^vdim` is only documented in the tutorials...

docs/src/devdocs/interpolations.md
KnutAM: https://ferrite-fem.github.io/Ferrite.jl/stable/topics/degrees_of_freedom/#Ordering-of-Dofs :stuck_out_tongue_winking_eye:

docs/src/devdocs/interpolations.md
termi-official: I just want to make sure that we have some cross-references between the pages, so users will have an easier time to find the relevant information they need beyond the examples themselves. :)

docs/src/devdocs/interpolations.md
termi-official: > [...] where order is the highest complete polynomial degree described by the interpolation on the reference shape.

I think with order we usually refer to the "convergence order" of the element in "compatible problems". For example, the lowest monomial appearing in a Q1rot elements is of order 2, but the interpolation order is just 1.

termi-official: > For the devdocs, I think it is reasonable to demand that users first are aquatinted with the public API / docs so that we don't have to repeat too much?

I can just partially agree here. I think repeating information is sometimes okay. However, we could at least add a reference to the relevant part of the public API.

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,51 @@
...
 # [Interpolations](@id devdocs-interpolations)
 
-## Type definitions
+All `Interpolation`s should subtype `Interpolation{shape, order}`,
+where `shape <: AbstractRefShape` is the reference shape for which
+the interpolation is defined and `order` is the characteristic interpolation
```
KnutAM, line 9:
```suggestion
the interpolation is defined and [`order`](@doc Ferrite.getorder) is the characteristic interpolation
```

docs/src/devdocs/interpolations.md
KnutAM: Added links to the public API documentation.

docs/src/devdocs/interpolations.md
KnutAM: #1149

docs/src/devdocs/interpolations.md
termi-official, line 1: Since we are improving on this section right now, I just want to quickly discuss whether we should also include some kind of reference on how to construct the interpolations in first place (e.g. https://finite-element.github.io/L2_fespaces.html or something else).

KnutAM: I think that would be nice, but I think that should be a separate PR. 
In the long term, how to add a new interpolation should become public API IMO, and then we should definitively have such a section in the topic guide. So it would be very nice to start that in the devdocs for now.

termi-official: > I think that would be nice, but I think that should be a separate PR.

Sure, whatever is easiest. :)

> In the long term, how to add a new interpolation should become public API IMO

Indeed, but I really really want to resolve the flexibility issues first (https://github.com/Ferrite-FEM/Ferrite.jl/issues/626) and we might dig a bit into whether the resulting API is really flexible enough to also support the fancy stuff.

KnutAM:
TODO: Include `get_direction` devdoc docstring (ref: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1159#issuecomment-2677600471) (thanks @gijswl)

docs/src/devdocs/interpolations.md
KnutAM: #1184

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,47 @@
...
+Ferrite.edgedof_interior_indices(::Interpolation)
+Ferrite.volumedof_interior_indices(::Interpolation)
+Ferrite.getnbasefunctions(::Interpolation)
+Ferrite.adjust_dofs_during_distribution(::Interpolation)
+```
 
-### Fallback methods applicable for all subtypes of `Interpolation`
+### For special interpolations
+#### Discontinuous interpolations
+For discontinuous interpolations, implementing the following methods might be required.
```
termi-official, line 25:
```suggestion
For discontinuous interpolations, implementing the following methods might be required to apply Dirichlet boundary conditions.
```

We should also add the exact conditions here under which we can apply Dirichlet in first place.

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,47 @@
...
+```
 
+#### Non-identity mapping
+For interpolations that have a non-identity mapping (see
+[Mapping of finite elements](@ref mapping_theory)), the
+mapping type must be specified.
+```@docs
+Ferrite.mapping_type
+Ferrite.get_direction
+```
```
termi-official, line 40: We might want to communicate users and devs that these dispatches are only needed if we want to use the FEValues infrastructure. Or do we need them also outside of this?

KnutAM: Feels a bit too fine-grained to specify what is needed for different use-cases, e.g. should we also say what is only required if you use the dof distribution? Currently, I don't have anything in mind, but perhaps it might be used outside as well in the future? 

(get_direction mentions that it is used for the mapping in the docstring)

docs/src/devdocs/interpolations.md
termi-official: No concrete text yet. Essentially right now these only work if we have
1. Nodal ansatz
2. We have $N_i(x) = 0$ for all ansatz functions which are not on the face

KnutAM: Actually, the nodal ansatz is not required if using `L2ProjectedDirichlet`, which still uses e.g. `dirichlet_facedof_indices`.

To avoid stalling on this addition, could we discuss and add this in a separate PR?

termi-official: Yea, let us open an issue to track and discuss these ideas first

KnutAM: #1186

termi-official (approved):
Can we just add a hint that this API might be subject to breaking changes?

termi-official (approved):
I think that is a great addition to the docs, guiding users in implementing their custom interpolations! Thanks for the work on this Knut.

Below a comment on some details which we might want to change in the future.

docs/src/devdocs/interpolations.md
```diff
@@ -1,12 +1,47 @@
 # [Interpolations](@id devdocs-interpolations)
 
-## Type definitions
+All `Interpolation`s should subtype `Interpolation{shape, order}`,
+where `shape <: AbstractRefShape` is the reference shape for which
+the interpolation is defined and [`order`](@ref Ferrite.getorder) is the characteristic interpolation
```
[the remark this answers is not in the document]
termi-official: Needs @fredrikekre . Introducing order actually dates back to ancient times. Is it still necessary for anything?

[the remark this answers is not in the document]
fredrikekre: Don't we already have some interpolations where order doesn't make sense? Perhaps they use `order = 1`. Could use `nothing` but doesn't really matter IMO.

Issue #1142, opened 2025-02-03
Docs: deadlinks to defelement.com in `interpolations.jl`


jfdev001:
In [src/interpolations.jl](https://github.com/Ferrite-FEM/Ferrite.jl/blob/1807f667086d12ec2b6a567264670980a309647c/src/interpolations.jl) there are dead links for sources of reference finite elements (e.g., [src/interpolations.jl#L670](https://github.com/Ferrite-FEM/Ferrite.jl/blob/1807f667086d12ec2b6a567264670980a309647c/src/interpolations.jl#L670))

```julia
function reference_shape_value(ip::Lagrange{RefQuadrilateral, 3}, ξ::Vec{2}, i::Int)
    # See https://defelement.com/elements/examples/quadrilateral-Q-3.html
    ...
```
These links could either (1) be removed (not ideal since obviously the information at the links inspired the function bodies), (2) should be replaced with a new source (using the same website now www.defelement.org or a different source), or (3) replaced with a wayback machine link (e.g., https://web.archive.org/web/20220817132305/https://defelement.com/elements/examples/quadrilateral-Q-3.html).

Option (3) is the easiest, and I can open a PR for this. Option (2) may not be feasible since the new website www.defelement.org does not seem to provide the functionals and basis functions for desired interpolation on an element (see e.g., https://defelement.org/elements/examples/quadrilateral-lagrange-equispaced-3.html). I could also open an issue in the [DefElement](https://github.com/DefElement/DefElement) repo since not listing the functionals and basis functions of shape functions on a particular element type is not ideal.

KnutAM:
Nice catch! 
Using the wayback temporarily and opening a pr/issue at defelement seems like the best option to me.

KnutAM:
Reopen as a reminder to update the links once https://github.com/DefElement/DefElement/issues/385 has been solved.

mscroggs:
... I've also added better redirecting to defelement.com so now (eg) https://defelement.com/elements/index.html redirects to https://defelement.org/elements/index.html rather than giving a 404 page

jfdev001:
> Reopen as a reminder to update the links once [DefElement/DefElement#385](https://github.com/DefElement/DefElement/issues/385) has been solved.

@KnutAM I can do this update now that the defelement website has been fixed. I'll do that this week (2025-02-17).

PR #1143, opened 2025-02-03
replaced all deadlinks with wayback links

jfdev001:
This PR address issue https://github.com/Ferrite-FEM/Ferrite.jl/issues/1142

Replaced dead links that provided information about the finite elements with the corresponding wayback links. This is still not ideal since in some cases, the wayback machine did not archive a related URL (e.g., the definition for the basis functions for the 1st order BubbleEnrichedLagrange on a triangular element is not accessible: https://web.archive.org/web/20250000000000*/https://defelement.com/elements/examples/triangle-bubble-enriched-lagrange-1.html). 

~A possible solution that I plan on trying this week is to add to `src/interpolations.jl` permalinks to the files from https://github.com/DefElement/DefElement that provide information on the interpolation definitions but whose HTML pages are not archived by the wayback machine. The DefElement files that provide this interpolation information should be findable in the commit history of https://github.com/DefElement/DefElement. The PR should remain in draft mode until this can be resolved.~

### DIFF
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -667,7 +667,7 @@ function reference_coordinates(::Lagrange{RefQuadrilateral, 3})
 end
 
 function reference_shape_value(ip::Lagrange{RefQuadrilateral, 3}, ξ::Vec{2}, i::Int)
-    # See https://defelement.com/elements/examples/quadrilateral-Q-3.html
+    # See https://web.archive.org/web/20220817132305/https://defelement.com/elements/examples/quadrilateral-Q-3.html
     # Transform domain from [-1, 1] × [-1, 1] to [0, 1] × [0, 1]
     ξ_x = (ξ[1] + 1) / 2
     ξ_y = (ξ[2] + 1) / 2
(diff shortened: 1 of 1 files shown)


KnutAM:
Oh, sorry if I misunderstood the issue, but it seems like all the shape functions are missing from defelement? Not just the 3rd order quadrilaterals? I cannot see them neither on computer nor phone...

Could be related to https://github.com/DefElement/DefElement/pull/382? But I don't know how their website is built, so not sure... 

Do you want to open at issue at defelement? (If not, I can do of course)

jfdev001:
Correct! The shape functions are missing from defelement in general. This appears to be the case for more than just the 3rd order polynomials. The wayback links show you the shape functions, so it's a temporary fix. 

I will open an issue at  https://github.com/DefElement/DefElement now.

src/interpolations.jl
```diff
@@ -1323,7 +1323,7 @@ end
 #######################################
 # Lagrange-Bubble RefTriangle order 1 #
 #######################################
-# Taken from https://defelement.com/elements/bubble-enriched-lagrange.html
+# Taken from https://web.archive.org/web/20240519081038/https://defelement.com/elements/bubble-enriched-lagrange.html
```
KnutAM, line 1326: First order shape functions were not available there, but we could use the vectorized version instead:
```suggestion
# Taken from https://web.archive.org/web/20230328191012/https://defelement.com/elements/examples/triangle-bubble-enriched-vector-Lagrange-1.html
```

KnutAM (commented):
Just validated that the links contain the correct information, with the change below, it should be good to go IMO.

src/interpolations.jl
jfdev001: Good catch, I didn't see this one!

PR #1150, opened 2025-02-14
Better tolerance for continuity tests

KnutAM:
Fixes the [test failure](https://github.com/Ferrite-FEM/Ferrite.jl/actions/runs/13327913031/job/37225209762) which was unrelated to that commit/pr. 

The components of function values should be between 0 and 1, so using an absolute tolerance *should* be more reliable I believe, while also being sufficient.

### DIFF
--- a/test/test_continuity.jl
+++ b/test/test_continuity.jl
@@ -68,7 +68,6 @@
         qr = QuadratureRule{RefShape}(zeros(length(ξs)), ξs)
         cv = CellValues(qr, ip_fun, ip_geo)
         reinit!(cv, cell2, cell_coords2)
-        # fun_vals2 = similar(fun_vals)
         ue2 = u[celldofs(dh, facet2[1])]
         for q_point in 1:getnquadpoints(cv)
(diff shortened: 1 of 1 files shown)


KnutAM:
> Okay, but the test fail sporadically, right?

Yes, but I believe the reason was that with "unlucky" random values, the tested value becomes small (as in that case), and that makes it more sensitive numerical errors when using a relative tolerance. 

I guess we could also print the accuracy of the found coordinate in case of test failure to see if this is the reason, or do you have some other idea?

KnutAM:
> Okay, but the test fail sporadically, right?

Hm, I tried running with 100 times lower tolerance (both abs and rel set to 10^-8), for 1000 times, and no test failures locally. 
Tried the specific case (`Nedelec{RefTriangle,2}` on `QuadraticTriangle`) 100k times, and still no failures with the lower references. 

Seems like a really unlucky case... 

Using 1e-9 tolerance gives failures, but that is expected. Only slightly lower abs tolerance accepted, so not sure if it will matter, but I still think abs is better so will merge in this one (to also get the cleanup).

PR #1151, opened 2025-02-14
ProjectedDirichlet BC for H(div) and H(curl)

KnutAM:
For H(div) and H(curl) interpolations, the dof is not associated with a point, but with facets and edges, respectively. However, the boundary conditions appear for the facets in the forms described in the [docstring](https://ferrite-fem.github.io/Ferrite.jl/previews/PR1151/reference/boundary_conditions/#Ferrite.WeakDirichlet). This implies that we cannot get a coordinate where we want to assign the dof, but instead we have to minimize the difference, not sure how much theory to include, but we could add for e.g. H(div) that our "problem" is,
```math
\begin{align*}
&\boldsymbol{q}(\boldsymbol{x}) \cdot \boldsymbol{n} = f \quad \text{ on } \Gamma, \quad \text{Not possible for arbitrary } f(\boldsymbol{x},...) \\
&\text{Instead}\\
&\boldsymbol{q}(\boldsymbol{x}) = \mathrm{argmin} U(\boldsymbol{q}(\boldsymbol{x})), \quad U(\boldsymbol{q}(\boldsymbol{x})) = \int_\Gamma \left[\boldsymbol{q}(\boldsymbol{x}) \cdot \boldsymbol{n} - f\right]^2\ \mathrm{d}\Gamma \\
&\text{Found as solution to the directional derivative in all possible directions,}\delta\boldsymbol{q}(\boldsymbol{x})\text{. I.e., for }\epsilon \rightarrow 0, \\
&0 = \frac{\partial}{\partial \epsilon} U(\boldsymbol{q}(\boldsymbol{x}) + \epsilon \delta\boldsymbol{q}(\boldsymbol{x}) = \int_\Gamma \left[\boldsymbol{q}(\boldsymbol{x}) \cdot \boldsymbol{n} - f\right]\delta\boldsymbol{q}(\boldsymbol{x}) \cdot \boldsymbol{n}\ \mathrm{d}\Gamma \\ 
&\text{Introducing the FE approximation for test and trial, we then get what we have in the docstring},\\
&\underbrace{\int_{\Gamma^f} \left[\delta\boldsymbol{N}^f_i \cdot \boldsymbol{n}^f\right]\left[\boldsymbol{N}^f_j \cdot \boldsymbol{n}^f\right]\ \mathrm{d}\Gamma}_{K^f_{ij}}\ a_j^f
= \underbrace{\int_{\Gamma^f} \left[\delta\boldsymbol{N}^f_i \cdot \boldsymbol{n}^f\right] q_n\ \mathrm{d}\Gamma}_{f^f_i}
\end{align*}
```
where the functional dependence, $f = f(\boldsymbol{x}, t, \boldsymbol{n})$, is skipped fore brevity. 

I also chose to introduce the internal `function_space(::Interpolation)::Val` function, as this was useful for dispatching. @termi-official: `RannacherTurek` and `CrouzeixRaviart` should be `L2`, right? 

<details> <summary> Side-note</summary>
An alternative name could be L2Dirichlet, as this could also be applied to $H^1$ spaces as well, and then it would actually be more accurate in the L2 sense. The problem with implementing that is that we get contributions from neighboring facets, and thus it either becomes a challenging problem to implement, or we need to solve a boundary FE problem on the global scale (probably easier and not so bad considering the sparsity) in order to apply the boundary conditions. But in most (probably all) practical cases, the additional "accuracy" in applying the BC is offset by the fact that an FE solution wouldn't be accurate enough if resolving the actual shape of the boundary prescribing function is important. However, if using $H^1$ spaces to describe e.g. fluxes, then using a standard Dirichlet condition would give the wrong total flux value if one uses e.g. linear shape functions but prescribe a quadratic velocity profile (e.g. in fluid problems...). 
</details>

### DIFF
--- a/docs/download_resources.jl
+++ b/docs/download_resources.jl
@@ -1,6 +1,7 @@
 # Download some assets necessary for docs/testing not stored in the repo
 import Downloads
 
+# Tutorials
 const directory = joinpath(@__DIR__, "src", "tutorials")
 mkpath(directory)
 
@@ -20,4 +21,15 @@ for (file, url) in [
     if !isfile(afile)
         Downloads.download(url, afile)
     end
+end
+
+# Topics
+const topics_directory = joinpath(@__DIR__, "src", "topics", "downloaded_assets")
+mkpath(topics_directory)
+
+for (file, url) in [
+        "L2ProjectedDirichlet.svg" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg",
+    ]
+    afile = joinpath(topics_directory, file)
+    isfile(afile) || Downloads.download(url, afile)
 end
--- a/docs/download_resources.jl
+++ b/docs/download_resources.jl
@@ -28,7 +28,7 @@ const topics_directory = joinpath(@__DIR__, "src", "topics", "downloaded_assets"
 mkpath(topics_directory)
 
 for (file, url) in [
-        "L2ProjectedDirichlet.svg" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg",
+        "ProjectedDirichlet.svg" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/ProjectedDirichlet.svg",
     ]
     afile = joinpath(topics_directory, file)
     isfile(afile) || Downloads.download(url, afile)
--- a/docs/src/topics/boundary_conditions.md
+++ b/docs/src/topics/boundary_conditions.md
@@ -100,7 +100,8 @@ and ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to 
 While normal `Dirichlet` boundary conditions assume the existence of such nodal support points,
 Ferrite provides the `ProjectedDirichlet`, which instead finds the degree of freedom values that
 minimizes the L2-distance between the prescribed function, ``f(\boldsymbol{x},t,\boldsymbol{n})``,
-and the finite element interpolation space. Although standard interpolations are not currently supported,
+and the finite element interpolation space, cf. [Bartels2004:ProjectedDirichlet](@cite).
+Although standard interpolations are not currently supported,
 the figure below illustrates well the difference between applying a standard `Dirichlet` condition and an
 `ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
 
--- a/docs/src/topics/boundary_conditions.md
+++ b/docs/src/topics/boundary_conditions.md
@@ -93,6 +93,21 @@ end
     Most examples make use of Dirichlet boundary conditions, for example [Heat
     Equation](@ref tutorial-heat-equation).
 
+## L2ProjectedDirichlet
+Some interpolations don't have nodal support points:
+``H(\mathrm{curl})`` interpolations, e.g., `Nedelec`, are associated to edges
+and ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to facets.
+While normal `Dirichlet` boundary conditions assume the existence of such nodal support points,
+Ferrite provides the `L2ProjectedDirichlet`, which instead finds the degree of freedom values that
+minimizes the L2-distance between the prescribed function, ``f(\boldsymbol{x},t,\boldsymbol{n})``,
+and the finite element interpolation space. Although standard interpolations are not currently supported,
+the figure below illustrates well the difference between applying a standard `Dirichlet` condition and an
+`L2ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
+
+![L2ProjectedDirichlet illustration](https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg)
+
+Here, we note that while the `Dirichlet` condition gives the correct value at the nodes, the `L2ProjectedDirichlet` gives
+a more accurate average boundary value (specifically the L2 projection of `f(x)` onto the finite element space).
 
 ## Neumann boundary conditions
 At the Neumann part of the boundary we know something about the gradient of the solution.
--- a/docs/src/topics/boundary_conditions.md
+++ b/docs/src/topics/boundary_conditions.md
@@ -104,7 +104,7 @@ and the finite element interpolation space. Although standard interpolations are
 the figure below illustrates well the difference between applying a standard `Dirichlet` condition and an
 `L2ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
 
-![L2ProjectedDirichlet illustration](https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg)
+![L2ProjectedDirichlet illustration](downloaded_assets/L2ProjectedDirichlet.svg)
 
 Here, we note that while the `Dirichlet` condition gives the correct value at the nodes, the `L2ProjectedDirichlet` gives
 a more accurate average boundary value (specifically the L2 projection of `f(x)` onto the finite element space).
--- a/docs/src/topics/boundary_conditions.md
+++ b/docs/src/topics/boundary_conditions.md
@@ -93,20 +93,20 @@ end
     Most examples make use of Dirichlet boundary conditions, for example [Heat
     Equation](@ref tutorial-heat-equation).
 
-## L2ProjectedDirichlet
+## ProjectedDirichlet
 Some interpolations don't have nodal support points:
 ``H(\mathrm{curl})`` interpolations, e.g., `Nedelec`, are associated to edges
 and ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to facets.
 While normal `Dirichlet` boundary conditions assume the existence of such nodal support points,
-Ferrite provides the `L2ProjectedDirichlet`, which instead finds the degree of freedom values that
+Ferrite provides the `ProjectedDirichlet`, which instead finds the degree of freedom values that
 minimizes the L2-distance between the prescribed function, ``f(\boldsymbol{x},t,\boldsymbol{n})``,
 and the finite element interpolation space. Although standard interpolations are not currently supported,
 the figure below illustrates well the difference between applying a standard `Dirichlet` condition and an
-`L2ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
+`ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
 
-![L2ProjectedDirichlet illustration](downloaded_assets/L2ProjectedDirichlet.svg)
+![ProjectedDirichlet illustration](downloaded_assets/ProjectedDirichlet.svg)
 
-Here, we note that while the `Dirichlet` condition gives the correct value at the nodes, the `L2ProjectedDirichlet` gives
+Here, we note that while the `Dirichlet` condition gives the correct value at the nodes, the `ProjectedDirichlet` gives
 a more accurate average boundary value (specifically the L2 projection of `f(x)` onto the finite element space).
 
 ## Neumann boundary conditions
--- a/docs/src/topics/boundary_conditions.md
+++ b/docs/src/topics/boundary_conditions.md
@@ -95,8 +95,8 @@ end
 
 ## ProjectedDirichlet
 Some interpolations don't have nodal support points:
-``H(\mathrm{curl})`` interpolations, e.g., `Nedelec`, are associated to edges
-and ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to facets.
+``H(\mathrm{curl})`` interpolations, e.g., `Nedelec`, are associated to edges and faces,
+while ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to facets.
 While normal `Dirichlet` boundary conditions assume the existence of such nodal support points,
 Ferrite provides the `ProjectedDirichlet`, which instead finds the degree of freedom values that
 minimizes the L2-distance between the prescribed function, ``f(\boldsymbol{x},t,\boldsymbol{n})``,
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1887,10 +1887,10 @@ function _update_weak_dbc!(
         dh::AbstractDofHandler, dofmapping::Dict{Int, Int}, dofcoefficients::Vector{Union{Nothing, DofCoefficients{T}}}, time::Real
     ) where {T}
     ip = function_interpolation(fv)
-    n = maximum(length, dirichlet_facetdof_indices(ip))
-    Kᶠ = zeros(n, n)
-    aᶠ = zeros(n)
-    fᶠ = zeros(n)
+    max_dofs_per_facet = maximum(length, dirichlet_facetdof_indices(ip))
+    Kᶠ = zeros(max_dofs_per_facet, max_dofs_per_facet)
+    aᶠ = zeros(max_dofs_per_facet)
+    fᶠ = zeros(max_dofs_per_facet)
     for fc in FacetIterator(dh, facets)
         reinit!(fv, fc)
         shape_nrs = dirichlet_facetdof_indices(ip)[getcurrentfacet(fv)]
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -155,7 +155,7 @@ A collection of constraints associated with the dof handler `dh`.
 """
 mutable struct ConstraintHandler{DH <: AbstractDofHandler, T}
     const dbcs::Vector{Dirichlet}
-    const wdbcs::Vector{L2ProjectedDirichlet}
+    const pdbcs::Vector{L2ProjectedDirichlet}
     const prescribed_dofs::Vector{Int}
     const free_dofs::Vector{Int}
     const inhomogeneities::Vector{T}
@@ -489,8 +489,8 @@ function update!(ch::ConstraintHandler, time::Real = 0.0)
             dbc.components, ch.dh, ch.bcvalues[i], ch.dofmapping, ch.dofcoefficients, time
         )
     end
-    for wdbc in ch.wdbcs
-        _update_weak_dbc!(ch.inhomogeneities, wdbc.f, wdbc.facets, wdbc.fv, wdbc.facet_dofs, ch.dh, ch.dofmapping, ch.dofcoefficients, time)
+    for pdbc in ch.pdbcs
+        _update_weak_dbc!(ch.inhomogeneities, pdbc.f, pdbc.facets, pdbc.fv, pdbc.facet_dofs, ch.dh, ch.dofmapping, ch.dofcoefficients, time)
     end
     # Compute effective inhomogeneity for affine constraints with prescribed dofs in the
     # RHS. For example, in u2 = w3 * u3 + w4 * u4 + b2 we allow e.g. u3 to be prescribed by
@@ -1878,11 +1878,11 @@ function add!(ch::ConstraintHandler, dbc::L2ProjectedDirichlet)
     return ch
 end
 
-function _add!(ch::ConstraintHandler, wdbc::L2ProjectedDirichlet, facet_dofs)
+function _add!(ch::ConstraintHandler, pdbc::L2ProjectedDirichlet, facet_dofs)
     # loop over all the faces in the set and add the global dofs to `constrained_dofs`
     constrained_dofs = Int[]
     cc = CellCache(ch.dh, UpdateFlags(; nodes = false, coords = false, dofs = true))
-    for (cellidx, facetidx) in wdbc.facets
+    for (cellidx, facetidx) in pdbc.facets
         reinit!(cc, cellidx)
         local_dofs = facet_dofs[facetidx]
         for d in local_dofs
@@ -1891,7 +1891,7 @@ function _add!(ch::ConstraintHandler, wdbc::L2ProjectedDirichlet, facet_dofs)
     end
 
     # save it to the ConstraintHandler
-    push!(ch.wdbcs, wdbc)
+    push!(ch.pdbcs, pdbc)
     for d in constrained_dofs
         add_prescribed_dof!(ch, d, NaN, nothing)
     end
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -67,7 +67,7 @@ end
 A `WeakDirichlet` condition enforces conditions for `field_name` on the boundary, `facets`, for
 non-nodal interpolations (e.g. ``H(\mathrm{div})`` or ``H(\mathrm{curl})``) in an weak sense,
 i.e. minimizing the L2 error between the function `f(x, t, n)` and the values described
-by the FE approximation on the facet, ``\Gamma^\mathrm{f}``. The arguments to the function are
+by the FE approximation on the facet, ``\Gamma^f``. The arguments to the function are
 the coordinate, ``\boldsymbol{x}``, the time, ``t``, and the facet normal, ``\boldsymbol{n}``.
 The quadrature rule is automatically created, but the default order, `qr_order = 2 * ip_order`
 (may be refined in future releases),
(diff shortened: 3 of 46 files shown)


termi-official:
> I also chose to introduce the internal function_space(::Interpolation)::Val function, as this was useful for dispatching. 

Yea, as already discussed before, I think having a trait system will be useful in the future.

> @termi-official: RannacherTurek and CrouzeixRaviart should be L2, right?

Yes.

src/interpolations.jl
```diff
@@ -450,6 +450,7 @@ struct DiscontinuousLagrange{shape, order} <: ScalarInterpolation{shape, order}
         return new{shape, order}()
     end
 end
+function_space(::DiscontinuousLagrange) = Val(:L2)
```
fredrikekre, line 453: Cleaner to have proper types?

KnutAM: Agreed! But maybe, 
```julia
struct FunctionSpace{space}
    FunctionSpace(space::Symbol) = new{space}()
end
```
is enough?

src/Dofs/ConstraintHandler.jl
```diff
@@ -1762,3 +1821,146 @@ function _condense_local!(
...
+    return nothing
+end
+
+function solve_weak_dbc!(aᶠ, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
+    fill!(Kᶠ, 0)
+    fill!(fᶠ, 0)
+    # Support varying number of facetdofs (for ref shapes with different facet types)
+    for i in (length(shape_nrs) + 1):size(Kᶠ, 1)
+        Kᶠ[i, i] = 1
+    end
```
KnutAM, line 1917: Currently, we don't have any H(div) or H(curl) interpolations with different number of shape functions associated with different facets. But theoretically, we could, hence this addition which is not tested. 

Removing it will case a singular matrix if such are implemented in the future, alternatively we could check and throw if `size(K,1) != length(shape_nrs)`.

src/Dofs/ConstraintHandler.jl
```diff
@@ -1762,3 +1821,146 @@ function _condense_local!(
...
+    return ch
+end
+
+
+function _update_weak_dbc!(
+        inhomogeneities::Vector{T}, f::Function, facets::AbstractVecOrSet{FacetIndex}, fv::FacetValues, facet_dofs::ArrayOfVectorViews,
+        dh::AbstractDofHandler, dofmapping::Dict{Int, Int}, dofcoefficients::Vector{Union{Nothing, DofCoefficients{T}}}, time::Real
+    ) where {T}
+    ip = function_interpolation(fv)
+    n = maximum(length, dirichlet_facetdof_indices(ip))
```
fredrikekre, line 1890: Is `n` the number of dofs related to a single facet? So for low orders `n = 1` and we solve a scalar eq?

fredrikekre (approved):
I haven't gone through the math but if it is sound and tests cover it then :+1:

termi-official:
Since I just saw it, should we rename `FunctionSpace` trait? Although technically correct, it might a bit misleading/ambiguous here. I think we rather want to describe something like conformity or continuity trait, right?

KnutAM:
> Since I just saw it, should we rename FunctionSpace trait? Although technically correct, it might a bit misleading/ambiguous here. I think we rather want to describe something like conformity or continuity trait, right?

Yes, seems like `dealii` agrees with you: https://dealii.org/current/doxygen/deal.II/classFiniteElementData.html#a0cd5f34c3ab828fac31004f3b52921a1 :)

src/Dofs/ConstraintHandler.jl
```diff
@@ -83,6 +135,7 @@ A collection of constraints associated with the dof handler `dh`.
 """
 mutable struct ConstraintHandler{DH <: AbstractDofHandler, T}
     const dbcs::Vector{Dirichlet}
+    const idbcs::Vector{WeakDirichlet}
```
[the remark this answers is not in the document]
KnutAM: Yes, absolutely (I think I had another name before, thanks for the detailed review and spotting of these things!)

src/Dofs/ConstraintHandler.jl
```diff
@@ -410,12 +463,18 @@ function update!(ch::ConstraintHandler, time::Real = 0.0)
         # g(x, t) = f(x) that discards the second parameter so that _update! can always call
         # the function with two arguments internally.
         wrapper_f = hasmethod(dbc.f, Tuple{get_coordinate_type(get_grid(ch.dh)), typeof(time)}) ? dbc.f : (x, _) -> dbc.f(x)
+        #TODO: Temporary way to get the ip, not safe as it just get the first sdh with field_name in it.
+        field_idx = find_field(ch.dh, dbc.field_name)
+        ip = getfieldinterpolation(ch.dh, field_idx)
         # Function barrier
         _update!(
-            ch.inhomogeneities, wrapper_f, dbc.facets, dbc.field_name, dbc.local_facet_dofs, dbc.local_facet_dofs_offset,
+            ch.inhomogeneities, wrapper_f, dbc.facets, dbc.local_facet_dofs, dbc.local_facet_dofs_offset,
```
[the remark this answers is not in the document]
KnutAM: Yes, I think the reason was that I tried to use the same function, but then decided it was cleaner to use `_update_weak_dbc!` instead. So this was a left-over, sorry about that, but I guess it makes sense to remove the argument, so leaving it in then.

src/Dofs/ConstraintHandler.jl
KnutAM: Yes. 
In general, we could do a lot of performance optimization here if required, but when I did some benchmark I didn't see that as a bottleneck in the Maxwell tutorial.

KnutAM:
> I haven't gone through the math but if it is sound and tests cover it then 👍

I believe it is sound, but would be good if someone can have a 2nd look on the tests [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/04f7cdb57c16c36cf65d26afd30c92b9e3f73cda/test/test_interpolations.jl#L427-L519) to make sure that I haven't done anything wrong there. 

When going through this again now, I realize that I should also test some moment-integrals, as currently only the total integral quantity is tested (e.g. the total flux over the boundary, but not the spatial distribution). Will update with that during the week.

fredrikekre (approved):
Is it too magical if `Dirichlet(::HcurlInterpolation)` adds a `WeakDirichlet`? Probably, but would be kind of nice to not have to change the BC when changing interpolations.

KnutAM:
> Is it too magical if Dirichlet(::HcurlInterpolation) adds a WeakDirichlet? Probably, but would be kind of nice to not have to change the BC when changing interpolations.

1) We might want to add `WeakDirichlet` to `:H1` interpolations in the future (see "side note" in the first comment). 
2) We might want to use `Dirichlet` for the method they have in dealii (@termi-official has some contacts that could help us figure out their exact implementation)

src/Dofs/ConstraintHandler.jl
termi-official: We also have the very same issue with H1/L2 interpolations and this property, so let us not introduce this assumption for now in parts of the code.

src/Dofs/ConstraintHandler.jl
```diff
@@ -1762,3 +1841,146 @@ function _condense_local!(
...
+
+function solve_weak_dbc!(aᶠ, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
+    fill!(Kᶠ, 0)
+    fill!(fᶠ, 0)
+    # Support varying number of facetdofs (for ref shapes with different facet types)
+    for i in (length(shape_nrs) + 1):size(Kᶠ, 1)
+        Kᶠ[i, i] = 1
+    end
+    integrate_weak_dbc!(Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
+    aᶠ .= Kᶠ \ fᶠ # Could be done non-allocating if required, using e.g. SMatrix
```
termi-official, line 1939: Should we cache the factorization of K for time-dependent problems (e.g. transient Maxwell)?

src/Dofs/ConstraintHandler.jl
```diff
@@ -61,6 +61,78 @@ function __to_components(c)
     return components
 end
 
+@doc raw"""
+    WeakDirichlet(field_name::Symbol, facets::AbstractSet{FacetIndex}, f::Function; qr_order = -1)
```
termi-official, line 65: Can we discuss the name for a moment? I think `WeakDirichlet` is misleading, because at least from the literature I know ,,weak bcs'' are usually introduced by adding energy terms (or weak form terms if we have no energy) to force the solution towards the right boundary condition, as done e.g. with Nitzsche method. However, the presented method here strongly enforces the boundary condition.

Can we merge this infrastructure into the `Dirichlet` struct directly and control the application by dispatching on the new conformity trait?

I also noticed that we are missing some docstring for the application of DBCs in L2 elements. :D

src/Dofs/ConstraintHandler.jl
termi-official: Also, do we know f there is any relation between unisolvence and the invertibility of K?

src/interpolations.jl
```diff
@@ -51,6 +51,11 @@ const InterpolationByDim{dim} = Interpolation{<:AbstractRefShape{dim}}
 abstract type ScalarInterpolation{refshape, order} <: Interpolation{refshape, order} end
 abstract type VectorInterpolation{vdim, refshape, order} <: Interpolation{refshape, order} end
 
+# Conformity, :H1, :Hdiv, :Hcurl, and :L2 currently implemented.
+struct Conformity{K}
+    Conformity(conformity::Symbol) = new{conformity}()
+end
```
termi-official, line 57: Out of curiosity: Is there a reason why we do this instead of four different structs?

src/Dofs/ConstraintHandler.jl
```diff
@@ -410,12 +483,18 @@ function update!(ch::ConstraintHandler, time::Real = 0.0)
         # g(x, t) = f(x) that discards the second parameter so that _update! can always call
         # the function with two arguments internally.
         wrapper_f = hasmethod(dbc.f, Tuple{get_coordinate_type(get_grid(ch.dh)), typeof(time)}) ? dbc.f : (x, _) -> dbc.f(x)
+        #TODO: Temporary way to get the ip, not safe as it just get the first sdh with field_name in it.
+        field_idx = find_field(ch.dh, dbc.field_name)
+        ip = getfieldinterpolation(ch.dh, field_idx)
```
termi-official, line 488: Unused code?

src/interpolations.jl
KnutAM: Seemed unnecessary to introduce the hierarchy below an `AbstractConformity` with several subtypes, when e.g. `Conformity{:H1}` reads quite good for the dispatch.

But could be changed too of course if that is considered better.

src/Dofs/ConstraintHandler.jl
KnutAM: > Can we merge this infrastructure into the Dirichlet struct directly and control the application by dispatching on the new conformity trait?

I think it would be wiser not to do that (ref: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1151#issuecomment-2665862101 above). But of course, we could consider changing `Dirichlet` to be a wrapper/input type, and then store separate structs internally depending on how we want to apply BCs.

src/interpolations.jl
termi-official: I was not even thinking about a type hierarchy here and was just curious. For me having the types directly available looks a bit more stable for users since the type accepts typos, but for the short names here it should be okay and these should be easy to find. @fredrikekre do you see any obvious fallacies with this pattern?

src/Dofs/ConstraintHandler.jl
KnutAM: > However, the presented method here strongly enforces the boundary condition.

You are right, ref [Bazilevs and Hughes (2007) Weak imposition of Dirichlet boundary conditions in fluid mechanics](https://www.sciencedirect.com/science/article/pii/S0045793005001258)

However, if the method here is used for an H1 interpolation, it would give a different result compared to the current Dirichlet which satisfies the prescribed value point-wise, whereas the present one only in a weak/integral sense.

How about going back to the original name - "IntegrableDirichlet"?

src/interpolations.jl
fredrikekre: Doesn't really matter since there will be four different concrete structs either way.

src/Dofs/ConstraintHandler.jl
termi-official: > However, if the method here is used for an H1 interpolation, it would give a different result compared to the current Dirichlet which satisfies the prescribed value point-wise, whereas the present one only in a weak/integral sense.

Oh, i see, it "smears out" the Boundary condition because we use Gauss-Legendre quadrature by default, right? The method should be exact when using Gauss-Lobatto quadrature tho.

src/Dofs/ConstraintHandler.jl
KnutAM: > Also, do we know f there is any relation between unisolvence and the invertibility of K?

From Slack discussion:
If the normal/tangential values on the boundaries are linearly independent, K will be invertible (of course assuming sufficiently high integration order). 
If one would introduce a modified BDM interpolation, changing
BDM element could be modified on edge 2 from
```julia
i == 3 && return Vec(-2x - 6y + 2, 4y) # N ⋅ n = (6y - 2)
i == 4 && return Vec(4x + 6y - 4, -2y) # N ⋅ n = (4 - 6y)
```
to
```julia
i == 3 && return Vec(-2x - 6y + 2, 4y) # N ⋅ n = (6y - 2)
i == 4 && return Vec(-2x^2 - 6y + 2, 4y) # N ⋅ n = (6y - 2)
```
such that i=4 should still be linearly independent form all other shape functions, but on edge 2 (x=0), 3 and 4 are the same.

But with this modification, it should fail the reference moment test.
Therefore, as long as an interpolation passes the reference moment tests, it should result in an invertible K.

src/Dofs/ConstraintHandler.jl
```diff
@@ -1762,3 +1841,146 @@ function _condense_local!(
...
+    return nothing
+end
+
+function solve_weak_dbc!(aᶠ, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
+    fill!(Kᶠ, 0)
+    fill!(fᶠ, 0)
+    # Support varying number of facetdofs (for ref shapes with different facet types)
+    for i in (length(shape_nrs) + 1):size(Kᶠ, 1)
+        Kᶠ[i, i] = 1
+    end
```
KnutAM, line 1937:
```suggestion
    # Supporting varying number of facetdofs (for ref shapes with different facet types) requires
    # for i in (length(shape_nrs) + 1):size(Kᶠ, 1)
    #     Kᶠ[i, i] = 1
    # end
    @assert length(shape_nrs) == size(Kᶠ, 1)
```

src/Dofs/ConstraintHandler.jl
KnutAM, line 488:
```suggestion
```

src/Dofs/ConstraintHandler.jl
KnutAM: Yes, assuming the Gauss-Lobatto has equally many points as the number of dofs, but with higher integration orders it doesn't matter what integration rule is used. I think of it like this: 

image

src/Dofs/ConstraintHandler.jl
KnutAM: Good spot!

KnutAM:
> I was asking myself if this infrastructure can help us with applying normal and tangential boundary conditions (e.g. static sliding contact)?

No, that would require redefining the coordinate system for the dofs, or introduce affine constraints. #1158 should make that quite acceptable to solve for users.

src/Dofs/ConstraintHandler.jl
termi-official: This is indeed quite interesting, I have not thought about non-matching functions here. So the new boundary condition enforcement should be closer to the actual function on our Dirichlet boundary, because the new method minimizes the distance between the analytical function and our function space, while the standard method is "just" interpolatory.

termi-official: I think we should add this sketch to the docs as it very precisely explains the difference to users.

src/interpolations.jl
KnutAM: Decided to change, I agree that it looks more standard, and the possible typo is a reasonable concern. The common type was only used for tests.

src/Dofs/ConstraintHandler.jl
KnutAM: https://ferrite-fem.github.io/Ferrite.jl/previews/PR1151/topics/boundary_conditions/#L2ProjectedDirichlet

docs/src/topics/boundary_conditions.md
```diff
@@ -93,6 +93,21 @@ end
...
+``H(\mathrm{curl})`` interpolations, e.g., `Nedelec`, are associated to edges
+and ``H(\mathrm{div})`` interpolations, e.g. `RaviartThomas`, are associated to facets.
+While normal `Dirichlet` boundary conditions assume the existence of such nodal support points,
+Ferrite provides the `L2ProjectedDirichlet`, which instead finds the degree of freedom values that
+minimizes the L2-distance between the prescribed function, ``f(\boldsymbol{x},t,\boldsymbol{n})``,
+and the finite element interpolation space. Although standard interpolations are not currently supported,
+the figure below illustrates well the difference between applying a standard `Dirichlet` condition and an
+`L2ProjectedDirichlet` condition when the prescribed function cannot be described by the chosen FE-interpolation.
+
+![L2ProjectedDirichlet illustration](https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg)
```
termi-official, line 107: This should be queried in the asset download script and reference the local figure, right?

src/Dofs/ConstraintHandler.jl
```diff
@@ -61,6 +61,78 @@ function __to_components(c)
     return components
 end
 
+@doc raw"""
+    L2ProjectedDirichlet(field_name::Symbol, facets::AbstractSet{FacetIndex}, f::Function; qr_order = -1)
```
termi-official, line 65: I assume we could also apply the H(curl)-conforming variant to ridges? If so, should we open a good-first-issue on this for potential contributors?

src/Dofs/ConstraintHandler.jl
```diff
@@ -61,6 +61,78 @@ function __to_components(c)
...
+```
+determining the coefficients to be prescribed, ``a_j^f``
+"""
+mutable struct L2ProjectedDirichlet
+    const f::Function
+    const facets::OrderedSet{FacetIndex}
+    const field_name::Symbol
+    const qr_order::Int
+    # Created during `add!`
+    fv::Union{Nothing, FacetValues}
```
termi-official, line 129: Just to make sure: This differs from Dirichlet in the sense that it only works on subdomains with the same element type (i.e. 3D no mixed grids), right? Should we do this split automatically in `add!` as a followup and add an error for now?

I assume that this is not covered yet because the 3D elements are not merged yet?

KnutAM: Will add support for that now that #1173 has been merged.

KnutAM: I don't know if that arises from some derivations (I thought so initially, but doesn't seem like it based on what I found when writing the following: https://ferrite-fem.github.io/Ferrite.jl/previews/PR798/tutorials/maxwell_good_bad_ugly/#Theoretical-background)

src/Dofs/ConstraintHandler.jl
KnutAM: I don't know if that arises from some derivations (I thought so initially, but doesn't seem like it based on what I found when writing the following: https://ferrite-fem.github.io/Ferrite.jl/previews/PR798/tutorials/maxwell_good_bad_ugly/#Theoretical-background)

docs/src/topics/boundary_conditions.md
KnutAM: Not sure what the best approach here is. Currently, only assets are downloaded for the tutorials, but I guess it makes sense to download it instead. Adding an svg to source control really bloats the number of lines etc, so not so nice.

Perhaps possible to use something like
```@example
using Downloads: download #hide
download(joinpath(@__DIR__, L2ProjectedDirichlet.svg), "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/L2ProjectedDirichlet.svg") #hide
nothing #hide
```
L2ProjectedDirichlet illustration

With #1037 would aid this by centralizing the asset location.

@fredrikekre: What are your views on how we should manage such assets in the non-tutorial part of the docs?

termi-official: > Adding an svg to source control [...]

Do we have git lfs configured? If so, then we can just add it via git lfs to treat it similar to binaries.

src/Dofs/ConstraintHandler.jl
KnutAM: Had already added support for this :D, now it is also tested.

docs/src/topics/boundary_conditions.md
KnutAM: Downloaded it and added it to a new `downloaded_assets` directory which is ignored by git.

src/Dofs/ConstraintHandler.jl
```diff
@@ -61,6 +61,78 @@ function __to_components(c)
...
+with ``\boldsymbol{q}(\boldsymbol{x}) = \boldsymbol{N}^f_j(\boldsymbol{x}) a_j^f``.
+Similar to for ``H(\mathrm{div})``, we find the minimum by setting the directional derivative equal to zero,
+which after inserting the FE-approximations yields the linear equation system,
+```math
+\underbrace{\int_{\Gamma^f} \left[\delta\boldsymbol{N}^f_i \times \boldsymbol{n}^f\right]\cdot\left[\boldsymbol{N}^f_j \times \boldsymbol{n}^f\right]\ \mathrm{d}\Gamma}_{K^f_{ij}}\ a_j^f
+= \underbrace{\int_{\Gamma^f} \left[\delta\boldsymbol{N}^f_i \times \boldsymbol{n}^f\right]\cdot \boldsymbol{f}(\boldsymbol{x}, t, \boldsymbol{n})\ \mathrm{d}\Gamma}_{f^f_i}
+```
+determining the coefficients to be prescribed, ``a_j^f``
+"""
+mutable struct L2ProjectedDirichlet
```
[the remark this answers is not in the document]
KnutAM: > I liked `WeakDirichlet`

Me too 😀 
The problem was that Hughes beat us with the name (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1151#discussion_r2042015156).

I guess so, just added L2 to be consistent with the L2Projector, but agree that ProjectedDirichlet reads better - will update.

KnutAM:
@lijas identified the corresponding functionality in deal.ii! 

For nodal and hdiv: https://www.dealii.org/current/doxygen/deal.II/namespaceVectorTools.html#a747e71d426643718d52eda904d46bb20

And for hcurl: https://www.dealii.org/current/doxygen/deal.II/group__constraints.html#ga190b302dc1ee7c89727c94c083d132d2

I will look closer if we are missing something since they have factored out the curl-functionality, but I haven't done that here.

KnutAM:
After going through the dealii docs, I've added the relevant reference here: https://ferrite-fem.github.io/Ferrite.jl/previews/PR1151/topics/boundary_conditions/#ProjectedDirichlet. 

For the H(curl), AFAIU this is related to @termi-official's comment regarding 3d, and they use a stepwise algorithm to reduce the equation system size when it couples (first solving dofs associated with edges, and then those associated with faces). This doesn't affect what is implemented in this PR, nor the user interface.

KnutAM:
Should be good from my side. 
Do you have any further comments/suggestions @termi-official or @fredrikekre?

docs/download_resources.jl
```diff
@@ -21,3 +22,14 @@ for (file, url) in [
         Downloads.download(url, afile)
     end
 end
+
+# Topics
+const topics_directory = joinpath(@__DIR__, "src", "topics", "downloaded_assets")
+mkpath(topics_directory)
+
+for (file, url) in [
+        "ProjectedDirichlet.svg" => "https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/ProjectedDirichlet.svg",
```
termi-official, line 31: The legend needs an update.

KnutAM: Fixed

termi-official (approved):
LGTM

> And for hcurl: https://www.dealii.org/current/doxygen/deal.II/group__constraints.html#ga190b302dc1ee7c89727c94c083d132d2

This is quite a smart strategy here. However, we need to implement EdgeValues and the related topological queries to reproduce this efficiently.

KnutAM:
Thanks for the comments and input @termi-official, @fredrikekre, and @lijas!

PR #1153, opened 2025-02-17
Change master/slave to main/constrained

KnutAM:
IMO we should avoid using the master/slave notation in the docs to make them more inclusive. Here is one suggestion for a different notation.

### DIFF
--- a/docs/src/topics/constraints.md
+++ b/docs/src/topics/constraints.md
@@ -79,14 +79,14 @@ be used to calculate the dependent DoFs. Ferrite has functionaliy for setting up
 ``\hat{\boldsymbol{K}}`` and ``\hat{\boldsymbol{f}}`` in an efficient way.
 
 !!! note "Limitations"
-    Ferrite currently cannot untangle constraints where a DoF is both *master* and *slave*
+    Ferrite currently cannot untangle constraints where a DoF is both a *main* and a *constrained*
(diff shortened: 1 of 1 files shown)


lijas (approved):

KristofferC:
This was merged very fast. How is this more inclusive? This is the standard terminology, no?

KnutAM:
> This was merged very fast.

Was just introduced in #1146, hence I considered fine to merge after @lijas approved. Sorry if that was too fast, I was not aware that it would be controversial to change this "tip" section. 

> This previous noemclature is the standard terminology, no? If anything, using new terms if anything is non inclusive because no one understands what things mean.

Across fields, there has been a drive to move away from this notation. I've also seen this notation being criticized at conferences.

KristofferC:
> Was just introduced in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1146, hence I considered fine to merge after @lijas approved.

That makes sense, I thought it had been in there longer.

> Across fields, there has been a drive to move away from this notation. I've also seen this notation being criticized at conferences.

There is a close to an arbitrary number of established terms that is possible to squeeze out some non-inclusivity if one really tries, "hanging/orphan/dummy nodes" just to name something. For now, I personally feel that the non-inclusivity of using non-established terms, leading to confusion/unclarity of our documentation, is real while the other type of non-inclusivity is more theoretical.

PR #1160, opened 2025-02-24
Fix wrong usage of shape_value_type

KnutAM:
Want to see test coverage first, and then should add test that catches this (unless I thought something wrong here - could you check @lijas?)

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -933,8 +933,8 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
     return _evaluate_at_grid_nodes(dh, u, fieldname)
 end
 
-_function_value_type(::ScalarInterpolation, ::AbstractVector{T}) where {T} = T
-_function_value_type(::VectorInterpolation{vdim}, ::AbstractVector{T}) where {vdim, T <: Number} = Vec{vdim, T}
+function_value_init(::ScalarInterpolation, ::AbstractVector{T}) where {T} = zero(T)
+function_value_init(::VectorInterpolation{vdim}, ::AbstractVector{T}) where {vdim, T <: Number} = zero(Vec{vdim, T})
 
 # Internal method that have the vtk option to allocate the output differently
 function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk} = Val(false)) where {T, vtk, sdim}
@@ -943,7 +943,7 @@ function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fie
     # Figure out the return type (scalar or vector)
     field_idx = find_field(dh, fieldname)
     ip = getfieldinterpolation(dh, field_idx)
-    RT = _function_value_type(ip, u)
+    RT = typeof(function_value_init(ip, u))
     if vtk
         # VTK output of solution field (or L2 projected scalar data)
         n_c = n_components(ip)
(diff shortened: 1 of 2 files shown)


lijas:
The function shape_value_type and your newly added _function_value_type does basically the same thing, so i dont understand what this fixes?

KnutAM:
The way I understand is that 
* `shape_value_type(ip, T)`: "Return the type of `shape_value(ip::Interpolation, ξ::Vec, ib::Int)`." (from docstring, should be `reference_shape_value` and `ξ::Vec{rdim, T}` AFAIU)
* `_function_value_type`: Returns the type of `sum(i -> reference_shape_value(ip, ξ, i) * u[i], 1:getnbasefunctions(ip))`

So the difference would occur if the eltype of u would be a tensor, because `shape_value_type(ip::ScalarInterpolation, T::Type{<:AbstractTensor})` is not defined, while `function_value_type(ip::ScalarInterpolation, ::AbstractVector{<:AbstractTensor})` is. 

I was surprised so perhaps we are never supposed reach this case in the code here (discovered it when working with #1161 where such a function became useful for the same purpose as used here).

lijas:
Ah I see, `_evaluate_at_grid_nodes` can evaluate with tensors also (not only the solution vector). Yes, then this makes sense.

fredrikekre:
Can `function_value_init` be used?

src/Dofs/DofHandler.jl
```diff
@@ -933,14 +933,17 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
     return _evaluate_at_grid_nodes(dh, u, fieldname)
 end
 
+_function_value_type(::ScalarInterpolation, ::AbstractVector{T}) where {T} = T
+_function_value_type(::VectorInterpolation{vdim}, ::AbstractVector{T}) where {vdim, T <: Number} = Vec{vdim, T}
```
KnutAM, line 937:
```suggestion
function_value_init(::ScalarInterpolation, ::AbstractVector{T}) where {T} = zero(T)
function_value_init(::VectorInterpolation{vdim}, ::AbstractVector{T}) where {vdim, T <: Number} = zero(Vec{vdim, T})
```

src/Dofs/DofHandler.jl
```diff
@@ -933,14 +933,17 @@ function evaluate_at_grid_nodes(dh::DofHandler, u::AbstractVector, fieldname::Sy
...
+
 # Internal method that have the vtk option to allocate the output differently
 function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fieldname::Symbol, ::Val{vtk} = Val(false)) where {T, vtk, sdim}
     # Make sure the field exists
     fieldname ∈ getfieldnames(dh) || error("Field $fieldname not found.")
     # Figure out the return type (scalar or vector)
     field_idx = find_field(dh, fieldname)
     ip = getfieldinterpolation(dh, field_idx)
-    RT = shape_value_type(ip, T)
+    RT = _function_value_type(ip, u)
```
KnutAM, line 946:
```suggestion
    RT = typeof(function_value_init(ip, u))
```

test/test_interpolations.jl
```diff
@@ -52,14 +52,16 @@ function test_interpolation_properties(ip::Interpolation{RefShape, FunOrder}) wh
         # Test A-D
         _test_interpolation_properties(dof_data, refshape_data)
 
-        # Test E: All base functions implemented.
+        # Test E: All base functions implemented and infers correct types
```
KnutAM, line 55:
```suggestion
        # Test E: All base functions implemented and infers correct types.
```

KnutAM (commented):
@fredrikekre's review comments

fredrikekre:
Cherry picked this into #1044 because it touched the same code and tests where overlapping a bit.

Issue #1159, opened 2025-02-23
Vector interpolations in 3 spatial dimensions


gijswl:
Hi all, I hope that opening an issue here is the right way to express my interest & get my question out there.

I see that support for H(curl) and H(div) interpolations is being added into Ferrite right now. I'd like to use these to (try to) implement a magnetoquasistatic FEM (e.g. using an $A$- or $A-\phi$ formulation) in 3D. Currently, however, the vector interpolations are only implemented for two-dimensional reference elements. Some necessary boundary conditions, etc. also seem to be under heavy development by @KnutAM .

When do you expect everything to be stable enough to attempt using Ferrite for this purpose?
I am, unfortunately, not familiar enough with the internals to define my own vector interpolations on `RefTetrahedron`s.

Looking forward, thanks!
Gijs

KnutAM:
Not sure when I will get around to implement more interpolations, is a bit depending on if I will need them myself I guess. But hopefully with the how-to devdocs in the workings (#1140) it should be easier to extend it yourself: https://ferrite-fem.github.io/Ferrite.jl/previews/PR1140/devdocs/interpolations/#devdocs-interpolations

gijswl:
Thanks for the quick reply. If there is sufficient documentation then I'd be more than happy to give it a go myself.
The biggest issue I see so far is that at least one function that must be defined for vector interpolations is not in the devdocs you link to: `get_direction` is used in the Piola mappings, but is not documented. Not sure if there are any others that I might run into.

I'll dive into it if I have some time, and post the progress here.

KnutAM:
> `get_direction` is used in the Piola mappings, but is not documented. Not sure if there are any others that I might run into.

Yes, there might be for sure be, but if you try and hit some just ask (here, or on slack in `# ferrite-fem`) I've added the `get_direction` as a todo in 1140, and if more are missing those should be added.

gijswl:
After some reading and playing around, I managed to get something that at least passes the unit tests: https://github.com/gijswl/Ferrite.jl/commit/321ff6d40a254d5393409bc210e3937589971d16, but I'm not yet convinced everything is correct. Any chance you could check to see if it looks reasonable?

During the process of implementing it, I ran into the following:
 - It was not immediately apparent that the edge ordering on DefElement and in Ferrite's RefTetrahedron were not the same. This was luckily caught by the tests that validate the DOF ordering. 
[...]

PR #1162, opened 2025-02-25
Implementation of Nedelec interpolation on tetrahedral and hexahedral elements

gijswl:
First attempt at implementing a `VectorInterpolation` on a three-dimensional reference element, based on the discussion in #1159 .

@KnutAM I am still unsure of `get_direction`. Right now it passes the continuity test, but I am not completely convinced that this is right. Perhaps I'm misunderstanding what that function is supposed to return.


src/interpolations.jl
```diff
@@ -1942,3 +1942,35 @@ function get_direction(::Nedelec{RefTriangle, 2}, j, cell)
...
+    i == 6 && return Vec(zero(x), -z, y) # DefElement φ0
+    throw(ArgumentError("no shape function $i for interpolation $ip"))
+end
+
+getnbasefunctions(::Nedelec{RefTetrahedron, 1}) = 6
+edgedof_interior_indices(::Nedelec{RefTetrahedron, 1}) = ((1,), (2,), (3,), (4,), (5,), (6,))
+facedof_indices(::Nedelec{RefTetrahedron, 1}) = ((1, 2, 3), (1, 4, 5), (2, 5, 6), (3, 4, 6))
+adjust_dofs_during_distribution(::Nedelec{RefTetrahedron, 1}) = false
+
+function get_direction(::Nedelec{RefTetrahedron, 1}, j, cell)
```
gijswl, line 1973: What is this expected to return?

test/test_interpolations.jl
```diff
@@ -398,7 +399,7 @@ end
             end
         end
 
-        function test_interpolation_functionals(::Val{:Hcurl}, ::Val{2}, ip::Interpolation)
+        function test_interpolation_functionals(::Val{:Hcurl}, ::Union{Val{2}, Val{3}}, ip::Interpolation)
```
gijswl, line 402: Is this a valid generalization, or do we need to implement additional tests for interpolation functionals with `rdim > 2`

src/interpolations.jl
```diff
@@ -1973,4 +1966,36 @@ adjust_dofs_during_distribution(::Nedelec{RefTetrahedron, 1}) = false
...
-end
\ No newline at end of file
+end
+
+
+# RefHexahedron, 1st order Lagrange
+# https://defelement.org/elements/examples/hexahedron-nedelec1-lagrange-1.html
+function reference_shape_value(ip::Nedelec{RefHexahedron, 1}, ξ::Vec{3}, i::Int)
+    x, y, z = ξ
+    nil = zero(x)
```
KnutAM, line 1976: Two alternative ways are, but they are all equivalent. 
```suggestion
function reference_shape_value(ip::Nedelec{RefHexahedron, 1}, ξ::Vec{3, T}, i::Int) where {T}
    x, y, z = ξ
    nil = zero(T)
```

```suggestion
function reference_shape_value(ip::Nedelec{RefHexahedron, 1}, ξ::Vec{3}, i::Int)
    x, y, z = ξ
    nil = zero(eltype(ξ))
```

src/interpolations.jl
```diff
@@ -1973,4 +1966,36 @@ adjust_dofs_during_distribution(::Nedelec{RefTetrahedron, 1}) = false
...
+
+    i == 1 && return Vec(0.125 * (y * z - y - z + 1), nil, nil)
```
KnutAM, line 1978: Ensures that it doesn't convert to `Float64` for `Float32` input. 
```suggestion
    i == 1 && return Vec((y * z - y - z + 1)/8, nil, nil)
```

src/interpolations.jl
KnutAM: It should return `+1` if the edge associated with shape function `j` is positive, and `-1` if negative. A positive edge is defined by the global node number of the first vertex being smaller than that of the second.

KnutAM (commented):
It surprised me that it looks like it passed some tests for the wrong definitions here. Would need to investigate a bit more to ensure that our tests cover sufficiently well.

termi-official:
> 8 dofs associated with the faces

I am not sure if dof distribution in 3D is ready for this.

gijswl:
For now I have no need of higher-order elements, but good to know that there may be a limitation in the dof distribution. Maybe that should also define the scope of this PR:
- [ ] 1st order Nedelec interpolation on `RefTetrahedron` and `RefHexahedron`.
- [ ] 1st order Raviart-Thomas interpolation on `RefTetrahedron` and `RefHexahedron`.
- [ ] Careful review of the interpolation unit test coverage for vector interpolations in 2 and 3 spatial dimensions.

src/interpolations.jl
KnutAM: I don't think we have defined what a positive face is in Ferrite, but maybe the definition that @fredrikekre suggested is already in place in some parts: 
A face is positive if, starting from the lowest node number, the next vertex' node number is smaller than the previous vertex number. 

Here, the vertices are defined by `Ferrite.faces(cell)`, e.g. for `cell =  Tetrahedron((3,5,4,7))`, we have 
`Ferrite.faces(Tetrahedron((3,5,4,2)) = ((3, 4, 5), (3, 5, 2), (5, 4, 2), (3, 2, 4))`

* face 1: `(3, 4, 5)` - negative (3 < 5)
* face 2: `(3, 5, 2)` - positive (3 > 2)
* face 3: `(5, 4, 2)` - negative (5 < 7)
* face 4: `(3, 2, 4)` - negative (3 < 4)

(Please correct me if I'm wrong with the convention @fredrik! ) 

So far, that was not required to use since we didn't have faces as facets for H(div) interpolations.

test/test_interpolations.jl
gijswl: I've made an attempt at validating Raviart-Thomas interpolation on tetrahedral and hexahedral elements, by computing the functional through integration over the faces. `quadgk` is not very well suited for that, so improvement is probably possible, but I couldn't find another method which allows the integration boundaries to depend on an integration variable (for the triangular faces).

src/interpolations.jl
gijswl: That appears to be right. Implementing this reasonsing (for `RaviartThomas{RefTetrahedron,1}` and `RaviartThomas{RefHexahedron,1}`), I get passing continuity tests.
I initially thought we should be able to default all the faces to positive, since the description of `Ferrite.faces` says the tuples it returns are oriented (normals pointing outward).

src/interpolations.jl
```diff
@@ -1839,6 +1839,64 @@ function get_direction(::RaviartThomas{RefTriangle, 2}, j, cell)
...
+end
+
+getnbasefunctions(::RaviartThomas{RefHexahedron, 1}) = 6
+edgedof_indices(ip::RaviartThomas{RefHexahedron, 1}) = edgedof_interior_indices(ip)
+facedof_indices(ip::RaviartThomas{RefHexahedron, 1}) = facedof_interior_indices(ip)
+edgedof_interior_indices(::RaviartThomas{RefHexahedron, 1}) = ((), (), (), (), (), (), (), (), (), (), (), ())
+facedof_interior_indices(::RaviartThomas{RefHexahedron, 1}) = ((1), (2), (3), (4), (5), (6))
+adjust_dofs_during_distribution(::RaviartThomas{RefHexahedron, 1}) = false
+
+function get_direction(::RaviartThomas{RefHexahedron, 1}, j, cell)
```
gijswl, line 1891: This is not very elegant

KnutAM: I think that is quite reasonable, but we could possibly abstract the logic to be more generally applicable. 

Just realized that we have 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/bd65884a06220ce006395ff23a581a5b1651483b/src/Grid/grid.jl#L763-L779

But that gives that "flipped" is positive according to the definition as I gave above - @termi-official, @AbdAlazezAhmed, @fredrikekre - is this intentional or did this use a different definition (essentially reversed)?

IMO we shouldn't use this function directly, but perhaps it would make more sense to do something like, 
```Julia
get_face_direction(cell, facenr) = get_face_direction(faces(cell)[facenr])

get_face_direction(facenodes::Tuple)
    min_idx = argmin(facenodes)
    if min_idx == 1
        positive = surface[2] < surface[end]
    elseif min_idx == length(surface)
        positive = surface[1] < surface[end - 1]
    else
        positive = surface[min_idx + 1] < surface[min_idx - 1]
    end
    return positive ? 1 : -1
end
```
or your code, which is equally good (except that I think `argmin` is cleaner (and potentially faster?) than `findmin`). 
 
Then we should call `get_face_direction` in `OrientationInfo`. 

We could do the same `get_edge_direction` abstraction as well, even if the logic is easier, maybe it makes the code easier to read to have this logic in the same place.

test/test_interpolations.jl
KnutAM: We could also use `FacetValues` I guess, so far I decided to avoid since to make the test purer and avoid using these before they have been tested. 

Will look at the code later this week and give feedback! 

Really nice work so far!

KnutAM:
> Careful review of the interpolation unit test coverage for vector interpolations in 2 and 3 spatial dimensions.

I will try to find time to look more carefully on this, so review might take some time... But really nice work!

gijswl:
> I will try to find time to look more carefully on this, so review might take some time... But really nice work!

Sure, take your time. Much appreciated!

KnutAM:
I tried resetting the direction function to 
```julia
function get_direction(::Nedelec{RefTetrahedron, 1}, j, cell)
     edge = edges(cell)[j]
     if(j == 3) # Edge 3 (v3, v1) is reversed compared to the "normal" definition, which has e[2] > e[1]
         return ifelse(edge[1] > edge[2], 1, -1)
     else
         return ifelse(edge[2] > edge[1], 1, -1)
     end
 end
```
as you had before, but then the continuity tests doesn't pass. This shows that the tests work since the direction is reversed for edge `j`. Did the test pass before? Did you change more regarding this interpolation that could affect this?

test/test_interpolations.jl
```diff
@@ -398,7 +409,34 @@ end
...
+                facet_coords = getindex.((Ferrite.reference_coordinates(ipg),), Ferrite.reference_facets(RefShape)[facetnr])
+                dof_inds = Ferrite.facetdof_interior_indices(ip)[facetnr]
+                ξ(s0, s1) = facet_coords[1] + (facet_coords[2] - facet_coords[1]) * s0 + (facet_coords[3] - facet_coords[1]) * s1
+                weighted_normal = reference_normals(RefShape)[facetnr] * reference_face_area(ip, facetnr)
+                for (facet_shape_nr, shape_nr) in pairs(dof_inds)
+                    moment_fun(s0, s1) = reference_moment(ip, s0, s1, facet_shape_nr)
+                    f(s0, s1) = moment_fun(s0, s1) * (reference_shape_value(ip, ξ(s0, s1), shape_nr) ⋅ weighted_normal)
+
+                    if (length(facet_coords) == 3)
+                        val, _ = quadgk(s0 -> quadgk(s1 -> f(s0, s1), 0, 1 - s0; atol = 1.0e-8)[1], 0, 1; atol = 1.0e-8) # TODO Replace quadgk by more suitable 2D cubature
```
gijswl, line 426: I have quite a nice implementation using HCubature and the Duffy transform (to transform the integral on a triangular domain to a rectangular domain). Will clean it up a bit and commit tomorrow

gijswl: HCubature can also be used to integrate on edges and rectangular faces, so that might replace QuadGK for all these tests?

test/test_interpolations.jl
```diff
@@ -398,7 +409,34 @@ end
...
 
-        function test_interpolation_functionals(::Val{:Hcurl}, ::Val{2}, ip::Interpolation)
+        # 3D, H(div)
+        function test_interpolation_functionals(::Val{:Hdiv}, ::Val{3}, ip::Interpolation)
+            RefShape = getrefshape(ip)
+            ipg = Lagrange{RefShape, 1}()
+            for facetnr in 1:nfacets(RefShape)
+                facet_coords = getindex.((Ferrite.reference_coordinates(ipg),), Ferrite.reference_facets(RefShape)[facetnr])
+                dof_inds = Ferrite.facetdof_interior_indices(ip)[facetnr]
+                ξ(s0, s1) = facet_coords[1] + (facet_coords[2] - facet_coords[1]) * s0 + (facet_coords[3] - facet_coords[1]) * s1
```
KnutAM, line 419:
```suggestion
                ξ(s0, s1) = facet_coords[1] + (facet_coords[2] - facet_coords[1]) * s0 + (facet_coords[end] - facet_coords[1]) * s1
```
Otherwise only true for triangular reference faces.

test/test_interpolations.jl
KnutAM: Yes, I think that would make sense!

test/test_interpolations.jl
```diff
@@ -373,57 +397,231 @@ end
...
+                for (edge_shape_nr, shape_nr) in pairs(dof_inds)
+                    f(s) = reference_moment(ip, s, edge_shape_nr) * (reference_shape_value(ip, ξ(s), shape_nr) ⋅ weighted_normal)
+                    @test integrate_edge(f) ≈ 1 # reference volume of RefLine
                 end
 
                 for shape_nr in 1:getnbasefunctions(ip)
                     shape_nr in dof_inds && continue # Already tested
                     for s in range(0, 1, 5)
-                        @test abs(reference_shape_value(ip, ξ(s), shape_nr) ⋅ weighted_normal) < 1.0e-10
+                        @test abs(reference_shape_value(ip, ξ(Vec(s)), shape_nr) ⋅ weighted_normal) < 1.0e-10
```
gijswl, line 459: Should this check still be here?

test/test_interpolations.jl
```diff
@@ -373,57 +403,214 @@ end
...
+            val, _ = hcubature(s -> f(Vec{2}(s)), [0, 0], [1, 1]; atol = 1.0e-8)
+            return val
+        end
+
+        parameterize_edge(edge_coords, s::Vec{1}) = edge_coords[1] + (edge_coords[2] - edge_coords[1]) * s[1]
+        parameterize_face(face_coords, s::Vec{2}) = face_coords[1] + (face_coords[2] - face_coords[1]) * s[1] + (face_coords[end] - face_coords[1]) * s[2]
+        # parameterize_volume(volume_coords, s::Vec{3}) = # TODO parameterization of volume (Tetrahedron, Hexahedron, Prism, Pyramid)
+
+        face_refshape(::Type{RefTetrahedron}, ::Int) = RefTriangle
+        face_refshape(::Type{RefHexahedron}, ::Int) = RefQuadrilateral
```
gijswl, line 446: Similar to `getfacerefshape` in `test_utils.jl#268-269`

test/test_interpolations.jl
```diff
@@ -373,57 +403,214 @@ end
...
+
+        reference_edge_weight(::Type{RefTriangle}) = 1.0
+        reference_edge_weight(::Type{RefQuadrilateral}) = 0.5
+        reference_face_weight(::Type{RefTriangle}) = 2.0
+        reference_face_weight(::Type{RefQuadrilateral}) = 0.25
```
gijswl, line 451: These weights are needed because the integration area of the parameterization in s and the area returned by `reference_face_area` are not the same (for 2D elements). For the 3D elements this is simply `1/reference_volume`. Perhaps there is a better way to get this consistent?

KnutAM: Yes, this didn't feel right, and realized that the reason is that using these functionals, we wouldn't get continuity. Therefore, I added #1173, which tests this continuity and shows that we need to scale the shape `reference_shape_value`s instead, and use different functionals considering our reference shapes (I guess technically not different functionals, but different edge/face parameterizations).

test/test_interpolations.jl
KnutAM: Why shouldn't it?
Did you add a test for this somewhere else?

KnutAM (commented):
I really like how this is developing @gijswl !

Unfortunately, I didn't get further than this, but as you might have seen, I've opened the related #1172 and #1173 to support the test implementation and coverage of assumptions. 

As this PR is currently growing, and since we try to keep the PRs as small and focused as possible, I didn't want to ask you to add a lot more in this, hence the added PRs.

Hope it is ok that I re-implemented `Nedelec{RefQuadrilateral,1}` in #1173 based on your implementation in this one to enable the testing.

gijswl:
Hi @KnutAM, thanks for all the feedback! I'll have a look at all your comments later this week.
Agreed that this PR might be growing a bit too large. Any proposal for where to draw the line?

KnutAM:
> Any proposal for where to draw the line?

As it is now, I think it is still manageable. But we shouldn't add more interpolations :D 
Part of the growing was my review comments on the `get_direction`, this part could be split out, but not really needed. 

If you want to, please take a look at #1173, probably makes sense to merge that one first, and then it can be used to test all interpolations you've implemented for `RefQuadrilateral` as well.

test/test_interpolations.jl
gijswl: I've updated this in [c83c5bc](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1162/commits/c83c5bc775363b59052939563d2c1d8b281b7ff1) for the triangle and quadrilateral elements. Re-deriving the shape functions for quadrilaterals showed that they were indeed off by a factor of 2. However, I don't see this for the tetrahedral/hexahedral elements. 
Do you see any reason why this integration should or should not be weighted (or conversely, that the shape functions should be weighted accordingly)? Is it desirable to test this by extending #1173 to 3D as well?

KnutAM: I think https://github.com/gijswl/Ferrite.jl/pull/1 should resolve this now. If you are happy with it and merge it, let's continue the discussions from there.

gijswl:
I am happy with the way this has turned out. I think it can be merged if there are no further comments!

KnutAM:
> I am happy with the way this has turned out. I think it can be merged if there are no further comments!

Great! I will try to find time to go carefully through again and check all assumptions. It looks like it will be a busy week (and I shouldn't procrastinate by looking at this one :D). But it does feel like we are now on safe ground with the way the current implementation passes tests.

KnutAM:
Hi @gijswl 
Now I finally got around to keep looking into this. With #1173 merged, could you update this PR by merging with master and add tests for mixed grids in 2d with the Raviart-Thomas elements too?

src/interpolations.jl
```diff
@@ -1852,10 +1874,84 @@ edgedof_interior_indices(::RaviartThomas{RefTriangle, 2}) = ((1, 2), (3, 4), (5,
...
+    shape_nr > 6 && return 1
+    edge_nr = (shape_nr + 1) ÷ 2
+    return get_edge_direction(cell, edge_nr)
+end
+
+# RefQuadrilateral, 1st order Lagrange
+edgedof_indices(ip::RaviartThomas{RefQuadrilateral}) = edgedof_interior_indices(ip)
+facedof_indices(ip::RaviartThomas{RefQuadrilateral}) = (ntuple(i -> i, getnbasefunctions(ip)),)
+
+# https://defelement.org/elements/examples/quadrilateral-raviart-thomas-lagrange-1.html
```
KnutAM, line 1887:
```suggestion
# RefQuadrilateral, 1st order Lagrange
# https://defelement.org/elements/examples/quadrilateral-raviart-thomas-lagrange-1.html
```

src/interpolations.jl
```diff
@@ -1852,10 +1874,84 @@ edgedof_interior_indices(::RaviartThomas{RefTriangle, 2}) = ((1, 2), (3, 4), (5,
...
+    i == 2 && return Vec(2 * x, 2 * (y - 1), 2 * z)
+    i == 3 && return Vec(2 * x, 2 * y, 2 * z)
+    i == 4 && return Vec(2 * (x - 1), 2 * y, 2 * z) # Changed sign, follow Ferrite's sign convention
+    throw(ArgumentError("no shape function $i for interpolation $ip"))
+end
+
+getnbasefunctions(::RaviartThomas{RefTetrahedron, 1}) = 4
+edgedof_interior_indices(::RaviartThomas{RefTetrahedron, 1}) = ntuple(_ -> (), 6)
+edgedof_indices(ip::RaviartThomas{RefTetrahedron, 1}) = edgedof_interior_indices(ip)
+facedof_interior_indices(::RaviartThomas{RefTetrahedron, 1}) = ((1), (2), (3), (4))
```
KnutAM, line 1922: This should have been caught by the tests, will fix in separate pr. 
```suggestion
facedof_interior_indices(::RaviartThomas{RefTetrahedron, 1}) = ((1,), (2,), (3,), (4,))
```

src/interpolations.jl
```diff
@@ -1852,10 +1874,84 @@ edgedof_interior_indices(::RaviartThomas{RefTriangle, 2}) = ((1, 2), (3, 4), (5,
...
+    i == 4 && return Vec(nil, (y + 1) / 8, nil) # Changed sign, follow Ferrite's sign convention
+    i == 5 && return Vec((x - 1) / 8, nil, nil) # Changed sign, follow Ferrite's sign convention
+    i == 6 && return Vec(nil, nil, (z + 1) / 8)
+    throw(ArgumentError("no shape function $i for interpolation $ip"))
+end
+
+getnbasefunctions(::RaviartThomas{RefHexahedron, 1}) = 6
+edgedof_interior_indices(::RaviartThomas{RefHexahedron, 1}) = ntuple(_ -> (), 12)
+edgedof_indices(ip::RaviartThomas{RefHexahedron, 1}) = edgedof_interior_indices(ip)
+facedof_interior_indices(::RaviartThomas{RefHexahedron, 1}) = ((1), (2), (3), (4), (5), (6))
```
KnutAM, line 1949:
```suggestion
facedof_interior_indices(::RaviartThomas{RefHexahedron, 1}) = ((1,), (2,), (3,), (4,), (5,), (6,))
```

PR #1164, opened 2025-02-28
Upgrade runic-pre-commit to v2.0.1

fredrikekre:


KnutAM:
On the surface this looks good to me, I assume this implies that the correct version of runic gets automatically installed? Or is manual creation of the runic environment for the correct version required? 

I can check that it works correctly next week if you want of course, but if not, this is good to go from my perspective. Great that this got fixed so fast!

PR #1169, opened 2025-03-06
Fix addindex!(A, ...) to allow adding to an index not in the sparsity pattern of A

alex-fu27:
In discussion #1167, I encountered the problem that `addindex!(A, v, i, j)` won't add to an index `(i,j)` if it isn't in the sparsity pattern of `A`. In this case, I suggest falling back to assignment: Falling back to `A[i,j] += v` would work, if `(i,j)` was not in the sparsity pattern in the first place, then `A[i,j]` must have been zero, so we can use `A[i,j] = v` instead.

### DIFF
--- a/src/arrayutils.jl
+++ b/src/arrayutils.jl
@@ -17,10 +17,9 @@ Equivalent to `A[i, j] += v` but more efficient.
 memory location for index `(i, j)` -- one time for the read, and one time for the write.
 This method avoids the double lookup.
 
-Zeros are ignored (i.e. if `iszero(v)`) by returning early. If the index `(i, j)` is not
-existing in the sparsity pattern of `A` this method throws a `SparsityError`.
-
-Fallback: `A[i, j] += v`.
+Zeros are ignored (i.e. if `iszero(v)`) by returning early.
+If the index `(i, j)` is not in the sparsity pattern of `A`, the entry must be zero and will be set to `v`.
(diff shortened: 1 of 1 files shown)


termi-official:
Sorry I do not fully understand the problem which you are trying to solve here. What is the problem with using https://ferrite-fem.github.io/Ferrite.jl/stable/reference/sparsity_pattern/#Ferrite.add_entry! here?

alex-fu27:
Basically, if your assembly generates a `SparseMatrixCSC` that has zeroes in it, and you do an operation, for example `A = M .+ dt .* K` for a transient problem, then this operation will remove the zero-valued entries from the `SparseMatrixCSC`'s sparsity pattern. If you then try to `apply!(::SparseMatrixCSC, ::ConstraintHandler)`, this will fail because `apply!` will try to modify those zero-valued entries through `addindex!`, which didn't allow that.

fredrikekre:
Thanks for the PR, but it is actually a feature that `addindex!` doesn't allow this because it is a very expensive operation and becomes a bit of a performance footgun.

It is unfortunate that broadcasting changes the sparsity pattern. It would have been nicer if the pattern of A was the union of the patterns of K and M in this example but probably that can't be changed. I have experimented a bit with this in https://github.com/fredrikekre/SpartanMatrices.jl where broadcasting is only allowed between matrices with identical patterns.

Issue #1110, opened 2024-11-13
`create_boundary....set` is broken for mixed grids


AbdAlazezAhmed:
Ran into it while testing #1108 , probably https://github.com/Ferrite-FEM/Ferrite.jl/blob/a38ed022dbc7914b344920698ddab7debf7f1e57/src/Grid/utils.jl#L168 will solve it

MWE:
```Julia
 using Ferrite

nodes = [Node((-1.0, 0.0)), Node((0.0, 0.0)), Node((1.0, 0.0)), Node((-1.0, 1.0)), Node((0.0, 1.0))];

cells = [
                   Quadrilateral((1, 2, 5, 4)),
                   Triangle((3, 5, 2)),
               ];

grid = Grid(cells, nodes);

topology = ExclusiveTopology(grid);

 filter_function(x) = x[1] > 0;

addboundaryvertexset!(grid, topology, "test_boundary_vertexset", filter_function);
```
```julia
ERROR: BoundsError: attempt to access Tuple{Tuple{Int64, Int64}, Tuple{Int64, Int64}, Tuple{Int64, Int64}} at index [4]
Stacktrace:
 [1] getindex(t::Tuple, i::Int64)
   @ Base ./tuple.jl:31
 [2] (::Ferrite.var"#_makeset#174"{…})(ff_nh::Ferrite.CollectionsOfViews.ArrayOfVectorViews{…})
   @ Ferrite ~/.julia/dev/Ferrite/src/Grid/utils.jl:168
 [3] _create_boundaryset(f::typeof(filter_function), grid::Grid{…}, top::ExclusiveTopology, ::Type{…}; all::Bool)
   @ Ferrite ~/.julia/dev/Ferrite/src/Grid/utils.jl:182
 [4] _create_boundaryset(f::Function, grid::Grid{…}, top::ExclusiveTopology, ::Type{…})
   @ Ferrite ~/.julia/dev/Ferrite/src/Grid/utils.jl:158
 [5] create_boundaryvertexset
   @ ~/.julia/dev/Ferrite/src/Grid/utils.jl:210 [inlined]
 [6] addboundaryvertexset!(grid::Grid{…}, top::ExclusiveTopology, name::String, f::Function; kwargs::@Kwargs{})
   @ Ferrite ~/.julia/dev/Ferrite/src/Grid/utils.jl:100
[...]

PR #1108, opened 2024-11-07
[WIP] Subdomain support for `InterfaceIterator`

AbdAlazezAhmed:
Also a bit better performance
<details>
<summary>some random benchmark:</summary>

```julia
julia> function foo(dh, top)
           for i in InterfaceIterator(dh, top)
           end
           end
foo (generic function with 1 method)

julia> using BenchmarkTools

julia> grid = generate_grid(Hexahedron, (20,20,20))
Grid{3, Hexahedron, Float64} with 8000 Hexahedron cells and 9261 nodes

julia> topology = ExclusiveTopology(grid);

julia> ip = DiscontinuousLagrange{RefHexahedron, 1}();

julia> dh = DofHandler(grid);

julia> add!(dh, :u, ip);

julia> close!(dh);

julia<050f226>> @benchmark foo($dh, $topology)
BenchmarkTools.Trial: 2131 samples with 1 evaluation.
 Range (min … max):  1.960 ms …  28.916 ms  ┊ GC (min … max): 0.00% … 89.13%
 Time  (median):     2.151 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.341 ms ± 739.360 μs  ┊ GC (mean ± σ):  1.66% ±  4.48%

   ▅▆▆█▅▄▃▁▁ ▂▃▄▃▅▅▄▃▃▃▂▂▁▁▁▁                                 ▁
  ▇█████████▇█████████████████▆▇▅▇▅▆▅▅▆▆▇▇█▆██▇▅▅▅▅▁▁▁▄▄▄▆▆▄▅ █
  1.96 ms      Histogram: log(frequency) by time      3.74 ms <

 Memory estimate: 1.86 MiB, allocs estimate: 42.

julia<master>> @benchmark foo($dh, $topology)
BenchmarkTools.Trial: 1605 samples with 1 evaluation.
 Range (min … max):  2.567 ms … 66.080 ms  ┊ GC (min … max): 0.00% … 93.99%
 Time  (median):     2.784 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.109 ms ±  1.686 ms  ┊ GC (mean ± σ):  4.78% ±  9.39%

  ▇█▇▆▆▄▁   ▃▅▅▅▄▄▃▃▁ ▁                  ▁▂▁                 ▁
[...]

PR #606, opened 2023-02-25
Extract boundary entities + getfaceedges helper

AbdAlazezAhmed:
fixes #595 (`addboundaryfaceset!`, `addboundaryedgeset!`, `addboundaryvertexset!` ) and adds helper functions to get face/edge edges/vertices (`getfaceedges`, `getfacevertices`, `getedgevertices`).

PR #1176, opened 2025-03-25
`addboundary*set()` errors for mixed grids

AbdAlazezAhmed:
It seems like it's iterating over neighborhoods of non-existing facets. Checking against `length(facets(cell))` should solve the issue.

### DIFF
--- a/src/Grid/utils.jl
+++ b/src/Grid/utils.jl
@@ -165,6 +165,7 @@ function _create_boundaryset(f::Function, grid::AbstractGrid, top::ExclusiveTopo
             cell_idx = ff_nh_idx[1]
             facet_nr = ff_nh_idx[2]
             cell = getcells(grid, cell_idx)
+            facet_nr > length(facets(cell)) && continue
             facet_nodes = facets(cell)[facet_nr]
             for (subentity_idx, subentity_nodes) in pairs(boundaryfunction(BI)(cell))
                 if Base.all(n -> n in facet_nodes, subentity_nodes)
--- a/src/Grid/utils.jl
+++ b/src/Grid/utils.jl
@@ -165,6 +165,8 @@ function _create_boundaryset(f::Function, grid::AbstractGrid, top::ExclusiveTopo
             cell_idx = ff_nh_idx[1]
             facet_nr = ff_nh_idx[2]
             cell = getcells(grid, cell_idx)
+            # `ff_nh::AbstractMatrix`, indexed by (cell_idx, facet_nr). For mixed grids,
+            # `facet_nr` can then become too large and should be skipped as the facet doesn't exist.
             facet_nr > length(facets(cell)) && continue
             facet_nodes = facets(cell)[facet_nr]
             for (subentity_idx, subentity_nodes) in pairs(boundaryfunction(BI)(cell))
--- a/test/test_grid_addboundaryset.jl
+++ b/test/test_grid_addboundaryset.jl
@@ -236,7 +236,7 @@
         @test getfacetset(grid, "test_boundary_facetset") == Ferrite.create_boundaryfacetset(grid, topology, filter_function)
     end
 
-    @testset "mixed grid" begin
+    @testset "mixed grid 3D" begin
         nodes = reshape([Node(Vec(x, y, z)) for x in -1:1, y in -1:1, z in 0:1], :)
         for (shape1, shape2) in ((Hexahedron, Wedge), (Wedge, Hexahedron))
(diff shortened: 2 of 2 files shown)


test/test_grid_addboundaryset.jl
```diff
@@ -61,6 +61,41 @@
...
+    function boundary_facets(::Type{Wedge}, ::Type{Hexahedron})
+        return Set(
+            [
+                FacetIndex(1, 1), FacetIndex(1, 2), FacetIndex(1, 4), FacetIndex(1, 5),
+                FacetIndex(2, 1), FacetIndex(2, 2), FacetIndex(2, 3), FacetIndex(2, 4), FacetIndex(2, 6),
+            ]
+        )
+    end
+
     #=
```
AbdAlazezAhmed, line 99: Should we do something regarding these tests? iirc the functions were removed.

test/test_grid_addboundaryset.jl
```diff
@@ -200,4 +235,14 @@
         addboundaryfacetset!(grid, topology, "test_boundary_facetset", filter_function)
         @test getfacetset(grid, "test_boundary_facetset") == Ferrite.create_boundaryfacetset(grid, topology, filter_function)
```
AbdAlazezAhmed, line 236: Also, are these are testing for `_addset!()` or something else?

KnutAM:
This is the fix from https://github.com/Ferrite-FEM/Ferrite.jl/pull/1108#discussion_r1841047951 to solve #1110 right? 

Just to be clear, I think it makes sense to fix this as a separate issue as done here - will take a look as soon as I find time! Thanks for fixing it!

test/test_grid_addboundaryset.jl
KnutAM: The first of these tests were introduced in #606 to test internal methods that supported the boundary element extraction, right? After [#789/#914](https://github.com/Ferrite-FEM/Ferrite.jl/commit/2ca4f941f76c2d350d485c337f1d892ab25de312#diff-13ae190cb008d749905405bddca2a9ce395b303e12c781186df0d908efd05a90) those internal methods are not used. So from my perspective they could be removed, at least nobody complained that those internal methods where removed.
But if they should be re-created, I guess we could keep them. In any case, worth a separate issue (or pr to remove tests or re-add functions if they have a use case IMO). 

Since the edgesets are now removed, I think the others could be removed in another pr

test/test_grid_addboundaryset.jl
KnutAM: Yes, since to make sure the sets are actually added. I guess a comment would have made sense :)

test/test_grid_addboundaryset.jl
```diff
@@ -200,4 +235,25 @@
...
     end
+
+    @testset "mixed grid 3D" begin
+        nodes = reshape([Node(Vec(x, y, z)) for x in -1:1, y in -1:1, z in 0:1], :)
+        for (shape1, shape2) in ((Hexahedron, Wedge), (Wedge, Hexahedron))
+            grid = Grid([generate_cell1(shape1), generate_cell2(shape2)], nodes)
+            topology = ExclusiveTopology(grid)
+            addboundaryfacetset!(grid, topology, "boundary", _ -> true)
+            @test getfacetset(grid, "boundary") == boundary_facets(shape1, shape2)
```
KnutAM, line 245: Is there a reason why we need to do this both combinations in 3D? Seems like explicitly creating the two 3D cells like for 2D would be cleaner than relying on the `generate_cell1` and `generate_cell2` methods, as well as the `boundary_facets`?

src/Grid/utils.jl
```diff
@@ -165,6 +165,7 @@ function _create_boundaryset(f::Function, grid::AbstractGrid, top::ExclusiveTopo
             cell_idx = ff_nh_idx[1]
             facet_nr = ff_nh_idx[2]
             cell = getcells(grid, cell_idx)
+            facet_nr > length(facets(cell)) && continue
```
KnutAM, line 168: We should probably add an explanation, something like
```suggestion
            # `ff_nh::AbstractMatrix`, indexed by (cell_idx, facet_nr). For mixed grids, 
            # `facet_nr` can then become to large and should be skipped as it doesn't exist. 
            facet_nr > length(facets(cell)) && continue
```
?

KnutAM (commented):
Some suggestions

KnutAM (approved):
LGTM, just remove added line.

PR #1177, opened 2025-03-25
Allow for views in `write_solution`

AbdAlazezAhmed:
`write_solution` currently requires vectors. This can be a bit annoying when having say a preallocated solution matrix with (ntstops * ndofs) and one wants to write the solution in a callback using a view.

This PR just allows for `AbstractVector`s and has some test to cover passing views.


termi-official (approved):

test/test_grid_dofhandler_vtk.jl
```diff
@@ -163,6 +163,32 @@ close(csio)
...
 end
 
+@testset "vtk export view" begin
+    grid = generate_grid(Hexahedron, (5, 5, 5))
+    dofhandler = DofHandler(grid)
+    ip = geometric_interpolation(Hexahedron)
+    add!(dofhandler, :temperature, ip)
+    add!(dofhandler, :displacement, ip^3)
+    close!(dofhandler)
+    u = zeros(ndofs(dofhandler))
```
KnutAM, line 173:
```suggestion
    u = rand(ndofs(dofhandler))
```

KnutAM (approved):
Since the new test is independent of testing grid/dofhandler, should we maybe move to https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/test_vtk_export.jl?

PR #1179, opened 2025-03-31
Update DefElement links

mscroggs:
Following https://github.com/DefElement/DefElement/pull/393, DefElement uses the [polynomial subdegree](https://defelement.org/ciarlet.html#The+degree+of+a+finite+element) to index the majority of its elements, rather than the [Lagrange superdegree](https://defelement.org/ciarlet.html#The+degree+of+a+finite+element).

This means that the links for a few examples have changed. This PR corrects those links


fredrikekre:
Thank you!

PR #1182, opened 2025-04-01
Fix PointEvalHandler on true subdomains

### LINKED ISSUE #1181: PointEvalHandler silently fails when evaluating at the boundary of true subdomains
mwe

```julia
using Ferrite

grid = generate_grid(Triangle, (4, 4), Vec((-2.0,-2.0)), Vec((2.0,2.0)))
addcellset!(grid, "hole", x->norm(x) ≤ 1.0)
dh = DofHandler(grid)
sdh = SubDofHandler(dh, getcellset(target_mesh_nonmatching, "hole"))
add!(sdh, :v, Lagrange{RefTriangle,1}())
close!(dh)

nodes = Vec{2, Float64}.([[0.0, -1.0], [0.0, 0.0], [-1.0, 0.0], [1.0, 0.0], [0.0, 1.0]])
ph = PointEvalHandler(grid, nodes; warn=true)
result = Ferrite.evaluate_at_points(ph, dh, zeros(ndofs(dh)))
#5-element Vector{Float64}:
# NaN
# 0.0
# NaN
# NaN
# NaN

VTKGridFile("peh-subdomain-mwe", grid) do vtk
    Ferrite.write_cellset(vtk, grid)
    Ferrite.write_solution(vtk, dh, result)
end
```

termi-official:
Fixes #1181 .

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -108,6 +108,14 @@ mutable struct DofHandler{dim, G <: AbstractGrid{dim}} <: AbstractDofHandler
     ndofs::Int
 end
 
+function is_defined_on_full_domain(dh::DofHandler)
+    num_cells = 0
+    for sdh in dh.subdofhandlers
+        num_cells += length(sdh.cellset)
+    end
+    return getncells(get_grid(dh)) == num_cells
+end
+
 """
     DofHandler(grid::Grid)
 
--- a/src/PointEvalHandler.jl
+++ b/src/PointEvalHandler.jl
@@ -5,7 +5,7 @@ Base.@kwdef struct NewtonLineSearchPointFinder{T}
 end
 
 """
-    PointEvalHandler(grid::Grid, points::AbstractVector{Vec{dim,T}}; kwargs...) where {dim, T}
+    PointEvalHandler(grid::Grid, points::AbstractVector{Vec{dim,T}} [, set = subdomains::Vector{OrderedSet{Int}}] ; kwargs...) where {dim, T}
 
 The `PointEvalHandler` can be used for function evaluation in *arbitrary points* in the
 domain -- not just in quadrature points or nodes.
@@ -33,6 +33,7 @@ struct PointEvalHandler{G, T <: Real}
     grid::G
     cells::Vector{Union{Nothing, Int}}
     local_coords::Vector{Union{Nothing, Vec{1, T}, Vec{2, T}, Vec{3, T}}}
+    constructed_on_subdomain::Bool
 end
 
 function Base.show(io::IO, ::MIME"text/plain", ph::PointEvalHandler)
@@ -58,7 +59,13 @@ end
 function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim, T}}; search_nneighbors = 3, warn::Bool = true, strategy = NewtonLineSearchPointFinder()) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
     cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn, strategy)
-    return PointEvalHandler(grid, cells, local_coords)
+    return PointEvalHandler(grid, cells, local_coords, false)
+end
+
+function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim, T}}, subdomains::Vector{OrderedSet{Int}}; search_nneighbors = 3, warn::Bool = true, strategy = NewtonLineSearchPointFinder()) where {dim, T}
+    node_cell_dicts = _get_node_cell_map(grid, subdomains)
+    cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn, strategy)
+    return PointEvalHandler(grid, cells, local_coords, true)
 end
 
 function _get_cellcoords(points::AbstractVector{Vec{dim, T}}, grid::AbstractGrid, node_cell_dicts::Dict{C, Dict{Int, Vector{Int}}}, search_nneighbors, warn, strategy::NewtonLineSearchPointFinder) where {dim, T <: Real, C}
@@ -202,6 +209,31 @@ function _get_node_cell_map(grid::AbstractGrid)
     return cell_dicts
 end
 
+function _get_node_cell_map(grid::AbstractGrid, subdomains::Vector{OrderedSet{Int}})
+    @inline function _check_contained_subdomain(i, subdomains)
+        for subdomain in subdomains
+            i ∈ subdomain && return true
+        end
+        return false
+    end
+    cells = getcells(grid)
+    C = eltype(cells) # possibly abstract
+    cell_dicts = Dict{Type{<:C}, Dict{Int, Vector{Int}}}()
+    ctypes = Set{Type{<:C}}(typeof(c) for c in cells)
+    for ctype in ctypes
+        cell_dict = cell_dicts[ctype] = Dict{Int, Vector{Int}}()
+        for (cellidx, cell) in enumerate(cells)
+            cell isa ctype || continue
+            _check_contained_subdomain(cellidx, subdomains) || continue
+            for node in cell.nodes
+                v = get!(Vector{Int}, cell_dict, node)
+                push!(v, cellidx)
+            end
+        end
+    end
+    return cell_dicts
+end
+
 """
     evaluate_at_points(ph::PointEvalHandler, dh::AbstractDofHandler, dof_values::Vector{T}, [fieldname::Symbol]) where T
     evaluate_at_points(ph::PointEvalHandler, proj::L2Projector, dof_values::Vector{T}) where T
@@ -226,6 +258,10 @@ function evaluate_at_points(
         ph::PointEvalHandler{<:Any, T1}, dh::AbstractDofHandler, dof_vals::AbstractVector{T2},
         fname::Symbol = find_single_field(dh)
     ) where {T1, T2}
+    if !is_defined_on_full_domain(dh) && !ph.constructed_on_subdomain
+        # @warn "DofHandler is defined on a true subdomain only, but the PointEvalHandler serached cells on the full grid. The evaluation assignment might be broken!" maxlog=1
+        error("!!!")
+    end
     npoints = length(ph.cells)
     # Figure out the value type by creating a dummy PointValues
     ip = getfieldinterpolation(dh, find_field(dh, fname))
--- a/src/PointEvalHandler.jl
+++ b/src/PointEvalHandler.jl
@@ -259,8 +259,7 @@ function evaluate_at_points(
         fname::Symbol = find_single_field(dh)
     ) where {T1, T2}
     if !is_defined_on_full_domain(dh) && !ph.constructed_on_subdomain
-        # @warn "DofHandler is defined on a true subdomain only, but the PointEvalHandler serached cells on the full grid. The evaluation assignment might be broken!" maxlog=1
-        error("!!!")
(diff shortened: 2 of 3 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -108,6 +108,14 @@ mutable struct DofHandler{dim, G <: AbstractGrid{dim}} <: AbstractDofHandler
...
 end
 
+function is_defined_on_full_domain(dh::DofHandler)
+    num_cells = 0
+    for sdh in dh.subdofhandlers
+        num_cells += length(sdh.cellset)
+    end
+    return getncells(get_grid(dh)) == num_cells
+end
+
```
kimauth, line 118: Wouldn't you have the same problem with a field that's defined on a subdomain, even if all cells are part of some `SubDofHandler`? In that case this check isn't sufficient, is it?

src/PointEvalHandler.jl
```diff
@@ -202,6 +209,31 @@ function _get_node_cell_map(grid::AbstractGrid)
     return cell_dicts
 end
 
+function _get_node_cell_map(grid::AbstractGrid, subdomains::Vector{OrderedSet{Int}})
```
kimauth, line 212: Do we need this to be its own function or could we just modify the original `_get_node_cell_map` to accept an optional keyword argument `cellset` that defaults to all cells? And why does `subdomains` need to be a vector here?

src/PointEvalHandler.jl
```diff
@@ -58,7 +59,13 @@ end
 function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim, T}}; search_nneighbors = 3, warn::Bool = true, strategy = NewtonLineSearchPointFinder()) where {dim, T}
     node_cell_dicts = _get_node_cell_map(grid)
     cells, local_coords = _get_cellcoords(points, grid, node_cell_dicts, search_nneighbors, warn, strategy)
-    return PointEvalHandler(grid, cells, local_coords)
+    return PointEvalHandler(grid, cells, local_coords, false)
+end
+
+function PointEvalHandler(grid::AbstractGrid{dim}, points::AbstractVector{Vec{dim, T}}, subdomains::Vector{OrderedSet{Int}}; search_nneighbors = 3, warn::Bool = true, strategy = NewtonLineSearchPointFinder()) where {dim, T}
```
kimauth, line 65: Would be nice if we could have a keword argument `cellset` here instead of `subdomains`, see my comment below.

kimauth (commented):
Thanks! I left a few questions, perhaps we can get a slightly nicer user interface 🙂

src/Dofs/DofHandler.jl
termi-official: Good catch, you are right. Let me fix that.

src/PointEvalHandler.jl
termi-official: > Do we need this to be its own function or could we just modify the original _get_node_cell_map to accept an optional keyword argument cellset that defaults to all cells?

I simply prefer to use multiple dispatch over kwargs/if-else constructs using isa, because it is easier to extend later when we add new features and feels cleaner.

> why does subdomains need to be a vector here?

The field to which the point is associated might span multiple subdomains (e.g. we have the cellsets named "A", "B" and "C", then we might want to search only on "B" and "C"). Think e.g. about discretizing the torso of some kind of animal, then each organ is potentially at least one subdomain. Does that make sense?

src/PointEvalHandler.jl
```diff
@@ -202,6 +209,31 @@ function _get_node_cell_map(grid::AbstractGrid)
...
+    end
+    cells = getcells(grid)
+    C = eltype(cells) # possibly abstract
+    cell_dicts = Dict{Type{<:C}, Dict{Int, Vector{Int}}}()
+    ctypes = Set{Type{<:C}}(typeof(c) for c in cells)
+    for ctype in ctypes
+        cell_dict = cell_dicts[ctype] = Dict{Int, Vector{Int}}()
+        for (cellidx, cell) in enumerate(cells)
+            cell isa ctype || continue
+            _check_contained_subdomain(cellidx, subdomains) || continue
```
KnutAM, line 227: Isn't this enough, or was a performance optimization required?
```suggestion
            any(set -> cellidx in set, subdomains) || continue
```

But if following other comments, would of course also remove the `any`...

src/PointEvalHandler.jl
```diff
@@ -202,6 +209,31 @@ function _get_node_cell_map(grid::AbstractGrid)
     return cell_dicts
 end
 
+function _get_node_cell_map(grid::AbstractGrid, subdomains::Vector{OrderedSet{Int}})
+    @inline function _check_contained_subdomain(i, subdomains)
+        for subdomain in subdomains
+            i ∈ subdomain && return true
+        end
+        return false
+    end
```
KnutAM, line 218:
```suggestion
```

src/PointEvalHandler.jl
KnutAM: Is the performance advantage sufficient to motivate the difference to other interfaces? E.g. that we don't allow users to pass a vector of facetsets to `Dirichlet`, but require their union. IMO the same is reasonable here, but we should allow an `Ferrite.AbstractVecOrSet{<:Integer}{<:Integer}`, allow defaulting to `1:getncells(grid)`. Then it would be the same function in both cases. I don't think a kwarg is required here since this is internal.

src/PointEvalHandler.jl
```diff
@@ -226,6 +258,9 @@ function evaluate_at_points(
         ph::PointEvalHandler{<:Any, T1}, dh::AbstractDofHandler, dof_vals::AbstractVector{T2},
         fname::Symbol = find_single_field(dh)
     ) where {T1, T2}
+    if !is_defined_on_full_domain(dh) && !ph.constructed_on_subdomain
+        @warn "DofHandler is defined on a true subdomain only, but the PointEvalHandler serached cells on the full grid. The evaluation assignment might be broken!" maxlog = 1
```
KnutAM, line 262:
```suggestion
        @warn "DofHandler is defined on a true subdomain only, but the PointEvalHandler searched cells on the full grid. The evaluation assignment might be broken!" maxlog = 1
```

src/PointEvalHandler.jl
KnutAM: I agree that a kwarg makes the user-experience better and we don't lock in on arg ordering. 

Even if an arg defaulting to `1:getncells(grid)` is alright from the convention that args say what and kwargs say how.

test/test_pointevaluation.jl
```diff
@@ -1,4 +1,4 @@
-using Ferrite, Test
+using Ferrite, Test, Logging, OrderedCollections
```
KnutAM, line 1: Why is this even here, I don't think we do this consistently?

KnutAM (commented):
Nice to get this fixed 🚀

src/PointEvalHandler.jl
termi-official: >Is the performance advantage sufficient to motivate the difference to other interfaces?

Sorry, but that does not make sense for me. First, I do not see what this has to do with performance advantages. Second, we already use Vector internally in `PointEvalHandler` (see https://github.com/Ferrite-FEM/Ferrite.jl/blob/6a14d7a2b9e4cbddcd43fc837defaa07fc4e123d/src/PointEvalHandler.jl#L238) and I just sticked to that design. Third, it does not make sense to have users providing the exact split subdomains in one part of the API (DofHandler), just merge these subdomains into a single set again to call into the PointEvalHandler.

All other handlers provide `add!` and `close!` and `PointEvalHandler` does not on the user-facing side. I might need to change the design of the PointEvalHandler interface to match the other interfaces in an attempt to resolve another bug Kim found https://github.com/Ferrite-FEM/Ferrite.jl/pull/1182/files/f4ea8b1a2aa2dc61d1beed2c0e60f91e38f07222#r2027088188 in this branch.

> We don't allow users to pass a vector of facetsets to Dirichlet, but require their union.

I am not sure if that is true. The following code works without problem, right?
```julia
ch = ...
add!(ch, Dirichlet(:v, ∂Ω1, x -> sin(x[1]))
add!(ch, Dirichlet(:v, ∂Ω2, x -> sin(x[1]))
close!(ch)
```
Or is there something I am missing that requires us to use:
```julia
ch = ...
add!(ch, Dirichlet(:v, union(∂Ω1, ∂Ω2) x -> sin(x[1]))
close!(ch)
```
? If that is a problem, can we fix it without merging these sets?

termi-official: I would even argue for simplicity here to just have some second constructor here `PointEvalHandler(dh::DofHandler, points::AbstractVector{Vec{dim,T}}, field_name::Symbol; ... )` which takes care of extracting the right cellsets from the DofHandler.

test/test_pointevaluation.jl
termi-official: I am not sure why these are missing from the test. Running only parts of the tests is right now quite painful, as we include some of the deps in the runner and sometime some of the deps in the file containing the testsets. Would you agree that it would be better if the different files containing the test sections would be self-contained, so we can run them directly for testing instead of modifying the test codes every time we only want to test part of the API?

src/PointEvalHandler.jl
KnutAM: > I am not sure if that is true. The following code works without problem, right?
> ```julia
> ch = ...
> add!(ch, Dirichlet(:v, ∂Ω1, x -> sin(x[1]))
> add!(ch, Dirichlet(:v, ∂Ω2, x -> sin(x[1]))
> close!(ch)
> ```

Sorry, maybe I misunderstood, but I thought you are proposing supporting the equivalent of
```julia
add!(ch, Dirichlet(:v, [∂Ω1, ∂Ω2], x -> sin(x[1]))
```

> First, I do not see what this has to do with performance advantages.

The only one I could see was avoiding creating a new set, which should be rather cheap in most cases. 

> Second, we already use Vector internally in PointEvalHandler

I don't follow this - there is no cellset/subdomain/subdomains internally now, right? 
I meant that instead of restricting the domains to be described by `OrderedSet`, we could also allow e.g. a range - but you are right that a `Vector` would be slow, so perhaps `cellset::Union{AbstractSet{<:Integer}, UnitRange{<:Integer}}` is better! 

> I might need to change the design of the PointEvalHandler interface to match the other interfaces

That sounds like a good idea to me - I just didn't think the passing of a Vector of sets was aligned with such an interface, the ability to `add!` would definitively be more Ferrite-like!

I also like the short-hand version with the fieldname :)

test/test_pointevaluation.jl
KnutAM: I totally agree that this can be very useful, I was just not aware that we have this as a convention to do! 

Would of course then be nicer if it would error if one doesn't do that, otherwise it is quite hard to make sure it is done consistently. This also includes e.g. including `test_utils.jl`. Perhaps it could rather make sense to have a file `initialize_tests.jl` that is included at the beginning of `runtests.jl`, and running individual tests can then always be done by including that file before the specific file (or snippets of code)? Of course a separate discussion...

termi-official: Yea we should probably follow something like this here https://github.com/SciML/OrdinaryDiffEq.jl/blob/1804a1bbc436c88580310316e69ff5a091f65eb1/test/runtests.jl , especially if we start expanding convergence rate tests and GPU test coverage.

src/PointEvalHandler.jl
termi-official: > I don't follow this - there is no cellset/subdomain/subdomains internally now, right?

Not directly, which is part of the problem. See https://github.com/Ferrite-FEM/Ferrite.jl/blob/6a14d7a2b9e4cbddcd43fc837defaa07fc4e123d/src/PointEvalHandler.jl#L237-L238

I will try to bring this PR over the finish line early next week.

termi-official:
This PR was a bit of a dead end. #1434 replaces this PR.

PR #1183, opened 2025-04-08
Allow for views (`AbstractVector`) in `evaluate_at_points!`

AbdAlazezAhmed:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/87b0888a542aaf9d24f36bb82cc1e9b85949e4dc/src/PointEvalHandler.jl#L225-L228
Accepts views, however,
https://github.com/Ferrite-FEM/Ferrite.jl/blob/87b0888a542aaf9d24f36bb82cc1e9b85949e4dc/src/PointEvalHandler.jl#L250-L257
does not.

### DIFF
--- a/test/test_pointevaluation.jl
+++ b/test/test_pointevaluation.jl
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
     return
 end
 
+function test_pe_views()
+    mesh = generate_grid(Quadrilateral, (3, 3))
+    perturb_standard_grid!(mesh, 1 / 10)
+    dof_vals = 1.0 * [0, 0, 1, 2, 6, 5, 3, 7, 4, 8, 10, 9, 11, 12, 14, 13, 15, 16, 0, 0]
+    points = [node.x for node in mesh.nodes[[6, 7, 11, 10]]] # same as nodes
+    vals_to = zeros(length(points) + 4)
+
+    dh = DofHandler(mesh)
+    add!(dh, :s, Lagrange{RefQuadrilateral, 1}()) # a scalar field
+    close!(dh)
+
+    ph = PointEvalHandler(mesh, points)
+
+    ip = Ferrite.get_func_interpolations(dh, :s)
+    Ferrite.evaluate_at_points!((@view vals_to[3:(end - 2)]), ph, dh, (@view dof_vals[3:(end - 2)]), :s, ip)
+    @test vals_to ≈ 1.0 * [0, 0, 6, 7, 11, 10, 0, 0]
+
+    return nothing
+end
+
 function _pointeval_dofhandler2_manual_projection(dh, csv, cvv, f_s, f_v)
     M = allocate_matrix(dh)
     f = zeros(ndofs(dh))
@@ -441,6 +461,10 @@ end
         test_pe_dofhandler()
         test_pe_dofhandler2(; three_dimensional = false)
         test_pe_dofhandler2(; three_dimensional = true)
+    end
+
+    @testset "inplace with views" begin
+        test_pe_views()
     end
 
     @testset "superparametric" begin
--- a/src/PointEvalHandler.jl
+++ b/src/PointEvalHandler.jl
@@ -203,8 +203,8 @@ function _get_node_cell_map(grid::AbstractGrid)
 end
 
 """
-    evaluate_at_points(ph::PointEvalHandler, dh::AbstractDofHandler, dof_values::Vector{T}, [fieldname::Symbol]) where T
-    evaluate_at_points(ph::PointEvalHandler, proj::L2Projector, dof_values::Vector{T}) where T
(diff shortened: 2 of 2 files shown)


AbdAlazezAhmed:
Also, shouldn't `_evaluate_at_points!` and `evaluate_at_points!` return `nothing`?

KnutAM:
> Also, shouldn't _evaluate_at_points! and evaluate_at_points! return nothing?

Why, that would be a breaking change, and normally mutating functions return the mutated argumented, consider e.g. `x = fill!(zeros(3), 1)`

AbdAlazezAhmed:
> > Also, shouldn't _evaluate_at_points! and evaluate_at_points! return nothing?
> 
> Why? That would be a breaking change, and normally mutating functions return the mutated argumented, consider e.g. `x = fill!(zeros(3), 1)`

IIUC the function would not be type-stable as the return type can potentially be abstract. I might be wrong tho.
Also, I think the in-place functions are not exported.

termi-official:
`evaluate_at_points!` is internal only right now.

AbdAlazezAhmed:
> > IIUC the function would not be type-stable as the return type can potentially be abstract.
> 
> It should specialize on the concrete type of the input that is mutated, that is not changed by allowing `::AbstractVector`.
> 
> > `evaluate_at_points!` is internal only right now.
> 
> True - so not breaking, but still not following the [standard convention](https://docs.julialang.org/en/v1/manual/style-guide/#bang-convention) IMO.
> 
> What is the motivation for this change?

Yes you're right, I was confused.

test/test_pointevaluation.jl
```diff
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
     return
 end
 
+function test_pe_views()
+    mesh = generate_grid(Quadrilateral, (3, 3))
+    perturb_standard_grid!(mesh, 1 / 10)
+    dof_vals = 1.0 * [0, 0, 1, 2, 6, 5, 3, 7, 4, 8, 10, 9, 11, 12, 14, 13, 15, 16, 0, 0]
```
KnutAM, line 196:
```suggestion
    dof_vals = Float64[0, 0, 1, 2, 6, 5, 3, 7, 4, 8, 10, 9, 11, 12, 14, 13, 15, 16, 0, 0]
```

test/test_pointevaluation.jl
```diff
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
...
+    points = [node.x for node in mesh.nodes[[6, 7, 11, 10]]] # same as nodes
+    vals_to = zeros(length(points) + 4)
+
+    dh = DofHandler(mesh)
+    add!(dh, :s, Lagrange{RefQuadrilateral, 1}()) # a scalar field
```
KnutAM, line 201:
```suggestion
    ip = Lagrange{RefQuadrilateral, 1}()
    add!(dh, :s, ip) # a scalar field
```

test/test_pointevaluation.jl
```diff
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
...
+    close!(dh)
+
+    ph = PointEvalHandler(mesh, points)
+
+    ip = Ferrite.get_func_interpolations(dh, :s)
```
KnutAM, line 206:
```suggestion
```

test/test_pointevaluation.jl
```diff
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
...
+    Ferrite.evaluate_at_points!((@view vals_to[3:(end - 2)]), ph, dh, (@view dof_vals[3:(end - 2)]), :s, ip)
```
KnutAM, line 207: Shouldn't `evaluate_at_points` be tested instead since that is public. 

Did this originate from a need to have the mutating version with a view. 

In that case, it would make sense IMO to test first that it works with `evaluate_at_points`, and then that the mutating version with a view also does the right thing and is allowed.

test/test_pointevaluation.jl
```diff
@@ -190,6 +190,26 @@ function test_pe_dofhandler()
...
+    @test vals_to ≈ 1.0 * [0, 0, 6, 7, 11, 10, 0, 0]
```
KnutAM, line 208:
```suggestion
    @test vals_to ≈ Float64[0, 0, 6, 7, 11, 10, 0, 0]
```

test/test_pointevaluation.jl
AbdAlazezAhmed: I needed it when using
https://github.com/JuliaHealth/Thunderbolt.jl/blob/499ab6548acaec751d239db2e1977898363e8151/src/ferrite-addons/transfer_operators.jl#L126

It's being used in a loop for each timestep and the solution is stored in an ntimesteps*ndofs matrix, thus I'm using views to access the solution at the nth timestep. I don't think it's the bottleneck but moving towards a non-allocating method would be nice.

PR #1185, opened 2025-04-18
Ensure that refshapes matches when creating FEValues

KnutAM:
This bit me giving silent wrong results
```julia
julia> ip = Lagrange{RefTriangle,1}()
Lagrange{RefTriangle, 1}()

julia> qr = QuadratureRule{RefQuadrilateral}(2);

julia> CellValues(qr, ip)
CellValues(scalar, rdim=2, and sdim=2): 4 quadrature points
 Function interpolation: Lagrange{RefTriangle, 1}()
Geometric interpolation: Lagrange{RefTriangle, 1}()^2
```
This PR throws an `ArgumentError` in this case.


termi-official:
I think we can add something analogue for InterfaceValues.

KnutAM:
> I think we can add something analogue for InterfaceValues.

This is already included. I implemented it on the level of `FunctionValues` and `GeometryMapping` to ensure that anything that uses this building blocks will give the error.

termi-official (approved):
LGTM. That should help tracking down issues in more complicated code bases.

PR #1189, opened 2025-04-26
Proposal: Split interpolation implementations into separate files

gijswl:
While working on the Nedelec and RT interpolations, I found it was quite difficult to move around the large `interpolations.jl` file (it being approx. 2000 lines). My proposal would be to put each interpolation family into its own file to improve maintainability.
Besides that, some interpolations were missing from the list of implemented interpolations. I've added these to the docstring of each family.

Thoughts?


termi-official:
I also already thought about this, but was not sure what a good policy to group them could look like, as too fine granularity can also be problematic if the number of interpolation keeps growing.

The implemented ones were added to the abstract type to improve visibility to what is implemented already.

KnutAM:
Split based on conformity could make sense. Ignore history would be good.

> The implemented ones were added to the abstract type to improve visibility to what is implemented already.

I've had a thought to define a tuple/vector of implemented interpolations. That could automate testing, enable documenting which cases are implemented etc.

gijswl:
I think it is most natural to split by FE space, as I proposed in the commit.
This is also the approach taken by [Gridap](https://github.com/gridap/Gridap.jl/tree/a8b502890a06382bd685e9d5f6775b643bf27fb1/src/ReferenceFEs) and [deal.ii](https://github.com/dealii/dealii/tree/master/source/fe).

Ignoring the commit for history seems like a good idea.

KnutAM:
For me, that split could be alright too, but it will as Dennis says lead to many small files which is a bit inconsistent with how the rest of the repo is organized. Perhaps @fredrikekre has some opinion here?

fredrikekre:
I don't mind the long file but this split is ok too.

src/Interpolations/Nedelec1.jl
```diff
@@ -0,0 +1,148 @@
+"""
```
fredrikekre, line 1: Perhaps Nedelec.jl? Both variants could be in the same file

src/interpolations.jl
fredrikekre, line 1: Rename to `src/Interpolations/Interpolations.jl`?

PR #1190, opened 2025-04-27
Add type assert when testing of dof indices functions

### LINKED ISSUE #1187: Test types of <entity>dof_xx_indicies for interpolations
See: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1162#discussion_r2060152846

KnutAM:
This ensures that silent implementation inconsistencies don't sneak in, ref https://github.com/Ferrite-FEM/Ferrite.jl/pull/1162#discussion_r2060152846

AFAIU the reason this could go through the regular usage tests is that an implementation working for `Tuple{Int}` should also work for `Int` in many cases, since an `Int` supports both indexing by index 1, `length`, and iteration.

Closes #1187

### DIFF
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -33,13 +33,13 @@ function test_interpolation_properties(ip::Interpolation{RefShape, FunOrder}) wh
         as_vector(t::Tuple) = collect(as_vector.(t))
         as_vector(i::Int) = i
         dof_data = (
-            vert = as_vector(Ferrite.vertexdof_indices(ip)),
-            edge = as_vector(Ferrite.edgedof_indices(ip)),
-            face = as_vector(Ferrite.facedof_indices(ip)),
-            edge_i = as_vector(Ferrite.edgedof_interior_indices(ip)),
-            face_i = as_vector(Ferrite.facedof_interior_indices(ip)),
-            vol_i = as_vector(Ferrite.volumedof_interior_indices(ip)),
-            n = getnbasefunctions(ip),
+            vert = as_vector(Ferrite.vertexdof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            edge = as_vector(Ferrite.edgedof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            face = as_vector(Ferrite.facedof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            edge_i = as_vector(Ferrite.edgedof_interior_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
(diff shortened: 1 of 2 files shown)


KnutAM:
Downstream test failure fixed in https://github.com/Ferrite-FEM/FerriteInterfaceElements.jl/pull/27

KnutAM:
This one should not be problematic to merge, but perhaps good to also have https://github.com/Ferrite-FEM/FerriteInterfaceElements.jl/pull/27 approved and ready to merge first, to avoid issues for anyone working on the Ferrite#master and FerriteInterfaceElements#main?

test/test_interpolations.jl
```diff
@@ -33,13 +33,13 @@ function test_interpolation_properties(ip::Interpolation{RefShape, FunOrder}) wh
...
-            edge_i = as_vector(Ferrite.edgedof_interior_indices(ip)),
-            face_i = as_vector(Ferrite.facedof_interior_indices(ip)),
-            vol_i = as_vector(Ferrite.volumedof_interior_indices(ip)),
-            n = getnbasefunctions(ip),
+            vert = as_vector(Ferrite.vertexdof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            edge = as_vector(Ferrite.edgedof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            face = as_vector(Ferrite.facedof_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            edge_i = as_vector(Ferrite.edgedof_interior_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            face_i = as_vector(Ferrite.facedof_interior_indices(ip)::NTuple{<:Any, NTuple{<:Any, Int}}),
+            vol_i = as_vector(Ferrite.volumedof_interior_indices(ip)::NTuple{<:Any, Int}),
```
termi-official, line 41: Note that we only require these to be iterable objects. Technically users can also return e.g. some kind of array.

KnutAM: True! We could be more flexible and make things work, but this follows the devdocs that we have.

Since an `Int` is iterable, that would also be allowed (and has obviously worked since I could correct a few cases with this PR). But as seen by https://github.com/Ferrite-FEM/FerriteInterfaceElements.jl/pull/27, that makes it easy to result in implementation bugs if different interpolations do different things. 

If we later include interpolations that support e.g. variable number of base functions, and thus they return e.g. `Vector{Vector{Int}}`, I would suggest widening the interface then.

KnutAM: > So the interface actually expects iterable collection types, right?

AFAIU the implementation works for iterables, but the interface is tuples of tuples.

PR #1192, opened 2025-05-01
Remove is_discontinuous

KnutAM:
Actually, it seems like `is_discontinuous` wasn't used prior to #867.

Is there a historical reason why this was in `InterpolationInfo`?

Update: This PR completely removes `is_discontinuous`, as the same information is now available via `conformity`.

### DIFF
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -302,10 +302,6 @@ end
         @test Ferrite.reference_coordinates(DiscontinuousLagrange{RefQuadrilateral, 0}()) ≈ [Vec{2, Float64}((0, 0))]
         @test Ferrite.reference_coordinates(DiscontinuousLagrange{RefTetrahedron, 0}()) ≈ [Vec{3, Float64}((1 / 4, 1 / 4, 1 / 4))]
         @test Ferrite.reference_coordinates(DiscontinuousLagrange{RefHexahedron, 0}()) ≈ [Vec{3, Float64}((0, 0, 0))]
-
-        # Test discontinuous interpolations related functions
-        d_ip = DiscontinuousLagrange{RefQuadrilateral, 1}()
-        ip = Lagrange{RefQuadrilateral, 1}()
     end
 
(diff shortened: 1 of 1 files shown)


termi-official:
I was also wondering when updating the InterpolationInfo. cc @AbdAlazezAhmed why was this added? I do not see it being used in the blame commit either.

AbdAlazezAhmed:
I think it was initially used in sparsity patterns in https://github.com/Ferrite-FEM/Ferrite.jl/pull/710/commits/24ebc14f520d4fef44296788490f28b313475927 and it was added to `InterpolationInfo` for type stability as in https://github.com/Ferrite-FEM/Ferrite.jl/pull/710#issuecomment-1556114899

src/interpolations.jl
```diff
@@ -436,13 +445,12 @@ nfacets(ip::InterpolationByDim{2}) = nedges(ip)
...
-    is_discontinuous(::Type{<:Interpolation})
+    is_discontinuous(ip::Interpolation)
 
-Checks whether the interpolation is discontinuous (i.e. `DiscontinuousLagrange`)
+Checks whether the interpolation is discontinuous (i.e. `DiscontinuousLagrange`),
+automatically implemented based on `conformity(ip)`.
 """
-is_discontinuous(ip::Interpolation) = is_discontinuous(typeof(ip))
-is_discontinuous(::Type{<:Interpolation}) = false
+is_discontinuous(ip::Interpolation) = is_discontinuous(conformity(ip))
```
termi-official, line 453: Now that we have a better granularity to describe the continuity, I would argue that this function should be removed, as it is really just the check that `conformity !=  H1`.

KnutAM: True, although we could also have e.g. H2, but for the vtk export that would probably be different anyways - I'll remove instead! 
Good to reduce some code sometimes 😄

termi-official: Yea, `is_discontinuous` was essentially the initial idea to distinguish the L2 space from H1 space without introducing the trait system (as at that time people argued against the trait system).

Issue #1195, opened 2025-05-06
Multiple facedofs per face


gijswl:
Hi all,

While we were working on the first-order H(curl) and H(div) interpolations in 3D, the limitation to one interior facedof was mentioned (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1162#issuecomment-2686204332). I found a similar warning here 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/c718c73751049e442f4f2349f5306c7f7e475f64/src/Dofs/DofHandler.jl#L440-L442
For higher-order interpolations, multiple interior facedofs and even volumedofs could be required. 

Is there any plan or ongoing work in this direction? 
What would be required to implement such functionality? If there is enough documentation to work with I would be happy to give it a shot again, but this feels like it might require a bit more knowledge of the internals than implementing those new interpolations.

termi-official:
xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/626 https://github.com/Ferrite-FEM/Ferrite.jl/issues/191 for very similar issues. It is not difficult to implement, but quite annoying in 3D. Essentially one needs to implement https://github.com/Ferrite-FEM/Ferrite.jl/blob/c718c73751049e442f4f2349f5306c7f7e475f64/src/Grid/grid.jl#L687-L714 to get orientation information (or recycle `InterfaceOrientationInfo`) and write down the permutation tables for all combinations of face orientation pairs of a given interpolation. The 2D version of what you need is implemented here https://github.com/Ferrite-FEM/Ferrite.jl/blob/c718c73751049e442f4f2349f5306c7f7e475f64/src/Dofs/DofHandler.jl#L609-L657 , which is so simple because there are just two orientations for edges. If you have more questions, then feel free to ask them here.

gijswl:
I've been trying to work through and understand the referenced paper (Scroggs et al. ACM Trans. Math. Softw. 48 (2022).)
So far it appears that the dof distribution depends on the type of interpolation. While permuting the dofs is possible for "simple" interpolations, some of them (notably vector interpolations) result in a linear transformation of the shape function space.

It would be nice if the permutation / transformation matrix could be calculated on the fly, similar to what is being done in `InterfaceValues` to match the orientation of the faces. In any case it seems the linear transformation would have to be applied during assembly, rather than permuting the dofs while distributing? Perhaps this should be part of `CellValues`?

KnutAM:
[...]

PR #1196, opened 2025-05-07
Use OrientationInfo instead of {Path,Surface}OrientationInfo everywhere

gijswl:
As far as I can see, `OrientationInfo` has the same functionality as `PathOrientationInfo` and `SurfaceOrientationInfo`, so it makes more sense to use only the former. This is preparation work for https://github.com/Ferrite-FEM/Ferrite.jl/issues/1195.

### DIFF
--- a/src/Grid/grid.jl
+++ b/src/Grid/grid.jl
@@ -664,58 +664,8 @@ end
 #################################
 # @TODO merge this code with into the logic in `ConstraintHandler`.
 
-"""
-    PathOrientationInfo
-
-Orientation information for 1D entities.
-
-The orientation for 1D entities is defined by the indices of the grid nodes
-associated to the vertices. To give an example, the oriented path
-```
-1 ---> 2
-```
-is called *regular*, indicated by `regular=true`, while the oriented path
-```
-2 ---> 1
-```
-is called *inverted*, indicated by `regular=false`.
-"""
-struct PathOrientationInfo
-    regular::Bool # Indicator whether the orientation is regular or inverted.
-end
-
-"""
-    SurfaceOrientationInfo
-
-Orientation information for 2D entities. Such an entity can be
-possibly flipped (i.e. the defining vertex order is reverse to the
-spanning vertex order) and the vertices can be rotated against each other.
-Take for example the faces
-```
-1---2 2---3
-| A | | B |
-4---3 1---4
-```
-which are rotated against each other by 90° (shift index is 1) or the faces
-```
-1---2 2---1
-| A | | B |
-4---3 3---4
-```
-which are flipped against each other. Any combination of these can happen.
-The combination to map this local face to the defining face is encoded with
-this data structure via ``rotate \\circ flip`` where the rotation is indiced by
-the shift index.
-    !!!NOTE TODO implement me.
-"""
-struct SurfaceOrientationInfo
-    #flipped::Bool
-    #shift_index::Int
-end
-
-
 @doc raw"""
-    InterfaceOrientationInfo
+    OrientationInfo
 
 Orientation information for 1D and 2D entities.
 The orientation is defined by the indices of the grid nodes
--- a/docs/src/devdocs/dofhandler.md
+++ b/docs/src/devdocs/dofhandler.md
@@ -9,8 +9,7 @@ during dof distribution, because the interpolations are not available as concret
 
 ```@docs
 Ferrite.InterpolationInfo
-Ferrite.PathOrientationInfo
-Ferrite.SurfaceOrientationInfo
+Ferrite.OrientationInfo
 ```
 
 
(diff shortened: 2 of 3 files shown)


termi-official:
@AbdAlazezAhmed do you remember what exactly was the reason to not use path and surface orientation info, but introduce a new InterfaceOrientationInfo struct?

I personally prefer a design where we keep orientation information for each dimension separate tho, because this should help developers in understanding what exactly is going on here and what exactly the involved assumptions/invariants for each dimension are.

termi-official:
So the plan from my side was to add different data types per dimension to control internal dispatches, which will be helpful to implement support for problems higher dimensions than 3 later on, as we just need to fill in the places with the dispatches.

gijswl:
> So the plan from my side was to add different data types per dimension to control internal dispatches

I actually like that better as well. I have no problem implementing that, but would be good to have confirmation about why the separate `OrientationInfo` exists.

AbdAlazezAhmed:
Seems like the plan was to replace them was `OrientationInfo`
https://github.com/AbdAlazezAhmed/Ferrite.jl/blob/9c60a0b6d8b27457da99c56e6cc4819ca74e7f19/src/Grid/grid.jl#L1453-L1458

Then it became `InterfaceOrientationInfo` storing relative orientation
https://github.com/AbdAlazezAhmed/Ferrite.jl/blob/5153fde45dedc5db9a3f63b72b11a69ecb18b950/src/Grid/grid.jl#L895-L899

Then back to `OrientationInfo` with a docs typo.
I think using a single `OrientationInfo` would be nice if we can fix a set of rules on how to define an orientation for n-dim facet
But yeah I think you can use `OrientationInfo` and `(Path/Surface)OrientationInfo` exchangeably

termi-official:
>  I think using a single OrientationInfo would be nice if we can fix a set of rules on how to define an orientation for n-dim facet

Not sure how helpful that is for keeping the code complexity manageable. For generic dimensions (=n) the best you can probably do is encoding the orientation directly as an m-dimensional bitvector.

gijswl:
Are the names deliberately chosen, or should they be updated to match the rest of the terminology:
- `PathOrientationInfo` -> `EdgeOrientationInfo`
- `SurfaceOrientationInfo` -> `FaceOrientationInfo`

termi-official:
I have chosen the names deliberately, as they precede the consistent rename to edge/face and facet/ridge. I have no strong feeling on renaming the orientation info structs.

KnutAM:
> I have no strong feeling on renaming the orientation info structs

I have a strong feeling they should be renamed 😄

src/Grid/grid.jl
```diff
@@ -674,117 +694,73 @@ associated to the vertices. To give an example, the oriented path
...
+is called *regular*, indicated by `flipped=true`, while the oriented path
 ```
 2 ---> 1
 ```
-is called *inverted*, indicated by `regular=false`.
+is called *inverted*, indicated by `flipped=false`.
 """
 struct PathOrientationInfo
-    regular::Bool # Indicator whether the orientation is regular or inverted.
+    flipped::Bool # Indicator whether the orientation is regular or inverted.
```
KnutAM, line 704: IMO it is less straight forward to store if the edge is flipped, than if it is `regular`. Personally, I would prefer `positive`, or perhaps we even can save simply as 
```julia
direction::Int8 # \pm 1
```
and return `Int8` from `get_face_direction` and `get_edge_direction`. Probably, we won't see any significant performance difference by using `Int8` vs `Int`, don't recall the profiles of the dof distribution now if this logic takes some significant portion of the time...

src/Grid/grid.jl
```diff
@@ -674,117 +694,73 @@ associated to the vertices. To give an example, the oriented path
...
+    return SurfaceOrientationInfo(flipped, shift_index)
 end
 
-function get_edge_direction(edgenodes::NTuple{2, Int})
-    positive = edgenodes[2] > edgenodes[1]
-    return ifelse(positive, 1, -1)
+
+"""
+    OrientationInfo(edgenodes::NTuple{2, Int})
+    OrientationInfo(facenodes::NTuple{N, Int})
```
KnutAM, line 756:
```suggestion
    orientation_info(edgenodes::NTuple{2, Int})
    orientation_info(facenodes::NTuple{N, Int})
```

src/Grid/grid.jl
termi-official: Reagardng the dof distribution time, it is mostly spend in the hash table lookups, as we do not materialize the mesh entities explicitly. But I think just a boolean is clearer here. Also, I would expect that `flipped = true` would mean `2--->1`.

gijswl: A "flip" is indeed the terminology used in the Scroggs paper to indicate that the nodes are swapped compared to the "normal" orientation. So, 
* `1 --> 2` is `flipped = false`
* `2 --> 1` is `flipped = true` 

This is the same way `OrientationInfo` already worked and in line with the paper. The previous definition of `regular::Bool` was opposite, of course. Can you check whether the tests make sense following this definition?

gijswl: I see now that the docstring was not updated yet.
Refer to the latest commit for the corrected definition.

PR #1199, opened 2025-05-13
2nd order Nedelec interpolation

gijswl:
As discussed in issue https://github.com/Ferrite-FEM/Ferrite.jl/issues/1195 I wanted to implement higher-order Hcurl interpolations. These require multiple degrees of freedom interior to the faces, which raises the issue of ensuring continuity between neighbouring elements when their local dofs do not match up.

The first commit (building on the work in PR #1196) includes the shape functions of `Nedelec{RefTetrahedron,2}` and minor additions to the interpolation tests to properly test this new interpolation. As a proof of concept I've implemented the facedof transformation inside `apply_mapping!`, but this need not be its final place. For convenience I've just commented out the restriction to a single facedof, although we would probably want to check if the correct face permutations/transformations are implemented.

As a result, the interpolation and continuity tests are passing, but I have my doubts about the actual continuity. Insofar as I can see, `test_continuity.jl` tests for continuity in the shape functions (i.e. the transformation `Nξ` to `Nx`), but does not consider `dNdx`, which is important for forming the `shape_curl` values during element assembly. @KnutAM do you have any comments on if we need to further transform `dNdx` and if this should be part of these tests?


KnutAM:
> Insofar as I can see, test_continuity.jl tests for continuity in the shape functions (i.e. the transformation Nξ to Nx), but does not consider dNdx, which is important for forming the shape_curl values during element assembly.

Just as we don't expect `dNdx` to be continuous for `H1Conformity`, we don't expect `curl(N(x))` to be continuous for `HcurlConformity`, that would require a different type of interpolation. HcurlConformity implies that $\mathrm{curl}(u)$ is bounded (i.e. in L2), but it may be discontinuous.

gijswl:
> we don't expect `curl(N(x))` to be continuous for `HcurlConformity`

Agreed, but since I am now applying a linear transformation to `Nξ` based on the orientation of the face, we should also make sure `dNdξ` is updated to account for the transformed orientation, right? Intuitively I feel like we have to do something more than apply the same transformation matrix to `dNdξ`, but I can't substantiate why.

Issue #1212, opened 2025-07-21
Missing dipatch for `toparaview()`


henrij22:
I had a missing dispatch when working with embedded finite elements (`refdim` != `griddim`). 

```julia
ERROR: MethodError: no method matching toparaview!(::SubArray{Float64, 1, Matrix{…}, Tuple{…}, true}, ::StaticArraysCore.SVector{2, Float64})
The function `toparaview!` exists, but no method is defined for this combination of argument types.
```

I'll link a PR

PR #1213, opened 2025-07-21
Fix missing dipatch for `toparaview()`

henrij22:
see issue #1212 

It only seems to happen when using the discontinous vtk writer, so maybe we can also fix the cause there


src/Export/VTK.jl
```diff
@@ -155,6 +155,10 @@ function toparaview!(v, x::SecondOrderTensor)
     tovoigt!(v, x)
     return v
 end
+function toparaview!(v, x::SVector{D}) where {D}
```
KnutAM:
```suggestion
function toparaview!(v::AbstractVector, x::SVector{D}) where {D}
```

KnutAM (changes requested):
Thanks! The new code looks good to me, but could you add a test-case for the embedded vtk-export that would catch this?

src/Export/VTK.jl
henrij22: Yes, I'll add a test.

test/test_vtk_export.jl
```diff
@@ -206,4 +206,22 @@
             end
         end
     end
+    @testset "discontinous_embedded" begin
+        mktempdir() do tmp
+            _grid = generate_grid(Line, (2,), Vec((0.0,)), Vec((1.0,)))
+            nodes = [(n.x[1], 0.0) |> Vec{2} |> Node for n in _grid.nodes]
+            grid = Grid(_grid.cells, nodes)
```
KnutAM: After #1214 is merged into master and this branch is updated of course:)  
```suggestion
            grid = generate_grid(Line, (2,), Vec((0.0, 0.0)), Vec((1.0, 0.5)))
```

test/test_vtk_export.jl
```diff
@@ -206,4 +206,22 @@
...
+
+            qr = QuadratureRule{RefLine}(2)
+            ip = Lagrange{RefLine, 1}()^2
+            dh = DofHandler(grid)
+            add!(dh, :u, ip)
+            close!(dh)
+
+            u = rand(ndofs(dh))
```
KnutAM:
```suggestion
            u = collect(range(0, 1, ndofs(dh)))
```
(Avoid `rand()` to enable sha-testing)

test/test_vtk_export.jl
```diff
@@ -206,4 +206,22 @@
...
+            VTKGridFile(joinpath(tmp, "output"), dh, write_discontinuous = true) do vtk
+                write_solution(vtk, dh, u)
+            end
```
KnutAM, line 222:
```suggestion
            end
            sha = bytes2hex(open(SHA.sha1, joinpath(tmp, "output") * ".vtu"))
            @test sha == # Insert generated sha above if vtk file looks correct.
```

KnutAM (approved):
Nice!

PR #1215, opened 2025-07-21
`check_isoparametric_boundaries`, `mass_qr`, and `cellcenter` for Pyramids and Wedges + `L2Projector` tests for Hexahedron, Tetrahedron, Pyramid, and Wedge.

AbdAlazezAhmed:
PointEvalHandler needed these functions to work. Also, L2 Projector is only tested with triangles and quads. I added a bit more general test using `Interpolations.jl` instead of manually calculating the test values for first order interpolations.


Project.toml
```diff
@@ -5,6 +5,7 @@ version = "1.1.0"
 [deps]
 EnumX = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
+Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
```
termi-official: extras?

test/test_l2_projection.jl
```diff
@@ -1,21 +1,17 @@
 # Tests a L2-projection of integration point values (to nodal values),
 # determined from the function y = 1 + x[1]^2 + (2x[2])^2
-function test_projection(order, refshape)
-    element = refshape == RefQuadrilateral ? Quadrilateral : Triangle
-    grid = generate_grid(element, (1, 1), Vec((0.0, 0.0)), Vec((1.0, 1.0)))
 
+function test_projection(order, elementtype)
+    refshape = getrefshape(elementtype)
+    dim = Ferrite.getrefdim(refshape)
+    grid = single_element_grid(elementtype)
```
lijas: Why did you change this to a single element grid? Previously we had triangles/tetrahedrons filled a square/cube.

test/test_l2_projection.jl
```diff
@@ -494,10 +512,15 @@ function test_l2proj_errorpaths()
 end
 
 @testset "Test L2-Projection" begin
-    test_projection(1, RefQuadrilateral)
-    test_projection(1, RefTriangle)
-    test_projection(2, RefQuadrilateral)
-    test_projection(2, RefTriangle)
+    for ref_shape in (
+                Quadrilateral,
```
lijas, line 529: These are not refshapes

test/test_l2_projection.jl
```diff
@@ -34,10 +30,31 @@ function test_projection(order, refshape)
     point_vars = project(proj, qp_values, qr)
     qp_values_matrix = reduce(hcat, qp_values)
     point_vars_2 = project(proj, qp_values_matrix, qr)
+
     if order == 1
         # A linear approximation can not recover a quadratic solution,
         # so projected values will be different from the analytical ones
-        ae = [f_approx(i) for i in 1:4]
+        qp_1D_coord = Float64[]
```
lijas: Add a small comment what you are trying to do here

lijas (commented):
Is it easy to implement the linear interpolation ourself and avoid importing the Interpolations package?

test/test_l2_projection.jl
AbdAlazezAhmed: My bad, I didn't notice it.

test/test_l2_projection.jl
AbdAlazezAhmed: I was trying to calculate the first order approximation values using linear extrapolation from the quadrature points instead of using precomputed hardcoded values. But since we use more than one element this won't work so I'll compute the values for other cell types and add them instead.

termi-official:
> PointEvalHandler needed these functions to work.

https://app.codecov.io/gh/Ferrite-FEM/Ferrite.jl/pull/1215?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Ferrite-FEM

The new functions are never called in the tests.

PR #1220, opened 2025-08-04
Add polyorder kwarg, and test that quadrature rules fulfils this

KnutAM:
#1200 revealed that we are not really clear on what `order` in `QuadratureRule` implies. Based on my investigations here, it implies the highest order of a complete polynomial that is exactly integrated for `RefSimplex`, `RefPrism`, and `RefPyramid`, but the number of quadrature points in each direction for `RefHypercube`. 

To allow requesting a certain accuracy (measured in what polynomial order that is exactly integrated), this PR adds an option to construct a quadrature rule as 
`QuadratureRule{RefShape}([T::Type{<:Number}], [quad_type::Symbol]; polyorder::Int)`, where the `polyorder` kwarg always is the highest order of a complete polynomial that is exactly integrated. 

This can then be used internally in #1200.

### DIFF
--- a/src/Quadrature/quadrature.jl
+++ b/src/Quadrature/quadrature.jl
@@ -12,10 +12,14 @@ using Base.Cartesian: @nloops, @ntuple, @nexprs
 
 """
     QuadratureRule{shape}([::Type{T},] [quad_rule_type::Symbol,] order::Int)
+    QuadratureRule{shape}([::Type{T},] [quad_rule_type::Symbol,]; polyorder::Int)
     QuadratureRule{shape}(weights::AbstractVector{T}, points::AbstractVector{Vec{rdim, T}})
 
 Create a `QuadratureRule` used for integration on the refshape `shape` (of type [`AbstractRefShape`](@ref)).
-`order` is the order of the quadrature rule.
+`order` is the order of the quadrature rule, which for hypercubes corresponds to the number of quadrature points
+in each direction, and for simplices, prisms, and pyramids to the polynomial order that is fully integrated.
+To control the order of the polynomial that is exactly integrated for all reference shapes, the keyword argument
+`polyorder` can be passed instead of the `order` argument.
 `quad_rule_type` is an optional argument determining the type of quadrature rule,
 currently the `:legendre` and `:lobatto` rules are implemented for hypercubes.
 For triangles up to order 8 the default rule is the one by `:dunavant` (see [Dun:1985:hde](@cite)) and for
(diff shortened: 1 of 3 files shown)


src/Quadrature/quadrature.jl
```diff
@@ -66,17 +70,40 @@ end
...
+function QuadratureRule{shape}(::Type{T}, args...; kwargs...) where {shape <: AbstractRefShape, T}
     quad_type = _default_quadrature_rule(shape)
+    return QuadratureRule{shape}(T, quad_type, args...)
+end
+function QuadratureRule{shape}(quad_type::Symbol, args...; kwargs...) where {shape <: AbstractRefShape}
+    return QuadratureRule{shape}(Float64, quad_type, args...; kwargs...)
+end
+
+# Passing polyorder kwarg:
+QuadratureRule{shape}(; kwargs...) where {shape <: AbstractRefShape} = QuadratureRule{shape}(Float64; kwargs...)
```
termi-official: I think this one is missing test coverage.

test/test_quadrules.jl
```diff
@@ -237,4 +237,90 @@ using Ferrite: reference_shape_value
...
+        # Define a struct that returns a complete polynomial function with random weights.
+        struct PolyFunction{dim}
+            weights::Vector{Float64}
+            exponents::Vector{NTuple{dim, Int}}
+        end
+
+        (f::PolyFunction{dim})(x::Vec{dim}) where {dim} = sum(w * sum(tuple(x...) .^ e) for (w, e) in zip(f.weights, f.exponents))
+
+        function PolyFunction{1}(order::Int)
+            return PolyFunction{1}(rand(order + 1), [(i,) for i in 0:order])
```
termi-official: I thought we want to remove random numbers from the tests? If we want to keep then, then we should at least add a mechanism to print them on test failure, so we can reproduce the error more easily.

termi-official (commented):
While I like the idea, I am not sure if adding kwargs is a good design for this (as this also seems to be some way forward to resolve the existing ambiguity in `order`). I guess that even with a good explanation, the difference between `order` and `polyorder` might be quite hard to understand for new users, especially with the existing ambiguitiy. Polyorder might also be a bit tricky as a name, if users come from some background where they also might use quadrature rules which also included non-polynomial functions.

May I suggest to rather introduce new `create_...` functions which return `QuadratureRule`s instead of adding more constructors?

src/Quadrature/quadrature.jl
KnutAM: Thanks, good catch - realized now it isn't required, covered above by the `QuadratureRule{shape}(args...; kwargs...)` version.

test/test_quadrules.jl
KnutAM: I didn't know we wanted to do that in general, but I guess in this case it does seem reasonable to avoid random numbers!

kreutzheidephil:
> My argument here is merely that I _think_ it is easier to comprehend (and discover) different functions than a single function with many different combinations of potentially incompatible kwargs.

Don't you think having this information in the constructor is more intuitive? If a new user wants play around with the accuracy of the quadrature rule, using the keyword arguments seems like a natural way to do so. If I was a new user I think the first place I would look for adjusting accuracy is in the constructor, I don't think I would look for a function which does this for me, because the constructor is that function. But maybe I am misunderstanding what you're suggesting :)

I also think throwing an error on non-polynomial rules is a nice feature, especially for newer users that don't have much knowledge in FEM.

PR #1223, opened 2025-08-11
Fix facetskeleton for mixed grids

KnutAM:
`facetskeleton` currently returns `FacetIndex` which doesn't exist on mixed grids. This PR fixes that, and makes the code for `_create_facet_skeleton` a bit more verbose to make it easier to follow. 

Adds a test which also formalizes the convention that was implemented, that the facet on the cell with the lowest cell index is picked.


AbdAlazezAhmed (approved):
LGTM, thanks!

Issue #1203, opened 2025-06-07
Relax error for 'function integrate_projected_dbc!(::HcurlConformity, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)'


lucasbanting:
I was trying out the 3D Nedelec elements in the new ferrite version. The function `function integrate_projected_dbc!(::HcurlConformity, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)` has ` throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))`. I tested commenting out the error, and I got correct results using zero-Dirichlet boundary conditions. Can this ArgumentError be removed, or be changed to warning instead?

I also think the ProjectedDirichlet docs should be updated, as it appears the required function signature is: `bc_fun(x, time, n)`, but the docs implied to me that a function `fun(x,time)` should be used. It seems that `bc_fun` could be implicitly created as `bc_fun(x, time, n) = cross(n, fun(x,time))` in the projected Dirichlet implementation.

KnutAM:
> I also think the `ProjectedDirichlet` docs should be updated, as it appears the required function signature is: `bc_fun(x, time, n)`, but the docs implied to me that a function `fun(x,time)` should be used. 

Where did it say that `fun(x,time)` should be used? This is how the rendered docstring looks to me. 

![Image](https://github.com/user-attachments/assets/b00370ff-b9cc-4094-8513-437b8db1c2e9)

> Can this ArgumentError be removed, or be changed to warning instead?

Yes, we could support a zero-valued bc without adding more advanced logic. Some options would be 
* Support f::Returns and check that `norm(f()) == 0` (more consistent with `Dirichlet`)
* Support f::Number and check that `norm(f) == 0` (would IMO be a nice addition also to `Dirichlet`, but adding just to `ProjectedDirichlet` would be a bit inconsistent). 

As a temporary workaround, probably the first option is the minimal possible change. If you want to do a PR, I can review it (but probably I won't have time before in a couple of weeks).

lucasbanting:
I was wrong about the documentation, thanks for the correction. I made a PR [!1232](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1232) to support the first option.

PR #1232, opened 2025-09-06
Relax warning for projected dirichlet for 3D hcurl

lucasbanting:
Relax warning for projected dirichlet by supporting `bc_fun::Returns` to support zero-dirichlet boundary conditions in 3D.

https://github.com/Ferrite-FEM/Ferrite.jl/issues/1203

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -2008,8 +2008,9 @@ function integrate_projected_dbc!(::HdivConformity, Kᶠ, fᶠ, bc_fun, fv, shap
 end
 
 function integrate_projected_dbc!(::HcurlConformity, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
-    if getrefdim(function_interpolation(fv)) == 3
-        throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))
+	if (getrefdim(function_interpolation(fv)) == 3) && !(bc_fun isa Returns && norm(bc_fun()) == 0)
+		@warn "ProjectedDirichlet for zero-dirichlet will work with bc_fun=Returns(zero(Vec{D}))."
+		throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))
     end
     for q_point in 1:getnquadpoints(fv)
         dΓ = getdetJdV(fv, q_point)
--- a/test/test_interpolations.jl
+++ b/test/test_interpolations.jl
@@ -753,12 +753,17 @@ end
             dh_L2, _ = _setup_dh_fv_for_bc_test(DiscontinuousLagrange{RefTriangle, 1}()^2; nel = 1, qr_order = 1)
             dh_Hcurl_3d, _ = _setup_dh_fv_for_bc_test(Nedelec{RefTetrahedron, 1}(); nel = 1, qr_order = 1)
             for dh in (dh_H1, dh_L2, dh_Hcurl_3d)
-                dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(zero(Vec{2})))
+                dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(zero(Vec{Ferrite.getrefdim(dh)}) .+ 1))
                 ch = add!(ConstraintHandler(dh), dbc)
                 @test_throws "ProjectedDirichlet is not implemented for" close!(ch)
             end
         end
 
+		@testset "ProjectedDirichlet Returns(0)" begin
+			dh_Hcurl_3d, _ = _setup_dh_fv_for_bc_test(Nedelec{RefTetrahedron, 1}(); nel = 1, qr_order = 1)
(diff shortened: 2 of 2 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -2008,8 +2008,10 @@ function integrate_projected_dbc!(::HdivConformity, Kᶠ, fᶠ, bc_fun, fv, shap
 end
 
 function integrate_projected_dbc!(::HcurlConformity, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
-    if getrefdim(function_interpolation(fv)) == 3
-        throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))
+	if (bc_fun isa Returns && norm(bc_fun()) == 0)
+		bc_fun = (x,t,n) -> 0*x
```
KnutAM: Can't we just use the `bc_fun::Returns` instead of redefining?

src/Dofs/ConstraintHandler.jl
```diff
@@ -2008,8 +2008,9 @@ function integrate_projected_dbc!(::HdivConformity, Kᶠ, fᶠ, bc_fun, fv, shap
 end
 
 function integrate_projected_dbc!(::HcurlConformity, Kᶠ, fᶠ, bc_fun, fv, shape_nrs, cell_coords, time)
-    if getrefdim(function_interpolation(fv)) == 3
-        throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))
+	if (getrefdim(function_interpolation(fv)) == 3) && !(bc_fun isa Returns && norm(bc_fun()) == 0)
+		@warn "ProjectedDirichlet for zero-dirichlet will work with bc_fun=Returns(zero(Vec{D}))."
+		throw(ArgumentError("ProjectedDirichlet is not implemented for 3D H(curl) conformity"))
```
KnutAM, line 2013:
```suggestion
		throw(ArgumentError("Non-zero-valued ProjectedDirichlet is not implemented for 3D H(curl) conformity. Use `bc_fun = Returns(zero(Vec{3}))` for that case."))
```

src/Dofs/DofHandler.jl
```diff
@@ -168,6 +168,14 @@ end
...
 get_grid(dh::DofHandler) = dh.grid
 
+"""
+    Ferrite.getrefdim(::DofHandler)
+
+Return the dimension of the degree of freedom handler.
+"""
+getrefdim(::DofHandler{dim, G}) where {dim, G} = dim
+getrefdim(::SubDofHandler{dh}) where {dh} = getrefdim(dh)
+
```
KnutAM, line 178:
```suggestion
```
The `dim` parameter in the `DofHandler` is the spatial dimension, and not the reference dimension.
It is also unrelated to this PR, so can be removed.

test/test_interpolations.jl
```diff
@@ -753,12 +753,17 @@ end
             dh_L2, _ = _setup_dh_fv_for_bc_test(DiscontinuousLagrange{RefTriangle, 1}()^2; nel = 1, qr_order = 1)
             dh_Hcurl_3d, _ = _setup_dh_fv_for_bc_test(Nedelec{RefTetrahedron, 1}(); nel = 1, qr_order = 1)
             for dh in (dh_H1, dh_L2, dh_Hcurl_3d)
-                dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(zero(Vec{2})))
+                dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(zero(Vec{Ferrite.getrefdim(dh)}) .+ 1))
```
KnutAM, line 756:
```suggestion
                dim = Ferrite.getspatialdim(dh.grid) # dim = sdim = vdim for this case
                dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(ones(Vec{dim})))
```

test/test_interpolations.jl
```diff
@@ -753,12 +753,17 @@ end
...
                 ch = add!(ConstraintHandler(dh), dbc)
                 @test_throws "ProjectedDirichlet is not implemented for" close!(ch)
             end
         end
 
+		@testset "ProjectedDirichlet Returns(0)" begin
+			dh_Hcurl_3d, _ = _setup_dh_fv_for_bc_test(Nedelec{RefTetrahedron, 1}(); nel = 1, qr_order = 1)
+            dbc = ProjectedDirichlet(:u, Set([FacetIndex(1, 1)]), Returns(zero(Vec{3})))
+            ch = add!(ConstraintHandler(dh_Hcurl_3d), dbc)
```
KnutAM, line 765:
```suggestion
            ch = close!(add!(ConstraintHandler(dh_Hcurl_3d), dbc))
```
We also need to test that the constraints are correctly enforced in this case. E.g. by checking that the tangential component is zero on multiple points on the boundary. Would also be good to check for at least two neighboring faces, since they share an edge which is the reason non-zero conditions cannot be implemented. Perhaps use `nel = 2` as well. We could also check directly that the dofs included in the `facedof_indices` for the faces in the provided `FacetIndex`es are zero as a first validation.

KnutAM (changes requested):
Nice work! 
Some comments and tests checking the functionality must be added.

PR #1233, opened 2025-09-19
Fix signature of `reference_shape_hessians_gradients_and_values!()`

henrij22:
This small fix allows us to overload the `reference_shape_hessians_gradients_and_values!()` for custom interpolations. It also harmonizes the signature with `reference_shape_gradients_and_values!()`.

### DIFF
--- a/src/interpolations.jl
+++ b/src/interpolations.jl
@@ -232,7 +232,7 @@ end
 Evaluate all shape function hessians, gradients and values of `ip` at once at the reference point `ξ`
 and store them in `hessians`, `gradients`, and `values`.
 """
-@propagate_inbounds function reference_shape_hessians_gradients_and_values!(hessians::AbstractVector, gradients::AbstractVector, values::AbstractVector, ip::Interpolation, ξ::Vec)
+@propagate_inbounds function reference_shape_hessians_gradients_and_values!(hessians::HAT, gradients::GAT, values::SAT, ip::IP, ξ::Vec) where {IP <: Interpolation, SAT <: AbstractArray, GAT <: AbstractArray, HAT <: AbstractArray}
     @boundscheck checkbounds(hessians, 1:getnbasefunctions(ip))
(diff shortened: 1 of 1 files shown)


KnutAM:
Can you provide an example when it didn't work before? 

It seems strange to allow `AbstractArray` instead of `AbstractVector` for this case, and this function should normally not be overloaded for a custom interpolation, but should work provided that the interpolation is implemented correctly.

src/interpolations.jl
```diff
@@ -232,7 +232,7 @@ end
 Evaluate all shape function hessians, gradients and values of `ip` at once at the reference point `ξ`
 and store them in `hessians`, `gradients`, and `values`.
 """
-@propagate_inbounds function reference_shape_hessians_gradients_and_values!(hessians::AbstractVector, gradients::AbstractVector, values::AbstractVector, ip::Interpolation, ξ::Vec)
+@propagate_inbounds function reference_shape_hessians_gradients_and_values!(hessians::HAT, gradients::GAT, values::SAT, ip::IP, ξ::Vec) where {IP <: Interpolation, SAT <: AbstractArray, GAT <: AbstractArray, HAT <: AbstractArray}
```
KristofferC, line 235: Why the new parameters (the `where` stuff)? They are not used in the function body?

KnutAM: I was just asking myself why we do that for the `reference_shape_gradients_and_values!`, I think that case is a typo/unnecessary usage of `where` and we should rather change that (and also to `AbstractVector`).

henrij22: I introduced the where stuff because it was used in the `gradient` method, and also thats what Julia proposed to me to circumvent the ambiguity. 

I'll however see if overloading `reference_shape_hessian_gradient_and_value` like Knut mentioned will work for me

KristofferC: They were originally written like that in https://github.com/Ferrite-FEM/Ferrite.jl/pull/819 but I see no reason for it.

PR #1252, opened 2025-11-24
More abstract vectors

termi-official:
This allows e.g. views or GPU vectors to be used.

### DIFF
--- a/test/test_grid_dofhandler_vtk.jl
+++ b/test/test_grid_dofhandler_vtk.jl
@@ -181,6 +181,14 @@ end
 
     @test getcells(grid, "cell_set") == [getcells(grid, 1)]
 
+    # cellnodes via empty DofHandler
+    nodeids = zeros(Int, 9)
+    dh = DofHandler(grid)
+    close!(dh)
+    Ferrite.cellnodes!(nodeids, dh, 1)
+    # Note that the return types typically differ (Vector vs Tuple)
+    @test all(nodeids .== Ferrite.get_node_ids(getcells(grid, 1)))
+
     # CellIterator on a grid without DofHandler
     grid = generate_grid(Triangle, (4, 4))
     n = 0
@@ -545,7 +553,6 @@ end
         Line((6, 7)),
     ]
     nodes = [Node(coord) for coord in zeros(Vec{2, Float64}, 18)]
-    grid = Grid(cells, nodes)
     @test_throws ErrorException ExclusiveTopology(grid)
     # topology = ExclusiveTopology(grid)
     # @test_throws ArgumentError Ferrite.facetskeleton(topology, grid)
@@ -584,7 +591,7 @@ end
     @test Set(Ferrite.getstencil(stars, quadgrid, VertexIndex(5, 4))) == Set([VertexIndex(4, 2), VertexIndex(4, 4), VertexIndex(5, 1), VertexIndex(5, 3), VertexIndex(7, 1), VertexIndex(7, 3), VertexIndex(8, 2), VertexIndex(8, 4), VertexIndex(4, 3), VertexIndex(5, 4), VertexIndex(7, 2), VertexIndex(8, 1)])
     @test Set(Ferrite.toglobal.((quadgrid,), Ferrite.getstencil(stars, quadgrid, VertexIndex(1, 1)))) == Set([1, 2, 5])
     @test Set(Ferrite.toglobal.((quadgrid,), Ferrite.getstencil(stars, quadgrid, VertexIndex(2, 1)))) == Set([2, 1, 6, 3])
-    @test Set(Ferrite.toglobal.((quadgrid,), Ferrite.getstencil(stars, quadgrid, VertexIndex(5, 4)))) == Set([10, 6, 9, 11, 14])
+    @test Set(Ferrite.toglobal(quadgrid, collect(Ferrite.getstencil(stars, quadgrid, VertexIndex(5, 4))))) == Set([10, 6, 9, 11, 14])
 
     face_skeleton = Ferrite.facetskeleton(topology, quadgrid)
     @test Set(face_skeleton) == Set(


src/Dofs/DofHandler.jl
```diff
@@ -204,13 +204,13 @@ Store the degrees of freedom that belong to cell `i` in `global_dofs`.
 
 See also [`celldofs`](@ref).
 """
-function celldofs!(global_dofs::Vector{Int}, dh::DofHandler, i::Int)
+function celldofs!(global_dofs::AbstractVector{Int}, dh::DofHandler, i::Int)
     @assert isclosed(dh)
     @assert length(global_dofs) == ndofs_per_cell(dh, i)
     unsafe_copyto!(global_dofs, 1, dh.cell_dofs, dh.cell_dofs_offset[i], length(global_dofs))
     return global_dofs
 end
```
KnutAM: I think allowing `AbstractVector` is not compatible with using `unsafe_copyto!` here?

src/Grid/grid.jl
```diff
@@ -584,14 +584,14 @@ end
 
 Mutate `x` to the coordinates of the cell corresponding to `idx` or `cell`.
 """
-@inline function getcoordinates!(x::Vector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
+@inline function getcoordinates!(x::AbstractVector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
     node_ids = get_node_ids(cell)
     @inbounds for i in 1:length(x)
         x[i] = get_node_coordinate(grid, node_ids[i])
     end
```
KnutAM, line 592: `1:length(x)` should be used for `AbstractVector` (and the use of `@inbounds` here seems highly questionable, seems like out of bounds can occur by just passing a too long vector `x`, but that is unrelated). 

Both could be fixed though with 
```suggestion
    node_ids = get_node_ids(cell)
    @boundscheck checkbounds(x, 1:length(node_ids))
    @inbounds for (i, node_id) in enumerate(node_ids)
        x[i] = get_node_coordinate(grid, node_id)
    end
```

src/Grid/grid.jl
```diff
@@ -604,12 +604,12 @@ Return the coordinate of the `n`th node in `grid`
...
+function cellnodes!(global_nodes::AbstractVector{Int}, grid::AbstractGrid, i::Int)
     cell = getcells(grid, i)
     _cellnodes!(global_nodes, cell)
     return global_nodes
 end
-function _cellnodes!(global_nodes::Vector{Int}, cell::AbstractCell)
+function _cellnodes!(global_nodes::AbstractVector{Int}, cell::AbstractCell)
     @assert length(global_nodes) == nnodes(cell)
     @inbounds for i in 1:length(global_nodes)
         global_nodes[i] = cell.nodes[i]
```
KnutAM: Same as `getcoordinates!` (needs `@boundscheck checkbounds` + `enumerate`) (and while at it, perhaps use `get_node_ids`?)

KnutAM (commented):
Just some quick comments due to slack-ping :)

src/Dofs/DofHandler.jl
termi-official: cc @fredrikekre I guess we need to variants here then? One on Vector with unsafe_copyto and one with just copyto. Is there a reason why this is not simply `copyto!`?

src/Grid/grid.jl
termi-official: Why adding a boundscheck here?

src/Dofs/DofHandler.jl
```diff
@@ -210,7 +210,16 @@ function celldofs!(global_dofs::Vector{Int}, dh::DofHandler, i::Int)
...
-function celldofs!(global_dofs::Vector{Int}, sdh::SubDofHandler, i::Int)
+function celldofs!(global_dofs::AbstractVector{Int}, dh::AbstractDofHandler, i::Int)
+    @assert isclosed(dh)
+    @assert length(global_dofs) == ndofs_per_cell(dh, i)
+    offset = sdh.cell_dofs_offset[i] - 1
+    @inbounds for j in 1:n
+        global_dofs[j] = dh.cell_dofs[offset + j]
+    end
+    return global_dofs
```
KnutAM: Why not
```suggestion
    return copyto!(global_dofs, 1, dh.cell_dofs, dh.cell_dofs_offset[i], length(global_dofs))
```
as I think you suggested previously? Was it slower? 

(And also, am I missing something or where is `n` defined? I guess we should add a test case for this one?)

termi-official: I just forgot and didn't realize that it was also not under test coverage. Should be fixed now.

src/Grid/grid.jl
```diff
@@ -584,14 +584,14 @@ end
 
 Mutate `x` to the coordinates of the cell corresponding to `idx` or `cell`.
 """
-@inline function getcoordinates!(x::Vector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
+@inline function getcoordinates!(x::AbstractVector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
     node_ids = get_node_ids(cell)
-    @inbounds for i in 1:length(x)
+    @inbounds for (i, node_id) in enumerate(node_ids)
         x[i] = get_node_coordinate(grid, node_ids[i])
```
KnutAM:
```suggestion
        x[i] = get_node_coordinate(grid, node_id)
```
The boundscheck is critical when using `x[i]` for `(i, _) in enumerate(node_ids)`, because now we also have the potential for out-of-bounds write. 

If we use `for i in eachindex(x)` we can only have out-of-bounds read in `get_node_coordinate` (currently this doesn't seem to propagate inbounds so we get error if `node_ids[i]` returns something outside the range of node numbers, which means this could sometimes pass and sometimes fail. If `get_node_coordinate` would propagate the inbounds (I think this would be valid to add a `@inline/@propagate_inbounds` to), we would get random noise back...

If you don't agree with the `@inbounds` analysis wrt. bounds-checking, that is also unrelated, so I can open a separate PR for that later. 

However, to avoid the potential issues of indexing into abstract vectors, we could simply do
```diff
- @inbounds for i in 1:length(x)
+ @inbounds for i in eachindex(x)::Base.OneTo
```
To guarantee one-based indexing and then that should be a sufficient change in this PR (if we handle the `@inbounds` issues elsewhere).

fredrikekre: `eachindex(x, node_ids)`?

KnutAM: Yes, but then `checkbounds` is required to avoid out-of-bounds write?

fredrikekre: IMO one should be able to rely on `eachindex` being correctly implemented and not giving out of bounds indices.

KnutAM: Right - I missed that the `x` was included, I just read `eachindex(node_ids)` 🤦 
That is (of course) correct.

termi-official: Done

termi-official: Reverted (see CI failure).

KnutAM: > IMO one should be able to rely on eachindex being correctly implemented and not giving out of bounds indices.

```julia
julia> eachindex((1,), (2,2))
Base.OneTo(2)
```

Was fixed for tuples in https://github.com/JuliaLang/julia/pull/48125, but that is not included in v1.12.4... 

However, I guess tuple + array is not supported, probably should though

src/Grid/grid.jl
termi-official: Checked.

src/Grid/grid.jl
```diff
@@ -584,14 +584,15 @@ end
 
 Mutate `x` to the coordinates of the cell corresponding to `idx` or `cell`.
 """
-@inline function getcoordinates!(x::Vector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
+@propagate_inbounds function getcoordinates!(x::AbstractVector{Vec{dim, T}}, grid::AbstractGrid, cell::AbstractCell) where {dim, T}
     node_ids = get_node_ids(cell)
-    @inbounds for i in 1:length(x)
-        x[i] = get_node_coordinate(grid, node_ids[i])
+    @boundscheck checkbounds(x, 1:length(node_ids))
+    @inbounds for (i, node_id) in enumerate(node_ids)
```
KnutAM:
```suggestion
    @boundscheck checkbounds(x, keys(node_ids))
    @inbounds for (i, node_id) in pairs(node_ids)
```
`keys(node_ids)` returns Base.OneTo so one less check + the most "correct" I believe.

src/Grid/grid.jl
```diff
@@ -604,16 +605,15 @@ Return the coordinate of the `n`th node in `grid`
...
     return global_nodes
 end
-function _cellnodes!(global_nodes::Vector{Int}, cell::AbstractCell)
+function _cellnodes!(global_nodes::AbstractVector{Int}, cell::AbstractCell)
     @assert length(global_nodes) == nnodes(cell)
-    @inbounds for i in 1:length(global_nodes)
-        global_nodes[i] = cell.nodes[i]
-    end
+    node_ids = get_node_ids(cell)
+    copyto!(global_nodes, node_ids)
```
KnutAM, line 616: While we went back and forth on this - the copyto is a bit slower (but I don't recall an application where this matters a lot?)
```julia
julia> Base.@propagate_inbounds function bar!(ids, t)
           @boundscheck checkbounds(ids, keys(t))
           @inbounds for (i, v) in pairs(t)
               ids[i] = v
           end
           return ids
       end;

julia> foo!(ids, t) = copyto!(ids, t);

julia> node_ids = ntuple(_ -> rand(1:1000), 27);

julia> global_nodes = zeros(Int, length(node_ids));

julia> @btime foo!($global_nodes, $node_ids); #copyto!
  10.552 ns (0 allocations: 0 bytes)

julia> @btime bar!($global_nodes, $node_ids); #ib loop
  4.125 ns (0 allocations: 0 bytes)
```
So fine to leave as is IMO

KnutAM (approved):
LGTM - some optional suggestions

src/Grid/grid.jl
termi-official: Essentially this comes down to https://github.com/JuliaLang/julia/issues/57003

termi-official:
CodeCov is still sad and I have no clue why. It looks like I hit a codecov bug here? Should we proceed merging anyway?

PR #1253, opened 2025-11-25
Use HyperHessians instead of ForwardDiff to compute the hessian in the Landau example

KristofferC:
I dusted off my four-year-old https://github.com/KristofferC/HyperHessians.jl, and it seems to still be quite good. Benchmarks run on 1.10 due to https://github.com/JuliaLang/julia/issues/60241 and https://github.com/JuliaLang/julia/issues/60243.

The package has had very little real world use but seems to work here at least and be quite a bit better.

``` 
# Fdiff
Single: 11.576176 seconds (1.00 M allocations: 2.434 GiB, 2.04% gc time)
Multi:   5.107421 seconds (994.12 k allocations: 2.397 GiB, 0.85% gc time)
Time for ∇²F! in multi: 0.195347 seconds (92.65 k allocations: 46.968 MiB)


# HyperHessians
Single:  8.298370 seconds (10.95 k allocations: 1.731 GiB, 2.51% gc time)
Multi:    3.529221 seconds (88.96 k allocations: 1.739 GiB, 0.37% gc time)
Time for ∇²F! in multi: 0.114557 seconds (2.65 k allocations: 282.797 KiB)
```


termi-official (approved):
After some discussion on slack I think I am convinced for now that the typical user won't find a way to trigger perturbation confusion when sticking to this example.

KristofferC:
Maybe I'll register the package and see if some people have an interest in poking on it before mergin thits.

 I would also like to get a reverse AD backend working on this to see what effect it has on performance. I started a DifferentiationInterface rewrite (https://github.com/Ferrite-FEM/Ferrite.jl/commit/17edffaca9d4f59a83a8621c02104d9d334f15ec#diff-bd098a9030f95d6512cf21c19a06a92acd264a6ca5b80f683ca379a0832abec5) but I got into troubles with every reverse AD backend I tried...

KristofferC:
The DI integration works fine and was very easy (great job) . It's just that no backend work... 😅

gdalle:
Thanks! What kind of error do you get with Enzyme?

wsmoses:
just for fun what if you use Enzyme from within reactant (the sret stuff above doesn't matter there)

wsmoses:
@KristofferC check out https://enzymead.github.io/Reactant.jl/dev/ ! x/ref https://enzymead.github.io/Reactant.jl/dev/tutorials/automatic-differentiation#automatic-differentiation

wsmoses:
yup!

tho ideally for performance you should compile once, and use the compiled object in the hot loop (jit is just shorthand for compile and immediately execute)

KristofferC:
This is now done.

PR #1254, opened 2025-11-25
Fix quadrature rule tests

termi-official:
... not testing the actual integration of the functions

### DIFF
--- a/test/test_quadrules.jl
+++ b/test/test_quadrules.jl
@@ -22,7 +22,7 @@ using Ferrite: reference_shape_value
             for order in orderrange
                 f = (x, p) -> sum([x[i]^p for i in 1:length(x)])
                 qr = QuadratureRule{shape}(rulename, order)
-                @test integrate(qr, (x) -> f(x, 2 * order - 1)) ≈ 0.0; atol = 1.0e-14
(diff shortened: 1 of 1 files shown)


test/test_quadrules.jl
```diff
@@ -59,7 +59,7 @@ using Ferrite: reference_shape_value
         dim = 3
         for order in orderrange
             qr = QuadratureRule{RefTetrahedron}(rulename, order)
-            @test integrate(qr, g) - 0.14 < 0.01
+            @test abs(integrate(qr, g) - 0.142857142857143) < 0.01 / order^2
```
KnutAM: Where is the number coming from? Could you add a comment how you got it? 

And I guess a more canonical test should be? 
```suggestion
            @test integrate(qr, g) \approx 0.142857142857143; atol = 0.01 / order^2
```
(Same suggestion with \approx for other tests)

termi-official: > Could you add a comment how you got it?

I do not understand why this is unclear, so I expaned the string explaining it on top of the testset further.

KnutAM: Sorry, I just didn't find that link when reviewing 😅 But anyways nice with the doi ref!

PR #1255, opened 2025-11-26
Add Lagrange{RefTetrahedron,3}

termi-official:


src/interpolations.jl
```diff
@@ -2132,3 +2207,69 @@ adjust_dofs_during_distribution(::Nedelec{RefHexahedron, 1}) = false
 function get_direction(::Nedelec{RefHexahedron, 1}, shape_nr, cell)
     return get_edge_direction(cell, shape_nr) # shape_nr = edge_nr
 end
+
+# TODO where to put this? I am a lazy man and I won't compute this by hand.
```
KnutAM: Seems like this should be a macro or generated function?

termi-official: Not 100% sure yet. This is for now just a template to get things started on higher order element generators, as we can simply generate these in a systematic way.

src/interpolations.jl
```diff
@@ -932,6 +933,80 @@ function reference_shape_value(ip::Lagrange{RefTetrahedron, 2}, ξ::Vec{3}, i::I
     throw(ArgumentError("no shape function $i for interpolation $ip"))
 end
 
+###################################
+# Lagrange RefTetrahedron order 3 #
+###################################
+getnbasefunctions(::Lagrange{RefTetrahedron, 3}) = 20
```
AbdAlazezAhmed, line 939: maybe 
```suggestion
getnbasefunctions(::Lagrange{RefTetrahedron, order}) where order = (order+1)*(order+2)*(order+3)÷6

```
would be beneficial in the future

src/interpolations.jl
AbdAlazezAhmed: Maybe 
```julia
@generated function do_magic(::Lagrange{RefTetrahedron, order}, ξ::Vec{3, Float64}, i::Int) where {order}

    N = 0
    for p₃ in 0:order
        for p₂ in 0:(order - p₃)
            for p₁ in 0:(order - p₃ - p₂)
                N += 1
            end
        end
    end

    @assert N == 20
    xs = reference_coordinates(Lagrange{RefTetrahedron, order})
    V = zeros(N, N)
    j = 0
    for p₃ in 0:order
        for p₂ in 0:(order - p₃)
            for p₁ in 0:(order - p₃ - p₂)
                j += 1
                for i in 1:N
                    x = xs[i]
                    V[i, j] = x[1]^p₁ * x[2]^p₂ * x[3]^p₃
                end
            end
        end
    end
    C = inv(V)' # These are our coefficients
    ex = :(    ξ_x = ξ[1];
    ξ_y = ξ[2];
    ξ_z = ξ[3];)
    for i in 1:N
        j = 0
        x = xs[i]
        ex_i = :(0.0)
        for p₃ in 0:order
            for p₂ in 0:(order - p₃)
                for p₁ in 0:(order - p₃ - p₂)
                    j += 1
                    ex_i = :($ex_i + $(C[i, j]) * ξ_x^$p₁ * ξ_y^$p₂ * ξ_z^$p₃)
                end
            end
        end
        ex = :($ex; i == $i && return $ex_i)
    end

    return ex
end
```
```julia
julia> @benchmark Ferrite.do_magic($ip, $(Vec(-0.0,-0.0,-0.0)),$(1))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  4.882 ns … 20.719 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.008 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.030 ns ±  0.389 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▅     █▂                                                   
  ▅█▆▃▂▂▄██▄▂▂▂▂▁▂▂▂▁▂▄▄▃▂▂▂▂▂▂▂▂▁▁▂▂▂▂▁▁▁▂▂▂▂▁▁▂▂▁▁▁▁▁▂▂▁▁▂ ▃
  4.88 ns        Histogram: frequency by time        5.86 ns <

julia> @benchmark Ferrite.reference_shape_value($ip, $(Vec(-0.0,-0.0,-0.0)),$(1))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  4.581 ns … 9.742 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.619 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.673 ns ± 0.209 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▅█                                                        
  ▂██▂▂▂▂▄▄▂▂▂▂▂▂▂▂▂▂▂▄▂▂▂▂▂▂▂▁▂▂▂▁▂▂▂▂▁▁▁▁▁▁▁▂▁▂▂▂▂▂▁▁▁▁▁▂ ▂
  4.58 ns        Histogram: frequency by time       5.53 ns <

 ```
 compiler vroom

src/interpolations.jl
termi-official: For the arbitrary order case yes.

src/interpolations.jl
termi-official: I moved everything here for the generator https://gist.github.com/termi-official/2d7f0c5e3cc8ea94921fab08395ab292 . Once I am advanced enough I will move a cleaner version of these into Forge.jl

KnutAM:
As discussed on slack, I think we should consider adding @AbdAlazezAhmed's `@generated` approach instead of more higher order manually implemented versions. At least define the following carefully

Can we have some rule-based system to decide which interpolations
1) Should have a manual (as current) implementations in base Ferrite
   *good for new devs and for anyone wanting to see how add new interpolations*
2) Should have a generic (possibly `@generated`) implementation in base Ferrite
   *good to allow less code and less chance of typos while providing a lot of functionality*
3) Should be implemented outside base Ferrite
   *reduce maintenance burden on base Ferrite*

PR #1257, opened 2025-12-09
Implement Hessians and Identity Mapping for embedded elements

henrij22:
**This does not correctly work at the moment**

This implements two things:
1. Compute Hessians for embedded elements
2. Apply geometry mapping to second derivatives. 

Linking also for reference : https://github.com/Ferrite-FEM/Tensors.jl/pull/188


src/FEValues/GeometryMapping.jl
```diff
@@ -170,13 +182,16 @@ end
...
     H = zero(otimes_returntype(eltype(x), typeof(J)))
     @inbounds for j in 1:n_basefuncs
         d2Mdξ2, dMdξ, _ = reference_shape_hessian_gradient_and_value(gip, ξ, j)
-        J += x[j] ⊗ dMdξ
-        H += x[j] ⊗ d2Mdξ2
+
+        J += otimes_helper(x[j], dMdξ)
+
+        # J += x[j] ⊗ dMdξ
+        # H += x[j] ⊗ d2Mdξ2
```
lijas, line 194: where did the computation of the hessian go?

src/FEValues/GeometryMapping.jl
```diff
@@ -170,13 +182,16 @@ end
 @inline function calculate_mapping(gip::ScalarInterpolation, ξ::Vec{rdim, T}, x::AbstractVector{<:Vec{sdim}}, ::Val{2}) where {T, rdim, sdim}
     n_basefuncs = getnbasefunctions(gip)
     @boundscheck checkbounds(x, Base.OneTo(n_basefuncs))
-    (rdim != sdim) && error("hessian for embedded elements not implemented (rdim=$rdim, sdim=$sdim)")
+    # (rdim != sdim) && error("hessian for embedded elements not implemented (rdim=$rdim, sdim=$sdim)")
```
lijas: We can remove this comment now .

test/test_cellvalues.jl
```diff
@@ -441,9 +441,83 @@
...
+
+            # Test embedded vs non-embedded
+            cv_ref2n = CellValues(qr, ip^2, ip^2; update_hessians = true)
+            coords_refn = [Vec{2}((x[1], x[2])) for x in Ferrite.reference_coordinates(ip)]
+            reinit!(cv_ref2n, coords_refn)
+
+            @test shape_hessian(cv_ref2, qp, 1)[1, 1, 1] ≈ shape_hessian(cv_ref2n, qp, 1)[1, 1, 1]
+            @test shape_hessian(cv_ref2, qp, 1)[1, 2, 1] ≈ shape_hessian(cv_ref2n, qp, 1)[1, 2, 1]
+            @test shape_hessian(cv_ref2, qp, 1)[2, 1, 1] ≈ shape_hessian(cv_ref2n, qp, 1)[2, 1, 1]
+            @test shape_hessian(cv_ref2, qp, 1)[2, 2, 1] ≈ shape_hessian(cv_ref2n, qp, 1)[2, 2, 1]
```
lijas, line 520: I think we can add tests for the function_hessian also, as we do [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/f99e20665346157208ecf872e4c176a7a37a575d/test/test_cellvalues.jl#L103):

```
                coords_nl = [x + rand(x) * 0.01 for x in coords] # add some displacement to nodes
                reinit!(cv, coords_nl)

                _ue_nl = [u_funk(coords_nl[i], V, G, H) for i in 1:n_basefunc_base]
                ue_nl = reinterpret(Float64, _ue_nl)

                for i in 1:getnquadpoints(cv)
                    xqp = spatial_coordinate(cv, i, coords_nl)
                    Hqp, Gqp, Vqp = Tensors.hessian(x -> function_value_from_physical_coord(func_interpol, coords_nl, x, ue_nl), xqp, :all)
                    @test function_value(cv, i, ue_nl) ≈ Vqp
                    @test function_gradient(cv, i, ue_nl) ≈ Gqp
                    if update_hessians
                        @test Ferrite.function_hessian(cv, i, ue_nl) ≈ Hqp
                    end
                end
```

Can you add a similar test in this @testset for hessians? (I am not sure if  function_value_from_physical_coord works with embedded elements)

src/FEValues/FunctionValues.jl
```diff
@@ -174,6 +174,39 @@ required_geo_diff_order(::CovariantPiolaMapping, fun_diff_order::Int) = 1 + fun_
...
 
+# aᵢBᵢⱼₖ = Cⱼₖ
+@inline function dothelper(A::SVector{sdim}, B::SArray{Tuple{sdim, rdim, rdim}}) where {rdim, sdim}
+    return SMatrix{rdim, rdim}((dot(A, B[:, j, k])  for j in 1:rdim, k in 1:rdim))
+end
+
+@inline function dothelper(A::SMatrix{vdim, sdim}, B::SArray{Tuple{sdim, rdim, rdim}}) where {vdim, rdim, sdim}
+    return SArray{Tuple{vdim, rdim, rdim}}(
+        (dothelper(A[i, :], B)[j, k] for i in 1:vdim, j in 1:rdim, k in 1:rdim)
+    )
```
lijas, line 187: I think this type of construction of SArrays is fine, but I dont know how efficiently the compiler can optimize it, maybe someone else can review this part :)

lijas (commented):
Nice!

Can you also add the hessian equation (for embeded elements) to the docs, similar to how we do it here: https://ferrite-fem.github.io/Ferrite.jl/stable/topics/FEValues/#Identity-mapping 

And add a reference to where it is taken from (or perhaps show the derivation :) )

src/FEValues/GeometryMapping.jl
henrij22: Aha, I only added it in above. This code then might not be tested atm. Also, what is the difference between `calculate_mapping(gip::ScalarInterpolation)` and `calculate_mapping(geo_mapping::GeometryMapping)`. They seem quite similar.

src/FEValues/FunctionValues.jl
henrij22: You were right, there are a lot of calls in there.
This implementation takes 24 ns on my system, I found a better version which only takes about 2 ns

test/test_cellvalues.jl
lijas: This might be more difficult than i first thought. For the embedded case, the function_value_from_physical_coord would need some signed distance search to find the parametric coord on the surface

henrij22:
@lijas  Thanks, okay then I have to take a second look. It's a bit tricky to see where the error could be, its either in the computation of the Hessian (of the geometry) or in the `apply_mapping()` function. 
I think we should have the AD test also somewhere in the test suite.

henrij22:
@lijas  So my idea to implement this was to use the existing derivation from https://ferrite-fem.github.io/Ferrite.jl/stable/topics/FEValues/#mapping_theory
and just calculate it with SMatrices. However there seems to be an issue that I can't quite solve. 
I tried it with einstein summation 

```julia
@tullio d2Ndx2[i, j] := Jinv[s, j] * d2Ndξ2[s, r] * Jinv[r, i]
@tullio d2Ndx2[i, j] += -dNdξ[r] * (Jinv[r, k] * H[k, p, s] * Jinv[p, i]) * Jinv[s, j]
```
However it does not pass your test ...
Edit: [in here](https://github.com/henrij22/Ferrite.jl/blob/7260a95f84158b430537947225ec274bc94457b4/src/FEValues/FunctionValues.jl#L257-L259)

lijas:
Cool, I guess you are using Kirchhoff-Love theory then if you need the Hessian? How are you getting C^1 continuity?

For fun, I was "vibe-mathing" with ChatGPT, and I _think_ it was able to derive the hessian. Perhaps an approach you can take if you dont find any good reference in the litterature haha.

KnutAM:
> Maybe if you are interested more, you can also drop me an email (my GH profile should have it) or DM, then we don't spam this thread haha

IMO please keep spamming :D (Or make a separate discussion/issue for shell implementations as this is something the current support is not very good for). These discussions are often super-useful for anyone wanting to implement these things in the future!

henrij22:
@KnutAM thanks for the heads up. The mixed tensors look great!
However I don't really know how to procede here. I've now implemented a transformation I think might be correct. At least for non-curved quadrilaterals I can verify it with the test lijas  proposed (not yet pushed), but for curved ones, the test fails, although I am not sure how to test this.

KnutAM:
I haven't worked much with shells, but I think some key issues (that you might already have hit/solved). If you already have derived things in this direction that you could share, that would be nice and perhaps we can push it in the right direction. I think it could be useful to see the complete formulation that you are aiming at (not necessarily in research but say the KL-formulation for curved shells):

* I think the theory assumes a curvelinear local coordinate system so standard index notation might not work (was googling a bit and found this master thesis https://publications.lib.chalmers.se/records/fulltext/250355/250355.pdf)
* The assumption in the derivations in the docs of [diffeomorphic](https://en.wikipedia.org/wiki/Diffeomorphism) doesn't hold, but I believe this is covered by considering the proper covariant/contravariant bases.

Issue #1229, opened 2025-08-28
Add DifferentiationInterface.jl usage to how-to docs.


KristofferC:
@fredrikekre  had a nice presentation today on JuliaCon on how to choose different AD frameworks using DifferentiationInterface.jl. I think that would be worthwhile to add to the how-to section in the docs.

KristofferC:
This is done

PR #1258, opened 2025-12-09
use DifferentiationInterface in the landau model

KristofferC:
Ref https://github.com/Ferrite-FEM/Ferrite.jl/issues/1229


docs/src/literate-gallery/landau.jl
termi-official, line 1: Should we move this now from Gallery to How-To and rename the title here to something like `Automatic Differentiable Assembly`?

Also, can we add some simple regression test here?

KristofferC:
This is done.

PR #1259, opened 2025-12-09
Add Ferrite.write_facetset

termi-official:


src/Export/VTK.jl
```diff
@@ -283,6 +284,23 @@ function write_nodeset(vtk, grid::AbstractGrid, nodeset::String)
     return vtk
 end
 
+"""
+    write_facetset(vtk::VTKGridFile, grid::AbstractGrid, facetset::String)
```
KnutAM: Would be nice to match the interface from `cellsets` allowing a vector/set of keys? 

Also, since this is not as trivial as cellsets, perhaps it could even be nice to have a lower-level interface providing just an `AbstractVecOrSet{<:FacetIndex}`, not requiring it to be stored in the grid?

src/Export/VTK.jl
```diff
@@ -283,6 +284,23 @@ function write_nodeset(vtk, grid::AbstractGrid, nodeset::String)
...
+
+Write nodal values of 1 for nodes of the faces in `facetset`, and 0 otherwise
+"""
+function write_facetset(vtk, grid::AbstractGrid, facetset::String)
+    z = zeros(getnnodes(grid))
+    for fi in getfacetset(grid, facetset)
+        for i in facets(getcells(grid, fi[1]))[fi[2]]
```
KnutAM: To make work for higher order elements, I think this is required. 
```suggestion
        cell = getcells(grid, fi[1]
        for cellnode in facetdof_indices(geometric_interpolation(cell))[fi[2]]
            i = get_node_ids(cell)[cellnode]
```

test/checksums.sha1
```diff
@@ -1,22 +1,22 @@
...
+f592c19830d0560bb29c736764e8af02b442e49b
 6c41cbbbcdbefda0afb6208295e6eab9cec6f812
-6d86368c529bf5cf7ae28e23a8094501d914d05f
+48125fdc01338ae420e04405699d23417babdae2
 11549a1498556572a8cfa052b3b06b660fbfe05e
-a44e2e480309dbf320a5bf044fd5e97bc73ab28d
+9cd7ab5fffd0c3acaf744d8cf721357e97c04267
 48528031570b719cd5683ef7fbceefe386dcb10b
-c21313b96210736201ad37e0cdc7a5958791041c
+ccc8b6567472b159321b4f096d1ce4555d39fac8
```
KnutAM, line 21: Have you manually checked that all of them produce the expected result?

termi-official: Yes, I have visually confirmed that the face export now works as expected on my end for all of these files. If you have a better way, then I am also happy to try.

src/Export/VTK.jl
```diff
@@ -283,6 +284,27 @@ function write_nodeset(vtk, grid::AbstractGrid, nodeset::String)
     return vtk
 end
 
+"""
+    write_facetset(vtk::VTKGridFile, grid::AbstractGrid, facetsetname::String [, facetset::AbstractVecOrSet{FacetIndex}])
+
+Write nodal values of 1 for nodes of the faces in `facetset`, and 0 otherwise
+"""
+function write_facetset(vtk, grid::AbstractGrid, facetsetname, facetset::AbstractVecOrSet{FacetIndex})
```
KnutAM: Missed this, should have facetsetname as the last to be consistent with the other funktions, right?

termi-official: I do not have a good intuition about what good argument orderings are. :D I changed it anyway in [69a74d5](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1259/commits/69a74d5944b4c9c6f93d04da4ef23de57d94a549)

PR #1265, opened 2025-12-29
avoid `Core.Box` in the package

KristofferC:
Uses https://github.com/JuliaLang/julia/pull/60478 to find `Core.Box` entries in the package and remove them. After this, there are no such closure boxes in the package.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1189,8 +1189,8 @@ function _add!(
             union!(all_node_idxs, nodes)
             for n in nodes
                 x = get_node_coordinate(grid, n)
-                min_x = Tx(i -> min(min_x[i], x[i]))
-                max_x = Tx(i -> max(max_x[i], x[i]))
+                min_x = Tx(min.(min_x.data, x.data))
+                max_x = Tx(max.(max_x.data, x.data))
             end
         end
         all_node_idxs_v = collect(all_node_idxs)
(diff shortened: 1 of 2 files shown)


src/Dofs/ConstraintHandler.jl
```diff
@@ -1189,8 +1189,8 @@ function _add!(
             union!(all_node_idxs, nodes)
             for n in nodes
                 x = get_node_coordinate(grid, n)
-                min_x = Tx(i -> min(min_x[i], x[i]))
-                max_x = Tx(i -> max(max_x[i], x[i]))
+                min_x = min.(min_x, x)
+                max_x = max.(max_x, x)
```
KnutAM: I think this converts to `Vector`?
```julia
julia> x = rand(Vec{3});

julia> min.(x, 2*x)
3-element Vector{Float64}:
 0.43556649868831676
 0.06759192439197814
 0.15171836180398857
```

Does `Tx(min.(min_x.data, x.data))` avoid the Core.Box? 
Or using `min_x = Ref(Tx(i -> typemax(eltype(Tx))))`?

KristofferC: Ah, yeah, we don't have broadcasting setup properly in Tensors.jl. I'll tweak it.

KnutAM: Or perhaps we should fix broadcasting properly in Tensors (even if breaking)?

PR #1270, opened 2026-01-19
Argyris element

lijas:
This is an implementation of the [Argyris](https://defelement.org/elements/examples/triangle-argyris-5.html) element, which is a triangular element with fifth-order basis functions that achieves \(C^1\) continuity across element boundaries (apparently, fifth order or higher is required to achieve \(C^1\) continuity). It is a Hermitian type element with "derivative"-dofs at the nodes. The element was invented in 1960 but is typically not used due to its high order and the complexity of the implementation of the isoparametric mapping.

The mapping of the shape functions is not as straightforward as the [identity mapping](https://ferrite-fem.github.io/Ferrite.jl/stable/topics/FEValues/#Identity-mapping) N^I_x = N^I_\xi. Instead, N_x is expressed as a linear combination of N_\xi through a sparse M-matrix, i.e. N_x = M\,N_\xi. Although I do not fully understand the underlying theory, the construction of the \(M\)-matrix is presented in [Kirby 2018](10.5802/smai-jcm.33) and [Domínguez 2008](https://doi.org/10.1145/1377612.1377620).

I think this element could be a quick and easy option for users to get \(C^1\) meshes, before exploring more advanced methods such as IGA.
 
- [x] Perhaps implement a solution to the plate equation in the gallery.


src/FEValues/FunctionValues.jl
```diff
@@ -272,3 +274,227 @@ end
...
+function apply_mapping!(funvals::Ferrite.FunctionValues{DO}, ::ArgyrisMapping, q_point::Int, mapping_values, cell, coords) where DO 
+    ip_geo = Ferrite.geometric_interpolation(cell)
+    @assert ip_geo isa Lagrange{RefTriangle, 1} "Only linear geometries allowed for Argyris interpolation"
+    @assert DO < 3
+
+    @show DO
+    
+    #Compute data required for the argyris mapping
+    #TODO: This can be done once per element, but we have to recompute for each quadrature point with the current setup
+    argyris_data = ArgyrisData(ip_geo, coords)
```
KnutAM: We could in `reinit!` instead add something like 
```julia
cell_mapping = initiate_mapping(geo_mapping, cell, x) # Cell-wise mapping calculations
for q_point in 1:getnquadpoints(cv)
    # Per q_point mapping calculations
    mapping = calculate_mapping(cell_mapping, geo_mapping, q_point, x)
    ...
end
```
and change 
```julia
struct MappingValues{CT, JT, HT, NT}
    cell::CT      # AbstractCell or Nothing
    J::JT # dx/dξ # Jacobian
    H::HT # dJ/dξ # Hessian
    extras::NT    # NamedTuple with extra mapping data for special mappings. Access with `basename(typeof(mapping_type(ip)))`
end
```
Not sure how this affects performance though...

KnutAM (commented):
Very cool! I'm sure we can adapt the `GeometryValues`/`MappingValues` if needed to make this more efficient. 

Before doing that, it could perhaps be nice to know what calculations would be needed for nonlinear geometry to make sure the infrastructure remains valid for such cases...

termi-official:
Really cool, totally appreciate this work!

However, I was already wondering if we do not overload CellValues with a bit too much stuff. Maybe we should consider for the elements with C1 mapping a different CellValues object (CellValues2 for the lack of creatifity atm)? Or is there a large overlap between these? If it is just a different geoemtry mapping, then we can also introduce a new geometry mapping struct for this case.

Regarding the Dirichlet boundry conditions, we could fall back for a Nitsche method for now if that is blocking.

src/FEValues/FunctionValues.jl
lijas: Yes, that would be a nice generalization and optimization. I will try to do some timings and if the reinit! of the Argyris element is not to bad compared to e.g. the element routine, this could be an optimization for the future.

src/FEValues/FunctionValues.jl
```diff
@@ -272,3 +274,227 @@ end
...
+#The shape functions on the physical element are related to the shape functions on reference element Nξ via a sparse matrix M: Nx = M*Nξ
+#Since M is sparse, we avoid creating the M-matrix and compute Nx directly.
+#For more information see: 
+# Robert C. Kirby. A general approach to transforming finite elements. The SMAI Journal of computational mathematics (2018)
+function apply_mapping!(funvals::Ferrite.FunctionValues{DO}, ::ArgyrisMapping, q_point::Int, mapping_values, cell, coords) where DO 
+    ip_geo = Ferrite.geometric_interpolation(cell)
+    @assert ip_geo isa Lagrange{RefTriangle, 1} "Only linear geometries allowed for Argyris interpolation"
+    @assert DO < 3
+
+    @show DO
```
KristofferC: just noting this

src/FEValues/FunctionValues.jl
```diff
@@ -272,3 +274,227 @@ end
...
+            Nx[3+o2] = N23x[2]
+            if DO > 0 # -- Shape gradient
+                dN23dξ = @SVector [dNdξ[2+o1], dNdξ[3+o2]]
+                dN23dx = J*dN23dξ + @SVector [m3[1]*dNdξ[b1] + m4[1]*dNdξ[b2],
+                                              m3[2]*dNdξ[b1] + m4[2]*dNdξ[b2]] 
+                dNdx[2+o1] = dN23dx[1] ⋅ Jinv
+                dNdx[3+o2] = dN23dx[2] ⋅ Jinv
+                if DO > 1 # -- Shape hessian
+                    d²N23dξ² = @SVector [d²Ndξ²[2+o1], d²Ndξ²[3+o2]]
+                    d²N23dx² = J*d²N23dξ² + @SVector [m3[1]*d²Ndξ²[b1] + m4[1]*d²Ndξ²[b1],
```
KristofferC: Is there a typo here (`d²Ndξ²[b1]` should be `d²Ndξ²[b2]`) and similar on the next line? Haven't done any derivation, but reasoning just from symmetry with e.g. 

```
dN23dx = J*dN23dξ + @SVector [m3[1]*dNdξ[b1] + m4[1]*dNdξ[b2],
```

where `b1` and `b2` are used.

src/FEValues/FunctionValues.jl
```diff
@@ -272,3 +274,227 @@ end
...
+    Jinv = DO > 0 ? inv(mapping_values.J) : nothing
+    H = DO > 1 ? mapping_values.H : nothing
+    Jinv2 = vdim==1 ? Jinv : otimesu(Jinv, Jinv)
+
+    t = @SVector [SVector{2}(argyris_data.t[i]) for i in 1:3]
+    τ = @SVector [SVector{3}(argyris_data.t[i][1]^2, 2*argyris_data.t[i][1]*argyris_data.t[i][2], argyris_data.t[i][2]^2) for i in 1:3]
+    J = SMatrix{2,2}(argyris_data.J)
+    Θ = @SMatrix [J[1,1]^2 J[1,2]*J[1,1] J[1,2]^2; 
+                  2*J[1,1]*J[2,1] J[1,2]*J[2,1] + J[1,1]*J[2,2] 2*J[2,2]*J[1,2]; 
+                  J[2,1]^2 J[2,1]*J[2,2] J[2,2]^2]
```
termi-official: Would https://github.com/Ferrite-FEM/Tensors.jl/pull/236 help here to use Tensors instead of StaticArrays?

src/interpolations.jl
```diff
@@ -2132,3 +2132,88 @@ adjust_dofs_during_distribution(::Nedelec{RefHexahedron, 1}) = false
...
+end
+
+conformity(::Argyris{RefTriangle, 5}) = H1Conformity() #TODO, change to H2Conformity
+mapping_type(::Argyris{RefTriangle,5}) = ArgyrisMapping()
+
+function get_direction(::Argyris{RefTriangle, 5}, shape_nr, cell)
+    (0 < shape_nr < 19) && return 1
+    edge_nr = shape_nr-18
+    return get_edge_direction(cell, edge_nr)
+end
```
termi-official:
```suggestion
end

```

termi-official (commented):
Can you also add an convergence rate test to show that everything is really implemented as it should be? I have implemented a simple test for most interpolations here https://github.com/Ferrite-FEM/Ferrite.jl/blob/master/test/integration/test_simple_scalar_convergence.jl which might be a starting point.

src/FEValues/FunctionValues.jl
lijas: Good catch. I have no tests for the hessian yet (other than solving the plate equations which visually gave reasonable results), so I should test the hessian aswell.

lijas:
@termi-official I tried adding the element to the convergence test but I could not get it to work, because (I think) the Dirichlet conditions prescribed both w and its gradient on the boundary. But yeah, I want to verify the implementation some way like this.

docs/src/gallery/index.md
```diff
@@ -20,6 +20,16 @@ Neumann boundary conditions and the method of manufactured solutions.
 
 ---
 
+#### [Plate equation](plate_equation.md)
```
termi-official: Maybe we can be specific here.

```suggestion
#### [Kirchhoff-Love Plate Equation](plate_equation.md)
```

src/FEValues/FunctionValues.jl
KristofferC: Has this been updated/fixed?

src/FEValues/FunctionValues.jl
```diff
@@ -289,3 +291,240 @@ end
...
+    n̂ = (Vec((1 / √2, 1 / √2)), Vec((-1.0, 0.0)), Vec((0.0, -1.0)))
+    t̂ = (Vec((1 / √2, -1 / √2)), Vec((0.0, 1.0)), Vec((-1.0, 0.0)))
+    B = zeros(Tensor{2, 2}, 3)
+    for i in 1:3
+        Ĝ = Tensor{2, 2}(hcat(n̂[i], t̂[i])')
+        G = Tensor{2, 2}(hcat(n[i], t[i])')
+        B[i] = Ĝ ⋅ J' ⋅ G'
+    end
+
+    return ArgyrisData(t, l, Tuple(B), J)
```
KnutAM: Should avoid allocation? 

```suggestion
    ts = (t1 / l1, t2 / l2, t3 / l3)
    n̂s = (Vec((1 / √2, 1 / √2)), Vec((-1.0, 0.0)), Vec((0.0, -1.0)))
    t̂s = (Vec((1 / √2, -1 / √2)), Vec((0.0, 1.0)), Vec((-1.0, 0.0)))
    B = map(ts, n̂s, t̂s) do t, n̂, t̂
        n = rotate(t, pi / 2)
        Ĝ = Tensor{2, 2}((n̂..., t̂...))' # Edit: hcat creates matrix, I think
        G = Tensor{2, 2}((n..., t...))' # this is equivalent but noalloc
        Ĝ ⋅ J' ⋅ G'
    end

    return ArgyrisData(ts, l, B, J)
```

test/test_continuity.jl
```diff
@@ -115,9 +115,14 @@
 
     continuity_function(ip::Interpolation) = continuity_function(Ferrite.conformity(ip))
     continuity_function(::Ferrite.H1Conformity) = ((v, _) -> v)
+    continuity_function(::Ferrite.H2Conformity) = ((∇v, n) -> ∇v ⋅ n)
```
KnutAM: This is not full $H^2$ conformity, why not testing the gradient directly?

lijas: Fixed.

src/FEValues/FunctionValues.jl
lijas: This works if I put the everything inside a function. This actually spead things up a bit.

src/FEValues/FunctionValues.jl
lijas: No, the hessian dofs contains d2Ndx2, d2Ndy2 and d2Ndxdy (no d2Ndydx), so a 3x3 matrix is required.

test/test_continuity.jl
KnutAM: Nice when making the test harder that it passes directly 😁

docs/src/reference/fevalues.md
```diff
@@ -46,8 +46,11 @@ getnbasefunctions(::Ferrite.AbstractValues)
 function_value
 function_gradient
 function_symmetric_gradient
+function_hessian
 function_divergence
 function_curl
+function_laplacian
+spatial_coordinate
```
KristofferC: Looks like this causes the doc build to fail.

docs/src/literate-gallery/plate_equation.jl
```diff
@@ -0,0 +1,170 @@
...
+            for j in 1:getnbasefunctions(facetvalues)
+                v = shape_value(facetvalues, iqp, j)
+                ke[i, j] += penalty * (w * v) * dV
+            end
+        end
+    end
+    return
+end
+
+# Here we create a standard assembly routine.
```
termi-official: Not so standard with the penalty approach on the boundary.

docs/src/literate-gallery/plate_equation.jl
```diff
@@ -0,0 +1,170 @@
...
+        getfacetset(grid, "top"),
+        getfacetset(grid, "bottom"),
+    )
+
+    for celldata in FacetIterator(dh, ∂Ω)
+        fill!(ke, 0.0)
+        reinit!(facetvalues, celldata)
+        bc_routine!(ke, facetvalues, penalty)
+        assemble!(assembler, celldofs(celldata), ke)
+    end
```
termi-official: Maybe we can briefly comment on this and remove the allocations due to the union?

docs/src/literate-gallery/plate_equation.jl
```diff
@@ -0,0 +1,170 @@
...
+E = 200.0e9       # Stiffness
+t = 0.01        # Thickness
+ν = 0.3         # Poisson's radtio
+penalty = 1.0e12  # Penalty stiffness
+D = (E * t^3) / (12 * (1 - ν^2)) # Flexural stiffness
+
+grid = generate_grid(Triangle, (20, 20), Vec((0.0, 0.0)), Vec((L, L)))
+
+# We use the Argyris interpolation as and FE approximation.
+ip = Argyris{RefTriangle, 5}()
```
termi-official, line 57: Are there any Argyris elements of other order or on other shapes?

docs/src/literate-gallery/plate_equation.jl
```diff
@@ -0,0 +1,170 @@
...
+
+# For the CellValues and FacetValues we need to requeest to update the hessians.
+qr = QuadratureRule{RefTriangle}(8)
+cellvalues = CellValues(qr, ip; update_hessians = true);
+
+fqr = FacetQuadratureRule{RefTriangle}(8)
+facetvalues = FacetValues(fqr, ip; update_hessians = true);
+
+# For the current BVP, There is a known analtyical solution (Navier's solution) that we can compare with.
+# This is a good testing strategy for PDE codes and known as the method of manufactured solutions.
```
termi-official: This is slightly misleading. Manufactured solutions typically refer to the case where we plug in a u and get some forcing terms out, which we then use for the assembly. Which is usually distinguished from hand-computed analytical solutions to specific BVPs.

src/FEValues/FunctionValues.jl
```diff
@@ -289,3 +291,244 @@ end
...
+
+    return ArgyrisData(t, l, B, J)
+end
+
+#Apply mapping for the Argyris element.
+#The shape functions on the physical element are related to the shape functions on reference element Nξ via a sparse matrix M: Nx = M*Nξ
+#Since M is sparse, we avoid creating the M-matrix and compute Nx directly.
+#For more information see:
+# Robert C. Kirby. A general approach to transforming finite elements. The SMAI Journal of computational mathematics (2018)
+function apply_mapping!(funvals::FunctionValues{DO}, ::ArgyrisMapping, q_point::Int, mapping_values, cell, coords) where {DO}
```
termi-official: Do we really need to introduce a wider signature or can we just use https://github.com/Ferrite-FEM/Ferrite.jl/blob/511e7150d40cc60171b561f5e6ab4a9632087c50/src/iterators.jl#L98 here ?

src/FEValues/FunctionValues.jl
```diff
@@ -289,3 +291,244 @@ end
...
+            if DO > 0
+                dNjdx = dNdξ[j] * B[i][1, 1]
+                dNdx[j] = dNjdx ⋅ Jinv
+                if DO > 1
+                    d²Njdx² = d²Ndξ²[j] * B[i][1, 1]
+                    d²Ndx²[j] = _hessian_helper(d²Njdx², dNdx[j], H, Jinv2)
+                end
+            end
+        end
+        #_M[b1,b1] = B[i][1,1]
```
termi-official:
```suggestion
```

src/FEValues/FunctionValues.jl
```diff
@@ -289,3 +291,244 @@ end
...
+        # row = (i-1)*6
+        # _M[1+row, 1+ row] = 1.0
+        # _M[1+row, b1_scalar] = m1
+        # _M[1+row, b2_scalar] = m2
+        # _M[(2:3) .+ row, (2:3) .+ row] = J
+        # _M[(2:3) .+ row, b1_scalar] = m3
+        # _M[(2:3) .+ row, b2_scalar] = m4
+        # _M[(4:6) .+ row, (4:6) .+ row] = Θ
+        # _M[(4:6) .+ row, b1_scalar] = m5
+        # _M[(4:6) .+ row, b2_scalar] = m6
```
termi-official:
```suggestion
```

src/interpolations.jl
```diff
@@ -2132,3 +2133,90 @@ adjust_dofs_during_distribution(::Nedelec{RefHexahedron, 1}) = false
...
+
+############################
+# Argyris triangle order 5 #
+############################
+# Dof order:
+# Nodes 1: δ, ∇x, ∇y, ∇∇xx, ∇∇xy, ∇∇yy
+# Nodes 2: δ, ∇x, ∇y, ∇∇xx, ∇∇xy, ∇∇yy
+# Nodes 3: δ, ∇x, ∇y, ∇∇xx, ∇∇xy, ∇∇yy
+# Edges 1-3: ∇_n (normal derivative, normal pointing outward)
+struct Argyris{refshape, order} <: ScalarInterpolation{refshape, order}
```
termi-official, line 2393: Isn't this fixed to
```suggestion
struct Argyris <: ScalarInterpolation{RefTriangle, 5}
```
?

docs/src/literate-gallery/plate_equation.jl
lijas: 👍

docs/src/literate-gallery/plate_equation.jl
lijas: I think the union is pretty clear from a tutorial perspective and we generally dont worry about allocations here. But I added a small comment here.

docs/src/literate-gallery/plate_equation.jl
lijas: Not really, but I think it is possible to define Argyris for quads aswell.

docs/src/literate-gallery/plate_equation.jl
lijas: 👍

src/FEValues/FunctionValues.jl
lijas: There is no CellCache in this function call so I have to pass coords.

src/FEValues/FunctionValues.jl
lijas: 👍

src/FEValues/FunctionValues.jl
lijas: 👍

src/interpolations.jl
lijas: I think there is only one Argyris element (triangle, order 5), but I kind of like the that you have to be explicit :) But maybe other people have different opinions.

docs/src/literate-gallery/plate_equation.jl
```diff
@@ -0,0 +1,173 @@
...
+                ke[i, j] += penalty * (w * v) * dV
+            end
+        end
+    end
+    return
+end;
+
+# Next, we assemble the contributions from the element plate stiffnesses and the stiffness arising from the penalty-based boundary constraint.
+function doassemble!(
+        cellvalues::CellValues, facetvalues::FacetValues, K::SparseMatrixCSC, f::Vector, dh::DofHandler, D::Float64, q0::Float64, penalty::Float64
```
KnutAM: drive-by but just make a bit shorter, use mutated values first, and avoid using the sparse arrays explicitly. When writing out `cellvalues` and `facetvalues`, I also think we don't need the types? 
```suggestion
        K, f, cellvalues::CellValues, facetvalues::FacetValues, dh::DofHandler, D, q0, penalty
```

src/FEValues/FunctionValues.jl
termi-official: Yea, the variable naming got unfortunately a bit messy. We sometimes use cell for CellCache and sometimes for AbstractCell.

docs/src/literate-gallery/plate_equation.png
termi-official, line 1: Assets go into `Ferrite.asset_url`. @KnutAM I think we never put instructions somewhere on how to upload assets there?

KnutAM: Should probably be in contributing.md too, but is described briefly in docstring in devdocs https://ferrite-fem.github.io/Ferrite.jl/stable/devdocs/documentation/#devdocs-documentation

lijas: This seems overly complicated for adding a picture for a tutorial. What is the reason for this? :) Why are not all tutorial figures downloaded from asset_url?

termi-official: > What is the reason for this? :) 

Git. Syncing becomes increasingly expensive, because you store the full binary stuff and touching it in some way means you have to store the previous and the new version. We have tanked the Ferrite-FEM Github memory contingent in the past already by not being careful here. An alternative is to use git lfs, but it this comes with its own issues.

> Why are not all tutorial figures downloaded from asset_url?

I was also wondering how we ended up with not being consistent here.

lijas:
I did some refactoring such that other Hermitian type elements can be implmeneted more easily in the future.

- The transformation matrix is computed and applied per element and not per quadrature point now. I store are transformation-matrix in the `FunctionValues `now.
- The Argyris element does not need to create a custom `ArgyrisMapping `anymore... the `IdentityMapping` is kept which is more mathematically correct.
- I refined the `reinit_need_cell` logic. Previously `reinit_need_cell` was only based on mapping-type, but interpolations with dofs with a "direction" also needs the cell.
- I rebased on #1493 to use dof-functional information

src/FEValues/FunctionValues.jl
```diff
@@ -179,19 +195,28 @@ end
 end
 
 # Identity mapping
-@inline function apply_mapping!(::FunctionValues{0}, ::IdentityMapping, ::Int, mapping_values, args...)
+@inline function apply_mapping!(funvals::FunctionValues{0}, ::IdentityMapping, q_point::Int, mapping_values, cell)
+    #Some elements also need to flip the direction of the dof, e.g. if they have normal gradient dofs.
+    @inbounds for j in 1:getnbasefunctions(funvals)
+        d = get_direction(funvals.ip, j, cell)
+        funvals.Nx[j, q_point] *= d
+    end
```
lijas, line 203: For standard lagrange elements, this seems to be all compiled away :) 

```
qr = QuadratureRule{RefTriangle}(2)
ip = Lagrange{Ferrite.RefTriangle, 1}()
cv = CellValues(qr, ip; update_gradients=false)

mapping_values = Ferrite.MappingValues(0, 0)
@code_llvm optimize=true Ferrite.apply_mapping!(
    cv.fun_values, Ferrite.IdentityMapping(), 1, mapping_values, Triangle((1,2,3))
)
```

```
; Function Signature: apply_mapping!(Ferrite.FunctionValues{0, Ferrite.Lagrange{Ferrite.RefSimplex{2}, 1}, Array{Float64, 2}, Nothing, Nothing, Nothing, Nothing, Nothing}, Ferrite.IdentityMapping, Int64, Ferrite.MappingValues{Int64, Int64}, Ferrite.Triangle)
;  @ /home/elias/code/Ferrite.jl/src/FEValues/FunctionValues.jl:198 within `apply_mapping!`
define void @"julia_apply_mapping!_36756"(ptr nocapture noundef nonnull readonly align 8 dereferenceable(16) %"funvals::FunctionValues", i64 signext %"q_point::Int64", ptr nocapture noundef nonnull readonly align 8 dereferenceable(16) %"mapping_values::MappingValues", ptr nocapture noundef nonnull readonly align 8 dereferenceable(24) %"cell::Triangle") #0 {
top:
;  @ /home/elias/code/Ferrite.jl/src/FEValues/FunctionValues.jl:204 within `apply_mapping!`
  ret void
}
```

PR #1271, opened 2026-01-19
Replace C_voigt by E_voigt for plane stress

MelanieInky:
There's an inconsistency (I think) here and running the code gives an error because C_voigt is defined instead of E_voigt, so I changed that.

### DIFF
--- a/docs/src/literate-tutorials/linear_elasticity.jl
+++ b/docs/src/literate-tutorials/linear_elasticity.jl
@@ -242,7 +242,7 @@ C = gradient(ϵ -> 2 * Gmod * dev(ϵ) + 3 * Kmod * vol(ϵ), zero(SymmetricTensor
 #md #     the elastic stiffness tensor should be changed to reflect this. The plane stress elasticity
 #md #     stiffness matrix in Voigt notation for engineering shear strains, is given as
 #md #     ```math
-#md #     \underline{\underline{\boldsymbol{E}}} = \frac{E}{1 - \nu^2}\begin{bmatrix}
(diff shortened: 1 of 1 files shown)


docs/src/literate-tutorials/linear_elasticity.jl
```diff
@@ -250,7 +250,7 @@ C = gradient(ϵ -> 2 * Gmod * dev(ϵ) + 3 * Kmod * vol(ϵ), zero(SymmetricTensor
 #md #     ```
 #md #     This matrix can be converted into the 4th order elastic stiffness tensor as
 #md #     ```julia
-#md #     C_voigt = Emod * [1.0 ν 0.0; ν 1.0 0.0; 0.0 0.0 (1-ν)/2] / (1 - ν^2)
+#md #     E_voigt = Emod * [1.0 ν 0.0; ν 1.0 0.0; 0.0 0.0 (1-ν)/2] / (1 - ν^2)
 #md #     C = fromvoigt(SymmetricTensor{4,2}, E_voigt)
```
KnutAM: Perhaps using `C_voigt` is better for consistency with `C`
```suggestion
#md #     C_voigt = Emod * [1.0 ν 0.0; ν 1.0 0.0; 0.0 0.0 (1-ν)/2] / (1 - ν^2)
#md #     C = fromvoigt(SymmetricTensor{4,2}, C_voigt)
```

KnutAM (commented):
Nice catch 🚀

Just a minor suggestion

MelanieInky:
Then I gather the plane stress elasticity stiffness matrix should be renamed as well in the details.

PR #1272, opened 2026-01-27
Some theoretical background for the L2 projector

termi-official:


src/L2_projection.jl
```diff
@@ -63,6 +63,19 @@ projected = project(proj, vals)
...
+### Background
+
+The operation done by the L2-projector can be interpreted as the Galerkin-projection of ``f`` onto ``u``:
+```math
+    \\int v (u - f) \\ \\mathrm{d}\\Omega = 0 \\quad \\forall v \\in U_h(\\Omega),
+```
+An alternative interpretation here is that we minimize the difference between ``f`` and ``u`` with the functional (induced by the ``L_2`` norm):
+```math
+    \\min_u \\int \\frac{1}{2} (u - f)^2 \\ \\mathrm{d}\\Omega \\, .
+```
```
KnutAM: Should we reverse the order and stating that the 1st equation is a result of setting the directional derivative of the 2nd equal to zero? 

Would it also make sense to introduce the FE-approximation explicitly above then to show the discrete equation system?

termi-official: >Should we reverse the order and stating that the 1st equation is a result of setting the directional derivative of the 2nd equal to zero?

Not necessarily. The first equation is the orthogonality and it just happens here that it is also the variation of the latter and thereby connected.

> Would it also make sense to introduce the FE-approximation explicitly above then to show the discrete equation system?

What do you mean?

KnutAM: > What do you mean?

```math
\underbrace{\int N_i N_j \ \mathrm{d}\Omega}_{M_{ij}}\ a_j = \underbrace{\int N_i f \ \mathrm{d}\Omega}_{f_i} \quad \forall v \in U_h(\Omega)
```
And then say that `M` is Cholesky factorizated upon close?

termi-official: Can do.

termi-official: Something like this maybe?

fredrikekre:
I still think theoretical background is Explanation (or what we call topic guides). I interpret reference as "what you need while coding", i.e. how to use the thing ("reference documentation serves the user who is at work").

docs/src/topics/projections.md
```diff
@@ -0,0 +1,30 @@
+# [Projections and Transfer](@id man-projection-transfer)
```
KnutAM, line 1:
```suggestion
# [Projection](@id man-projection-transfer)
```
Since we only have projection for now?

docs/src/topics/projections.md
```diff
@@ -0,0 +1,30 @@
...
+
+This section is concerned with projection and transfer techniques implemented in Ferrite.
```
KnutAM, line 3:
```suggestion
Ferrite can [`project`](@ref) values stored in the quadrature points onto a finite element interpolation, making it suitable for e.g. plotting. Currently, [L2 projection](@ref man-l2projection) is implemented.
```

docs/src/topics/projections.md
```diff
@@ -0,0 +1,30 @@
...
+
+## [L2 Projection](@id man-l2projection)
+
+A commonly used projection technique in finite element problems is the L2 projection which is implemented in Ferrite as the `[L2Projector](@ref)`.
+This technique is concerned with finding the best field for some function, or after numerical quadrature for some given data points.
+To be mathematically more precise, the operation done by the L2-projector can be interpreted as the Galerkin-projection of ``f`` onto ``u``:
```
KnutAM, line 9:
```suggestion
This method finds the closest (in the L2 norm) projection of some function onto a finite element interpolation. In practice, the function is often described by its values in the quadrature points and numerical integration is used directly.

The L2-projection is the Galerkin-projection of the function ``f`` onto the function space ``U(\Omega)``, i.e. find `u \in U(\Omega)` such that
```

docs/src/topics/projections.md
```diff
@@ -0,0 +1,30 @@
...
+\int v u \ \mathrm{d}\Omega = \int v f \ \mathrm{d}\Omega \quad \forall v \in U(\Omega),
+```
+where we can see that applying a numerical quadrature rule to the right hand side shows us that we just need some samples of ``f`` at the quadrature points for this technique to work.
+
+After finite element discretization the resulting problem to solve becomes
+```math
+\underbrace{\int N_i N_j \ \mathrm{d}\Omega}_{M_{ij}}\ u^h_j = \underbrace{\int N_i f \ \mathrm{d}\Omega}_{b_i}
+```
+where ``u^h_j`` denotes the unknown dofs of the finite-dimensional discrete space to project onto (e.g. Lagrange polynomials) with ansatz and test functions ``N``.
+``b_i`` denotes the right hand side of the finite-dimensional discrete linear system. For vectorized problems we simply apply the projection component-wise.
```
KnutAM, line 30:
```suggestion
``b_i`` denotes the right hand side of the finite-dimensional discrete linear system. The L2 projection of vector- or tensor-valued quantities onto scalar-valued finite element spaces is computed by projecting component-wise.
```

To prep for adding details from https://github.com/Ferrite-FEM/Ferrite.jl/pull/1161 here.

src/L2_projection.jl
```diff
@@ -34,11 +34,11 @@ different cell sets with `add!` before `close!`ing the projector,
...
 The `L2Projector` acts as the integrated left hand side of the projection equation:
-Find projection ``u \\in U_h(\\Omega) \\subset L_2(\\Omega)`` such that
+Find projection ``u \\in U(\\Omega) \\subset L_2(\\Omega)`` such that
 ```math
-\\int v u \\ \\mathrm{d}\\Omega = \\int v f \\ \\mathrm{d}\\Omega \\quad \\forall v \\in U_h(\\Omega),
+\\int v u \\ \\mathrm{d}\\Omega = \\int v f \\ \\mathrm{d}\\Omega \\quad \\forall v \\in U(\\Omega),
 ```
-where ``f \\in L_2(\\Omega)`` is the data to project. The function space ``U_h(\\Omega)``
+where ``f \\in L_2(\\Omega)`` is the data to project. The function space ``U(\\Omega)``
 is the finite element approximation given by the interpolations `add!`ed to the `L2Projector`.
```
KnutAM, line 42: Perhaps make it cleaner to not have it at all here then?

And if not removing, regarding the change from `U_h` to `U`, shouldn't it be `U_h` here because we say that it is the finite element approximation, whereas in the the initial part of the topic guide it is described as a general function space which later is restricted (discretized) with finite elements? 

```suggestion
a function space. See the corresponding [topic guide](@ref man-l2projection)
for the mathematical details. 
To define the function space, add interpolations for
different cell sets with `add!` before `close!`ing the projector,
see the example below.
```

docs/src/topics/projections.md
termi-official: We also have the PointEvalHandler as a transfer function. I just didn't wanted to include too many changes in one PR.

KnutAM: OK, I read transfer as transferring between grids. But if following the API-section, this should still be two separate sub-headings then?

termi-official: > I read transfer as transferring between grids

Yes. This is one examplaric use-case of the PointEvalHandler. Does that make sense?

> should still be two separate sub-headings then?

This is the heading for the full section. We also have this for example in the `Grid and AbstractGrid` section.


I am still open for improvements on the name. Postprocessing is misleading tho.

KnutAM: > Postprocessing is misleading tho.

True, but that also a problem in the API section then, and keeping these in sync I think is better? And if we decide on a good way to split we do it for both? 

> Yes. This is one examplaric use-case of the PointEvalHandler. Does that make sense?

True, but another is plotting along a line as put in the postprocessing how-to, and I think finding that functionality as a transfer could be hard for users? 

> This is the heading for the full section.

Yes, wasn't against a general heading (just that moving it to include export, if we do that). I'm also not sure what the best name here would be. 

> I just didn't wanted to include too many changes in one PR.

Sure, that makes sense (but hard to see here, so IMO better add a general header then :) )

termi-official:
I think this is a bigger change than I initially anticipated. I just wanted to upstream here some stuff from the Ferrite paper. Probably easiest if this is picked up after we have fixed the docs.

Issue #1133, opened 2025-01-20
assemble! for non-square J missing


termi-official:
`assemble!` is missing a dispatch for insertion of dofs for non-square element matrices.

KnutAM:
Should work for the basic assembler? Or did that get lost? 
https://github.com/Ferrite-FEM/Ferrite.jl/pull/471

I don't think we can create sparsity patterns for non-square element matrices, so one has to use that one AFAIK.

termi-official:
> Should work for the basic assembler? Or did that get lost?

Does not work for CSC.

> I don't think we can create sparsity patterns for non-square element matrices, so one has to use that one AFAIK.

We can create arbitrary sparsity patterns with the new API. Also note that the element matrix being non-square does not have to do much with the global sparsity pattern. 

xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/816

KnutAM:
True! Somehow forgot😅

Issue #968, opened 2024-06-01
Extend functionallity of assemble! method


Joroks:
Currently, assemble! assumes, that `Ke` is a Matrix. In some situations, the user might want to assemble single values into the matrix. This could look something like this:

```julia
function assemble!(a::Ferrite.Assembler{T}, rowdofs::Int, coldofs::Int, Ke::T) where {T}
    push!(a.V, Ke)
    push!(a.I, rowdofs)
    push!(a.J, coldofs)
end
```
Furthermore, in the Project I'm currently working on, I know that my element matrices have a predictable pattern of zero entries (Because the entry for each node is essentially a diagonal Matrix, I would end up with 2/3 structual zeros in the final assembled Matrix). It might be worthwhile to include a keword argument `dropzeros`  that automatically removes these zero entries:

```julia
function assemble!(a::Ferrite.Assembler{T}, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix{T}; dropzeros::Bool=false) where {T}
    @assert(size(Ke,1) == length(rowdofs))
    @assert(size(Ke,2) == length(coldofs))

    for I in CartesianIndices(Ke)
        dropzeros && Ke[I] == 0 && continue

        (i,j) = Tuple(I)

        push!(a.V, Ke[I])
        push!(a.I, rowdofs[i])
        push!(a.J, coldofs[j])
    end
end
```

This way, I wouldn't have to call dropzeros on the assembled matrix, which would probably be slightly faster and would allocate less memory during assembly.

If this is something worth including, I can get to work implementing this and creating a pull request. However, this would be my first time contributing to a project and I'd first have to figure out how that acctually works 😅

fredrikekre:
Hi, first off, is there a reason you are using the "old" assembler instead of using the newer one based on `create_sparsity_pattern`? The latter one is already much more flexible and more performant I would say.

For example, with the keyword argument `coupling` you can specify the coupling between degrees of freedom on the element level. You can use it to specify coupling between different fields (if you have more than one) or different components (if you have a vector-valued problem) or simply dof-by-dof. You can think of the `coupling` argument as a "template" for the element matrices you will assemble. For example, if you know you only have diagonal entries, and the element matrix is 4x4 you can pass
```julia
coupling = [
    true false false false
    false true false false
    false false true false
    false false false true
]
```
> It might be worthwhile to include a keword argument dropzeros that automatically removes these zero entries:

[...]

PR #1279, opened 2026-02-06
Add assembly into rectangular sparse matrices

termi-official:
Closes https://github.com/Ferrite-FEM/Ferrite.jl/issues/1133 (and https://github.com/Ferrite-FEM/Ferrite.jl/issues/968 ?)

### DIFF
--- a/test/test_assembler_extensions.jl
+++ b/test/test_assembler_extensions.jl
@@ -103,7 +103,7 @@ using SparseArrays, LinearAlgebra
         @test all(f[rdofs] .== 2fe)
 
         # CSRAssembler: Assemble rectangular part in quadratic matrix
-        K = SparseMatrixCSR(6, 6, [K.colptr..., 7, 7, 7], K.rowval, K.nzval)
+        K = SparseMatrixCSR{1}(6, 6, K.rowptr, K.colval, K.nzval)
         assembler = start_assemble(K, f)
         rdofs = [1, 4, 6]
         cdofs = [1, 3]


src/assembler.jl
```diff
@@ -253,20 +261,29 @@ end
...
 Assemble the element stiffness matrix `Ke` (and optional force vector `fe`) into the global
 stiffness (and force) in `A`, given the element degrees of freedom `dofs`.
 
-This is equivalent to `K[dofs, dofs] += Ke` and `f[dofs] += fe`, where `K` is the global
-stiffness matrix and `f` the global force/residual vector, but more efficient.
+This is equivalent to `K[rowdofs, coldofs] += Ke` and `f[rowdofs] += fe` efficiently, or
+equivalently this is applying `K[dofs, dofs] += Ke` and `f[dofs] += fe` efficiently, where
+`K` is the global stiffness matrix and `f` the global force/residual vector.
 """
 assemble!(::AbstractAssembler, ::AbstractVector{<:Integer}, ::AbstractMatrix, ::AbstractVector)
```
KnutAM, line 277: I think (1) we should start with the standard (square) case and (2) that it becomes clearer if we separate the two docstrings. Also, I think the `@docs` attachment/filter was already wrong but even more important here that it will be attached to any call to `assemble!` with an abstract assembler. 

```suggestion
Assemble the square element stiffness matrix `Ke` (and optional force vector `fe`) into the global
stiffness (and force) in `A`, given the element degrees of freedom `dofs`.

This is equivalent to `K[dofs, dofs] += Ke` and `f[dofs] += fe`, where `K` is the global stiffness matrix and `f` the global force/residual vector, but more efficient.

    assemble!(A::AbstractAssembler, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix)
    assemble!(A::AbstractAssembler, rowdofs::AbstractVector{Int}, coldofs::AbstractVector{Int}, Ke::AbstractMatrix, fe::AbstractVector)
    
Assemble the element stiffness matrix `Ke` (and optional force vector `fe`) into the global
stiffness (and force) in `A`, given the element row degrees of freedom, `rowdofs`, and element column degrees of freedom, `coldofs`. 

This is equivalent to `K[rowdofs, coldofs] += Ke` and `f[rowdofs] += fe`, but more efficient.
"""
assemble!(::AbstractAssembler, args...)
```

test/test_assemble.jl
```diff
@@ -52,6 +55,17 @@
...
+    I = [1, 1, 4, 4, 6, 6]
+    J = [1, 3, 1, 3, 1, 3]
+    V = zeros(length(I))
+    K = sparse(I, J, V)
+    assembler = start_assemble(K)
+    rdofs = [1, 4, 6]
+    cdofs = [1, 3]
+    Ke = rand(length(rdofs), length(cdofs))
+    assemble!(assembler, rdofs, cdofs, Ke)
+    @test (K[rdofs, cdofs] .== Ke) |> all
```
KnutAM: Just to keep consistent code-style since we usually (I think) don't use this style here? 
```suggestion
    @test all(K[rdofs, cdofs] .== Ke)
```

KnutAM: Good to check that we really add and not just replace?
```suggestion
    assemble!(assembler, rdofs, cdofs, Ke)
    assemble!(assembler, rdofs, cdofs, Ke)
    @test (K[rdofs, cdofs] .== 2 * Ke) |> all
```

And perhaps add a test including `f` as well to have complete coverage of cases?

test/test_assembler_extensions.jl
```diff
@@ -84,6 +85,17 @@ using SparseArrays, LinearAlgebra
...
+        I = [1, 1, 4, 4, 6, 6]
+        J = [1, 3, 1, 3, 1, 3]
+        V = zeros(length(I))
+        K = sparsecsr(I, J, V)
+        assembler = start_assemble(K)
+        rdofs = [1, 4, 6]
+        cdofs = [1, 3]
+        Ke = rand(length(rdofs), length(cdofs))
+        assemble!(assembler, rdofs, cdofs, Ke)
+        @test (K[rdofs, cdofs] .== Ke) |> all
```
KnutAM: As above
```suggestion
        @test all(K[rdofs, cdofs] .== Ke)
```
Also regarding the other tests

KnutAM (commented):
Just some minor comments regarding docs and tests, but looks good to me. 

Could you post a benchmark in the PR comparing master and PR for assembling a constant `Ke` and `fe` with some grid, just so we document that there are no performance regressions (I assume there isn't?)

test/test_assemble.jl
```diff
@@ -52,6 +55,17 @@
     K, _ = finish_assemble(a)
     @test (K[rdofs, cdofs] .== Ke) |> all
 
+    I = [1, 1, 4, 4, 6, 6]
+    J = [1, 3, 1, 3, 1, 3]
+    V = zeros(length(I))
+    K = sparse(I, J, V)
```
KnutAM, line 62: Maybe test for a symmetric matrix case too just to provide full test coverage?

KnutAM:
> This looks correct to me. The block format is neither CSR nor CSC.

But the underlying matrix is one of them? For me it was just not clear if both are supported (perhaps I should just try but it is late 😄)

test/test_assemble.jl
termi-official: Strictly rectangular matrices cannot be symmtric by definition.

KnutAM: Right :D 
In that case should we error then?

termi-official: Check

src/assembler.jl
```diff
@@ -254,19 +268,29 @@ end
...
+
+Assemble the element stiffness matrix `Ke` (and optional force vector `fe`) into the global
+stiffness (and force) in `A`, given the element row degrees of freedom, `rowdofs`, and element column degrees of freedom, `coldofs`.
+This is equivalent to `K[rowdofs, coldofs] += Ke` and `f[rowdofs] += fe`, but more efficient.
 """
 assemble!(::AbstractAssembler, ::AbstractVector{<:Integer}, ::AbstractMatrix, ::AbstractVector)
 
 @propagate_inbounds function assemble!(A::AbstractAssembler, dofs::AbstractVector{<:Integer}, Ke::AbstractMatrix, fe::Union{AbstractVector, Nothing} = nothing)
-    return _assemble!(A, dofs, Ke, fe, false)
+    size(Ke, 1) == size(Ke, 2) || throw(ArgumentError("Ke is rectangular, but only a single dof range is provided. Please call assemble!(A, rowdofs, coldofs, Ke, fe) instead."))
```
KnutAM: To match docstring and avoid confusion with `dof_range`. 
```suggestion
    size(Ke, 1) == size(Ke, 2) || throw(ArgumentError("Ke is rectangular, but only a single `dofs` vector is provided. Please call assemble!(A, rowdofs, coldofs, Ke, fe) instead."))
```

src/assembler.jl
```diff
@@ -196,8 +200,10 @@ Assembler for symmetric sparse matrix with CSC storage type.
 struct SymmetricCSCAssembler{Tv, Ti, MT <: Symmetric{Tv, <:AbstractSparseMatrixCSC{Tv, Ti}}} <: AbstractCSCAssembler
     K::MT
     f::Vector{Tv}
-    permutation::Vector{Int}
-    sorteddofs::Vector{Int}
+    rowpermutation::Vector{Int}
+    colpermutation::Vector{Int}
+    sortedrowdofs::Vector{Int}
+    sortedcoldofs::Vector{Int}
```
KnutAM: Just to document why these are included even if not needed? 
```suggestion
    rowpermutation::Vector{Int} # Symmetric assembly doesn't need separate row and
    colpermutation::Vector{Int} # col permutation and dofs, but simplifies code reuse
    sortedrowdofs::Vector{Int}  # reuse with non-symmetric cases. sortedrowdofs and
    sortedcoldofs::Vector{Int}  # rowpermutation always aliased to sortedcoldofs and colpermutation.
```

src/assembler.jl
```diff
@@ -239,11 +245,19 @@ start_assemble(K::Union{AbstractSparseMatrixCSC, Symmetric{<:Any, <:AbstractSpar
...
 function start_assemble(K::AbstractSparseMatrixCSC{T}, f::Vector = T[]; fillzero::Bool = true, maxcelldofs_hint::Int = 0) where {T}
     fillzero && (fillzero!(K); fillzero!(f))
-    return CSCAssembler(K, f, zeros(Int, maxcelldofs_hint), zeros(Int, maxcelldofs_hint))
+    if size(K, 1) == size(K, 2)
+        permutation = zeros(Int, maxcelldofs_hint)
+        sorteddofs = zeros(Int, maxcelldofs_hint)
+        return CSCAssembler(K, f, permutation, permutation, sorteddofs, sorteddofs)
+    else
+        return CSCAssembler(K, f, zeros(Int, maxcelldofs_hint), zeros(Int, maxcelldofs_hint), zeros(Int, maxcelldofs_hint), zeros(Int, maxcelldofs_hint))
+    end
```
KnutAM: Is this aliasing needed for performance? I think it should be valid to assemble a rectangular contribution to an overall square matrix?

termi-official: >Is this aliasing needed for performance?

As you asked to change it to not introduce any diff in perf, yes. Otherwise we will sort twice, which is a few percent slower.

src/assembler.jl
```diff
@@ -283,53 +307,62 @@ Sorts the dofs into a separate buffer and returns it together with a permutation
...
     K = matrix_handle(A)
-    @boundscheck checkbounds(K, dofs, dofs)
+    @boundscheck checkbounds(K, rowdofs, coldofs)
 
     # We assume that the input dofs are not sorted, because the cells need the dofs in
     # a specific order, which might not be the sorted order. Hence we sort them.
     # Note that we are not allowed to mutate `dofs` in the process.
-    sorteddofs, permutation = _sortdofs_for_assembly!(A.permutation, A.sorteddofs, dofs)
+    sortedcoldofs, colpermutation = _sortdofs_for_assembly!(A.colpermutation, A.sortedcoldofs, coldofs)
+    sortedrowdofs, rowpermutation = if A.sortedcoldofs !== A.sortedrowdofs
```
KnutAM: Would this regain the [above](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1279#discussion_r2872360982) discussed performance regression of not aliasing? 
```suggestion
    sortedrowdofs, rowpermutation = if coldofs !== rowdofs
```

termi-official: Good point. Lets go with this proposition then.

Issue #1280, opened 2026-02-09
Allocations in `InterfaceIterator(dh, topo)`  (can be avoided with `InterfaceCache`)


yolhan83:
When iterating over interfaces using `InterfaceIterator(dh, topo)` I observe significant heap allocations.
In contrast, iterating over `facetskeleton(topo, grid)` + `InterfaceCache(dh)` yields ~0 allocations after warm-up.
## Minimal Reproducible Example (MWE)
```julia
using Ferrite
function loop_bad(dh, topo)
    s = 0
    for ic in InterfaceIterator(dh, topo)  # allocs here ( iterate on flamegraph)
        s += length(interfacedofs(ic))  
    end
    return s
end
function loop_cached(skel, neigh, icache)
    s = 0
    for facet in skel # no allocs
        nbhs = @inbounds neigh[facet[1], facet[2]]
        isempty(nbhs) && continue
        Ferrite.reinit!(icache, facet, nbhs[1])
        s += length(icache.dofs)         
    end
    return s
end
function main(; nx=50, ny=50, order=1)
    grid = generate_grid(Quadrilateral, (nx,ny), Vec{2}((0.0,0.0)), Vec{2}((1.0,1.0)))
    topo = ExclusiveTopology(grid)
    ip = DiscontinuousLagrange{RefQuadrilateral,order}()
    dh = DofHandler(grid); add!(dh, :u, ip); close!(dh)
    skel   = collect(facetskeleton(topo, grid))               
    neigh  = Ferrite.get_facet_facet_neighborhood(topo, grid)     
    icache = InterfaceCache(dh)
    loop_bad(dh, topo)
    loop_cached(skel, neigh, icache)
    bad    = @allocated loop_bad(dh, topo)
    cached = @allocated loop_cached(skel, neigh, icache)
    println((bad=bad, cached=cached))
end
main()
```
output :
```julia 
(bad = 719888, cached = 0)
``` 
## Expected result
Ideally, `InterfaceIterator` should allow a **non-allocating** path, similar to what `InterfaceCache` provides.
```julia
julia> versioninfo()
Julia Version 1.12.4
Commit 01a2eadb04 (2026-01-06 16:56 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 20 × 12th Gen Intel(R) Core(TM) i7-12700H
  WORD_SIZE: 64
[...]

PR #1281, opened 2026-02-09
simplify state signature and inline in InterfaceIterator

yolhan83:
Attempt at solving https://github.com/Ferrite-FEM/Ferrite.jl/issues/1280, 
mwe : 
```julia
using Ferrite
function loop_bad(dh, topo)
    s = 0
    for ic in InterfaceIterator(dh, topo)  # allocs here ( iterate on flamegraph)
        s += length(interfacedofs(ic))  
    end
    return s
end
function loop_cached(skel, neigh, icache)
    s = 0
    for facet in skel # no allocs
        nbhs = @inbounds neigh[facet[1], facet[2]]
        isempty(nbhs) && continue
        Ferrite.reinit!(icache, facet, nbhs[1])
        s += length(icache.dofs)         
    end
    return s
end
function main(; nx=50, ny=50, order=1)
    grid = generate_grid(Quadrilateral, (nx,ny), Vec{2}((0.0,0.0)), Vec{2}((1.0,1.0)))
    topo = ExclusiveTopology(grid)
    ip = DiscontinuousLagrange{RefQuadrilateral,order}()
    skel   = collect(facetskeleton(topo, grid))               
    neigh  = Ferrite.get_facet_facet_neighborhood(topo, grid)     
    icache = InterfaceCache(dh)
    loop_bad(dh, topo)
    loop_cached(skel, neigh, icache)
    bad    = @allocated loop_bad(dh, topo)
    cached = @allocated loop_cached(skel, neigh, icache)
    println((bad=bad, cached=cached))
end
main()
```
before : `(bad = 719888, cached = 0)` getting worse with nx , ny and order
after : `(bad = 960, cached = 0)` only getting bigger with order

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -370,13 +370,14 @@ end
         facet_a = skeleton[i]; i += 1
         neighbors = neighborhood[facet_a[1], facet_a[2]]
         isempty(neighbors) && continue
-        length(neighbors) > 1 && error("multiple neighboring faces not supported yet")
-        reinit!( ii.cache, facet_a, neighbors[1])
-        return  ii.cache, i
+        length(neighbors) > 1 && error("multiple neighboring facets not supported yet")
+        facet_b = neighbors[1]
+        reinit!(ii.cache, facet_a, facet_b)
+        return ii.cache, i
     end
     return nothing
 end
-@inline Base.iterate(ii::InterfaceIterator) = Base.iterate(ii, 1)
+@inline Base.iterate(ii::InterfaceIterator) = iterate(ii, 1)
 
 # Iterator interface for CellIterator/FacetIterator
 const GridIterators{C} = Union{CellIterator{C}, FacetIterator{C}, InterfaceIterator{C}}


src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
-            continue
-        end
-        neighbors = neighborhood[facet_a[1], facet_a[2]]
-        length(neighbors) > 1 && error("multiple neighboring faces not supported yet")
-        facet_b = neighbors[1]
-        reinit!(ii.cache, facet_a, facet_b)
-        return (ii.cache, state)
+@inline function Base.iterate(ii::InterfaceIterator{<:Any,<:Grid{sdim}}, i::Int) where {sdim}
+    neighbors = get_facet_facet_neighborhood(ii.topology, ii.grid)# TODO: This could be moved to InterfaceIterator constructor (potentially type-instable for non-union or mixed grids)
+    skel::Vector{FacetIndex} = facetskeleton(ii.topology, ii.grid)
```
termi-official: Do we need the type restriction?

yolhan83: When trying outside Ferrite I got a weird ambiguity issue but seems like when inside its ok to remove it, may be due to my multiple tests though

src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
-        facet_a, state = it
-        if isempty(neighborhood[facet_a[1], facet_a[2]])
-            continue
-        end
-        neighbors = neighborhood[facet_a[1], facet_a[2]]
-        length(neighbors) > 1 && error("multiple neighboring faces not supported yet")
-        facet_b = neighbors[1]
-        reinit!(ii.cache, facet_a, facet_b)
-        return (ii.cache, state)
+@inline function Base.iterate(ii::InterfaceIterator, i::Int)
```
KnutAM: Does the inline make a difference? Seems like this could increase compilation times?

src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
+    neighbors = get_facet_facet_neighborhood(ii.topology, ii.grid)# TODO: This could be moved to InterfaceIterator constructor (potentially type-instable for non-union or mixed grids)
```
KnutAM: Avoid rename?

src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
+    skel::Vector{FacetIndex} = facetskeleton(ii.topology, ii.grid)
```
KnutAM: `skeleton`?

termi-official:
> would be cool to have a way to give the interface cache to the interface iterator

This would be rather uncontroversion to add I think (cc @KnutAM what do you think?)

> I also want to implement a pure hyperbolic RKDG exemple and a LDG one

That would also be very interesting to showcase how to do that in Ferrite!

src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
+    while i <= length(skel)
+        fa = skel[i]; i += 1
```
KnutAM: `facet_a`?

src/iterators.jl
```diff
@@ -363,23 +363,21 @@ function InterfaceIterator(
...
+        nbhs = neighbors[fa[1], fa[2]]
```
KnutAM: `nbhs` should be called `neighbors` IMO

src/iterators.jl
termi-official:
```suggestion
    neighbors = get_facet_facet_neighborhood(ii.topology, ii.grid) # TODO: This could be moved to InterfaceIterator constructor (potentially type-instable for non-union or mixed grids)
```
Runic complains here and below at the extra new line.

src/iterators.jl
KnutAM:
```suggestion
@inline function Base.iterate(ii::InterfaceIterator, i::Integer)
```

KnutAM:
> This would be rather uncontroversion to add I think (cc @KnutAM what do you think?)

I think the main point would be to be able to iterate over the interfaces without the iterator, like we can wit e.g. `CellCache`,
```julia
cell = CellCache(..)
for cellnr in set
    reinit!(cell, cellnr)
…
```

But then some way to do
```Julia
set = create_interfaceset(…) # new
ic = InterfaceCache(…)
for facet_pair in set
    reinit!(ic, facet_pair)
…
```

src/iterators.jl
yolhan83: Yes it's basicly a must now in julia for all iterators

yolhan83: cf https://discourse.julialang.org/t/why-does-my-iterable-allocate/86216

src/iterators.jl
yolhan83: rename of ?

KnutAM: neighborhood to neighbors

yolhan83: oh ok

src/iterators.jl
```diff
@@ -363,24 +363,22 @@ function InterfaceIterator(
...
+@inline function Base.iterate(ii::InterfaceIterator, i::Integer)
     neighborhood = get_facet_facet_neighborhood(ii.topology, ii.grid) # TODO: This could be moved to InterfaceIterator constructor (potentially type-instable for non-union or mixed grids)
-    while true
-        it = iterate(facetskeleton(ii.topology, ii.grid), state...)
-        it === nothing && return nothing
-        facet_a, state = it
-        if isempty(neighborhood[facet_a[1], facet_a[2]])
-            continue
-        end
+    skeleton::Vector{FacetIndex} = facetskeleton(ii.topology, ii.grid)
```
KnutAM:
```suggestion
    skeleton = facetskeleton(ii.topology, ii.grid)
```

src/iterators.jl
```diff
@@ -363,24 +363,22 @@ function InterfaceIterator(
...
+    IC = ii.cache
```
KnutAM:
```suggestion
```
Or at least call it `cache` since `IC` looks like a type, but I think it is easier to just use `ii.cache` those two places it is being used.

src/iterators.jl
```diff
@@ -363,24 +363,22 @@ function InterfaceIterator(
...
+    j = i
```
KnutAM: Why do this, and not just use `i`? 
To me this made it harder to understand the code.

src/iterators.jl
```diff
@@ -363,24 +363,22 @@ function InterfaceIterator(
...
+    while j <= length(skeleton)
+        facet_a = skeleton[j]; j += 1
         neighbors = neighborhood[facet_a[1], facet_a[2]]
+        isempty(neighbors) && continue
         length(neighbors) > 1 && error("multiple neighboring faces not supported yet")
-        facet_b = neighbors[1]
-        reinit!(ii.cache, facet_a, facet_b)
-        return (ii.cache, state)
+        reinit!(IC, facet_a, neighbors[1])
```
KnutAM:
```suggestion
        reinit!(ii.cache, facet_a, neighbors[1])
```

src/iterators.jl
```diff
@@ -363,24 +363,22 @@ function InterfaceIterator(
...
+        return IC, j
```
KnutAM:
```suggestion
        return ii.cache, j
```

src/iterators.jl
KnutAM: Is this to avoid the `::Union{Nothing, Vector{FacetIndex}`? Does this make a performance difference? If so, I think we should rather add it in `facetskeleton`.

KnutAM (commented):
A few more comments

src/iterators.jl
yolhan83: no perf penalty

KnutAM (approved):
I'm wondering if all the performance issue would be fixed by just using `@inline`?

But anyways I think this updated code is easier to follow and for sure much faster! Thanks a lot! 

Will merge once CI passes

PR #1283, opened 2026-02-11
add a DG SW model exemple

yolhan83:
This PR adds a shallow water RKDG exemple to showcase how this can be done in Ferrite.
I won't go further on this description since everything is in the md, please suggest anykind of clarification.

### DIFF
--- a/docs/src/literate-tutorials/dg_shallow_water.jl
+++ b/docs/src/literate-tutorials/dg_shallow_water.jl
@@ -308,8 +308,8 @@ close!(dh)
 
 nd = ndofs(dh)
 
-qr = QuadratureRule{RefQuadrilateral}(:lobatto, order+1)
-facet_qr = FacetQuadratureRule{RefQuadrilateral}(:lobatto, order+1)
+qr = QuadratureRule{RefQuadrilateral}(:lobatto, order + 1)
+facet_qr = FacetQuadratureRule{RefQuadrilateral}(:lobatto, order + 1)
 
 cv = CellValues(qr, ip)
 fv = FacetValues(facet_qr, ip)


docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,560 @@
...
+    getfacetset(grid, "bottom"),
+)
+
+order = 1
+ip = DiscontinuousLagrange{RefQuadrilateral, order}()
+
+dh = DofHandler(grid)
+add!(dh, :h, ip)
+add!(dh, :hu, ip)
+add!(dh, :hv, ip)
```
termi-official: Since I do not know if you are fully aware, but please note that we can do something like

```suggestion
iph = DiscontinuousLagrange{RefQuadrilateral, order}()
ipv = DiscontinuousLagrange{RefQuadrilateral, order}()^2

dh = DofHandler(grid)
add!(dh, :h, iph)
add!(dh, :v, ipv)
```

and express the equations below directly in terms of tensor operations. This is just for your information, no need to take any action here from my side.

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,560 @@
...
+        for qp in 1:getnquadpoints(cv)
+            dE = getdetJdV(cv, qp)
+            for i in 1:nb
+                Φᵢ = shape_value(cv, qp, i)
+                for j in 1:nb
+                    Φⱼ = shape_value(cv, qp, j)
+                    m = (Φᵢ * Φⱼ) * dE
+                    Me[i, j] += m
+                    Me[i + nb, j + nb] += m
+                    Me[i + 2nb, j + 2nb] += m
```
termi-official: The safe way to extract these is to use the function `dof_range` and then use the respective indices as for example in https://github.com/Ferrite-FEM/Ferrite.jl/blob/f245839f9d5ac0ba0a3aae9841e0a157b4b3e735/docs/src/literate-tutorials/incompressible_elasticity.jl#L139-L148 .

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,560 @@
...
+    Me = zeros(ndpc, ndpc)
+
+    asM = start_assemble(M)
+
+    for cell in CellIterator(dh)
+        Ferrite.reinit!(cv, cell)
+        dofs = celldofs(cell)
+        fill!(Me, 0.0)
+
+        for qp in 1:getnquadpoints(cv)
```
termi-official, line 334: We typically function barrier this quadrature part to get better performance.

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,560 @@
...
+# ### Time derivative: `dU = M^{-1} R(U)`
+function ode!(dU, U, param, t)
+    fill!(dU, 0.0)
+    volume_integral!(dU, U, param, t)
+    interface_integral!(dU, U, param, t)
+    boundary_facets!(dU, U, param, t)
+
+    lin = param.lin
+    lin.b = dU
+    copyto!(dU, solve!(lin).u)
```
termi-official: Note that we do not need to solve on every `ode!` call. You can invert the mass matrix during assembly (see above). This way we obtain simply
```suggestion
    mul!(dU, M, dU2)
```
where you compute the action of your PDE discretization into the buffer dU2 (instead of into dU).

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,560 @@
...
+                    Φⱼ = shape_value(cv, qp, j)
+                    m = (Φᵢ * Φⱼ) * dE
+                    Me[i, j] += m
+                    Me[i + nb, j + nb] += m
+                    Me[i + 2nb, j + 2nb] += m
+                end
+            end
+        end
+
+        assemble!(asM, dofs, Me)
```
termi-official:
```suggestion
        assemble!(asM, dofs, inv(Me))
```

termi-official (commented):
First feedback from my side.

> [...] some tutos didnt work on my pc [...]

What failure do you get?

yolhan83:
> First feedback from my side.
> 
> > [...] some tutos didnt work on my pc [...]
> 
> What failure do you get?

I get : 
```julia

┌ Error: failed to run `@example` block in docs\src\tutorials\stokes-flow.md:459-461
│ ```@example stokes-flow
│ main()
│ ```
│   exception =
│ 
│    Stacktrace:
│      [1] error(s::String)
│        @ Base .\error.jl:44
│      [2] open
│        @ C:\Users\yolha\.julia\artifacts\649584fb27316a4395dbb47fde904c83626feae5\lib\gmsh.jl:113 [inlined]
│      [3] togrid(filename::String; domain::String)
│        @ FerriteGmsh C:\Users\yolha\.julia\packages\FerriteGmsh\MH6r7\src\FerriteGmsh.jl:185
│      [4] togrid
│        @ C:\Users\yolha\.julia\packages\FerriteGmsh\MH6r7\src\FerriteGmsh.jl:176 [inlined]
│      [5] (::Main.var"__atexample__named__stokes-flow".var"#setup_grid##0#setup_grid##1")(dir::String)
│        @ Main.var"__atexample__named__stokes-flow" .\stokes-flow.md:213
│      [6] mktempdir(fn::Main.var"__atexample__named__stokes-flow".var"#setup_grid##0#setup_grid##1", parent::String; prefix::String)
│        @ Base.Filesystem .\file.jl:936
│      [7] mktempdir (repeats 2 times)
│        @ .\file.jl:932 [inlined]
│      [8] setup_grid(h::Float64)
│        @ Main.var"__atexample__named__stokes-flow" .\stokes-flow.md:210
│      [9] main()
│        @ Main.var"__atexample__named__stokes-flow" .\stokes-flow.md:462
│     [10] top-level scope
│        @ stokes-flow.md:460
│     [11] eval(m::Module, e::Any)
│        @ Core .\boot.jl:489
│     [12] #61
│        @ C:\Users\yolha\.julia\packages\Documenter\HdXI4\src\expander_pipeline.jl:856 [inlined]
│     [13] cd(f::Documenter.var"#61#62"{Module, Expr}, dir::String)
│        @ Base.Filesystem .\file.jl:101
│     [14] (::Documenter.var"#59#60"{Documenter.Page, Module, Expr})()
│        @ Documenter C:\Users\yolha\.julia\packages\Documenter\HdXI4\src\expander_pipeline.jl:855
│     [15] (::IOCapture.var"#12#13"{Type{InterruptException}, Documenter.var"#59#60"{Documenter.Page, Module, Expr}, IOContext{Base.PipeEndpoint}, IOContext{Base.PipeEndpoint}, Base.TTY, Base.TTY})()
│        @ IOCapture C:\Users\yolha\.julia\packages\IOCapture\Y5rEA\src\IOCapture.jl:170
│     [16] with_logstate(f::IOCapture.var"#12#13"{Type{InterruptException}, Documenter.var"#59#60"{Documenter.Page, Module, Expr}, IOContext{Base.PipeEndpoint}, IOContext{Base.PipeEndpoint}, Base.TTY, Base.TTY}, logstate::Base.CoreLogging.LogState)  
│        @ Base.CoreLogging .\logging\logging.jl:542
│     [17] with_logger(f::Function, logger::Base.CoreLogging.ConsoleLogger)
│        @ Base.CoreLogging .\logging\logging.jl:653
│     [18] capture(f::Documenter.var"#59#60"{Documenter.Page, Module, Expr}; rethrow::Type, color::Bool, passthrough::Bool, capture_buffer::IOBuffer, io_context::Vector{Any})
│        @ IOCapture C:\Users\yolha\.julia\packages\IOCapture\Y5rEA\src\IOCapture.jl:167
│     [19] runner(::Type{Documenter.Expanders.ExampleBlocks}, node::MarkdownAST.Node{Nothing}, page::Documenter.Page, doc::Documenter.Document)
│        @ Documenter C:\Users\yolha\.julia\packages\Documenter\HdXI4\src\expander_pipeline.jl:854
└ @ Documenter C:\Users\yolha\.julia\packages\Documenter\HdXI4\src\utilities\utilities.jl:47
```

docs/src/literate-tutorials/dg_shallow_water.jl
yolhan83: Yes I saw that too but I'm less used to it, hope its fine this way

docs/src/literate-tutorials/dg_shallow_water.jl
yolhan83: for the assembly I'm not sure its important, but yeah maybe for the rest I should add elementwise / interfacewise and facetwise functions

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,540 @@
...
+            ΦᵢL = shape_value(iv, qp, i; here = true)
+            ΦᵢR = shape_value(iv, qp, i + nb; here = false)
+            Ri[dr_h[i]] += (-ΦᵢL * fhat[1]) * dF
+            Ri[dr_qx[i]] += (-ΦᵢL * fhat[2]) * dF
+            Ri[dr_qy[i]] += (-ΦᵢL * fhat[3]) * dF
+            Ri[dr_h[i] + ndpc] += (ΦᵢR * fhat[1]) * dF
+            Ri[dr_qx[i] + ndpc] += (ΦᵢR * fhat[2]) * dF
+            Ri[dr_qy[i] + ndpc] += (ΦᵢR * fhat[3]) * dF
+        end
+    end
```
termi-official: @AbdAlazezAhmed can you take a look at this part here? Any chance that we miss some interface for this, so users do not need to manually index this stuff? Also, aren't we missing some 1/h term?

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,540 @@
...
+    nb = getnbasefunctions(cv)
+    Ferrite.reinit!(cv, cell)
+    dofs = celldofs(cell)
+    Ue = @view U[dofs]
+    coords = getcoordinates(cell)
+    for qp in 1:getnquadpoints(cv)
+        dE = getdetJdV(cv, qp)
+        h = function_value(cv, qp, Ue, 1:nb)
+        qx = function_value(cv, qp, Ue, (nb + 1):2nb)
+        qy = function_value(cv, qp, Ue, (2nb + 1):3nb)
```
termi-official:
```suggestion
        h = function_value(cv, qp, Ue, dr_h)
        qx = function_value(cv, qp, Ue, dr_qx)
        qy = function_value(cv, qp, Ue, dr_qy)
```

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,540 @@
...
+# ### Volume integral on each element
+function element_volume_integral!(R, cell, U, dr, param)
+    dr_h, dr_qx, dr_qy = dr
+    Re = param.Re
+    fill!(Re, 0.0)
+    cv = param.cv
+    nb = getnbasefunctions(cv)
+    Ferrite.reinit!(cv, cell)
+    dofs = celldofs(cell)
+    Ue = @view U[dofs]
```
termi-official: We should leave a quick comment here that this is only okay for DG, because `celldofs` is guaranteed to index a continuous chunk of memory.

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,540 @@
...
+function interface_integral!(R, ic, U, dr, param)
+    dr_h, dr_qx, dr_qy = dr
+    Ri = param.Ri
+    fill!(Ri, 0.0)
+    iv = param.iv
+    nb = getnbasefunctions(param.cv)
+    ndpc = 3 * nb
+    Ferrite.reinit!(iv, ic)
+    idofs = interfacedofs(ic)
+    Ui = @view U[idofs]
```
termi-official: It is more performant to copy the relevant dofs into a continuous buffer instead of using views to relieve some pressure on the memory in the loop below.

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,540 @@
...
+        qy = function_value(fv, qp, u_e, dr_qy)
+        Uin = Vec{3}((h, qx, qy))
+        X = spatial_coordinate(fv, qp, coords)
+        Ubc = bound_condition(Uin, X, tag, t, n)
+        fhat = LF(Uin, Ubc, n)
+        @inbounds for i in 1:nb
+            Φᵢ = shape_value(fv, qp, i)
+            Rb[dr_h[i]] -= (Φᵢ * fhat[1]) * dF
+            Rb[dr_qx[i]] -= (Φᵢ * fhat[2]) * dF
+            Rb[dr_qy[i]] -= (Φᵢ * fhat[3]) * dF
```
termi-official, line 455: Correct me if I am wrong, but don't we need some element size term here to enforce the Dirichlet condition weakly?

yolhan83: I'm using a ghost element method for the week forcing at the boundary, so the code is exactly similar to the inner interface beside that we dont compute the dofs for the ghost element and we preomcute what we want or wish at the outer trace. Maybe you're thinking of a strong one with a forcing term of order 1/size but that can lead to non-physical solution in this framework. It is ok when DG is used on parabolic / eliptic pdes though.

docs/src/literate-tutorials/dg_shallow_water.jl
yolhan83: I found that weird using it too especially, I thought it should be either
```julia
hL = function_value(iv, qp, Ui, dr_h; here=true) 
hR = function_value(iv, qp, Ui, dr_h; here=false)
```
or
```julia
hL = function_value(iv, qp, Ui, dr_h) 
hR = function_value(iv, qp, Ui, dr_h .+ ndpc)
```
but having to do both seems weird. Maybe differencing Ui_here and Ui_there since the beggening would be better ?

docs/src/tutorials/index.md
```diff
@@ -178,3 +178,11 @@ Galerkin Infrastructure For the finite element toolbox
 Ferrite.jl"](https://summerofcode.withgoogle.com/programs/2023/projects/SLGbRNI5).
 
 **Keywords**: scalar-valued solution, Dirichlet boundary conditions, Discontinuous Galerkin, Interior penalty
+
+---
+
+#### [Tutorial 14: Discontinuous Galerkin shallow water equation (DG + RK)](@ref tutorial-swe-dg)
+
+This tutorial guides you through solving the **two-dimensional shallow water equations** (a nonlinear **hyperbolic** conservation law) using a **Discontinuous Galerkin (DG)** spatial discretization and an **explicit Runge–Kutta** time integrator. The tutorial introduces the hyperbolic DG workflow in Ferrite: assembling **volume terms** with `CellIterator`, coupling elements through **numerical fluxes** on **interfaces** via `InterfaceIterator`/`InterfaceValues` (two-sided traces with `here=true/false`), and treating boundary conditions using `FacetIterator`/`FacetValues`. A robust **Rusanov (local Lax–Friedrichs)** flux is used, and the semi-discrete system is advanced in time with `OrdinaryDiffEq`. The output is written as a sequence of VTK frames, ready to be combined into a GIF animation.
```
KristofferC: >  The output is written as a sequence of VTK frames

Is this true?

yolhan83: I removed the code to save the VTK, how to add it without it being executed in the CI ?
```julia
isdir("datas") || mkdir("datas")
for i in eachindex(sol.u)
   ui = sol.u[i]
   VTKGridFile("./datas/swe_$(i)", dh) do vtk
       write_solution(vtk, dh, ui)   
   end
end
```

termi-official: We typically add
```
if isdefined(Main, :is_ci) #hide
    IS_CI = Main.is_ci     #hide
else                       #hide
    IS_CI = false          #hide
end                        #hide
nothing                    #hide
```
on top of the file and then solve only a coarse problem on CI, e.g. via
```
if IS_CI                                                                                            #hide
    gmsh.option.setNumber("Mesh.MeshSizeFromCurvature", 20)                                         #hide
    gmsh.option.setNumber("Mesh.MeshSizeMax", 0.15)                                                 #hide
end   
```

I would just include the IO here (for the CI to make sure it executes at least). Also, we typically use the TimeChoiceIterator over the solution interface to also save some memory.

docs/src/literate-tutorials/dg_shallow_water.jl
termi-official: Interesting, thanks for sharing this. Do you know some paper analyzing this? I have not read the theory part at the top yet, bit if it is not explained there yet, could you add the reference there?

yolhan83: The clearest I know is this one : 
The Runge–Kutta discontinuous Galerkin method for conservation laws V: multidimensional systems](https://www.sciencedirect.com/science/article/pii/S0021999198958922 
by B Cockburn and CW Shu its quite old 1998 but its just too well explained.

Of course can't do that without citing a friend who did his phd at the same time as I did, 
Poussel, Camille, Mehmet Ersoy, and Frederic Golay. "Wetting and drying treatments with mesh adaptation for shallow water equations using a Runge–Kutta discontinuous Galerkin method." International Journal for Numerical Methods in Fluids 97.5 (2025): 692-712.

won't cite myself since I'm more on the modelling side but I used this method for every papers I did (with a twist for advection-diffusion problem)

yolhan83: also if you want the theoric part to be clearer I can do that but I thought I would go really fast to the code since the point stay Ferrite

docs/src/tutorials/index.md
yolhan83: """
Also, we typically use the TimeChoiceIterator over the solution interface to also save some memory.
"""
can you add that as a suggestion so that I can merge it to this PR ?

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,559 @@
...
+# \mathbf{R}_{\mathrm{int}}(\mathbf{U})
+# +
+# \mathbf{R}_{\partial}(\mathbf{U}),
+# ```
+#
+# where:
+# - $M$ is the (block) **mass matrix** with entries $M_{ij}^E = \int_E \Phi_i^E \Phi_j^E\,dE$,
+# - $\mathbf{R}_{\mathrm{vol}}$ contains the volume contributions $\int_E (S(U)\Phi_i^E + F(U):\nabla\Phi_i^E)\,dE$,
+# - $\mathbf{R}_{\mathrm{int}}$ contains interior interface flux contributions,
+# - $\mathbf{R}_{\partial}$ contains boundary flux contributions (using boundary states).
```
termi-official: This line breaks for some reason after the `$`

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,559 @@
...
+# \tfrac{c}{2}\left(U^+ - U^-\right),
+# ```
+#
+# where $c$ is an estimate of the maximum wave speed in direction $n$.
+# For the shallow water equations one may use
+# ```math
+# c(U,n) = |u\cdot n| + \sqrt{g h}.
+# ```
+#
+# ### Semi-discrete system
```
termi-official, line 167: We should emphasize here somewhere that we use the form where we invert $M$ and apply the action of the inverse to $f$.

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,559 @@
...
+add!(dh, :qx, ip)
+add!(dh, :qy, ip)
+close!(dh)
+
+nd = ndofs(dh)
+
+qr = QuadratureRule{RefQuadrilateral}(2 * order + 1)
+facet_qr = FacetQuadratureRule{RefQuadrilateral}(2 * order + 1)
+
+cv = CellValues(qr, ip)
```
termi-official, line 314: Please note that we have implemented the Gauss-Lobatto quadrature, which diagonalizes the mass matrix for Lagrange ansatz functions (such that we obtain a spectral discretization).

docs/src/literate-tutorials/dg_shallow_water.jl
termi-official: I always struggle with a good balance for how much theory should be stated in the docs. @AbdAlazezAhmed what would you put in the docs for this example?

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,559 @@
...
+    isdir("datas") || mkdir("datas")
+    for i in eachindex(sol.u)
+        ui = sol.u[i]
+        VTKGridFile("./datas/swe_$(i)", dh) do vtk
+            write_solution(vtk, dh, ui)
+        end
+    end
+end
+
+# **Further important ingredients (not covered here).** For realistic shallow-water simulations, one typically adds: (i) a **positivity / slope limiter** to prevent non-physical negative water depths and to control oscillations near sharp gradients; (ii) a dedicated treatment of **dry states (wetting–drying / desiccation)** to handle cells where `h → 0` robustly (e.g. thresholds, modified fluxes, and consistent momentum handling); (iii) an **automatic CFL timestep selection**, e.g.
```
termi-official: Can we make this part a regular enumeration?

docs/src/literate-tutorials/dg_shallow_water.jl
termi-official: We also might want to put this tutorial before the Navier-Stokes tutorial, because it shows OrdinaryDiffEq withouth the funky parts to strongly enforce the Dirichlet boundary condition and discontinuity handling.

docs/src/tutorials/index.md
termi-official: Please take a look at https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/ns_vs_diffeq/ for this. I have there something along the lines of
```julia
pvd = paraview_collection("vortex-street")
for (step, (u, t)) in enumerate(intervals(integrator))
    VTKGridFile("vortex-street-$step", dh) do vtk
        write_solution(vtk, dh, u)
        pvd[t] = vtk
    end
end
vtk_save(pvd);
```

docs/src/literate-tutorials/dg_shallow_water.jl
yolhan83: thats really nice indeed, I wonder, how do we take adventage of it ? 
for instance here 
```julia
    for qp in 1:getnquadpoints(iv)
        n = getnormal(iv, qp; here = true)
        dF = getdetJdV(iv, qp)
        hL = function_value(iv, qp, Ui, dr_h, dr_h .+ ndpc; here = true)
        qxL = function_value(iv, qp, Ui, dr_qx, dr_qx .+ ndpc; here = true)
        qyL = function_value(iv, qp, Ui, dr_qy, dr_qy .+ ndpc; here = true)
        hR = function_value(iv, qp, Ui, dr_h, dr_h .+ ndpc; here = false)
        qxR = function_value(iv, qp, Ui, dr_qx, dr_qx .+ ndpc; here = false)
        qyR = function_value(iv, qp, Ui, dr_qy, dr_qy .+ ndpc; here = false)
        UL = Vec{3}((hL, qxL, qyL))
        UR = Vec{3}((hR, qxR, qyR))
        fhat = LF(UL, UR, n)
        for i in 1:nb
            ΦᵢL = shape_value(iv, qp, i; here = true)
            ΦᵢR = shape_value(iv, qp, i + nb; here = false)
            Ri[dr_h[i]] += (-ΦᵢL * fhat[1]) * dF
            Ri[dr_qx[i]] += (-ΦᵢL * fhat[2]) * dF
            Ri[dr_qy[i]] += (-ΦᵢL * fhat[3]) * dF
            Ri[dr_h[i] + ndpc] += (ΦᵢR * fhat[1]) * dF
            Ri[dr_qx[i] + ndpc] += (ΦᵢR * fhat[2]) * dF
            Ri[dr_qy[i] + ndpc] += (ΦᵢR * fhat[3]) * dF
        end
```
for each point there is only one i per facet where the Phi are not zeros same here
```julia
        dE = getdetJdV(cv, qp)
        h = function_value(cv, qp, Ue, dr_h)
        qx = function_value(cv, qp, Ue, dr_qx)
        qy = function_value(cv, qp, Ue, dr_qy)
        Uval = Vec{3}((h, qx, qy))
        Fx, Fy = F(Uval)
        X = spatial_coordinate(cv, qp, coords)
        SS = S(Uval, X)
        for i in 1:nb
            Φᵢ = shape_value(cv, qp, i)
            ∇Φᵢ = shape_gradient(cv, qp, i)
            contrib = (Fx * ∇Φᵢ[1] + Fy * ∇Φᵢ[2] + SS * Φᵢ) * dE
            Re[dr_h[i]] += contrib[1]
            Re[dr_qx[i]] += contrib[2]
            Re[dr_qy[i]] += contrib[3]
        end
    end
```
where I don't actually need to sum SS for all i

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,566 @@
...
+        assemble!(asM, dofs, inv(Me))
+    end
+    return M
+end
+M = allocate_matrix(dh)
+assemble_mass_matrix!(M, dh, cv)
+droptol!(M, 1.0e-12)
+M = Diagonal(M)
+# ### Volume integral on each element
+function element_volume_integral!(R, cell, U, dr, param)
```
lijas, line 357: Element rotine becomes:

```
function element_volume_integral!(R, cell, U, param)
    Re = param.Re
    fill!(Re, 0.0)
    cv = param.cv
    nb = getnbasefunctions(cv)
    Ferrite.reinit!(cv, cell)
    dofs = celldofs(cell)
    Ue = @view U[dofs] # Note: For performance reasons, we don’t recommend using this outside the DG framework; here, the DOFs are guaranteed to be stored contiguously in memory.
    coords = getcoordinates(cell)
        dE = getdetJdV(cv, qp)
        Uval = function_value(cv, qp, Ue)
        F = F_function(Uval)
        X = spatial_coordinate(cv, qp, coords)
        SS = S(Uval, X)
        for i in 1:nb
            Φᵢ = shape_value(cv, qp, i)
            ∇Φᵢ = shape_gradient(cv, qp, i)
            Re[i] += (F ⊡ ∇Φᵢ+ SS ⋅ Φᵢ) * dE
        end
    end
    Ferrite.assemble!(R, dofs, Re)
    return nothing
end
```

with
```

# Define double contraction operator for StaticArrays matrices
⊡(A::StaticMatrix, B::StaticMatrix) = sum(A .* B)
```

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,566 @@
...
+qr = QuadratureRule{RefQuadrilateral}(:lobatto, order + 1)
+facet_qr = FacetQuadratureRule{RefQuadrilateral}(:lobatto, order + 1)
+
+cv = CellValues(qr, ip)
+fv = FacetValues(facet_qr, ip)
+iv = InterfaceValues(facet_qr, ip)
+
+#-
+# ### Mass matrix (block-diagonal per element)
+function assemble_mass_matrix!(M, dh, cv)
```
lijas, line 320: Mass matrix routine simplifies to:

```
function assemble_mass_matrix!(M, dh, cv)
    nb = getnbasefunctions(cv)
    Me = zeros(nb, nb)

    asM = start_assemble(M)
    for cell in CellIterator(dh)
        Ferrite.reinit!(cv, cell)
        dofs = celldofs(cell)
        fill!(Me, 0.0)

            dE = getdetJdV(cv, qp)
            for i in 1:nb
                Φᵢ = shape_value(cv, qp, i)
                for j in 1:nb
                    Φⱼ = shape_value(cv, qp, j)
                    m = (Φᵢ ⋅ Φⱼ) * dE
                    Me[i, j] += m
                end
            end
        end

        assemble!(asM, dofs, inv(Me))
    end
    return M
end
```

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,566 @@
...
+        end
+    end
+    Ferrite.assemble!(R, dofs, Re)
+    return nothing
+end
+#-
+
+# ### Interface integral on each interface
+function interface_integral!(R, ic, U, dr, param)
+    dr_h, dr_qx, dr_qy = dr
```
lijas, line 392:
```
function interface_integral!(R, ic, U, param)
    Ri = param.Ri
    fill!(Ri, 0.0)
    iv = param.iv
    nb = getnbasefunctions(param.cv)
    Ferrite.reinit!(iv, ic)
    idofs = interfacedofs(ic)
    Ui = @view U[idofs]
    for qp in 1:getnquadpoints(iv)
        n = getnormal(iv, qp; here = true)
        dF = getdetJdV(iv, qp)
        UL = function_value(iv, qp, Ui; here = true)
        UR = function_value(iv, qp, Ui; here = false)
        
        #Ujmp = function_value_jump(iv, qp, U)
        #c = max(max_speed(UL, n), max_speed(UR, n))
        #fhat = 0.5(F(UL)⋅n + F(UR)⋅n) - c*Ujmp
        
        fhat = LF(UL, UR, n)
        for i in 1:nb
            Φᵢ = 2*shape_value_jump(iv, qp, i)
            Ri[i] += (Φᵢ ⋅ fhat) * dF
        end
    end
    Ferrite.assemble!(R, idofs, Ri)
    return nothing
end
```

docs/src/literate-tutorials/dg_shallow_water.jl
```diff
@@ -0,0 +1,566 @@
...
+
+#-
+# ### Physics: bathymetry, flux, source, numerical flux
+const g = 9.81
+
+bat(X) = 0.5 * exp(-(X[1] - 5)^2 - (X[2] - 5)^2)
+batpx(X) = -2 * (X[1] - 5) * 0.5 * exp(-(X[1] - 5)^2 - (X[2] - 5)^2)
+batpy(X) = -2 * (X[2] - 5) * 0.5 * exp(-(X[1] - 5)^2 - (X[2] - 5)^2)
+
+function F(U)
```
lijas, line 206: We need F to ouput and 3x2 matrix, which also matches your equations

```
function F(U)
    h, qx, qy = U
    ux = qx / h
    uy = qy / h
    p = g * h * h / 2
    return @SMatrix [qx       qy;
              qx*ux+p  qy*ux;
              qx*uy    qy*uy+p]
end
```

lijas (commented):
Nice work! I dont know anything about the shallow water equations, but from reading the introduction it seems like the fields are grouped in to a vector U = [h, qx, qy] ? In that case, I think it would be better to treat this as a vector problem instead of three scalar equations. The code will become cleaner and follow the equations much closer. Below I will give suggestion on how to update parts of the code, but you have to update other parts of the code accordingly... This requires some refactoring but I think it can be worth it.

docs/src/literate-tutorials/dg_shallow_water.jl
yolhan83: Really nice thank you I will update afaik, I remember now why I didn't do that, I was a bit scared to mix SMatrix with Tensors but since you think it's all right 👍

termi-official: If we will merge https://github.com/Ferrite-FEM/Tensors.jl/pull/236 in the future then mixing the SMatrix in here also won't be necessary anymore. :)

Issue #1034, opened 2024-07-31
Linear Shell tutorial uses internals


KristofferC:
https://github.com/Ferrite-FEM/Ferrite.jl/blob/f7f2592bd26e8005a739cfe19169097c0146e269/docs/src/literate-tutorials/linear_shell.jl#L267

Should this be a Ferrite function?

KnutAM:
xref #842

KnutAM:
And #997 actually suggests that we should include this function (and more) in Ferrite, I think that would be a good option!

termi-official:
> https://github.com/Ferrite-FEM/Ferrite.jl/blob/f7f2592bd26e8005a739cfe19169097c0146e269/docs/src/literate-tutorials/linear_shell.jl#L267
> 
> Should this be a Ferrite function?

My take on this would be that we expose all of these functions to public, so users can write custom FEValues.

PR #1286, opened 2026-02-16
Use MixedTensors instead of StaticArrays for embedded

### LINKED ISSUE #1289: Wrong integration weight for embedded facetvalues
For an embedded line, we don't have `getdetJdV(fv::FacetValues, 1) ≈ 1`.

Fixed by #1286, but note here in case that takes some time to get merged.

KnutAM:
~Builds on https://github.com/Ferrite-FEM/Tensors.jl/pull/236~ (merged)
Requires Tensors v1.7


src/FEValues/FunctionValues.jl
```diff
@@ -179,17 +167,11 @@ required_geo_diff_order(::CovariantPiolaMapping, fun_diff_order::Int) = 1 + fun_
...
-# Hotfix to get the dots right for embedded elements until mixed tensors are merged.
-# Scalar/Vector interpolations with sdim == rdim (== vdim)
-@inline dothelper(A, B) = A ⋅ B
-# Vector interpolations with sdim == rdim != vdim
-@inline dothelper(A::SMatrix{vdim, dim}, B::Tensor{2, dim}) where {vdim, dim} = A * SMatrix{dim, dim}(B)
-# Scalar interpolations with sdim > rdim
-@inline dothelper(A::SVector{rdim}, B::SMatrix{rdim, sdim}) where {rdim, sdim} = B' * A
-# Vector interpolations with sdim > rdim
-@inline dothelper(B::SMatrix{vdim, rdim}, A::SMatrix{rdim, sdim}) where {vdim, rdim, sdim} = B * A
+# TODO: Have Tensors.jl support pinv?
```
termi-official: I would argue here with a yes, as we also need the pseudo inverse in other functions (e.g. point finder).

docs/Project.toml
```diff
@@ -26,3 +26,9 @@ TaskLocalValues = "ed4db957-447d-4319-bfb6-7fa9ae7ecf34"
 Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
+
+[compat]
+Optim = "1"
+
+[sources]
+Tensors = {url = "https://github.com/Ferrite-FEM/Tensors.jl.git", rev="kam/MixedTensors"}
```
termi-official: TODO remove this (left here so we do not forget)

docs/Project.toml
```diff
@@ -26,3 +26,9 @@ TaskLocalValues = "ed4db957-447d-4319-bfb6-7fa9ae7ecf34"
 Tensors = "48a634ad-e948-5137-8d70-aa71f2a747f4"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
+
+[compat]
+Optim = "1"
```
termi-official: Why is here just an isolated compat entry?

termi-official (commented):
Quite satisfactory to see all the hacks to get embedded stuff going being removed!

Can you also update the shell example to use the new interface? This way we can see how it pans out in practice (xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/1034).

docs/Project.toml
KnutAM: To not forget it - also in `Project.toml` for the package!

docs/Project.toml
KnutAM: https://github.com/JuliaNLSolvers/Optim.jl/issues/1234

KnutAM:
> Can you also update the shell example to use the new interface? This way we can see how it pans out in practice (xref https://github.com/Ferrite-FEM/Ferrite.jl/issues/1034).

I don't really know that example in detail. However, when looking over it I don't see any uses of `StaticArray`, so directly I don't think this PR makes it possible to avoid internals. This PR just avoids having to use `StaticArray`s for mixed-cases and tensors for non-mixed. I imagine (but don't know) that this would also be helpful for creating a `ShellValues` object or similar that could help having a nice infrastructure, but that is out of scope for this PR.

src/FEValues/FunctionValues.jl
```diff
@@ -5,35 +5,17 @@
...
-
-
-# Vector, vdim != sdim != rdim (TODO: Use Vec/Tensor if (s|r)dim <= 3?)
-typeof_N(::Type{T}, ::VectorInterpolation{vdim}, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, vdim, sdim, rdim} = SVector{vdim, T}
-typeof_dNdx(::Type{T}, ::VectorInterpolation{vdim}, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, vdim, sdim, rdim} = SMatrix{vdim, sdim, T, vdim * sdim}
-typeof_dNdξ(::Type{T}, ::VectorInterpolation{vdim}, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, vdim, sdim, rdim} = SMatrix{vdim, rdim, T, vdim * rdim}
-typeof_d2Ndx2(::Type{T}, ::VectorInterpolation{vdim}, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, vdim, sdim, rdim} = SArray{Tuple{vdim, sdim, sdim}, T, 3, vdim * sdim * sdim}
-typeof_d2Ndξ2(::Type{T}, ::VectorInterpolation{vdim}, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, vdim, sdim, rdim} = SArray{Tuple{vdim, rdim, rdim}, T, 3, vdim * rdim * rdim}
+typeof_dNdx(::Type{T}, ::ScalarInterpolation, ::VectorizedInterpolation{sdim, <:AbstractRefShape{rdim}}) where {T, sdim, rdim} = Vec{sdim, T}
```
termi-official, line 9: Do we want to leave some comment about the purpose of this code and/or in the dev docs?

src/FEValues/FunctionValues.jl
```diff
@@ -179,17 +159,11 @@ required_geo_diff_order(::CovariantPiolaMapping, fun_diff_order::Int) = 1 + fun_
...
-@inline dothelper(A, B) = A ⋅ B
-# Vector interpolations with sdim == rdim != vdim
-@inline dothelper(A::SMatrix{vdim, dim}, B::Tensor{2, dim}) where {vdim, dim} = A * SMatrix{dim, dim}(B)
-# Scalar interpolations with sdim > rdim
-@inline dothelper(A::SVector{rdim}, B::SMatrix{rdim, sdim}) where {rdim, sdim} = B' * A
-# Vector interpolations with sdim > rdim
-@inline dothelper(B::SMatrix{vdim, rdim}, A::SMatrix{rdim, sdim}) where {vdim, rdim, sdim} = B * A
+# TODO: Have Tensors.jl support pinv?
+@inline function calculate_Jinv(J::MixedTensor2{dim1, dim2}) where {dim1, dim2}
+    Js = SMatrix{dim1, dim2}(J.data)
```
termi-official: I guess there is no easy way to eliminate this StaticArrays dependency here? Or can we hard-code the Moore-Penrose pseudoinverse (or any other peusoinverse) here?

test/test_interpolations.jl
```diff
@@ -305,19 +305,6 @@ end
     end
 
     @testset "Correctness of AD of embedded interpolations" begin
-        ip = Lagrange{RefHexahedron, 2}()^3
-        ξ = rand(Vec{3, Float64})
-        for I in 1:getnbasefunctions(ip)
```
termi-official, line 310: Should we leave some test against finite differences here?

KnutAM: That would be a different test, the previous test just checked that the result from the path for static arrays (applied to the non-embedded case) matched that of Tensors.

That would be a different (unrelated IMO) addition: Then we should have a test for these directly for all interpolations, not just for the embedded case.

src/FEValues/FunctionValues.jl
KnutAM: I'll add a comment in the code, just used below so should be enough I think.

src/FEValues/FunctionValues.jl
KnutAM: I don't think it is important to eliminate the dependency, since Tensors.jl have that anyways. But solving this should be in `Tensors.jl` anyways (https://github.com/Ferrite-FEM/Tensors.jl/issues/241)

termi-official: > I don't think it is important to eliminate the dependency

I do not agree here. Any redundant dependency adds maintenance overhead. Especially now that we expand the ecosystem a bit a breaking release upstream can introduce unnecessary blocks.

termi-official: To settle this, here https://github.com/Ferrite-FEM/Tensors.jl/pull/244 is a faster implementation anyway which we should use. Turns out StaticArrays is not really efficient here as it does a full SVD.

src/FEValues/FunctionValues.jl
```diff
@@ -179,17 +159,13 @@ required_geo_diff_order(::CovariantPiolaMapping, fun_diff_order::Int) = 1 + fun_
...
-# Scalar interpolations with sdim > rdim
-@inline dothelper(A::SVector{rdim}, B::SMatrix{rdim, sdim}) where {rdim, sdim} = B' * A
-# Vector interpolations with sdim > rdim
-@inline dothelper(B::SMatrix{vdim, rdim}, A::SMatrix{rdim, sdim}) where {vdim, rdim, sdim} = B * A
+@inline function calculate_Jinv(
+        J::Union{ #MixedTensor2{sdim, rdim}
+            MixedTensor2{2, 1}, MixedTensor2{3, 1}, MixedTensor2{3, 2},
+        }
+    )
+    return inv(tdot(J)) ⋅ (J)' # Optimized pinv for invertible `J'⋅J`
```
termi-official: Maybe we should add a bit of context if someone stumbles on this.
```suggestion
    # Optimized Moore-Penrose pseudo inverse.
    # We assume that `J'⋅J` is invertible. This is a reasonable for non-degenerate elements.
    return inv(tdot(J)) ⋅ (J)' 
```

termi-official (approved):
LGTM. Thanks for working on this! Really awesome to strip away our nitty gritty hacks to get temporary support for embedded stuff and now get full Tensors.jl level support with the mixed tensor API!

PR #1290, opened 2026-02-26
Add storage eltype to abstract assembler types

KnutAM:
This makes it possible to infer the types of element matrices to allocate when given an assembler, i.e. 

```julia
function doassemble!(assembler::Ferrite.AbstractAssembler{T}, cv, dh) where {T}
    n = ndofs_per_cell(dh)
    Ke = zeros(T, n, n)
    for cell in CellIterator(dh)
        reinit!(cell, cv)
        element_routine!(Ke, cv)
        assemble!(assembler, celldofs(cell), Ke)
    end
end
```

### DIFF
--- a/test/test_assemble.jl
+++ b/test/test_assemble.jl
@@ -56,9 +56,9 @@ import LinearAlgebra: Symmetric
     @test all(K[rdofs, cdofs] .== Ke)
 
     # CSCAssembler: assemble with different row and col dofs
+    I = [1, 1, 4, 4, 6, 6]
+    J = [1, 3, 1, 3, 1, 3]
     for T in (Float32, Float64)
-        I = [1, 1, 4, 4, 6, 6]
-        J = [1, 3, 1, 3, 1, 3]
         V = zeros(T, length(I))
         K = sparse(I, J, V)
         f = zeros(T, 6)
@@ -129,7 +129,8 @@ import LinearAlgebra: Symmetric
     end
 
     # Error paths
-    assembler = start_assemble(Kc, fc)
+    K = sparse(I, J, zeros(length(I)))
+    assembler = start_assemble(K, zeros(size(K, 1)))
     @test_throws BoundsError assemble!(assembler, [11, 1, 2, 3], rand(4, 4))
     @test_throws BoundsError assemble!(assembler, [11, 1, 2, 3], rand(4, 4), rand(4))
     @test_throws BoundsError assemble!(assembler, [11, 1, 2], rand(4, 4))


src/assembler.jl
```diff
@@ -1,14 +1,14 @@
-abstract type AbstractAssembler end
-abstract type AbstractCSCAssembler <: AbstractAssembler end
-abstract type AbstractCSRAssembler <: AbstractAssembler end
+abstract type AbstractAssembler{Tv} end
```
termi-official, line 1: Why should this only contain the value type and not also the index type?

Also, we might at least want to leave some internal docstring on the rationale of why we want to include the value type in the abstract type here for future devs.

Also, should we then put this type into the assembly devdocs?

KnutAM: > Why should this only contain the value type and not also the index type?

I didn't include it since for a dense matrix there is no index type.

And even for sparse matrices, (I don't think), you want to provide `celldofs` as `Int32` even if the index type of the matrix is `Int32`?

> Also, should we then put this type into the assembly devdocs?

Yes - will add this! I think it makes sense to put the rationale in the devdocs too.

termi-official: Technically possible to use Int32 to boost memory access (i.e. you can query 2x the amount of memory in the same time), which gives you in the case of memory-bound assembly routines a nice speedup. But yes, in general we want Int64 (as Int32 is limited to 2^31-1 $\approx$ 2.000.000.000 dofs), which seems to also work out for large scale problems so far.

test/test_assemble.jl
```diff
@@ -56,21 +56,24 @@ import LinearAlgebra: Symmetric
...
-    rdofs = [1, 4, 6]
-    cdofs = [1, 3]
-    Ke = rand(length(rdofs), length(cdofs))
-    fe = rand(length(rdofs))
-    assemble!(assembler, rdofs, cdofs, Ke, fe)
-    assemble!(assembler, rdofs, cdofs, Ke, fe)
-    @test_throws ArgumentError assemble!(assembler, rdofs, Ke, fe) # Not in sparsity pattern
-    @test all(K[rdofs, cdofs] .== 2Ke)
-    @test all(f[rdofs] .== 2fe)
+    for T in (Float32, Float64)
```
termi-official, line 61: Should we have a similar test for the block and CSR assembler? We are essentially also testing for a regression on the assembly capability into Float32, right?

KnutAM: > We are essentially also testing for a regression on the assembly capability into Float32, right?

Yes, that was a low-hanging fruit, so added it that way. 


Added to the CSR too now, but for the block the tests felt like it would test too much (since they rely on `alocate_matrix` and didn't want to change the test interface there). For the change introduced here, I think this should be fine.

PR #1294, opened 2026-03-05
avoid using `threadid` in landau example, instead use OhMyThreads + ChunkSplitters

KristofferC:
Indexing by `threadid` is not really valid. I removed the `calcall` because it felt kind of pointless to just have hanging there.

### DIFF
--- a/docs/src/literate-gallery/landau.jl
+++ b/docs/src/literate-gallery/landau.jl
@@ -47,9 +47,9 @@ struct ModelParams{V, T}
     G::T
 end
 
-# ### ThreadCache
+# ### TaskCache
 # This holds the values that each task will use during the assembly.
-struct ThreadCache{CV, T, DIM, F <: Function, GC <: GradientConfig, HC <: HessianConfig}
+struct TaskCache{CV, T, DIM, F <: Function, GC <: GradientConfig, HC <: HessianConfig}
     cvP::CV
     element_indices::Vector{Int}
     element_dofs::Vector{T}
@@ -60,7 +60,7 @@ struct ThreadCache{CV, T, DIM, F <: Function, GC <: GradientConfig, HC <: Hessia
     gradconf::GC
     hessconf::HC
 end
-function ThreadCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpotential)
+function TaskCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpotential)
     element_indices = zeros(Int, dpc)
     element_dofs = zeros(dpc)
     element_gradient = zeros(dpc)
@@ -69,23 +69,23 @@ function ThreadCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpot
     potfunc = x -> elpotential(x, cvP, modelparams)
     gradconf = GradientConfig(potfunc, zeros(dpc), Chunk{12}())
     hessconf = HessianConfig(potfunc, zeros(dpc), Chunk{4}())
-    return ThreadCache(cvP, element_indices, element_dofs, element_gradient, element_hessian, element_coords, potfunc, gradconf, hessconf)
+    return TaskCache(cvP, element_indices, element_dofs, element_gradient, element_hessian, element_coords, potfunc, gradconf, hessconf)
 end
 
 # ## The Model
 # Everything is combined into a model. The caches are pre-allocated (one per task)
 # and indexed by chunk index during assembly.
-mutable struct LandauModel{T, DH <: DofHandler, CH <: ConstraintHandler, TC <: ThreadCache}
+mutable struct LandauModel{T, DH <: DofHandler, CH <: ConstraintHandler, TC <: TaskCache}
     dofs::Vector{T}
     dofhandler::DH
     boundaryconds::CH
-    threadindices::Vector{Vector{Int}}
+    colors::Vector{Vector{Int}}
     caches::Vector{TC}
 end
 
 function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elpotential, ntasks) where {DIM, T}
     grid = generate_grid(Tetrahedron, gridsize, left, right)
-    threadindices = Ferrite.create_coloring(grid)
+    colors = create_coloring(grid)
 
     qr = QuadratureRule{RefTetrahedron}(2)
     ipP = Lagrange{RefTetrahedron, 1}()^3
@@ -108,8 +108,8 @@ function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elp
 
     dpc = ndofs_per_cell(dofhandler)
     cpc = length(grid.cells[1].nodes)
-    caches = [ThreadCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for _ in 1:ntasks]
-    return LandauModel(dofvector, dofhandler, boundaryconds, threadindices, caches)
+    caches = [TaskCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for _ in 1:ntasks]
+    return LandauModel(dofvector, dofhandler, boundaryconds, colors, caches)
 end
 
 # utility to quickly save a model
@@ -140,9 +140,9 @@ end
 # This calculates the total energy of the grid.
 function F(dofvector::Vector{T}, model) where {T}
     out = zero(T)
-    for indices in model.threadindices
+    for indices in model.colors
         partial = OhMyThreads.@tasks for (ichunk, range) in enumerate(chunks(indices; n = length(model.caches)))
-            @set reducer = +
+            OhMyThreads.@set reducer = +
             cache = model.caches[ichunk]
             local_energy = zero(T)
             for i in range
@@ -161,7 +161,7 @@ end
 # so assembly is safe without locks.
 function ∇F!(∇f::Vector{T}, dofvector::Vector{T}, model::LandauModel{T}) where {T}
     fill!(∇f, zero(T))
-    for indices in model.threadindices
+    for indices in model.colors
         OhMyThreads.@tasks for (ichunk, range) in enumerate(chunks(indices; n = length(model.caches)))
             cache = model.caches[ichunk]
             for i in range
@@ -179,7 +179,7 @@ function ∇²F!(∇²f::SparseMatrixCSC, dofvector::Vector{T}, model::LandauMod
     dh = model.dofhandler
     ntasks = length(model.caches)
     assemblers = [start_assemble(∇²f; fillzero = (i == 1)) for i in 1:ntasks]
-    for indices in model.threadindices
+    for indices in model.colors
         OhMyThreads.@tasks for (ichunk, range) in enumerate(chunks(indices; n = ntasks))
             cache = model.caches[ichunk]
             for i in range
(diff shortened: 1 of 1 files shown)


docs/src/literate-gallery/landau.jl
```diff
@@ -47,7 +48,7 @@ struct ModelParams{V, T}
 end
 
 # ### ThreadCache
```
KnutAM:
```suggestion
# ### TaskCache
```

docs/src/literate-gallery/landau.jl
```diff
@@ -47,7 +48,7 @@ struct ModelParams{V, T}
...
-# This holds the values that each thread will use during the assembly.
+# This holds the values that each task will use during the assembly.
 struct ThreadCache{CV, T, DIM, F <: Function, GC <: GradientConfig, HC <: HessianConfig}
```
KnutAM:
```suggestion
struct TaskCache{CV, T, DIM, F <: Function, GC <: GradientConfig, HC <: HessianConfig}
```

docs/src/literate-gallery/landau.jl
```diff
@@ -72,16 +73,17 @@ function ThreadCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpot
...
 
 # ## The Model
-# everything is combined into a model.
+# Everything is combined into a model. The caches are pre-allocated (one per task)
+# and indexed by chunk index during assembly.
 mutable struct LandauModel{T, DH <: DofHandler, CH <: ConstraintHandler, TC <: ThreadCache}
     dofs::Vector{T}
     dofhandler::DH
     boundaryconds::CH
     threadindices::Vector{Vector{Int}}
```
KnutAM:
```suggestion
    colors::Vector{Vector{Int}}
```

docs/src/literate-gallery/landau.jl
```diff
@@ -72,16 +73,17 @@ function ThreadCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpot
...
-    threadcaches::Vector{TC}
+    caches::Vector{TC}
 end
 
-function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elpotential) where {DIM, T}
+function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elpotential, ntasks) where {DIM, T}
     grid = generate_grid(Tetrahedron, gridsize, left, right)
     threadindices = Ferrite.create_coloring(grid)
```
KnutAM:
```suggestion
    colors = create_coloring(grid)
```

docs/src/literate-gallery/landau.jl
```diff
@@ -106,7 +108,7 @@ function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elp
 
     dpc = ndofs_per_cell(dofhandler)
     cpc = length(grid.cells[1].nodes)
-    caches = [ThreadCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for t in 1:Threads.maxthreadid()]
+    caches = [ThreadCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for _ in 1:ntasks]
     return LandauModel(dofvector, dofhandler, boundaryconds, threadindices, caches)
```
KnutAM:
```suggestion
    return LandauModel(dofvector, dofhandler, boundaryconds, colors, caches)
```

docs/src/literate-gallery/landau.jl
```diff
@@ -106,7 +108,7 @@ function LandauModel(α, G, gridsize, left::Vec{DIM, T}, right::Vec{DIM, T}, elp
 
     dpc = ndofs_per_cell(dofhandler)
     cpc = length(grid.cells[1].nodes)
-    caches = [ThreadCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for t in 1:Threads.maxthreadid()]
+    caches = [ThreadCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for _ in 1:ntasks]
```
KnutAM:
```suggestion
    caches = [TaskCache(dpc, cpc, copy(cvP), ModelParams(α, G), elpotential) for _ in 1:ntasks]
```

docs/src/literate-gallery/landau.jl
```diff
@@ -119,75 +121,77 @@ function save_landau(path, model, dofs = model.dofs)
...
 end
 
-# This calculates the total energy calculation of the grid
+# This calculates the total energy of the grid.
 function F(dofvector::Vector{T}, model) where {T}
-    outs = fill(zero(T), Threads.maxthreadid())
-    @assemble! begin
-        outs[Threads.threadid()] += cache.element_potential(eldofs)
+    out = zero(T)
+    for indices in model.threadindices
```
KnutAM:
```suggestion
    for indices in model.colors
```

docs/src/literate-gallery/landau.jl
```diff
@@ -119,75 +121,77 @@ function save_landau(path, model, dofs = model.dofs)
...
-# The gradient calculation for each dof
+# The gradient calculation for each dof.
+# The grid coloring ensures no two tasks within a color share dofs,
+# so assembly is safe without locks.
 function ∇F!(∇f::Vector{T}, dofvector::Vector{T}, model::LandauModel{T}) where {T}
     fill!(∇f, zero(T))
-    @assemble! begin
-        ForwardDiff.gradient!(cache.element_gradient, cache.element_potential, eldofs, cache.gradconf)
-        @inbounds assemble!(∇f, cache.element_indices, cache.element_gradient)
+    for indices in model.threadindices
```
KnutAM:
```suggestion
    for indices in model.colors
```

docs/src/literate-gallery/landau.jl
```diff
@@ -119,75 +121,77 @@ function save_landau(path, model, dofs = model.dofs)
...
 # The Hessian calculation for the whole grid
 function ∇²F!(∇²f::SparseMatrixCSC, dofvector::Vector{T}, model::LandauModel{T}) where {T}
-    assemblers = [start_assemble(∇²f) for t in 1:Threads.maxthreadid()]
-    @assemble! begin
-        ForwardDiff.hessian!(cache.element_hessian, cache.element_potential, eldofs, cache.hessconf)
-        @inbounds assemble!(assemblers[Threads.threadid()], cache.element_indices, cache.element_hessian)
+    dh = model.dofhandler
+    ntasks = length(model.caches)
+    assemblers = [start_assemble(∇²f; fillzero = (i == 1)) for i in 1:ntasks]
+    for indices in model.threadindices
```
KnutAM:
```suggestion
    for indices in model.colors
```

docs/src/literate-gallery/landau.jl
```diff
@@ -72,16 +73,17 @@ function ThreadCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpot
 end
 
 # ## The Model
-# everything is combined into a model.
+# Everything is combined into a model. The caches are pre-allocated (one per task)
+# and indexed by chunk index during assembly.
 mutable struct LandauModel{T, DH <: DofHandler, CH <: ConstraintHandler, TC <: ThreadCache}
```
KnutAM:
```suggestion
mutable struct LandauModel{T, DH <: DofHandler, CH <: ConstraintHandler, TC <: TaskCache}
```

docs/src/literate-gallery/landau.jl
KnutAM:
```suggestion
function TaskCache(dpc::Int, nodespercell, cvP::CellValues, modelparams, elpotential)
```

KnutAM:
```suggestion
    return TaskCache(cvP, element_indices, element_dofs, element_gradient, element_hessian, element_coords, potfunc, gradconf, hessconf)
```

docs/src/literate-gallery/landau.jl
```diff
@@ -119,75 +121,77 @@ function save_landau(path, model, dofs = model.dofs)
...
-# This calculates the total energy calculation of the grid
+# This calculates the total energy of the grid.
 function F(dofvector::Vector{T}, model) where {T}
-    outs = fill(zero(T), Threads.maxthreadid())
-    @assemble! begin
-        outs[Threads.threadid()] += cache.element_potential(eldofs)
+    out = zero(T)
+    for indices in model.threadindices
+        partial = OhMyThreads.@tasks for (ichunk, range) in enumerate(chunks(indices; n = length(model.caches)))
+            @set reducer = +
```
KnutAM: Would this make sense to be extra clear where this macro comes from? (I assume it is from here)
```suggestion
            OhMyThreads.@set reducer = +
```

docs/src/literate-gallery/landau.jl
```diff
@@ -255,7 +259,7 @@ G = V2T(1.0e2, 0.0, 1.0e2)
 α = Vec{3}((-1.0, 1.0, 1.0))
 left = Vec{3}((-75.0, -25.0, -2.0))
 right = Vec{3}((75.0, 25.0, 2.0))
-model = LandauModel(α, G, (50, 50, 2), left, right, element_potential)
+model = LandauModel(α, G, (50, 50, 2), left, right, element_potential, Threads.nthreads())
```
KnutAM, line 263: Would be nice to add some quick tests to check that we don't make unintended changes (both here and for future)?
```suggestion
dh = model.dofhandler               #hide
ddf = allocate_matrix(dh)           #hide
df = zeros(ndofs(dh))               #hide
a = collect(range(0, 1, ndofs(dh))) #hide
@test F(a, model) ≈ ??              #hide
∇F!(df, a, model)                   #hide
@test norm(df) ≈ ??                 #hide
∇²F!(ddf, a, model)                 #hide
@test norm(ddf) ≈ ??                #hide
```

KristofferC:
Updated based on review comments. We will see if the result test passes everywhere.

docs/src/literate-gallery/landau.jl
```diff
@@ -119,75 +121,77 @@ function save_landau(path, model, dofs = model.dofs)
...
 
-# This calculates the total energy calculation of the grid
+# This calculates the total energy of the grid.
 function F(dofvector::Vector{T}, model) where {T}
-    outs = fill(zero(T), Threads.maxthreadid())
-    @assemble! begin
-        outs[Threads.threadid()] += cache.element_potential(eldofs)
+    out = zero(T)
+    for indices in model.colors
+        partial = OhMyThreads.@tasks for (ichunk, range) in enumerate(chunks(indices; n = length(model.caches)))
```
termi-official, line 144: Is there any advantage of using chunk splitters vs just using ranges?

KristofferC: I think `chunks` just returns a vector of ranges basically

docs/src/literate-gallery/landau.jl
```diff
@@ -255,12 +259,16 @@ G = V2T(1.0e2, 0.0, 1.0e2)
...
+model = LandauModel(α, G, (50, 50, 2), left, right, element_potential, Threads.nthreads())
 
 save_landau("landauorig", model)
-@time minimize!(model)
+@time res = minimize!(model)
+@assert Optim.converged(res)
 save_landau("landaufinal", model)
 
+using Test # src
+@test Optim.minimum(res) ≈ -10858.806775 # src
```
termi-official, line 270: Isn't the tolerance here a bit tight? I.e. can we really guarantee machine precision, such that in the future we won't see failures here popping up due to changes in Ferrite (or e.g. changes in Optim.jl).

termi-official (approved):
Other than one more note this PR is good from my side.

docs/src/literate-gallery/landau.jl
KristofferC: This is using approx so not machine precision? I already cut off a bunch of decimals here from the answer I got locally.

KnutAM: I guess the approx tolerance is quite tight for an optimization run (which is why I originally suggested just to test the different assembly runs), but I'm also fine merging this as is and then adopt the tests if we see random failures in the future (should be fine see manually that it fails with small changes now that we have a reference solution). Of course, we could do 
```suggestion
@test Optim.minimum(res) ≈ -10858.807f0 # src
```
to reduce the precision 😄

KristofferC: We can mess with it if we see there is an actual problem in the future?

Issue #1261, opened 2025-12-10
Coloring vtk export tests fails on prerelease


KnutAM:
Should be checked what changes if the test failure [here](https://github.com/Ferrite-FEM/Ferrite.jl/actions/runs/20091851467/job/57640909679) is an issue or not

PR #1295, opened 2026-03-05
make grid coloring not depend on iteration order of dicts and sets

KristofferC:
Resolves https://github.com/Ferrite-FEM/Ferrite.jl/issues/1261

### DIFF
--- a/src/Grid/coloring.jl
+++ b/src/Grid/coloring.jl
@@ -1,6 +1,6 @@
 # Incidence matrix for element connections in the grid
 function create_incidence_matrix(g::AbstractGrid, cellset = 1:getncells(g))
-    cell_containing_node = OrderedDict{Int, Set{Int}}()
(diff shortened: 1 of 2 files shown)


src/Grid/coloring.jl
```diff
@@ -1,6 +1,6 @@
 # Incidence matrix for element connections in the grid
 function create_incidence_matrix(g::AbstractGrid, cellset = 1:getncells(g))
-    cell_containing_node = Dict{Int, Set{Int}}()
+    cell_containing_node = OrderedDict{Int, Set{Int}}()
```
KnutAM: Is this change required? Or is it for speed? 

The final `incidence_matrix` is unique for a given grid, so should be independent on the iteration order, right?

KristofferC: Not needed, I've reverted this.

PR #1302, opened 2026-03-11
FastSparsityPattern

KnutAM:
I noticed during benchmarking that our general sparsity pattern is quite slow for larger systems. After some AI-assisted research I found a faster algorithm that is applicable in standard cases. Currently, it is only implemented for the standard cell coupling, but it should be possible to extend to interface coupling and constraints as well.

Update - **suggested plan:**
1) This is ready for review and can be merged from my PoV. `FastSparsityPattern` is **internal** and used as fastpath when possible. This gives a good speedup for standard cases.
2) We can successively add features with this algorithm (see [comment](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1302#issuecomment-4076932850) below)
3) It could potentially replace the current `SparsityPattern`, but it is too soon to say which strategy will be more performant for the more complicated sparsity patterns that the current `SparsityPattern` supports.

<details> <summary> Old notes </summary>
However, it is a two-pass algorithm, requiring that all information is there when doing the first pass. So while not implemented by this PR, my idea for closest approach to the current interface would be that every item is `add!`ed as for the current `SparsityPattern`, but that these steps are stored, and then we could call `close!` to actually create the pattern.

Need to add tests and docs. My idea is that this should replace `allocate_matrix(MatrixType, dh)` for `MatrixType` `SparseMatrixCSC` and `SparseMatrixCSR`, and the general pattern is used for more complicated cases. 

Timing results for allocating the default `SparseMatrixCSC` (script included below):
```julia
Triangle: nel = 10^4
ndofs_per_cell = 3, ndofs ≈ 10 k
SparsityPattern (master):   0.001921 seconds (43 allocations: 5.581 MiB)
FastSparsityPattern (PR):   0.000564 seconds (39 allocations: 3.251 MiB)
ndofs_per_cell = 18, ndofs ≈ 121 k
SparsityPattern (master):   0.066033 seconds (142 allocations: 129.310 MiB)
FastSparsityPattern (PR):   0.017080 seconds (39 allocations: 105.907 MiB)

Triangle: nel = 10^5
ndofs_per_cell = 3, ndofs ≈ 100 k
SparsityPattern (master):   0.028829 seconds (55 allocations: 26.643 MiB)
FastSparsityPattern (PR):   0.006261 seconds (39 allocations: 31.220 MiB)
ndofs_per_cell = 18, ndofs ≈ 1 M
SparsityPattern (master):   0.702183 seconds (1.07 k allocations: 1.264 GiB, 0.51% gc time)
FastSparsityPattern (PR):   0.178801 seconds (39 allocations: 1.029 GiB, 0.74% gc time)

Triangle: nel = 10^6
ndofs_per_cell = 3, ndofs ≈ 1 M
SparsityPattern (master):   0.309288 seconds (177 allocations: 222.022 MiB)
FastSparsityPattern (PR):   0.065926 seconds (39 allocations: 337.079 MiB)
ndofs_per_cell = 18, ndofs ≈ 12 M
SparsityPattern (master):  12.537659 seconds (10.32 k allocations: 12.493 GiB, 1.37% gc time)
FastSparsityPattern (PR):   2.231395 seconds (39 allocations: 10.279 GiB, 1.05% gc time)

Hexahedron: nel = 10^4
ndofs_per_cell = 8, ndofs ≈ 12 k
SparsityPattern (master):   0.008291 seconds (52 allocations: 16.573 MiB)
FastSparsityPattern (PR):   0.001680 seconds (39 allocations: 13.688 MiB)
ndofs_per_cell = 108, ndofs ≈ 364 k
SparsityPattern (master):   1.904070 seconds (2.42 k allocations: 2.681 GiB, 1.97% gc time)
FastSparsityPattern (PR):   0.327835 seconds (39 allocations: 2.030 GiB, 0.37% gc time)

Hexahedron: nel = 10^5
ndofs_per_cell = 8, ndofs ≈ 103 k
SparsityPattern (master):   0.084524 seconds (130 allocations: 90.648 MiB, 0.61% gc time)
FastSparsityPattern (PR):   0.017978 seconds (39 allocations: 75.798 MiB)
ndofs_per_cell = 108, ndofs ≈ 3 M
SparsityPattern (master):  29.977094 seconds (21.53 k allocations: 24.126 GiB, 1.08% gc time)
FastSparsityPattern (PR):   5.577470 seconds (39 allocations: 18.241 GiB, 0.85% gc time)

Hexahedron: nel = 10^6
ndofs_per_cell = 8, ndofs ≈ 1 M
SparsityPattern (master):   0.863224 seconds (813 allocations: 885.615 MiB, 5.52% gc time)
FastSparsityPattern (PR):   0.179126 seconds (39 allocations: 767.470 MiB, 0.51% gc time)
ndofs_per_cell = 51, ndofs ≈ 11 M
SparsityPattern (master):  76.650587 seconds (43.36 k allocations: 47.272 GiB, 1.04% gc time)
FastSparsityPattern (PR):  16.670520 seconds (39 allocations: 34.905 GiB, 0.29% gc time)
```

<details> <summary> Benchmarking script </summary>

```julia
using Ferrite

function timeit(dh; check, compile)
    if compile
        allocate_matrix(dh)
        allocate_matrix_pr(dh)
    end
    print("SparsityPattern (master): ")
    GC.gc() # Ensure fair comparison
    K1 = @time allocate_matrix(dh)
    
    print("FastSparsityPattern (PR): ")
    GC.gc()
    K2 = @time allocate_matrix_pr(dh)
    if check
        # Validate that they are the same
        map!(identity, K1.nzval, 1:length(K1.nzval))
        map!(identity, K2.nzval, 1:length(K2.nzval))
        K1 == K2 || error("Results are not the same")
    end
    return nothing
end

for CT in (Triangle, Hexahedron)
    for nel in (10^4, 10^5, 10^6)
        RefShape = getrefshape(CT)
        dim = Ferrite.getrefdim(CT)
        nel_i = round(Int, nel^(1//dim))
        grid = generate_grid(CT, ntuple(_->nel_i, dim))
        println("$CT: nel = 10^", round(Int, log10(nel)))
        # Case 1, single 1st order scalar interpolation (few dofs per cell)
        dh = close!(add!(DofHandler(grid), :u, Lagrange{RefShape, 1}()))
        dofstr(n) = n > 10^6 ? string(n ÷ 10^6, " M") : (n > 10^3 ? string(n ÷ 10^3, " k") : string(n))
        println("ndofs_per_cell = ", ndofs_per_cell(dh), ", ndofs ≈ ", dofstr(ndofs(dh)))
        timeit(dh; compile = nel == 10^4, check = false)

        # Case 2, multiple and higher order interpolations (more dofs per cell)
        dh = DofHandler(grid)
        order = nel == 10^6 && CT == Hexahedron ? 1 : 2 # 2nd order vector for large 3d grid not feasible
        add!(dh, :u, Lagrange{RefShape, order}()^dim)
        add!(dh, :p, Lagrange{RefShape, 2}())
        close!(dh)
        println("ndofs_per_cell = ", ndofs_per_cell(dh), ", ndofs ≈ ", dofstr(ndofs(dh)))
        timeit(dh; compile = nel == 10^4, check = false)
        println()
    end
end
```
</details>

</details>

### DIFF
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -134,7 +134,7 @@ function sort_rows!(sp::FastSparsityPattern, rowrange::UnitRange)
         i1 = sp.rowptr[row]
         i2 = sp.rowptr[row + 1] - 1
         if i1 < i2
-            sort!(view(sp.colidx, i1:i2))
+            sort!(view(sp.colidx, i1:i2); alg = QuickSort)
         end
     end
     return sp
@@ -145,15 +145,11 @@ function sort_rows_threaded!(
         ntasks = max(min(Threads.nthreads() * 100, getnrows(sp) ÷ 1000), 1)
     )               # Otherwise, 100 per thread for load balancing
     nrows = getnrows(sp)
-    ΔN = nrows ÷ ntasks
-    Base.Experimental.@sync begin
-        for taskid in 1:ntasks
-            Threads.@spawn begin
-                first_idx = 1 + ΔN * (taskid - 1)
-                last_idx = min(first_idx + ΔN - 1, nrows)
-                sort_rows!(sp, first_idx:last_idx)
-            end
-        end
+    ΔN = (nrows + 1) ÷ ntasks
+    Threads.@threads for taskid in 1:ntasks
+        first_idx = 1 + ΔN * (taskid - 1)
+        last_idx = min(first_idx + ΔN - 1, nrows)
+        sort_rows!(sp, first_idx:last_idx)
     end
     sp.is_colidx_sorted = true
     return sp
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -123,12 +123,6 @@ function _allocate_matrix(::Type{SparseMatrixCSR{1, Tv, Ti}}, sp::FastSparsityPa
     return SparseMatrixCSR{1}(getnrows(sp), getncols(sp), sp.rowptr, sp.colidx, nzval)
 end
 
-function sort_rows!(sp)
-    sort_rows!(sp, 1:getnrows(sp))
-    sp.is_colidx_sorted = true
-    return sp
-end
-
 function sort_rows!(sp::FastSparsityPattern, rowrange::UnitRange)
     @inbounds for row in rowrange
         i1 = sp.rowptr[row]
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -145,7 +145,7 @@ function sort_rows_threaded!(
         ntasks = max(min(Threads.nthreads() * 100, getnrows(sp) ÷ 1000), 1)
     )               # Otherwise, 100 per thread for load balancing
     nrows = getnrows(sp)
-    ΔN = (nrows + 1) ÷ ntasks
+    ΔN = cld(nrows, ntasks)
     Threads.@threads for taskid in 1:ntasks
         first_idx = 1 + ΔN * (taskid - 1)
         last_idx = min(first_idx + ΔN - 1, nrows)
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -1,7 +1,7 @@
 module FerriteSparseMatrixCSR
 
 using Ferrite, SparseArrays, SparseMatricesCSR
-import Ferrite: AbstractSparsityPattern, CSRAssembler
+import Ferrite: AbstractSparsityPattern, CSRAssembler, FastSparsityPattern, getnrows, getncols
 import Base: @propagate_inbounds
 
 # Could be generalized if https://github.com/JuliaSparse/SparseArrays.jl/pull/546 is merged
@@ -110,6 +110,53 @@ function _allocate_matrix(::Type{SparseMatrixCSR{1, Tv, Ti}}, sp::AbstractSparsi
     end
     S = SparseMatrixCSR{1}(Ferrite.getnrows(sp), Ferrite.getncols(sp), rowptr, colval, nzval)
     return S
+end
+
+## ================= ##
+# FastSparsityPattern #
+## ================= ##
+
+function _allocate_matrix(::Type{SparseMatrixCSR{1, Tv, Ti}}, sp::FastSparsityPattern{Ti}, sym::Bool) where {Tv, Ti}
+    sym && throw(ArgumentError("FastSparsityPattern does not support symmetric matrices yet"))
+    sp.is_colidx_sorted || sort_rows_threaded!(sp) # Require sorted rows
+    nzval = zeros(Tv, length(sp.colidx))
+    return SparseMatrixCSR{1}(getnrows(sp), getncols(sp), sp.rowptr, sp.colidx, nzval)
+end
+
+function sort_rows!(sp)
+    sort_rows!(sp, 1:getnrows(sp))
+    sp.is_colidx_sorted = true
+    return sp
+end
+
+function sort_rows!(sp::FastSparsityPattern, rowrange::UnitRange)
+    @inbounds for row in rowrange
+        i1 = sp.rowptr[row]
+        i2 = sp.rowptr[row + 1] - 1
+        if i1 < i2
+            sort!(view(sp.colidx, i1:i2))
+        end
+    end
+    return sp
+end
+
+function sort_rows_threaded!(
+        sp::FastSparsityPattern, # Default ΔN ≥ 1000 and `n_tasks ≥ 1`
+        ntasks = max(min(Threads.nthreads() * 100, getnrows(sp) ÷ 1000), 1)
+    )               # Otherwise, 100 per thread for load balancing
+    nrows = getnrows(sp)
+    ΔN = nrows ÷ ntasks
+    Base.Experimental.@sync begin
+        for taskid in 1:ntasks
+            Threads.@spawn begin
+                first_idx = 1 + ΔN * (taskid - 1)
+                last_idx = min(first_idx + ΔN - 1, nrows)
+                sort_rows!(sp, first_idx:last_idx)
+            end
+        end
+    end
+    sp.is_colidx_sorted = true
+    return sp
 end
 
 end
--- a/src/Dofs/sparsity_pattern.jl
+++ b/src/Dofs/sparsity_pattern.jl
@@ -681,3 +681,195 @@ function _allocate_matrix(::Type{SparseMatrixCSC{Tv, Ti}}, sp::AbstractSparsityP
     S = SparseMatrixCSC(getnrows(sp), getncols(sp), colptr, rowval, nzval)
     return S
 end
+
+## ================== ##
+# FastSparsityPattern #
+## ================== ##
+
+# Full `AbstractSparsityPattern` interface not supported
+mutable struct FastSparsityPattern{Ti} <: AbstractSparsityPattern
+    const rowlen::Vector{Ti} # Number of stored entries in each row
+    const marker::Vector{Ti} # Marker if column has been "visited" by certain row
+    const rowptr::Vector{Ti} # Index of stored entries at the start of each row
+    const colidx::Vector{Ti} # colidx[i] gives the column number of the ith stored entry
+    is_colidx_sorted::Bool   # Is colidx sorted for each row
+end
+function FastSparsityPattern(::Type{Ti}, ncols, nrows) where {Ti <: Integer}
+    rowlen = zeros(Ti, nrows)
+    marker = zeros(Ti, ncols)
+    rowptr = Vector{Ti}(undef, nrows + 1)
+    colidx = Vector{Ti}(undef, 0) # To be resized later
+    return FastSparsityPattern(rowlen, marker, rowptr, colidx, false)
+end
+
+FastSparsityPattern(dh::DofHandler) = FastSparsityPattern(Int, dh)
+function FastSparsityPattern(::Type{Ti}, dh::DofHandler) where {Ti}
+    sp = FastSparsityPattern(Ti, ndofs(dh), ndofs(dh))
+    # Step 1: Create cell_dof_views::ArrayOfVectorViews (would be nice in `DofHandler` directly)
+    ncells = getncells(dh.grid)
+    indices = copy(dh.cell_dofs_offset)
+    push!(indices, length(dh.cell_dofs) + 1)
+    cell_dofs_views = ArrayOfVectorViews(indices, dh.cell_dofs, LinearIndices((ncells,)))
+    # Step 2: Define mapping rownr to cells
+    row_to_cells = create_row_to_cells(cell_dofs_views, sp)
+    # Step 3: Count how many cols stored for each row
+    count_row_sizes!(sp, row_to_cells, cell_dofs_views)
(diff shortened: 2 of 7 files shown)


KnutAM:
> didn't like the add_entry! interface

yes, that would require some accumulator for all those. 
Should be manageable with using the ArrayOfVectorViews and the addition of close! to the interface, such that `sp` doesn’t need to be valid after each

src/Dofs/sparsity_pattern.jl
```diff
@@ -681,3 +681,151 @@ function _allocate_matrix(::Type{SparseMatrixCSC{Tv, Ti}}, sp::AbstractSparsityP
...
+                    sp.colidx[pos] = col
+                    pos += 1
+                end
+            end
+        end
+    end
+    return sp
+end
+
+allocate_matrix(sp::FastSparsityPattern) = allocate_matrix(SparseMatrixCSC, sp)
```
termi-official, line 849: Since there is no benefit in the short-call `function allocate_matrix(::Type{MatrixType}, dh::DofHandler, args...; kwargs...) where {MatrixType}` of using the generic sparsity pattern, shoudn't we make this pattern the default for CSC and CSR matrices?

ext/FerriteSparseMatrixCSR.jl
```diff
@@ -112,4 +112,51 @@ function _allocate_matrix(::Type{SparseMatrixCSR{1, Tv, Ti}}, sp::AbstractSparsi
...
+    return sp
+end
+
+function sort_rows_threaded!(
+        sp::FastSparsityPattern, # Default ΔN ≥ 1000 and `n_tasks ≥ 1`
+        ntasks = max(min(Threads.nthreads() * 100, getnrows(sp) ÷ 1000), 1)
+    )               # Otherwise, 100 per thread for load balancing
+    nrows = getnrows(sp)
+    ΔN = nrows ÷ ntasks
+    Base.Experimental.@sync begin
```
termi-official: Isn't this already a base feature for quite a while?
```suggestion
    Base.@sync begin
```
At least I could not find in the Julia docs why we should have the `Experimental` module here.

termi-official (commented):
Don't we miss test coverage here or am I missing something?

Also, could you store the benchmark in the benchmarks folder for future reference (when we will revive that folder).

ext/FerriteSparseMatrixCSR.jl
KristofferC: They do different things. The experimental one throws if one of the tasks inside the block throws. 

```
help?> Base.Experimental.@sync

  Wait until all lexically-enclosed uses of @async, @spawn, Distributed.@spawnat and
  Distributed.@distributed are complete, or at least one of them has errored. The first
  exception is immediately rethrown. It is the responsibility of the user to cancel any
  still-running operations during error handling.

  !!! Note This is different to @sync in that errors from wrapped tasks are thrown
  immediately, potentially before all tasks have returned.
```

termi-official: Is it intended that this function is not reachable with the docs search https://docs.julialang.org/en/v1/ ?

termi-official: And thanks for the elaboration!

KristofferC: I am not sure what the "policy" is when it comes to docs for the experimental stuff.

KnutAM: I've just copied this from @fredrikekre's threaded assembly tutorial branch, and use it in FerriteAssembly too. I'm guessing `Experimental.@sync` just gives easier-to-interpret error messages and maybe has some performance advantages. Is this correct, @KristofferC or @fredrikekre ?

KnutAM: Actually - this is not needed at all, updating to using the `Threads.@threads` since there is no buffer - this allocates less and is equally fast.

src/Dofs/sparsity_pattern.jl
KnutAM: > Since there is no benefit in the short-call function allocate_matrix(::Type{MatrixType}, dh::DofHandler, args...; kwargs...) where {MatrixType} 

There is - you can pass `args = (ch,)`, and several `kwargs...` that aren't supported by `FastSparsityPattern`

KnutAM: An update is coming up though that will make the fast path being the default though

ext/FerriteSparseMatrixCSR.jl
```diff
@@ -112,4 +112,47 @@ function _allocate_matrix(::Type{SparseMatrixCSR{1, Tv, Ti}}, sp::AbstractSparsi
...
+    sp.is_colidx_sorted || sort_rows_threaded!(sp) # Require sorted rows
+    nzval = zeros(Tv, length(sp.colidx))
+    return SparseMatrixCSR{1}(getnrows(sp), getncols(sp), sp.rowptr, sp.colidx, nzval)
+end
+
+function sort_rows!(sp)
+    sort_rows!(sp, 1:getnrows(sp))
+    sp.is_colidx_sorted = true
+    return sp
+end
```
termi-official: This function is not used.

src/Dofs/sparsity_pattern.jl
```diff
@@ -413,6 +413,7 @@ arguments `args` and keyword arguments `kwargs`.
     copy(K)`) instead.
 """
 function allocate_matrix(::Type{MatrixType}, dh::DofHandler, args...; kwargs...) where {MatrixType}
+    _can_use_fastsp(args...; kwargs...) && return allocate_matrix(MatrixType, FastSparsityPattern(dh, args...; kwargs...))
```
termi-official: Maybe we also need to add a check that the dispatch `allocate_matrix(MatrixType, FastSparsityPattern(dh, args...; kwargs...))` exists.

src/Dofs/sparsity_pattern.jl
```diff
@@ -681,3 +682,185 @@ function _allocate_matrix(::Type{SparseMatrixCSC{Tv, Ti}}, sp::AbstractSparsityP
...
+    # Step 4: Build the rowptr (indices for s)
+    build_rowptr!(sp)
+    fill_colidx!(sp, row_to_cells, cell_dofs_views)
+    return sp
+end
+
+getncols(sp::FastSparsityPattern) = length(sp.marker)
+getnrows(sp::FastSparsityPattern) = length(sp.rowlen)
+
+function create_celldofs(dh::DofHandler)
```
termi-official, line 760: Why do we need this transformation? This step looks redundant, as we usually work directly on the cell dofs.

src/Dofs/sparsity_pattern.jl
termi-official: Oh wait, shouldn't the function signature actually have `MatrixType <: SparseMatrixCSC`? Otherwise this callback might cause some very interesting trouble.

test/test_sparsity_patterns.jl
```diff
@@ -331,3 +332,29 @@ end
         @test_throws ErrorException add_sparsity_entries!(p, dh, ch_bad; keep_constrained = false)
     end
 end
+
+@testset "FastSparsityPattern" begin
```
termi-official, line 337: Maybe we should also add a path to cover the case when we have a non Int data type.

src/Dofs/sparsity_pattern.jl
KnutAM: It is a lot less convenient (and a bit slower), since that relies on looking up the number of dofs in the current cell which first checks which subdofhandler the cell belongs to and checks how many dofs per cell for that subdofhandler. 

This is why I commented above that it would be nice to store this in the dofhandler directly (replacing the current `cell_dofs_offset` and `cell_dofs` fields). But I would leave this to a separate PR. This would simplify the code (and make it faster) in other parts of the codebase too. 

With this - we simply have `celldofs[cellnr]` giving a view of the dofs for that cell which is faster than creating a view based on the current fields.

ext/FerriteSparseMatrixCSR.jl
KnutAM: Yep, if we are ok with having the threaded version only then we can remove this. Wasn't sure since all other ops in Ferrite are single threaded. Alternatively, we could check if it is running on a single thread that this should be used then to avoid spawning tasks in that case.

termi-official: Is there any disadvantage for having only a threaded version?

KnutAM: If single thread then slightly slower for small matrices. I can’t think of a use case where it matters though, so will delete

src/Dofs/sparsity_pattern.jl
KnutAM: done

test/test_sparsity_patterns.jl
KnutAM: done

src/Dofs/sparsity_pattern.jl
termi-official: > This would simplify the code (and make it faster) in other parts of the codebase too.

But this would just be true in the case of having iterations over the dof handler instead of having iterations over the subdofhandler, as recommended.

termi-official: I understand that we will need something like this for performant p-adaptivity support, but I do not see why we should pay the overhead for the currently working cases.

KnutAM: > But this would just be true in the case of having iterations over the dof handler instead of having iterations over the subdofhandler, as recommended.

No, the current code for `celldofs!` is essentially
```julia
function celldofs!(global_dofs::AbstractVector{Int}, dh::AbstractDofHandler, i::Int)
    sdhidx = dh.cell_to_subdofhandler[i]
    N = if sdhidx ∉ 1:length(dh.subdofhandlers)
        0 # Dof handler is just defined on a subdomain
    else
        ndofs_per_cell(dh.subdofhandlers[sdhidx])
    end # Edit: copy-paste-bug
    return copyto!(global_dofs, 1, dh.cell_dofs, dh.cell_dofs_offset[i], N)
end
```

With the new storage format (stores 1 more integer), we get 
```julia
function celldofs!(global_dofs::AbstractVector{Int}, dh::AbstractDofHandler, i::Int)
    copyto!(global_dofs, dh.cell_dofs[i])
end
```
(This can be made more efficient by using the copyto with offsets based on length and the underlying vector if we want to). 

But even more attractive, we don't need the mutating version in many cases since we can simply call `celldofs(dh, i)` which can be implemented as `celldofs(dh, i) = dh.cell_dofs[i]`.


When doing operations that don't involve the type of cells, there is no need to loop over subdofhandlers separately, and just looping over all cells will be faster (since the only difference is the number of dofs).

src/Dofs/sparsity_pattern.jl
```diff
@@ -681,3 +687,188 @@ function _allocate_matrix(::Type{SparseMatrixCSC{Tv, Ti}}, sp::AbstractSparsityP
...
 end
+
+## ================= ##
+# FastSparsityPattern #
+## ================= ##
+
+# Full `AbstractSparsityPattern` interface not supported
+# For now, this remains a non-public fast-path for `allocate_matrix(MatrixType, dh)`,
+# but can be extended in the future or potentially replace `SparsityPattern`.
+# See #1302 for details.
```
termi-official: Can we make this a doc string and mark it as strictly internal for now? We might find a way to make the general sparsity pattern faster in the future and remove this again.

termi-official (approved):
I left one more comment which we might want to resolve here. LGTM otherwise.

Issue #397, opened 2021-10-21
Use ReadVTK.jl for VTK output testing


fredrikekre:
We can probably use https://github.com/trixi-framework/ReadVTK.jl for testing VTK output instead of the SHA1 hashing strategy.

PR #1303, opened 2026-03-12
Add function to write all facetsets and fix cellset export docs

termi-official:
I have this function in essentially all of my downstream packages.

### DIFF
--- a/src/Export/VTK.jl
+++ b/src/Export/VTK.jl
@@ -285,6 +285,7 @@ function write_nodeset(vtk, grid::AbstractGrid, nodeset::String)
 end
 
 """
+    write_facetset(vtk::VTKGridFile, grid::AbstractGrid)
     write_facetset(vtk::VTKGridFile, grid::AbstractGrid, facetsetname::String)
     write_facetset(vtk::VTKGridFile, grid::AbstractGrid, facetset::AbstractVecOrSet{FacetIndex}, facetsetname::String)
 
@@ -305,17 +306,25 @@ function write_facetset(vtk, grid::AbstractGrid, facetset::AbstractVecOrSet{Face
     return vtk
 end
 write_facetset(vtk, grid::AbstractGrid, facetsetname::String) = write_facetset(vtk, grid, getfacetset(grid, facetsetname), facetsetname)
+function write_facetset(vtk, grid::AbstractGrid)
+    for name in keys(getfacetsets(grid))
+        write_facetset(vtk, grid, name)
+    end
+    return vtk
+end
 
 """
     write_cellset(vtk, grid::AbstractGrid)
     write_cellset(vtk, grid::AbstractGrid, cellset::String)
-    write_cellset(vtk, grid::AbstractGrid, cellsets::Union{AbstractVector{String},AbstractSet{String})
+    write_cellset(vtk, grid::AbstractGrid, cellsets::Union{AbstractVector{String}, AbstractSet{String})
 
 Write all cell sets in the grid with name according to their keys and
 celldata 1 if the cell is in the set, and 0 otherwise. It is also possible to
 only export a single `cellset`, or multiple `cellsets`.
+
+If you want to write per cell data directly, please see [`write_cell_data`](@ref).
 """
-function write_cellset(vtk, grid::AbstractGrid, cellsets = keys(getcellsets(grid)))
+function write_cellset(vtk, grid::AbstractGrid, cellsets::AbstractVecOrSet{String} = keys(getcellsets(grid)))
     z = zeros(getncells(grid))
     for cellset in cellsets
         fill!(z, 0)
--- a/test/test_vtk_export.jl
+++ b/test/test_vtk_export.jl
@@ -141,7 +141,27 @@
             @test bytes2hex(open(SHA.sha1, fname2 * ".vtu")) == testhash
         end
     end
+    @testset "write_facetsets" begin
+        # More tests in `test_grid_dofhandler_vtk.jl`
+        mktempdir() do tmp
+            grid = generate_grid(Quadrilateral, (2, 2))
 
+            v1 = VTKGridFile(joinpath(tmp, "g1"), grid) do vtk::VTKGridFile
+                @test Ferrite.write_facetset(vtk, grid, "left") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "bottom") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "right") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "top") === vtk
+            end
+            @test v1 isa VTKGridFile
+
+            v2 = VTKGridFile(joinpath(tmp, "g2"), grid) do vtk::VTKGridFile
+                @test Ferrite.write_facetset(vtk, grid) === vtk
+            end
+            @test v2 isa VTKGridFile
+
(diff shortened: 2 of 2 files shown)


src/Export/VTK.jl
```diff
@@ -305,17 +306,25 @@ function write_facetset(vtk, grid::AbstractGrid, facetset::AbstractVecOrSet{Face
     return vtk
 end
 write_facetset(vtk, grid::AbstractGrid, facetsetname::String) = write_facetset(vtk, grid, getfacetset(grid, facetsetname), facetsetname)
+function write_facetset(vtk, grid::AbstractGrid)
+    for name in keys(getfacetsets(grid))
+        write_facetset(vtk, grid, name)
```
KnutAM, line 311: n many cases there might be an overlap in names between different types of sets.

So when I've needed this, I've ended up prepending by "facetset_" * name. For consistency, we should do the same with cellsets then, but perhaps ok to have those without prefix to not change the output?

termi-official: We never did this. If users want to have the facetset_ prefix, they can name the sets so themselves, right?

KnutAM: Note - was wrong regarding cellsets - these are different data and can co-exist with the same name.

However, the clash comes with `Ferrite.write_nodeset`. This will silently fail. For Abaqus, facets are described by node sets, which are therefore also added as facetsets in `FerriteMeshParser`.

The main issue is that there is no check for overwritten outputs, so the following
```julia
julia> grid = generate_grid(Quadrilateral, (10,10));
julia> addnodeset!(grid, "left", x -> x[1] < 0);
julia> VTKGridFile("test1", grid) do vtk
           Ferrite.write_facetset(vtk, grid, "left")
           Ferrite.write_nodeset(vtk, grid, "left")
       end;
julia> VTKGridFile("test2", grid) do vtk
           Ferrite.write_nodeset(vtk, grid, "left")
           Ferrite.write_facetset(vtk, grid, "left")
       end;
```
produces different results.

However, since we have neither `write_facetset` or `write_nodeset` for all stored sets at the moment, we could easily have the default prefixes IMO without being breaking while being consistent.

My suggestion would be
```julia
write_facetset(vtk, grid::AbstractGrid; prefix = "facetset_")
write_nodeset(vtk, grid::AbstractGrid; prefix = "nodeset_")
```
I'd prefer having the non-empty default `prefix` there, but we could of course have `prefix = ""` and add a warning that it will overwrite if there are other (node/facet)sets with the same name.

termi-official: > For Abaqus, facets are described by node sets [...]

Luckily we are using Ferrite here. :P I do not bother too much how software X handles their boundaries tbh. I also do not see any advantage in adding, in addition to the facetsets, an additional set with exactly the same name containign all the nodes on these facetsets. This additional data is essentially just dead weight.

termi-official: There is no sane way of checking if a field exists already in VTK so the warning you ask for should go directly into WriteVTK.

termi-official: Okay tracking this further down, having mutiple data fields with the same name is allowed by the file specification (https://docs.vtk.org/en/latest/vtk_file_formats/vtkxml_file_format.html). Paraview deviates here requiring unique names.

KnutAM: I think we at minimum should provide the `prefix = ""` kwarg, and add to the docstring that this can be used to differentiate different types of set with the same name.
 
> There is no sane way of checking if a field exists already in VTK so the warning you ask for should go directly into WriteVTK.

I think the key point is that we provide several different entrypoints to `WriteVTK.vtk_point_data`, which is not obvious to the user. Even if it says in the docs that it writes nodal data, this is an easy caveat which might cost users quite some debug time.

KnutAM: But let's not this discussion stop progress :D
We can merge this as-is, and add the prefix kwarg later.

test/test_vtk_export.jl
```diff
@@ -141,7 +141,27 @@
...
+    @testset "write_facetsets" begin
+        # More tests in `test_grid_dofhandler_vtk.jl`
+        mktempdir() do tmp
+            grid = generate_grid(Quadrilateral, (2, 2))
+
+            v1 = VTKGridFile(joinpath(tmp, "g1"), grid) do vtk::VTKGridFile
+                @test Ferrite.write_facetset(vtk, grid, "left") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "bottom") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "right") === vtk
+                @test Ferrite.write_facetset(vtk, grid, "top") === vtk
```
KnutAM, line 153: Failure due to this is so annoying. To make sure we would need change in the Grid to `facetsets::OrderedDict{String, OrderedSet{FacetIndex}}` just to make testing more stable.

Here I think we could equally well do
```suggestion
                for name in keys(Ferrite.getfacetsets(grid))
                    @test Ferrite.write_facetset(vtk, grid, name) === vtk
                end
```

test/test_vtk_export.jl
```diff
@@ -141,7 +141,27 @@
...
+            end
+            @test v1 isa VTKGridFile
 
+            v2 = VTKGridFile(joinpath(tmp, "g2"), grid) do vtk::VTKGridFile
+                @test Ferrite.write_facetset(vtk, grid) === vtk
+            end
+            @test v2 isa VTKGridFile
+
+            @test bytes2hex(open(SHA.sha1, joinpath(tmp, "g1.vtu"))) == bytes2hex(open(SHA.sha1, joinpath(tmp, "g2.vtu")))
```
termi-official, line 162: Again the same "issue" as before, as the test relies on the dict insertion order, which is not preserved in the new Julia release (and we are not allowed to depend on this in first place). @KristofferC how are we supposed to test this? Loop over all permutations for the sets in v1?

KnutAM (commented):
Forgot to click submit for review :D

test/test_vtk_export.jl
KnutAM: Should be fine if we use the same dict - but I think this is a problem with the sha-testing?

termi-official: The test here should be "test whether all sets with the correct names and entries have been written out", independent of the order. Here we explicitly have a dependency on the order of the write.

test/test_vtk_export.jl
termi-official: Yea I thought about this, but I do not like this solution a lot.

KnutAM: Why not? Because it is similar to the implementation? The only change compared to the original is relying on that getfacetsets works and that the expected sets are stored in the grid.

test/test_vtk_export.jl
KnutAM: #397 would enable solving this but I wouldn’t let that block this pr.

test/test_vtk_export.jl
termi-official: You rely on implementation details for the test. This is not really different from what we do right now for the dict driectly and we are just shifting the problem around. We should actually test that all fields have been written exactly once, independent of the write order.

KnutAM: In that case I would propose to remove two of the sets and check that exactly one of the two possible remaining permutations are stored. Personally I don’t think this is necessary because the parts you rely on are things tested elsewhere, but before ReadVTK testing is in place I think we should just do it easy here.

PR #1305, opened 2026-03-13
Change in notation of (test/ trial) functions and shape_value

Avina-cK:
In all changed files: Change in notation of test functions: δu, v -> ϕ
In transient_heat_equation.jl:
- Assembly function of K,M,f merged to one.
- Added reference to fe_into and heat_equation.jl


docs/src/topics/fe_intro.md
```diff
@@ -47,16 +47,16 @@ simplicity we will consider only constant conductivity $k$.
 ## Weak form
 
 The solution to the equation above is usually calculated from the corresponding weak
-form. By multiplying the equation with an arbitrary *test function* $\delta u$, integrating
+form. By multiplying the equation with an arbitrary *test function* $\varphi$, integrating
```
termi-official, line 50: I definitely see the point made by the contributor here @KnutAM . Since we explain it here from the Bubnov-Galerkin perspective I see that the notation $\delta u$ can indeed be a bit confusing. It is definitely not wrong to use $\delta u$ in this context. Would you agree with changing the notation here?

I am just not sure about the choice of $\varphi$, because we mix different alphabets for closely related symbols. Despite $v$ being visually close to $u$, should we rather opt for $v$ to still highlight the closeness of these objects?

docs/src/literate-tutorials/transient_heat_equation.jl
```diff
@@ -35,22 +35,29 @@
 # ```
 # The semidiscrete weak form is given by
 # ```math
-# \int_{\Omega}v \frac{\partial u}{\partial t} \ \mathrm{d}\Omega + \int_{\Omega} k \nabla v \cdot \nabla u \ \mathrm{d}\Omega = \int_{\Omega} f v \ \mathrm{d}\Omega,
+# \int_{\Omega}\delta u \frac{\partial u}{\partial t} \ \mathrm{d}\Omega + \int_{\Omega} k \nabla \delta u \cdot \nabla u \ \mathrm{d}\Omega = \int_{\Omega} f \delta u \ \mathrm{d}\Omega,
```
termi-official, line 38: Maybe I misunderstood you, but didn't you wanted to change these to be consistent with the heat example? Before changing these, let us wait for a quick feedback on https://github.com/Ferrite-FEM/Ferrite.jl/pull/1305/changes#r2934003577 .

Avina-cK: 😅right. I must have missed that part. Once we get feedback on which symbol would be better, I'll update the lot and fix the two broken reference links.

docs/src/topics/fe_intro.md
KnutAM: > δu can indeed be a bit confusing

Yes, but I also think that for most users it will make things easier if we are always consistent that test function for `x` is `δx`, even if mathematically the variation and test functions are different. Alternatively, we should have another consistent scheme (but that would require something like `φᵤ` or `uₜ` to have a good scheme for coupled problems). Using the subscripts quickly becomes confusing though, since we typically want to reserve these for indices (e.g. old value in time-dependent problems)

Specifically using `φ` is not good, since we used that for the shape functions in the intro to FEM (we use `N` in other places).

In the code, we are currently mixing using the notation for the test function and function itself with the test shape function and shape function. (I'm no fan of that but others are :D)

### My preferred notation 
For a field `x`,
* Function `x` (obviously) ($x \approx N_i a_i$)
* Test function `δx` (but highlight in the intro to FEM that we use this even though we argue from the "Bubnov-Galerkin perspective") ($\delta x \approx \delta N_i c_i$)
* Use `N` in code for `shape_value` for the function interpolation
* Use `δN` in code for `shape_value` for the test function interpolation
* In code, add letter `Nx` / `δNx` in coupled problems (and $N^x$ / $\delta N^x$ in math notation)
* For assembly, we can subscript these, e.g. `∇δNᵢ⋅D⋅∇Nⱼ` or `∇δNuᵢ⋅D⋅∇Npⱼ`


Of course, we can exchange `N -> φ`, but easier to type `N ` and consistent with the internal implementation in `FunctionValues`. 
`\varphi` and `\phi` also renders a bit differently depending, sometimes `\varphi` looks like `\phi`, so we shouldn't use both for different things.

KnutAM: And noting @fredrikekre's comment here: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1304#issuecomment-4055009281 - I guess that would be ok **if** we note that we (ab)use the notation `δx` for the test function even though it is not necessarily a variation.

And regarding the transient heat problem specifically, I think this can be derived from a potential so it can actually be considered a variation, but this would fail for other cases such as partially saturated porous media.

I guess we could introduce a different notation instead of `δ`, e.g. `𝒯` (`\scrT`), but I'm guessing this would be more confusing for most users even if we are consistent?

Avina-cK:
All (except for one) examples' code sections are now using the following notation:
- N in code for shape_value for the function interpolation
- δN in code for shape_value for the test function interpolation 
- For assembly, subscripted these, e.g. ∇δNᵢ⋅D⋅∇Nⱼ or ∇δNuᵢ⋅D⋅∇Npⱼ

docs/src/literate-tutorials/ns_vs_diffeq.jl
termi-official, line 1: This tutorial is not fully translated to the new syntax (see the pressure blocks).

docs/src/literate-tutorials/reactive_surface.jl
```diff
@@ -105,13 +105,13 @@ function assemble_element_mass!(Me::Matrix, cellvalues::CellValues)
         dΩ = getdetJdV(cellvalues, q_point)
         ## Loop over test shape functions
         for i in 1:n_basefuncs
-            δuᵢ = shape_value(cellvalues, q_point, i)
+            δNᵢ = shape_value(cellvalues, q_point, i)
```
termi-official, line 108: Any reason why we keep `δ` here specifically, but not in the other tutorial?

docs/src/literate-tutorials/transient_heat_equation.jl
```diff
@@ -35,27 +35,34 @@
...
 # ```
 # If we assemble the discrete operators, we get the following algebraic system:
 # ```math
 # \mathbf{M} \mathbf{u}_{n+1} + Δt \mathbf{K} \mathbf{u}_{n+1} = Δt \mathbf{f} + \mathbf{M} \mathbf{u}_{n}
 # ```
+# Here, `M` refers to the so called mass matrix, which always occurs in time related terms, i.e.
+# ```math
+# M_{ij} = \int_{\Omega} \delta u_i \, u_j \ \mathrm{d}\Omega,
+# ```
+# where $u_j$ and $\delta u_i$ are the shape functions of the trial and test functions, respectively.
```
termi-official: I think this sentence needs a bit further clarification. Also, what is the reason to deviate from the new rule here and use u/delta u instead of N?

docs/src/literate-tutorials/dg_heat_equation.jl
```diff
@@ -240,15 +240,15 @@ function assemble_interface!(Ki::Matrix, iv::InterfaceValues, μ::Float64)
...
         for i in 1:getnbasefunctions(iv)
             ## Multiply the jump by the negative normal to get the definition from the theory section.
-            δu_jump = shape_value_jump(iv, q_point, i) * (-normal)
-            ∇δu_avg = shape_gradient_average(iv, q_point, i)
+            δNᵢ_jump = shape_value_jump(iv, q_point, i) * (-normal)
+            ∇δNᵢ_avg = shape_gradient_average(iv, q_point, i)
             ## Loop over trial shape functions
             for j in 1:getnbasefunctions(iv)
                 ## Multiply the jump by the negative normal to get the definition from the theory section.
                 u_jump = shape_value_jump(iv, q_point, j) * (-normal)
```
termi-official: Is there a reason to keep that `u` here?

docs/src/literate-tutorials/heat_equation.jl
```diff
@@ -130,13 +130,13 @@ close!(ch)
...
 # `u`. We get all of these from `cellvalues`.
 #
-# !!! note "Notation"
-#     Comparing with the brief finite element introduction in [Introduction to FEM](@ref),
-#     the variables `δu`, `∇δu` and `∇u` are actually $\phi_i(\textbf{x}_q)$, $\nabla
-#     \phi_i(\textbf{x}_q)$ and $\nabla \phi_j(\textbf{x}_q)$, i.e. the evaluation of the
-#     trial and test functions in the quadrature point ``\textbf{x}_q``. However, to
-#     underline the strong parallel between the weak form and the implementation, this
-#     example uses the symbols appearing in the weak form.
+#!!! note "Notation"
```
termi-official: The indentation here is gone for some reason

termi-official (commented):
Some drive-by comments on stuff I saw found when going over the proposed changes. No need to take action yet.

docs/src/literate-tutorials/transient_heat_equation.jl
Avina-cK: Ah. I forgot to change the mathematical equation but I've done that now!

docs/src/literate-tutorials/ns_vs_diffeq.jl
Avina-cK: 😅 My mistake, I switched it to review way too soon. I'm a bit new to Git and contributing; thanks for your patience.

termi-official: Absolutely not problem! :) Take your time. And no worries, we all understand that the first steps to open source contributions have a bit of a barrier due all to the nitty-gritty details like familiarizing one-self with git, documentation conventions, ... .

Avina-cK:
All (except linearshell) tutorials have now been updated. I'm not sure I know how to update the linearshell one, so I left it as is. It is partially already using the 'N' notation and I think it is okay as is.

A small section has been added in fe_intro describing the new notation.

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -323,14 +323,14 @@ function doassemble!(cellvalues::CellValues, K::SparseMatrixCSC, dh::DofHandler,
         for q_point in 1:getnquadpoints(cellvalues)
             dΩ = getdetJdV(cellvalues, q_point)
             for i in 1:n_basefuncs
-                δεi = shape_symmetric_gradient(cellvalues, q_point, i)
+                δNεᵢ = shape_symmetric_gradient(cellvalues, q_point, i)
```
termi-official, line 326: This is now inconsistent with the other tutorials.

I am fine with either using δεi here and commenting on the symbol being the symmetric gradient (as common in literature), or using ∇ˢʸᵐN as in the elasticity example. Any comments on that @KnutAM ?

docs/src/literate-tutorials/dg_heat_equation.jl
```diff
@@ -240,15 +240,15 @@ function assemble_interface!(Ki::Matrix, iv::InterfaceValues, μ::Float64)
...
-            δu_jump = shape_value_jump(iv, q_point, i) * (-normal)
-            ∇δu_avg = shape_gradient_average(iv, q_point, i)
+            δNᵢ_jump = shape_value_jump(iv, q_point, i) * (-normal)
+            ∇δNᵢ_avg = shape_gradient_average(iv, q_point, i)
             ## Loop over trial shape functions
             for j in 1:getnbasefunctions(iv)
                 ## Multiply the jump by the negative normal to get the definition from the theory section.
-                u_jump = shape_value_jump(iv, q_point, j) * (-normal)
-                ∇u_avg = shape_gradient_average(iv, q_point, j)
+                Nⱼ = shape_value_jump(iv, q_point, j) * (-normal)
```
termi-official, line 248: This is not correct.

```suggestion
                Nⱼ_jump = shape_value_jump(iv, q_point, j) * (-normal)
```

docs/src/literate-tutorials/heat_equation.jl
```diff
@@ -132,11 +132,9 @@ close!(ch)
 #
 # !!! note "Notation"
 #     Comparing with the brief finite element introduction in [Introduction to FEM](@ref),
-#     the variables `δu`, `∇δu` and `∇u` are actually $\phi_i(\textbf{x}_q)$, $\nabla
+#     the variables `δNᵢ`, `∇δNᵢ` and `∇Nⱼ` are actually $\phi_i(\textbf{x}_q)$, $\nabla
 #     \phi_i(\textbf{x}_q)$ and $\nabla \phi_j(\textbf{x}_q)$, i.e. the evaluation of the
-#     trial and test functions in the quadrature point ``\textbf{x}_q``. However, to
-#     underline the strong parallel between the weak form and the implementation, this
-#     example uses the symbols appearing in the weak form.
+#     trial and test functions in the quadrature point ``\textbf{x}_q``.
```
termi-official, line 137: Reading the tutorial again there is now a bit of a disconnect between the weak form and the implementation. Can we refer to the theory intro (Introduction to FEM) in the theory section in the begin of this file with a brief summary of how the fem discretization with `N` relates to the weak form with `u`? Please take a look at the linear elasticity tutorial ("Finite element form") for some guidance on what is missing here.

docs/src/literate-tutorials/incompressible_elasticity.jl
```diff
@@ -137,14 +137,14 @@ function assemble_up!(Ke, fe, cell, cellvalues, facetvalues_u, grid, mp, t, dofr
         end
 
         for (iₚ, Iₚ) in pairs(dofrange_p)
-            δp = shape_value(cellvalues.p, q_point, iₚ)
+            δNpᵢ = shape_value(cellvalues.p, q_point, iₚ)
```
termi-official, line 140: The new notation here feels similarly confusing as the old one in this tutorial.

Is there any established notation for mixed problems in elasticity we could adopt @KnutAM ?

docs/src/literate-tutorials/ns_vs_diffeq.jl
```diff
@@ -265,10 +265,10 @@ function assemble_mass_matrix(cv::MultiFieldCellValues, M::SparseMatrixCSC, dh::
             ## Remember that we assemble a vector mass term, hence the dot product.
             ## There is only one time derivative on the left hand side, so only one mass block is non-zero.
             for i in 1:n_basefuncs_v
-                φᵢ = shape_value(cv.v, q_point, i)
+                Nvᵢ = shape_value(cv.v, q_point, i)
```
termi-official, line 268: Similar issue as in the incompressible elasticity example. Also, the derivation on top must be updated.
However, I think it is fine to deviate from using N in more advanced tutorials. We just need to clarify the used symbols in the tutorial header. From my understanding the primary issues are
1. we use e.g. u ambiguously for the solution, function value and shape function
2. it seems to be confusing what the test function is
right?

docs/src/literate-tutorials/plasticity.jl
```diff
@@ -211,11 +211,11 @@ function assemble_cell!(Ke, re, cell, cellvalues, material, ue, state, state_old
 
         dΩ = getdetJdV(cellvalues, q_point)
         for i in 1:n_basefuncs
-            δϵ = shape_symmetric_gradient(cellvalues, q_point, i)
-            re[i] += (δϵ ⊡ σ) * dΩ # add internal force to residual
+            δN_ϵ = shape_symmetric_gradient(cellvalues, q_point, i)
```
termi-official, line 214: This should be made consistent with the linear elasticity tutorial.

docs/src/literate-tutorials/porous_media.jl
```diff
@@ -122,11 +122,11 @@ function element_routine!(Ke, re, material::Elastic, cv::CellValues, a, args...)
         ϵ = function_symmetric_gradient(cv, q_point, a)
         σ = material.C ⊡ ϵ
         for i in 1:n_basefuncs
-            δ∇N = shape_symmetric_gradient(cv, q_point, i)
-            re[i] += (δ∇N ⊡ σ) * dΩ
+            ∇δNᵢ = shape_symmetric_gradient(cv, q_point, i)
```
termi-official, line 125: We might want to add `ˢʸᵐ` here and below to underline the symmetry.

docs/src/literate-tutorials/transient_heat_equation.jl
```diff
@@ -35,27 +35,34 @@
...
 # which yields:
 # ```math
-# \int_{\Omega} v\, u_{n+1}\ \mathrm{d}\Omega + \Delta t\int_{\Omega} k \nabla v \cdot \nabla u_{n+1} \ \mathrm{d}\Omega = \Delta t\int_{\Omega} f v \ \mathrm{d}\Omega + \int_{\Omega} v \, u_{n} \ \mathrm{d}\Omega.
+# \int_{\Omega} \delta u\, u_{n+1}\ \mathrm{d}\Omega + \Delta t\int_{\Omega} k \nabla \delta u \cdot \nabla u_{n+1} \ \mathrm{d}\Omega = \Delta t\int_{\Omega} f \delta u \ \mathrm{d}\Omega + \int_{\Omega} \delta u \, u_{n} \ \mathrm{d}\Omega.
 # ```
 # If we assemble the discrete operators, we get the following algebraic system:
 # ```math
 # \mathbf{M} \mathbf{u}_{n+1} + Δt \mathbf{K} \mathbf{u}_{n+1} = Δt \mathbf{f} + \mathbf{M} \mathbf{u}_{n}
 # ```
+# Here, `M` refers to the so called mass matrix, which always occurs in time related terms, i.e.
```
termi-official, line 49:
```suggestion
# Here, `M` refers to the so called mass matrix, which typically occurs in time related terms, i.e.
```

Just for ODE systems in mass-matrix form. General DAE systems can have more involved terms.

docs/src/literate-tutorials/transient_heat_equation.jl
```diff
@@ -107,76 +109,48 @@ close!(ch)
 update!(ch, 0.0);
 
 # ### Assembling the linear system
-# As in the heat equation example we define a `doassemble!` function that assembles the diffusion parts of the equation:
-function doassemble_K!(K::SparseMatrixCSC, f::Vector, cellvalues::CellValues, dh::DofHandler)
-
+# As in [Tutorial 1: Heat equation](heat_equation.md), we define a `doassemble!` function that assembles the
+# diffusion and diffusive parts of the equation:
```
termi-official, line 113:
```suggestion
# diffusion and mass parts of the equation:
```

docs/src/literate-tutorials/transient_heat_equation.jl
```diff
@@ -107,76 +109,48 @@ close!(ch)
...
+function doassemble!(K::SparseMatrixCSC, M::SparseMatrixCSC, f::Vector, cellvalues::CellValues, dh::DofHandler)
     n_basefuncs = getnbasefunctions(cellvalues)
     Ke = zeros(n_basefuncs, n_basefuncs)
+    Me = zeros(n_basefuncs, n_basefuncs)
     fe = zeros(n_basefuncs)
-
+    #Initiate assembler for matrices and vector
```
termi-official, line 119: This is how you make the comments visible in the generated markdown file. Also needs fixes below.

```suggestion
    ## Initiate assembler for matrices and vector
```

docs/src/literate-tutorials/transient_heat_equation.jl
termi-official: (I know this is a pre-existing tissue, but I just noticed it)

docs/src/topics/fe_intro.md
```diff
@@ -169,6 +169,11 @@ being the chosen approximation when changing from the integral to the finite sum
 For an example of the implementation to solve a heat problem with Ferrite check out [this
 thoroughly commented example](@ref tutorial-heat-equation).
 
+### Notation (tutorials)
+For a function, `u` such that $u \approx \sum N_i a_i$, the corresponding test function is represented as `δu` such that $\delta u \approx \sum \delta N_i c_i$. In code, `N` / `δN` denote the `shape_value` for the function interpolation.
```
termi-official: Maybe we want to clarify the notation further, as $\delta$ is sometimes used as the variation operator (similar as $\frac{d}{dt}$ for time derivatives).
```suggestion
For a function, `u` such that $u \approx \sum N_i a_i$, the corresponding test function is represented as `δu`, even if the problem does not have variational structure, such that $\delta u \approx \sum \delta N_i c_i$. In code, `N` / `δN` denote the `shape_value` for the function interpolation.
```

docs/src/topics/fe_intro.md
```diff
@@ -169,6 +169,11 @@ being the chosen approximation when changing from the integral to the finite sum
...
+In coupled problems, the function's name is appended to the shape functions, i.e., `Nu` / `δNu`.
```
termi-official, line 174: I think we should discuss this a bit (cc @KnutAM ).

docs/Project.toml
```diff
@@ -22,6 +22,7 @@ Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 OrdinaryDiffEqRosenbrock = "43230ef6-c299-4910-a778-202eb28ce4ce"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
+Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
```
termi-official, line 25: You should add revise to you global env.

docs/Manifest.toml
termi-official, line 1: This should be reverted.

termi-official (commented):
Thanks for working on improving the tutorials and making them more approachable @Avina-cK ! I appreciate it.

Some things to discuss here before we proceed. While I see that some tutorials benefit from the changes, I am not sure about the mixed problems here.

docs/src/literate-tutorials/computational_homogenization.jl
KnutAM: I think the easiest is using that from the elasticity, or alternatively simply writing `sym_∇δN` to avoid overusing the superscripts as these are annoying to type...

docs/src/literate-tutorials/incompressible_elasticity.jl
KnutAM: Don't think so, e.g. [Carter and Booker (1989)](https://www.sciencedirect.com/science/article/pii/0045794989901697) use different symbols for each field (N and X), in MoFEM's tutorial they use $\boldsymbol{B}$ for the B-matrix for displacements and $\boldsymbol{N}_p$ for the pressure shape functions (but formulate all without indices). 

I don't find it particularly confusing (but probably because I suggested it 😄 ), what is missing is that there is no theory where it should say that we have the FE form like this:
```math
\begin{align*}
K_{ij}^{uu} &= \int_\Omega \mathrm{grad}(\delta \boldsymbol{N}_i^u)^\mathrm{sym,dev} : \left[2G\ \mathrm{grad}(\boldsymbol{N}_j^u)^\mathrm{sym,dev}\right]\ \mathrm{d}\Omega \\
K_{ij}^{up} &=  -\int_\Omega \mathrm{div}(\delta \boldsymbol{N}_i^u)\ N_j^p\ \mathrm{d}\Omega \\
K_{ij}^{pu} &=  -\int_\Omega \delta N_i^p\ \mathrm{div}(\boldsymbol{N}_j^u)\ \mathrm{d}\Omega \\
K_{ij}^{pp} &=  \int_\Omega \frac{\delta N_i^p N_j^p}{K} \mathrm{d}\Omega
\end{align*}
```

docs/src/topics/fe_intro.md
termi-official: Discussion in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1305#discussion_r3200429903 .

docs/src/literate-tutorials/computational_homogenization.jl
termi-official: My view on the things is different here. You write the code here once, but it is read a lot of times, so even if we take a little bit more time to type, it is worth the time invest if the result has improved readability.

I do not have a strong opinion here tho.

KnutAM: > You write the code here once, but it is read a lot of times, so even if we take a little bit more time to type

I was not thinking of the one typing the tutorial code, but that often the pattern often becomes a template for users making their own implementations, and then this can be annoying.

Avina-cK: I think that it is okay to leave it as `δεi`, since it's common in literature, or type it out as `∇ˢʸᵐN`. At least on my Julia IDE setup, I can just type \^sym and press tab to automatically get `ˢʸᵐ`. 

As a student of Mathematics, I find superscripts to be easier to read and match with on-paper maths (compared to multi-symbol regular script variables). It's probably also why I was keen on the ϕ notation (which is used in most of my classes), but I agree with the phi / varphi rendering issue.

docs/src/literate-tutorials/incompressible_elasticity.jl
Avina-cK: The addition of explicitly mentioned theory does greatly help! Would using this superscript notation (`Nᵖⱼ`) be better than `Npⱼ`?
I can try expanding on a few tutorials' theory, but I won't touch the advanced ones (since I'm not yet knowledgeable about them).

docs/src/literate-tutorials/ns_vs_diffeq.jl
Avina-cK: Exactly. From all the discussion, here's what I gather we have currently chosen:

There's some PDE: 𝒟u = f

- `uₕ` is the discrete trial function (that'll hopefully turn into the solution). At a point, x, 
```math 
u (x) \approx u_h = \sum_i{(Nu)_i \hat{u}_i}
```
where `Nuᵢ` is the `shape_value`, `ûᵢ` is the weight value at `i` and `uₕ` is the `function_value` at `x`.

- The corresponding discrete test function is `δuₕ`. At a point, x, 
```math 
\delta u (x) \approx \delta u_h = \sum_i{(\delta Nu)_i \hat{(\delta u )}_i} 
``` 
where `δNuᵢ` is the `shape_value`, `δûᵢ` is the weight value at `i` and `δuₕ` is the `function_value` at `x`.

termi-official: I know it is probably a typo here, but to be sure we are on the same page, we have

$$u(x) ≈ u_h(x)  = \sum_i N^{(u)}_i(x)\hat{u}_i$$

where $\hat{u}_i$ is the weight of you (not necessarily nodal) ansatz function with index $i$.

PR #1306, opened 2026-03-17
avoid creating unnecessary views in `InterfaceValues`

### LINKED ISSUE #1301: `function_(gradient/value)_(jump/average)` allocates when `u` is a view
It seems like getting a view of a view that's not indexed by a range allocate. I ran into this when doing error estimation for AMR, it had a lot of allocations and I didn't suspect this could be the issue back then.
 
```julia
using Ferrite

using BenchmarkTools

grid = generate_grid(Hexahedron, (1,2,1));

ip = Lagrange{RefHexahedron, 1}();

fqr = FacetQuadratureRule{RefHexahedron}(2);

iv = Ferrite.InterfaceValues(fqr, ip);

reinit!(iv, first(InterfaceIterator(grid)));

u = rand(100);

ui = @view u[rand(1:100, 16)];

@benchmark function_gradient_jump($iv, $1, $ui)
```
```julia
BenchmarkTools.Trial: 10000 samples with 987 evaluations per sample.
 Range (min … max):  49.479 ns …  48.324 μs  ┊ GC (min … max):  0.00% … 99.78%
 Time  (median):     63.649 ns               ┊ GC (median):     0.00%
 Time  (mean ± σ):   95.356 ns ± 569.371 ns  ┊ GC (mean ± σ):  17.39% ±  4.06%

  ▅█▇▄▃▃▁▁            ▃▄▄▅▅▄▅▅▄▃▂▁▁▁▁▁                         ▂
  █████████▇▇▅▇██▇▇▆▆▆██████████████████▇█▇▇█▇▆▅▄▆▅▄▄▄▅▄▄▄▅▄▄▅ █
  49.5 ns       Histogram: log(frequency) by time       170 ns <

 Memory estimate: 256 bytes, allocs estimate: 4.
```
```julia
ui = @view u[40:56];

@benchmark function_gradient_jump($iv, $1, $ui)
```
```julia
BenchmarkTools.Trial: 10000 samples with 995 evaluations per sample.
 Range (min … max):  26.543 ns … 547.550 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     28.528 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   28.915 ns ±  10.339 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   ▁▁  ▂▄▁  ▄▆▂  ▆█▅   ▁                              ▁        ▁
  ▆██▂▃███▃▃███▅████▇▇▆██▆▅▆▆█▆▆▆▆▇▆▅▆▅▆▅▅▄▄▅▅▇█▅▅▃▅▅▇█▄▃▄▅▃▇█ █
  26.5 ns       Histogram: log(frequency) by time      33.9 ns <

```
```julia
ui = @view u[collect(40:56)];

@benchmark function_gradient_jump($iv, $1, $ui)
```
```julia

BenchmarkTools.Trial: 10000 samples with 986 evaluations per sample.
 Range (min … max):  49.354 ns …  48.101 μs  ┊ GC (min … max):  0.00% … 99.72%
 Time  (median):     91.593 ns               ┊ GC (median):     0.00%
 Time  (mean ± σ):   97.412 ns ± 570.019 ns  ┊ GC (mean ± σ):  16.81% ±  3.94%

  ▆█▇▅▃▃▂▁             ▂▅▄▅▆▅▅▆▆▅▃▂▂▂▂▁▁▁                      ▂
  █████████▇▇▆▇███▇▆▇▇▆█████████████████████▇█▇▇▆▇▆▆▆▅▄▆▅▃▅▃▂▄ █
  49.4 ns       Histogram: log(frequency) by time       163 ns <

 Memory estimate: 256 bytes, allocs estimate: 4.
```
This might not be an issue if one uses a vector instead of views, I just wanted to document this (and maybe disallow views to be passed to a function that creates views).

KristofferC:
fixes #1301

### DIFF
--- a/src/FEValues/common_values.jl
+++ b/src/FEValues/common_values.jl
@@ -180,7 +180,7 @@ function function_value(fe_v::AbstractValues, q_point::Int, u::AbstractVector, d
     @boundscheck checkbounds(u, dof_range)
     @boundscheck checkquadpoint(fe_v, q_point)
     val = function_value_init(fe_v, u)
-    @inbounds for (i, j) in pairs(dof_range)
+    @inbounds for (i, j) in enumerate(dof_range)
         val += shape_value(fe_v, q_point, i) * u[j]
     end
     return val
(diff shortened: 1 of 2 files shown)


src/FEValues/common_values.jl
```diff
@@ -220,20 +220,21 @@ function function_gradient(fe_v::AbstractValues, q_point::Int, u::AbstractVector
...
-    @inbounds for (i, j) in pairs(dof_range)
+    @inbounds for (i, j) in enumerate(dof_range)
         grad += shape_gradient(fe_v, q_point, i) * u[j]
     end
     return grad
 end
 
 # TODO: Deprecate this, nobody is using this in practice...
-function function_gradient(fe_v::AbstractValues, q_point::Int, u::AbstractVector{<:Vec})
+function function_gradient(fe_v::AbstractValues, q_point::Int, u::AbstractVector{<:Vec}, dof_range = eachindex(u))
```
KnutAM, line 230: What path made this active, @KristofferC?

Do we use this somewhere currently, or was it just in a test?

KristofferC: The calls where I changed the view to instead pass the dofs should dispatch to here, but if this argument was not present, it dispatched to the function above (the one not specialized for `<:Vec`). 

This was to fix a test failure.

PR #1310, opened 2026-03-21
Introduce isconstrained bitvector in ConstraintHandler

KnutAM:
This makes it easier and faster to check if a dof is constrained. 
Previous check was `haskey(ch.dof_mapping, dofnr)`, now just `ch.isconstrained[dofnr]`.

This speeds up `zero_out_rows!` and makes creating `free_dofs` easier (and a tiny bit faster). 

Benchmarks
```julia
# close!
4.963 ms (59 allocations: 25.78 MiB) # PR
5.075 ms (57 allocations: 25.41 MiB) # master

# apply!
293.966 ms (0 allocations: 0 bytes) # PR
1.333 s (0 allocations: 0 bytes)    # master
```

<details> <summary> benchmark code </summary>

```julia
using Ferrite, BenchmarkTools
grid = generate_grid(Hexahedron, (100,100,100));
dh = close!(add!(DofHandler(grid), :u, Lagrange{RefHexahedron, 1}()^3));
@btime close!(ch) setup = (ch = add!(ConstraintHandler(dh), Dirichlet(:u, getfacetset(grid,"left"), Returns(zero(Vec{3}))))) evals = 1;

ch = close!(add!(ConstraintHandler(dh), Dirichlet(:u, getfacetset(grid,"left"), Returns(zero(Vec{3})))));
K = allocate_matrix(dh); 
f = zeros(ndofs(dh));
@btime apply!($K, $f, $ch);
```

Extra runs for close!:

```julia
# PR 
  5.007 ms (59 allocations: 25.78 MiB)
  5.025 ms (59 allocations: 25.78 MiB)
  5.052 ms (59 allocations: 25.78 MiB)
  
# master
  5.100 ms (57 allocations: 25.41 MiB)
  5.121 ms (57 allocations: 25.41 MiB)
  5.111 ms (57 allocations: 25.41 MiB)
```

</details>

### DIFF
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -67,7 +67,7 @@ function Ferrite.zero_out_rows!(K::SparseMatrixCSR, ch::ConstraintHandler)
 end
 
 function Ferrite.zero_out_columns!(K::SparseMatrixCSR, ch::ConstraintHandler)
-    @boundscheck checkbounds(ch.isconstrained, Base.OneTo(size(K, 2)))
+    @boundscheck checkbounds(ch.isconstrained, axes(K, 2))
     colval = K.colval
     nzval = K.nzval
     return @inbounds for (i, col) in pairs(colval)
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -913,7 +913,7 @@ function zero_out_columns!(K::AbstractSparseMatrixCSC, ch::ConstraintHandler) # 
 end
 
 function zero_out_rows!(K::AbstractSparseMatrixCSC, ch::ConstraintHandler)
-    @boundscheck checkbounds(ch.isconstrained, Base.OneTo(size(K, 1)))
+    @boundscheck checkbounds(ch.isconstrained, axes(K, 1))
     rowval = K.rowval
     nzval = K.nzval
     @inbounds for (i, row) in pairs(rowval)


src/Dofs/ConstraintHandler.jl
termi-official, line 1: For symmetry purposes I would also add the optimization to `zero_out_columns!`. And what about the local condensation?

src/Dofs/ConstraintHandler.jl
```diff
@@ -259,21 +260,15 @@ isclosed(ch::ConstraintHandler) = ch.closed
...
-    resize!(out, n_entries - n_diff)
-    diff_ind = out_ind = 1
-    for i in 1:n_entries
-        if diff_ind ≤ n_diff && i == diff[diff_ind]
-            diff_ind += 1
-        else
-            out[out_ind] = i
-            out_ind += 1
-        end
+function set_freedofs!(free_dofs::Vector{Int}, isconstrained::BitVector, num_dofs::Int, num_prescribed::Int)
```
termi-official: The signature should be
```suggestion
function _set_free_dofs!(free_dofs::Vector{Int}, isconstrained::BitVector, num_dofs::Int, num_prescribed::Int)
```
to not drift from the used conventions.

src/Dofs/ConstraintHandler.jl
KnutAM: Good point! Added this for `CSR`. For CSC, there is no optimization to be added as we are looping over the `prescribed_dofs` directly.

Also added for the local condensation.

ext/FerriteSparseMatrixCSR.jl
```diff
@@ -67,10 +67,11 @@ function Ferrite.zero_out_rows!(K::SparseMatrixCSR, ch::ConstraintHandler)
 end
 
 function Ferrite.zero_out_columns!(K::SparseMatrixCSR, ch::ConstraintHandler)
+    @boundscheck checkbounds(ch.isconstrained, axes(K, 2))
     colval = K.colval
     nzval = K.nzval
-    return @inbounds for i in eachindex(colval, nzval)
-        if haskey(ch.dofmapping, colval[i])
+    return @inbounds for (i, col) in pairs(colval)
```
KristofferC, line 73: `pairs` here feels odd. Should it be `enumerate`?

src/Dofs/ConstraintHandler.jl
```diff
@@ -913,11 +913,12 @@ function zero_out_columns!(K::AbstractSparseMatrixCSC, ch::ConstraintHandler) #
 end
 
 function zero_out_rows!(K::AbstractSparseMatrixCSC, ch::ConstraintHandler)
+    @boundscheck checkbounds(ch.isconstrained, axes(K, 1))
     rowval = K.rowval
     nzval = K.nzval
-    @inbounds for i in eachindex(rowval, nzval)
-        if haskey(ch.dofmapping, rowval[i])
-            nzval[i] = 0
+    @inbounds for (i, row) in pairs(rowval)
```
KristofferC, line 919: Same w.r.t `pairs`

ext/FerriteSparseMatrixCSR.jl
KnutAM: My interpretation here is that `colval` and `nzval` have the same indices, hence it should be `pairs`. Of course, in practice for `SparseMatrixCSR` both `colval` and `nzval` are `Vector`s, so it doesn't matter.

src/Dofs/ConstraintHandler.jl
KnutAM: As above, but here the storage types in `AbstractSparseMatrixCSC` could be offset arrays, and then I assume that it should be `pairs` since we assume `rowval` and `nzval` have the same indexing?

ext/FerriteSparseMatrixCSR.jl
KristofferC: Ok, I think that is fair.

src/Dofs/ConstraintHandler.jl
KristofferC: 👍

PR #1291, opened 2026-02-27
Minimal GPU example which preserves the element routine

termi-official:
This PR has a fundamentally different design from the previous PRs. Here I try to preserve the assembly routine as-is and modify the assembly loop around them. This PR is really just an MWE to show the direction for discussion purposes.

The main vehicle for this PR is a SOA transformation of the CellCache and CellValues objects. I actually like it, but it might not be the best one performance-wise.

Closes #628 .

## TODOs

* [x] Subdomain support
* [x] Move device constraint handler to KA extension
* [x] Remove `Ferrite.reinit!(cc_i::ImmutableCellCache, cellid::Integer)`

## Follow-up PRs

* Make CSCAssembler/CSRAssembler GPU compatible . Right now the blocker is the runtime resize call, which I do not know how to get rid of.

PR #1319, opened 2026-04-12
Review on #1291

KnutAM:
Was easier to make changes directly, as I needed to understand better how everything works with the GPU and Adapt.jl interface. 

Will also add comments on #1291.


termi-official:
Thanks Knut. I think I have incorporated all changes from this PR into the GPU PR.

PR #1324, opened 2026-04-21
Format embedded Julia in docstrings and markdown files with Runic

fredrikekre:


KnutAM:
Nice! I think it makes perfect sense to use the same requirements in the docstrings.

I guess it would be to restrictive to apply within code blocks as well in the docstrings/documentation (.md) text? Would be nice for consistency though!

fredrikekre:
> I guess it would be to restrictive to apply within code blocks as well in the docstrings/documentation (.md) text? Would be nice for consistency though!

Done

KnutAM (approved):
LGTM

PR #1325, opened 2026-04-28
Make shape_symmetric_gradient work for PointValues

henrij22:
The function `shape_symmetric_gradient` did not work for me for PointValues. 

Now the following code works: 
```julia

using Ferrite

ne = 2
grid = generate_grid(Quadrilateral, (ne, ne), Vec(0.0, 0.0), Vec((1.0, 1.0)))
cc = CellCache(grid)
reinit!(cc, 1)

ip = Lagrange{RefQuadrilateral, 1}()^2

cv = CellValues(qr, ip)
pv = PointValues(cv)

reinit!(cv, cc)
reinit!(pv, cc.coords, Vec((0.5, 0.5)))
shape_symmetric_gradient(cv, 1, 1)
shape_symmetric_gradient(pv, 1, 1)
```

Also added test for `shape_symmetric_gradient` and  `symmetric_gradient`


KnutAM:
Nice, seems like this has been missed!

However, feels like we could/should define 
```julia
shape_symmetric_gradient(fev::AbstractValues, qp::Integer, i::Integer) = symmetric(shape_gradient(fev, qp, i))
```
in `common_values.jl`, instead of the current symmetric occurring that the `FunctionValues` level? Similar as is done for `shape_divergence` and `shape_curl`.

henrij22:
Good point, I've put the evlauation in common_values.jl and removed it from point- and cellvalues

I think there is a pre-commit check fail, that is not related to this PR?

src/FEValues/common_values.jl
KnutAM: Can add the default implementation here

KnutAM: And remove from

https://github.com/Ferrite-FEM/Ferrite.jl/blob/efa551d6e6c353e9c0447fdf1e86409ed8be700e/src/FEValues/FacetValues.jl#L96

https://github.com/Ferrite-FEM/Ferrite.jl/blob/efa551d6e6c353e9c0447fdf1e86409ed8be700e/src/FEValues/CellValues.jl#L103

?

KnutAM (commented):
> I think there is a pre-commit check fail, that is not related to this PR?

Should not be related, not sure what's going on here - seemed to pass on @fredrikekre's last pr which updated the runic checks?

fredrikekre:
I guess this https://github.com/Ferrite-FEM/Ferrite.jl/blob/efa551d6e6c353e9c0447fdf1e86409ed8be700e/.github/workflows/Check.yml#L17 should be `SKIP: runic,runic-md` now but it is weird that it fails.

KnutAM:
> but it is weird that it fails.

I deleted the cache: "pre-commit-3||6cd8084397da75e58f6d6387e234fe470a36bdbe740bae6dd0cce921aee96fb8", and then it work since it forces re-install of Runic. 

I think this errors surfaces because of the update in #1314, as `julia-actions/cache` v3 now don't delete the cache on failed jobs, so earlier rerunning the job would make this re-install. 

Could it be that there really should be an `instantiate` or smth similar added to the pre-commit workflow @fredrikekre ?

PR #1327, opened 2026-04-30
Untangle affine constraints in close!

kreutzheidephil:
The goal here is to allow the user to define nested affine constraints. An untangling is then performed in `close!`. To do this I proceed by finding all nested constraints to construct the system `A * a_c = C * a_f + g`. Here `A` and `C` are sparse matrices, as that is expected from nested constraints. I first construct `A` then update the `ch.dofcoefficients` to then construct `C` and `g`. Using the LU decomposition of `A`, I the solve the system and update the `ch.dofcoefficients` and `ch.affine_inhomogeneities` accordingly.

I have put most of the functions I added into a new file as `ConstraintHandler.jl` is already a bit long and would benefit from a little clean up.

I would love some feedback. I would particularly like to discuss the following points:
1. What error should be thrown when `A` is singular. I have just written something so that it is not empty but am not sure if it is useful.
2. Anything regarding the test.
3. Is it necessary to save the decomposition of `A` to make the `AffineConstraints` update-able or am I missing a different workaround? I ask this as I am interested in doing this in the near future.
4. Function names.


fredrikekre:
I haven't looked at your implementation, but IIRC deal.ii just does a fixed point iteration to substitute dofs equation by equation, i.e. starting with
```
u1 = u2 + u3
u2 = u4 + u5
u5 = 4
```
replace `u2` in first:
```
u1 = u4 + u5 + u3
u2 = u4 + u5
u5 = 4
```
replace `u5` in first (or maybe move on to the second equation and come back?):
```
u1 = u4 + 4 + u3
u2 = u4 + u5
u5 = 4
```
replace `u5` in second:
```
u1 = u4 + 4 + u3
u2 = u4 + 4
u5 = 4
```
etc until convergence.

termi-official:
Just note that the described iteration does not converge in the case of cyclic constraints.

kreutzheidephil:
I do it differently. Using your example I would move `u2` to the LHS and then solve the system:

$$
\begin{bmatrix}
1 & -1 \\
0 & 1
\end{bmatrix}
\begin{bmatrix}
u1 \\
u2
\end{bmatrix} =
\begin{bmatrix}
1 & 0 & 0 \\
0 & 1 & 1
\end{bmatrix}
\begin{bmatrix}
u3 \\
u4 \\
u5
\end{bmatrix}
$$

for `u1` and `u2`.

@KnutAM and I discussed this a bit and came to the conclusion that using this approach was probably easier and less error prone. As the fixed point approach is essentially a Gauss elimination (?) which might require a lot of possibly difficult-to-read code. Furthermore, if a lot of nested constraints appear then a sparse linear solve is probably faster?

KnutAM:
> Just note that the described iteration does not converge in the case of cyclic constraints.

This is how @kreutzheidephil and I reasoned too, and kindof nice that you just get a singular exception automatically instead of handling a lot of special cases that is needed before eventually throwing an error. The only question I have is if this strategy is sufficiently fast or if a special algorithm is required to be more efficient.

Edit: @kreutzheidephil was faster :D

termi-official:
> As the fixed point approach is essentially a Gauss elimination (?)

Nope.

> The only question I have is if this strategy is sufficiently fast or if a special algorithm is required to be more efficient.

We might want to make this modular so users can choose. I think for generic affine constraints this 
(solve) is the best we can do. But usually we have a lot extra structure to efficiently resolve the constraints.

kreutzheidephil:
> We might want to make this modular so users can choose.

@termi-official could you elaborate a bit on this?

termi-official:
Fredriks algorithm (see e.g. https://arxiv.org/pdf/1905.04033) is typically more efficient than forming the system and doing a linear solve if your affine constraints are due to some AMR alg, where we can guarantee non-cyclic dependencies.

kreutzheidephil:
So maybe the way forward would be to make a function `untangle_affine_constraints!` available to users. It can then be called before `close!` in case they have nested constraints and don't mind waiting for a linear solve to untangle them. Then we can leave `close!` free for a more efficient algorithm in the future?

KnutAM:
> So maybe the way forward would be to make a function untangle_affine_constraints available to users.

Great idea! Can probably have a syntax like `untangle_constraints!(ch; algorithm::Bool)`? 

In order to be able to make this backwards compatible with periodic bcs, we probably want to call this automatically if a periodic bc has been added. If it can be done at essentially zero cost (for non-entangled cases), we could probably always call this, and then users can call `untangle_constraints!` before `close!` if they want an optimized version.

fredrikekre:
I think untangling should happen automatically but we can surface the algo choice as a keword argument to close when/if we have more algos implemented. Don't worry about that in this PR.

kreutzheidephil:
Just some quick comments on the last commit.

1. I have renamed `isnested` to `istangled` to consistently use the tangled terminology. Maybe the term nested is better. Discussing with @KnutAM yesterday he made the remark that tangled could be considered to include cyclic constraints, while nested is without cyclic.
2. I have not bench marked anything. However, I do the same test to see if constraints are tangled that was also previously implemented, meaning in the case that no constraints are tangled it will not be any slower than before.

test/test_constraints.jl
```diff
@@ -1397,12 +1397,48 @@ end # testset
...
-    @testset "error paths" begin
+    # Test when affine constraints weren't allowed to be nested
+    #
+    # @testset "error paths" begin
+    #     ch = ConstraintHandler(dh)
+    #     add!(ch, AffineConstraint(1, [2 => 1.0], 0.0))
+    #     add!(ch, AffineConstraint(2, [3 => 1.0], 0.0))
+
+    #     @test_throws ErrorException("nested affine constraints currently not supported") close!(ch)
+    # end # subtestset
```
termi-official:
```suggestion
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -310,19 +311,10 @@ function close!(ch::ConstraintHandler)
...
-            i == 0 && continue
-            icoeffs = ch.dofcoefficients[i]
-            if !(icoeffs === nothing || isempty(icoeffs))
-                error("nested affine constraints currently not supported")
-            end
-        end
+
+    if istangled(ch) # untangle affine constraints
+        @debug @warn "untangling nested and cyclic affine constraints"
+        untangle_constraints!(ch)
```
termi-official, line 341: Maybe we can be more explicit with the name here.

```suggestion
    if istangled(ch)
        @debug @warn "untangling nested and cyclic affine constraints"
        untangle_affine_constraints!(ch)
```

Also, why warning?

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,231 @@
...
+
+    # TODO: maybe add possibility to warn user (if user chooses through kwarg) if `A` is ill conditioned
+
+    luA = try
+        LinearAlgebra.lu(A; check = true)
+    catch e
+        if e isa LinearAlgebra.SingularException
+            throw(
+                "the affine constraints are nested and untangling them results in " *
+                    "ill defined constraints. A possibility to avoid this is to guarantee that " *
```
termi-official:
```suggestion
                    "cyclic constraints. A possibility to avoid this is to guarantee that " *
```
Pretty sure this is what we want to say here. Probably nothing we or the user can do here. Note that cyclic essentially reduces to linear dependend. Or do you have more cases in mind?

kreutzheidephil: I just thought if we have for example `u_1 = ...` and `u_2 = 1e-16*u_1 ` we will get a ill conditioned matrix during untangling, possibly so ill conditioned that the LU decomp fails. Seeing as this could happen without cyclic constraints I wanted to leave it general.

src/Dofs/ConstraintHandler.jl
kreutzheidephil: > Maybe we can be more explicit with the name here.

I agree 
> Also, why warning?

I just thought it might be helpful when debugging if the user has complex constraints and is getting unexpected solutions but maybe unnecessary.

src/Dofs/ConstraintHandler.jl
```diff
@@ -357,9 +349,12 @@ constraint if true.
 """
 function add_prescribed_dof!(ch::ConstraintHandler, constrained_dof::Int, inhomogeneity, dofcoefficients = nothing)
     @assert(!isclosed(ch))
+    @assert(constrained_dof ≤ ndofs(ch.dh))
     i = get(ch.dofmapping, constrained_dof, 0)
+    # TODO: It would be cool not to override and just add regardless
+    # and then untangle in close! This however not done yet.
```
lijas, line 379: Just a comment regarding this. In the future, we can consider adding constraints as a "flat" vector instead, where dont specify which node should be master/slave (e.g u1 + u2 - 3 = 0 instead of u1 = -u2 + 3), and then let the untangling-algorithm pick the best set of constrained nodes. I asked Gemini, and there seems to exist such an algorithm called Rank-Revealing QR Decomposition with Column Pivoting. I played around with it yesterday, and it seems to work really nicely. But perhaps out of scope for this PR :)

kreutzheidephil: This sounds really cool :) 
Possibly a nice way to generalize the constraints more too

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,231 @@
...
+                return true
+            end
+        end
+    end
+    return false
+end
+
+function _assign_new_index!(d::Dict{Int, Int}, k::Int)
+    return if !haskey(d, k)
+        d[k] = maximum(values(d); init = 0) + 1
```
KristofferC: This makes the algorithm O(n^2) since it computes a `maximum` inside the loop. We are only adding (never deleting) things so something like

```julia
function _assign_new_index!(d::Dict{Int, Int}, key::Int)
    return get!(d, key, length(d) + 1)
end
```

should be fine

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,231 @@
...
+    C, g, affine_fdof_ordering = _create_rhs_affine_constraint_matrices(ch, affine_equation_ordering)
+
+    # TODO: maybe add possibility to warn user (if user chooses through kwarg) if `A` is ill conditioned
+
+    luA = try
+        LinearAlgebra.lu(A; check = true)
+    catch e
+        if e isa LinearAlgebra.SingularException
+            throw(
+                "the affine constraints are tangled and untangling them results in " *
```
KristofferC: this should throw some type of exception, `ArgumentError`?

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,231 @@
...
+                    "ill defined constraints. A possibility to avoid this is to guarantee that " *
+                    "the constraints are not tangled before calling close!"
+            )
+        else
+            rethrow(e)
+        end
+    end
+
+    C .= LinearAlgebra.ldiv(luA, C)
```
KristofferC: `LinearAlgebra.ldiv(luA, C)` gives a dense matrix, so if you have many constraints you are going to blow up memory here. Also, instead of `C .= ` I would use `copy!`

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,231 @@
...
+Untangle the constraints in `ch`. This function will error if the constraints are not tangled.
+"""
+function untangle_constraints!(ch::ConstraintHandler)
+    # Incase user calls this? Maybe not relevant if not exported?
+    @assert istangled(ch) "the constraints are not tangled"
+    A, affine_equation_ordering, _, dofcoeffs_to_remove = _create_lhs_affine_constraint_matrix(ch)
+
+    # update ch.dofcoefficients so that they can be used to construct `C`
+    for (k, v) in dofcoeffs_to_remove
+        deleteat!(ch.dofcoefficients[k], v)
```
KristofferC: This mutates `ch` but the operation can then fail later giving a corrupted and mutated `ch`. It is probably better to only mutate `ch` ones you have the final values you are going to update it with.

KristofferC:
This probably does not have to be done in this PR, but doing a full solve for all constraints is a bit overkill. Having had to work on MTK.jl etc what I think they do is to create groups of "strongly connected components" by walking constrained dofs and see what constrained dofs they couple to. So in the simplest case:

```
u₁ = 2u₂ + u₅
u₂ = 3u₃ + 4
u₃ = u₆ - 1
```

you have

```
u₁ ──→ u₂ ──→ u₃
```

(u₆ and u₅ are leaves). In this case you don't have to factorize at all you can just solve for the constrained variables and "plug in" the result everywhere.

For e.g.

```
u₁ = 0.5u₂ + u₅
u₂ = 0.25u₁ + u₆ + 1
```

you have a cycle `u₁ ──→ u₂  ──→ u₁` and have to solve a 2x2 system.

And then the mixed case

```
u₁ = 2u₂ + u₇
u₂ = 0.5u₃ + u₅
u₃ = 0.25u₂ + u₆
```

you get 

```
u₁ ──→ u₂ ⇄ u₃
```

so here you can solve `u₂` and `u₃` and then substitute into `u₁`.

So you get one factorization for every SCC instead of a big global one.

----------

Also, one thing I have had need for is when you have cohesive elements and multiple polyhedra (like grains) that touch in a Y shape, and you want to tie all the u-dofs there together, basically `u1 = u2 = u3`. I wonder if we could have something like

```
add!(ch, EqualityConstraint([1, 2, 3, 4]))
```

that does this automatically for you.

src/Dofs/untangling_affine_constraints.jl
kreutzheidephil: I think that is probably the most appropriate :)

src/Dofs/untangling_affine_constraints.jl
kreutzheidephil: That is true, I guess one way to get around that would be to construct the result column wise. I would assume it would take longer but at least it wouldn't blow up memory.

src/Dofs/untangling_affine_constraints.jl
kreutzheidephil: Oh yes I missed this bug, thank you

kreutzheidephil:
There have been a couple of suggestions on how to make the untangling more efficient. One thing that I think is possibly overlooked is the extension of the `AffineConstraints` to allow time dependent inhomogeneities. Assume we have the following tangled constraints with time dependent inhomogeneities:
```
u_1 = 2u_2 + f_1(t)
u_2 = -u_1 + u_4 + f_2(t)
u_3 = 5u_2 + 3
```
If we write the untangled solution as in the current approach (this PR) we get

`u_c = A^-1 * C * u_f + A^-1 * g(t)`.

Concretely, for the system above, this is
```
u_1 = -2u_4 - f_1(t) - f_2(t)
u_2 = -u_4 - f_1(t) - f_2(t)
u_3 = -5u_4 - 5f_1(t) - 5f_2(t) + 3
```
so the inhomogeneities "mix". If we want to update the `ch.inhomogeneities` in `update!` we need to save `A^-1` as its LU (or similar) or the information it contains, as it will be required in each update. I think a sparse LU is an efficient way to save and then apply this information in `update!`. The approach in this PR builds it for free essentially, other approaches it is potentially something that would have to be built alongside the untangling. I guess a possible way to circumvent all this is to implement the approach suggested by @lijas https://github.com/Ferrite-FEM/Ferrite.jl/pull/1327#discussion_r3278862479, so that we avoid the tangling from the beginning, as a forward and backward substitution in every `update!` call is probably not desirable.

It would be nice to discuss this a little.

src/Dofs/ConstraintHandler.jl
```diff
@@ -334,20 +335,10 @@ function close!(ch::ConstraintHandler)
...
-            i == 0 && continue
-            icoeffs = ch.dofcoefficients[i]
-            if !(icoeffs === nothing || isempty(icoeffs))
-                @debug println("Nested affine constraint detected: master dof $d of $coeffs is itself affinely constrained")
-                error("nested affine constraints currently not supported")
-            end
-        end
+
+    if istangled(ch) # untangle affine constraints
+        @debug @warn "untangling tangled and cyclic affine constraints"
```
KnutAM, line 340:
```suggestion
        @debug @info "untangling tangled and cyclic affine constraints"
```

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,264 @@
+"""
+    untangle_constraints!(ch::ConstraintHandler)
+
+Untangle the constraints in `ch`. This function will error if the constraints are not tangled.
+"""
+function untangle_constraints!(ch::ConstraintHandler)
+    # Incase user calls this? Maybe not relevant if not exported?
+    @assert istangled(ch) "the constraints are not tangled"
```
KnutAM, line 8:
```suggestion
function _untangle_constraints!(ch::ConstraintHandler)
```
Document as internal (and add `_` prefix), so we don't need to do the check twice. If we want a public version, we can have `untangle_constraints!` that first checks and then untangles, but the internal one can assume tangled.

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,264 @@
...
+            push!(coeffs, (dof => C.nzval[nz_i]))
+        end
+    end
+    return dc
+end
+
+"""
+    istangled(ch::ConstraintHandler)
+
+Check if the constraint handler has any tangled dofs.
```
KnutAM, line 219: Define what is implied by tangled in the docstring?

src/Dofs/ConstraintHandler.jl
```diff
@@ -382,9 +373,13 @@ constraint if true.
 """
 function add_prescribed_dof!(ch::ConstraintHandler, constrained_dof::Integer, inhomogeneity, dofcoefficients = nothing)
     @assert(!isclosed(ch))
+    @assert(constrained_dof ≤ ndofs(ch.dh))
     i = get(ch.dofmapping, constrained_dof, 0)
+    # TODO: It would be cool not to override and just add regardless
+    # and then untangle in close! This however not done yet.
+    # If you're interested in this see https://github.com/Ferrite-FEM/Ferrite.jl/pull/1327#discussion_r3278862479
```
KnutAM, line 380:
```suggestion
```

src/Dofs/untangling_affine_constraints.jl
```diff
@@ -0,0 +1,264 @@
...
+
+    new_dofcoefficients = deepcopy(ch.dofcoefficients)
+    # update new_dofcoefficients so that they can be used to construct `C`
+    for (k, v) in dofcoeffs_to_remove
+        deleteat!(new_dofcoefficients[k], v)
+    end
+    C, g, affine_fdof_ordering = _create_rhs_affine_constraint_matrices(ch, new_dofcoefficients, affine_equation_ordering)
+
+    # TODO: maybe add possibility to warn user (if user chooses through kwarg) if `A` is ill conditioned
+
```
KnutAM, line 19:
```suggestion
```

PR #1333, opened 2026-05-12
Introduce AbstractCellCache

lijas:
With these changes, FerriteIGA.jl can use L2Projector without any additional code. It is able to do this by extending the CellIterator/CellCache "interface". 

It would be nice to have a documented interface for the CellIterator in the future so costum grids can use it.

### DIFF
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -242,7 +242,7 @@ end
     `CellIterator` is stateful and should not be used for things other than `for`-looping
     (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
 """
-struct CellIterator{CC <: CellCache, IC <: IntegerCollection}
+struct CellIterator{CC, IC <: IntegerCollection}
     cc::CC
     set::IC
 end
--- a/src/iterators.jl
+++ b/src/iterators.jl
@@ -14,6 +14,8 @@ UpdateFlags(; nodes::Bool = true, coords::Bool = true, dofs::Bool = true) =
 ## CellCache ##
 ###############
 
+abstract type AbstractCellCache end
+
 """
     CellCache(grid::Grid)
     CellCache(dh::AbstractDofHandler)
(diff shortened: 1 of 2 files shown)


src/iterators.jl
```diff
@@ -242,7 +242,7 @@ end
     `CellIterator` is stateful and should not be used for things other than `for`-looping
     (e.g. broadcasting over, or collecting the iterator may yield unexpected results).
 """
-struct CellIterator{CC <: CellCache, IC <: IntegerCollection}
+struct CellIterator{CC, IC <: IntegerCollection}
```
termi-official: This change does not work.

lijas: 😞 
```
Got exception outside of a @test
  MethodError: CellIterator(::DofHandler{2, Grid{2, Quadrilateral, Float64}}, ::OrderedSet{Int64}) is ambiguous.
  
  Candidates:
    CellIterator(gridordh::Union{DofHandler, Grid}, set::Union{Nothing, AbstractSet{<:Integer}, AbstractVector{<:Integer}})
      @ Ferrite ~/work/Ferrite.jl/Ferrite.jl/src/iterators.jl:250
    CellIterator(cc::CC, set::IC) where {CC, IC<:(Union{AbstractSet{var"#s5"}, AbstractVector{var"#s5"}} where var"#s5"<:Integer)}
      @ Ferrite ~/work/Ferrite.jl/Ferrite.jl/src/iterators.jl:246
```

KnutAM: Looks like a Julia bug to me? The union should be more specific than Any? 

Edit: Aha, no bug - missed the nothing in one of the unions - can just add a dispatch for nothing specifically somewhere to use all cells?

termi-official: Or we introduce some AbstractCellCache type. This should also help here.

KnutAM: How should this interface with ImmutableCellCache in #1291 ? Both cellcaches but different API? Will we need 
* AbstractCellCache
  * AbstractMutableCellCache
    * CellCache
  * AbstractImmutableCellCache
    * ImmutableCellCache

?

termi-official: We never defined APIs for most of our types. Assembler, Dofhandler, FEValues... all subtypes there have different APIs. Which is okay given that they serve different purposes.

Issue #1331, opened 2026-05-05
Atomic assembler


termi-official:
We should add a way to make the assemblers `assemble` call atomic if users request so. Some geometries come with topologies which are inherently bad to color.

PR #1337, opened 2026-05-18
Add extension point for custom assembly ops

termi-official:
This allows us to simply add stuff like #1331 .

The alternative design here is to add the operation as a field to the assembler struct instead of passing it into `assemble!`.

For the first tutorial I get on the branch

```julia
julia> include("src/literate-tutorials/heat_equation.jl")
  69.279 μs (20 allocations: 4.49 KiB)
Test Passed
```

and on master

```julia
julia> include("src/literate-tutorials/heat_equation.jl")
  69.420 μs (20 allocations: 4.49 KiB)
Test Passed
```

### DIFF
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -14,8 +14,9 @@ end
         K::SparseMatrixCSR, Ke::AbstractMatrix,
         rowdofs::AbstractVector, sortedrowdofs::AbstractVector, rowpermutation::AbstractVector,
         coldofs::AbstractVector, sortedcoldofs::AbstractVector, colpermutation::AbstractVector,
-        sym::Bool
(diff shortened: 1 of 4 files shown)


KristofferC:
This may be useful, but for #1331 I think I would prefer a simple `atomic=true/false` kwarg.

termi-official:
> This may be useful, but for #1331 I think I would prefer a simple `atomic=true/false` kwarg.

I see the point. But there might also be other options users might want to apply, if e.g. the hardware does not come with support for atomics. We can try to detect this and switch between atomics and e.g. spin-locks, but not sure if it is worth the effort.

I am not too happy with the current design either, so if you have an idea for how to improve it, I am open for suggestions.

termi-official:
Superseeded by #1417

Issue #704, opened 2023-05-10
Merge entity orientation information of grid and constraint handler


termi-official:
We should merge the constraint handler entity orientation api (https://github.com/Ferrite-FEM/Ferrite.jl/blob/90a5889cdecdea44389ba8a30b096d3a691e34a3/src/Dofs/ConstraintHandler.jl#L882-L887) with the grid orientation api (https://github.com/Ferrite-FEM/Ferrite.jl/blob/90a5889cdecdea44389ba8a30b096d3a691e34a3/src/Grid/grid.jl#L1385-L1432). Both do essentially the same.

termi-official:
There is more to consider soon:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/743#discussion_r1363705812
https://github.com/Ferrite-FEM/Ferrite.jl/pull/743#discussion_r1363712185

PR #1343, opened 2026-06-10
🤖 :  Support multiple nodal dofs per face in 3D dof distribution

### LINKED ISSUE #1195: Multiple facedofs per face

KristofferC:
I wanted to try the new Fable model on something real and I came to think about this TODO. I am not sure it is wise (but maybe it is) to merge this but at least it can be used as inspiration.

Fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/1195

Bot commit message:

----------

🤖 

 Resolve the long-standing TODO in the dof distribution about more than
  one interior dof per face in 3D, which requires taking the relative
  rotation between local faces into account, not just the direction as
  for edges.

  `sortface` now returns an `OrientationInfo` (rotation shift + flip of
  the local face relative to the canonical face spanned by the sorted
  vertex tuple), replacing the empty `SurfaceOrientationInfo` placeholder.
  `permute_and_push!` uses it to map the interior face dofs, which are
  assumed to be placed on a regular lattice (triangular for triangle
  faces, tensor grid for quadrilateral faces), from the canonical storage
  order to the local face orientation, following the methodology of
  Scroggs et al. The lattice enumeration convention matches the existing
  interior node ordering of `Lagrange{RefTriangle, 4/5}` and
  `Lagrange{RefQuadrilateral, 3}`, and is documented in the
  `permute_and_push!` docstring. As before, no adjustment is done for
  cells with reference dimension 2, where the "face" is the interior of
  the cell itself (and not necessarily a lattice, e.g.
  `RaviartThomas{RefTriangle, 2}`).

  Add `Lagrange{RefTetrahedron, 3}` and `Lagrange{RefTetrahedron, 4}`,
  implemented via barycentric lattice multi-indices, where order 4 is the
  first interpolation with multiple interior dofs per face.

  Tests:
  - Brute-force geometric verification of the local-to-canonical index
    mapping for all 6 triangle / 8 quadrilateral face orientations.
  - Two-tet shared-face test checking that every global dof maps to the
    same physical location from both cells, for all 24 x 24 vertex
    orderings (also for a vectorized interpolation).
  - New interpolations added to the interpolation property tests, the
    facet continuity tests (including checking that interior face dofs
    follow the documented lattice enumeration), CellValues tests, and the
    Poisson convergence tests.

----------


src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,133 @@ so the unique representation is always a tuple length 3.
...
+# References
+ - [Scroggs2022](@cite) Scroggs et al. ACM Trans. Math. Softw. 48 (2022).
 """
-@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, ::SurfaceOrientationInfo, adjust_during_distribution::Bool, rdim::Int)
-    if rdim == 3 && adjust_during_distribution && length(dofs) > 1
-        error("Dof distribution for interpolations with multiple dofs per face not implemented yet.")
-    end
+@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, orientation::OrientationInfo, adjust_during_distribution::Bool, nfacenodes::Int, rdim::Int)
+    # TODO Investigate if we can somehow pass the interpolation into this function in a
+    # typestable way.
```
termi-official: For context: The idea here was to have a function `permute_and_push!(celldofs, dofs, orientation, must_adjust, interpolation)`. This way users can dispatch their high order own interpolations.

Also, this function can silently give the wrong result for high order non-Lagrange interpolations.

src/Grid/grid.jl
```diff
@@ -700,38 +700,8 @@ struct PathOrientationInfo
...
-"""
-struct SurfaceOrientationInfo
-    #flipped::Bool
-    #shift_index::Int
-end
-
-
 @doc raw"""
-    InterfaceOrientationInfo
+    OrientationInfo
```
termi-official: This is related to https://github.com/Ferrite-FEM/Ferrite.jl/issues/704 , but I think it goes in the wrong direction. The distinction here has been introduced to not extend the orientation info struct when implementing >3D elements, so we can add something like `VolumeOrientationInfo`. Just extending the struct can also give really hard to track down bugs when trying to move a 3D implementation to a 4D implementation.

I think the right way forward is to consequently use PathOrientationInfo (above) in 1D and SurfaceOrientationInfo (which is essentially this struct here) in 2D.

Does that make sense?

src/interpolations.jl
```diff
@@ -929,6 +931,87 @@ function reference_shape_value(ip::Lagrange{RefTetrahedron, 2}, ξ::Vec{3}, i::I
...
 end
 
+######################################
+# Lagrange RefTetrahedron order 3, 4 #
+######################################
+
+const Lagrange3DTet34 = Union{
+    Lagrange{RefTetrahedron, 3},
+    Lagrange{RefTetrahedron, 4},
+}
```
termi-official, line 959: xref https://github.com/Ferrite-FEM/Ferrite.jl/pull/1255#issuecomment-3722839070

termi-official (commented):
Left some comments here to consider.

The lattice operations look legitimate but we need to expand the test suite to ensure that they really work fine.

src/interpolations.jl
KristofferC: Feels a bit like early bike shedding, when the interpolations are actually added to Ferrite, the exact method as to how they should be defined can be discussed.

src/Grid/grid.jl
KristofferC: Kind of, but it becomes kind of inconsistent with `InterfaceOrientationInfo` which is a single struct for different dimensions (you can see the two `0` const arguments in its constructor now). Anyway, I pushed a commit that changed it to the suggestion here.

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
+# References
+ - [Scroggs2022](@cite) Scroggs et al. ACM Trans. Math. Softw. 48 (2022).
 """
-@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, ::SurfaceOrientationInfo, adjust_during_distribution::Bool, rdim::Int)
-    if rdim == 3 && adjust_during_distribution && length(dofs) > 1
-        error("Dof distribution for interpolations with multiple dofs per face not implemented yet.")
-    end
+@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, orientation::SurfaceOrientationInfo, adjust_during_distribution::Bool, interior_facedofs_on_lattice::Bool, nfacenodes::Int, rdim::Int)
+    # TODO Investigate if we can somehow pass the interpolation into this function in a
+    # typestable way (instead of relying on the interior_facedofs_on_lattice trait).
```
termi-official, line 792: To ensure this can be customized in the future, can we explore if there is a sane way to pass the interpolation or some other object down here to control the dispatch? I am primarily thinking about downstream libraries providing new interpolations which do not follow tensor product orderings on the faces.

KristofferC: I'd rather not prematurely generalize and abstract things until the concrete need is there. I'm sure these type of refactorings can be done at that point.

KnutAM:
Just regarding naming etc. if something becomes public here, we should remove the path/surface in favour of edge/face to be consistent in the naming (path/surface were introduced before facet was introduced)

src/Dofs/DofHandler.jl
```diff
@@ -578,13 +576,13 @@ for the object (vertex, face) then simply return those, otherwise create new dof
...
 
-function add_face_dofs(cell_dofs::Vector{Int}, cell::AbstractCell, facedict::Dict, nfacedofs::Vector{Int}, nextdof::Int, adjust_during_distribution::Bool, n_copies::Int)
+function add_face_dofs(cell_dofs::Vector{Int}, cell::AbstractCell, facedict::Dict, nfacedofs::Vector{Int}, nextdof::Int, adjust_during_distribution::Bool, interior_facedofs_on_lattice::Bool, n_copies::Int)
     for (fi, face) in pairs(faces(cell))
         nfacedofs[fi] > 0 || continue # skip if no dof on this vertex
         sface, orientation = sortface(face)
         @debug println("\t\tface #$sface, $orientation")
         nextdof, dofs = get_or_create_dofs!(nextdof, nfacedofs[fi], n_copies, facedict, sface)
-        permute_and_push!(cell_dofs, dofs, orientation, adjust_during_distribution, getrefdim(cell)) # TODO: passing rdim of cell is temporary, simply to check if facedofs are internal to cell
+        permute_and_push!(cell_dofs, dofs, orientation, adjust_during_distribution, interior_facedofs_on_lattice, length(face), getrefdim(cell))
```
termi-official:
```suggestion
        nfacevertices = length(face) # `faces` returns the vertices of the geometric entity only
        permute_and_push!(cell_dofs, dofs, orientation, adjust_during_distribution, interior_facedofs_on_lattice, nfacevertices, getrefdim(cell))
```

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
-
-    !!!TODO Investigate if we can somehow pass the interpolation into this function in a typestable way.
+# References
+ - [Scroggs2022](@cite) Scroggs et al. ACM Trans. Math. Softw. 48 (2022).
 """
-@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, ::SurfaceOrientationInfo, adjust_during_distribution::Bool, rdim::Int)
-    if rdim == 3 && adjust_during_distribution && length(dofs) > 1
-        error("Dof distribution for interpolations with multiple dofs per face not implemented yet.")
-    end
+@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, orientation::SurfaceOrientationInfo, adjust_during_distribution::Bool, interior_facedofs_on_lattice::Bool, nfacenodes::Int, rdim::Int)
```
termi-official:
```suggestion
@inline function permute_and_push!(cell_dofs::Vector{Int}, dofs::StepRange{Int, Int}, orientation::SurfaceOrientationInfo, adjust_during_distribution::Bool, interior_facedofs_on_lattice::Bool, nfacevertices::Int, rdim::Int)
```

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
+    # typestable way (instead of relying on the interior_facedofs_on_lattice trait).
     n_copies = step(dofs)
     @assert n_copies > 0
-    for dof in dofs
-        for i in 1:n_copies
-            push!(cell_dofs, dof + (i - 1))
+    ndofs = length(dofs)
+    if rdim == 3 && adjust_during_distribution && ndofs > 1
+        interior_facedofs_on_lattice || error("Dof distribution for an interpolation with multiple dofs on a face shared between 3D cells requires the interior face dofs to be placed on a regular lattice; this interpolation has not opted in, see `Ferrite.interior_facedofs_on_lattice` and `Ferrite.permute_and_push!`.")
+        if nfacenodes == 3 # triangular face
```
termi-official:
```suggestion
        if nfacevertices == 3 # triangular face
```

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
+            q = _triangle_lattice_order(ndofs)
+            for t2 in 0:q, t1 in 0:(q - t2)
+                dof = dofs[_canonical_facedof_index_triangle(t1, t2, q, orientation)]
+                for i in 1:n_copies
+                    push!(cell_dofs, dof + (i - 1))
+                end
+            end
+        elseif nfacenodes == 4 # quadrilateral face
```
termi-official:
```suggestion
        elseif nfacevertices == 4 # quadrilateral face
```

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
+                error("$ndofs interior dofs on a quadrilateral face do not make up a regular lattice.")
+            end
+            for j in 0:(m - 1), i in 0:(m - 1)
+                dof = dofs[_canonical_facedof_index_quadrilateral(i, j, m, orientation)]
+                for k in 1:n_copies
+                    push!(cell_dofs, dof + (k - 1))
+                end
+            end
+        else
+            error("Faces with $nfacenodes vertices are not supported.")
```
termi-official:
```suggestion
            error("Faces with $nfacevertices vertices are not supported.")
```

src/Dofs/DofHandler.jl
```diff
@@ -713,39 +711,136 @@ so the unique representation is always a tuple length 3.
...
+    return c2 * (q + 1) - c2 * (c2 - 1) ÷ 2 + c1 + 1
+end
+
+# Compute the canonical (storage) index of the interior face dof at local lattice position
+# (i, j) of a quadrilateral face, with i, j ∈ {0, ..., m - 1} counted along the local
+# v1 → v2 and v1 → v4 directions, respectively. The local lattice maps to the canonical
+# lattice by the dihedral transformation that takes the local vertex at position u to
+# canonical position σ(u): a rotation by -shift_index quarter turns, followed by a diagonal
+# reflection (swapping positions 2 and 4) if the face is flipped.
+function _canonical_facedof_index_quadrilateral(i::Int, j::Int, m::Int, orientation::SurfaceOrientationInfo)
```
termi-official, line 865: To see if this works as expected in the fully integrated workflow, and as a regression check, should we add `Lagrange{RefHexahedron,3}`?

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Next] - xxxx-xx-xx
 
+### Added
+ - New interpolations `Lagrange{RefTetrahedron, 3}` and `Lagrange{RefTetrahedron, 4}`.
+   ([#1343])
+ - Dof distribution now supports interpolations with multiple dofs on faces shared between
+   cells (e.g. `Lagrange{RefTetrahedron, 4}`) by taking the relative orientation
+   (rotation and flip) of the face into account. Previously this errored. ([#1343])
```
termi-official:
```suggestion
 - Dof distribution now supports interpolations with multiple nodal dofs on faces shared between
   cells (e.g. `Lagrange{RefTetrahedron, 4}`) by taking the relative orientation
   (rotation and flip) of the face into account. The interpolation has to follow a specific tensor product ordering of the dofs, as described in the devdocs. ([#1343]).
```

termi-official (commented):
For the sake of not letting this contribution disappear, let us try to push it forward towards merging it, as this is a feature I have been asked about a few times already. We can make the internals of the face dof structure nicer in subsequent PRs.

KristofferC:
added two commits based on the review

src/interpolations.jl
```diff
@@ -1152,6 +1153,99 @@ function reference_shape_value(ip::Lagrange{RefHexahedron, 2}, ξ::Vec{3, T}, i:
 end
 
 
+##################################
+# Lagrange RefHexahedron order 3 #
+##################################
+# Tricubic tensor-product interpolation. The 64 nodes sit on the regular 4×4×4 lattice of
+# the reference hexahedron, each node being a tensor product of the equispaced 1D order-3
```
termi-official, line 1163: This is bad. We already made this mistake for higher order Triangles. Equispaced nodes no not work very well (the interpolation error oscillates and suffers from Runges phenomenon). The easiest fix is to use the 1D Lobatto points (the same as in our Lobatto quadrature) to build the 3D interpolation. This can tank some simulations pretty heart, because your approximation error can diverge with a good looking solution.

There are also tetrahedral Lagrange interpolations using non-equispaced points (e.g. obtained via the warp-and-blend technique by Tim Warburton) which work better. They are essentially the Lobatto points on the edges smoothly extended into the interior.

KristofferC: For order 3 the difference is really not that large I think. And there is some value in having it be consistent with the other interpolations. But I change it anyway (deal.ii apparently use these). But then we should probably be consistent with the order 2 as well? Could be done in another PR maybe

termi-official: > But then we should probably be consistent with the order 2 as well?

For order 2 and 1 the Lobatto points are equispaced already.

KristofferC: Yeah, sorry, meant dim 2.

termi-official: Oh, I see. I also thought about that, but this could be considered a breaking change, so I hesitated. The reason why this can be seen breaking is simply that it can change the simulation output considerably for those who depend on it right now.

KristofferC: I felt like I had to revert the Lobatto change because if you are in a `SubDofHandler` situation and attach a Q3 shell and a Q3 solid you will then associate different physical locations with the same Dof. I think we should introduce these 3d and 2d changes together (or maybe with completely new names) As a follow-up to that I made it so that 2D face dofs use the same orientation as 3d faces.

termi-official: > or maybe with completely new names

I would not even provide higher order interpolations with uniform nodal placement to users in first place, because these are dangerous to use. Should we then introduce a different name in this PR already and deprecate the high order uniform Lagrange 2D?

KristofferC: Since we have the 2d uniform I don't think it is too bad to add the 3d one and then do a wholesale switch for both of them at the same time when we figure out the best way of doing that.

docs/src/devdocs/interpolations.md
```diff
@@ -30,6 +30,14 @@ Ferrite.dirichlet_facedof_indices(::Interpolation)
 Ferrite.dirichlet_edgedof_indices(::Interpolation)
 ```
 
+#### Multiple dofs per face in 3D
+For interpolations with more than one interior dof on a face shared between 3D cells, the
+interior face dofs must be placed on a regular lattice and the following must be implemented
+to opt in to the dof distribution permutation (see [`Ferrite.permute_and_push!`](@ref)).
+```@docs
+Ferrite.interior_facedofs_on_lattice(::Interpolation)
```
termi-official, line 38: Should we mark this as experimental in the hope that users won't rely on it?

termi-official (changes requested):
Before we merge this, can we make a PR with higher order quadrature rules on tets and add these to the tests? With order 4 rules user's won't be able to use `Lagrenge{RefTet,3/4}` anyway.

docs/src/devdocs/interpolations.md
KristofferC: it's in devdocs, don't think it need an explicit experimental, or?

termi-official: Let us think about the following scenario: A user wants to add a new interpolation. IMHO the user will end up either in the dev-docs and start putting something together, or he will first end up in the interpolations.jl file, depending on how he searches. Worst case he has more than 1 dof per face, maybe because he is implementing Legendre support (instead of Lagrange), so he will hit this. How should we guide him to proceed?

KristofferC:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/1389 for quad rule.

KristofferC:
I think I am reaching my limit with this PR, happy for someone else to push this over the finish line 👍

termi-official:
Should be good to go from my side.

termi-official (approved):

KristofferC:
<3 thanks for all the help!

PR #1348, opened 2026-06-15
Update Optim.jl version in docs as Optim#1234 is resolved

KnutAM:
With `Optim.jl` v2.2.1, the landau example now works again (was restricted to `Optim.jl` v1 before). 

xref: https://github.com/JuliaNLSolvers/Optim.jl/issues/1234#issuecomment-4706006666

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.12.6"
 manifest_format = "2.0"
-project_hash = "6d0dc5950af997fe06d71214d77ad65f439e8f72"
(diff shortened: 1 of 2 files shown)


KnutAM:
Had to limit diffeq version, had to stop fixing that separate issue, but came as far as 

* `autodiff = false` to `autodiff = AutoFiniteDiff()`
* `integrator = init(...; verbose = true)` to `integrator = init(...; verbose = DEVerbosity())` (or smth else, but this was the easiest
* `intervals`: Not defined error - stopped here...

termi-official:
I will handle the update. I just forgot about doing it yet.

termi-official (approved):

PR #1351, opened 2026-06-17
Bump Version 1.4.1

termi-official:


KnutAM:
Needs also update to changelog headers. 

Guess the [no CI] in #1350 was a bit too confident :D 
I forgot about the pre-commit checks...

termi-official:
> Guess the [no CI] in #1350 was a bit too confident :D

Yea, I forgot that these also have pre-commit checks. :D I guess we should introduce some `[skip test]` marker for such cases?

KnutAM:
Not 100 % (feel I should now this :D), but should the docs Manifest also be updated so that it has v1.4.1 in it? On the other hand, the hash would still be different after the squash commit?

termi-official:
> should the docs Manifest also be updated so that it has v1.4.1 in it?

Yes.

> On the other hand, the hash would still be different after the squash commit?

We use the path as source, so this is fortunately not a concern.

KnutAM (approved):

PR #1352, opened 2026-06-17
test that the package has no `Core.Box`es

KristofferC:
To avoid regressing on https://github.com/Ferrite-FEM/Ferrite.jl/pull/1265. This test is only active on nightly.


termi-official:
Any reason we didn't merge this?

KristofferC:
probably because we do not run CI on nightly so it wasn't really testing anything at this moment

Issue #1320, opened 2026-04-17
Problems with periodic BCs


DRollin:
I ran into two issues when using `PeriodicDirichlet` which I want to document here:
1. They are not affected by `update!`
    When you know that `AffineConstraint`s are not updated, this is not surprising, but in my opinion it would be good to have it updated.
   (I wanted to use that to solve the same system with different inhomogeneities.)
   Since @kreutzheidephil wanted to take a look at the `AffineConstraint`s, he might be able to solve this.
2. When defining the boundary, the approaches `∂Ω = ["left" => "right",  "bottom" => "top"]` and 
    ```
        img = union(getfacetset(grid, "left"),  getfacetset(grid, "bottom"))
	mir = union(getfacetset(grid, "right"), getfacetset(grid, "top"))
	∂Ω = collect_periodic_facets(grid, mir, img) 
    ```
     Do not give the same result (for the first version weird things are happening at the corners).
    Again, I am in contact with @kreutzheidephil to find a solution, but wanted to document the problem here.

lijas:
For the second point, the reason they give different results is because 

```
periodic = PeriodicDirichlet(
    :u,
    ["left" => "right", "bottom" => "top"],
    [1, 2]
)
```
also locks one of the corner nodes to remove the rigid body motion. The approach with collect_periodic_facets does not do this, and the user has to remove rigid body rotations/translations themself some way. This is poorly documented I think, and should be updated someway. My suggestion is to:
1. add an option when creating a periodic consraint `PeriodicDirichlet(...; remove_rigid_body_motion=true`). 
2. or for the user to define themself what corner node to lock, via a normal Dirichlet constraint on one of the corner nodes. The problem with this approach is that I think the problem may become overconstrained in some cases (a node is both master and slave). This couldbe solved by implementing a "untangling" algorithm of the constraints in the close!() function.

kreutzheidephil:
Hi @lijas 
correct me if I am wrong, but isn't the locking of the solution in corners legacy code, i.e. [here](https://github.com/Ferrite-FEM/Ferrite.jl/blob/efa551d6e6c353e9c0447fdf1e86409ed8be700e/src/Dofs/ConstraintHandler.jl#L1174-L1182)? If that is the case then wouldn't what @DRollin mentioned be a bug, as no `depwarn` is thrown?

lijas:
[...]

PR #1353, opened 2026-06-18
Add time dependent PeriodicDirichlet

### LINKED ISSUE #1332: Periodic / affine constraints don't allow time dependent inhomogeneities
I just wanted to document this. @DRollin mentioned this in #1320 but I wanted to explicitly open an issue just for this to facilitate discussion and to add a MWE.

```
using Ferrite
dh = DofHandler(grid)
add!(dh, :u, ip)
close!(dh)
ch = ConstraintHandler(dh)
∂Ω = collect_periodic_facets(grid, getfacetset(grid, "left"), getfacetset(grid, "right"))
pdbc = PeriodicDirichlet(:u, ∂Ω, (x, t) -> t)
add!(ch, pdbc)
close!(ch)
println("ch.inhomogeneities before update!: $(ch.inhomogeneities)")
update!(ch, 1.0)
println("ch.inhomogeneities after update!:  $(ch.inhomogeneities)")
```
output: 
```
julia> 
ch.inhomogeneities before update!: [0.0, 0.0]
ch.inhomogeneities after update!:  [0.0, 0.0]
```

This leads to a difficult-to-diagnose silent errors when solving nonlinear problems with pseudo time stepping for example.

lijas:
Adds an update function for PeriodicDirichlet's so it is possible to have time dependent inhomogeneties.

I dont know how common it is but seems to be requested in #1320 

It also seems like PeriodicDirichlet's with inhomogeneties were untested, so this adds some more tests for that.

Closes #1332

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Next] - xxxx-xx-xx
 
 ### Documentation
-
 - Landau example in code gallery now shows how to use DifferentiationInterface and HyperHessians as a backend. (#1345)
+
+### Added
+ - `PeriodicDirichlet` can now have time dependent inhomogenities. ([#1353])
 
 ## [v1.4.1] - 2026-06-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Landau example in code gallery now shows how to use DifferentiationInterface and HyperHessians as a backend. (#1345)
 
 ### Added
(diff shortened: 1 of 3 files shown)


KnutAM:
I think it would be nice to have this more general and not restricted to periodic BCs, but to have something like
https://github.com/Ferrite-FEM/Ferrite.jl/discussions/1336#discussioncomment-16950944
(named: https://github.com/Ferrite-FEM/Ferrite.jl/discussions/1336#discussioncomment-16952724)

lijas:
We can probably add something like that aswell for other cases, but i think it is better if the constraint (PeriodicDirichlet in this case) that "owns"/created the affine constraints should handle its update (instead of passing it to some other struct). 

Also, i dont se how your approach can pass the coordinate argument in f(x,t) because that requires looping over the faces.

KnutAM:
> Also, i dont se how your approach can pass the coordinate argument in f(x,t) because that requires looping over the faces.

That’s the `args[i]` part

Issue #1357, opened 2026-06-26
Docs: typo in Stokes Flow tutorial?


SeSodesa:
In the [Stokes Flow tutorial](https://ferrite-fem.github.io/Ferrite.jl/stable/tutorials/stokes-flow/#Geometry-and-mesh-generation-with-Gmsh.jl), there is the following line related to storing a transformation matrix into a Gmsh file:

https://github.com/Ferrite-FEM/Ferrite.jl/blob/93ffff1c0a80f0b96c5c90769c218d997eac214c/docs/src/literate-tutorials/stokes-flow.jl#L195

Shouldn't cosine of $- pi / 2$ be equal to $0$? Julia seems to think so at least:
```julia
julia> cos(-pi/2)
6.123233995736766e-17
```
Have not yet checked for what else might need changing but I though I'd raise the issue now.

KristofferC:
Yes should be `sin`. For symmetry, might make sense to also annotate the two `cos(-pi/2)` values (0)

PR #1360, opened 2026-06-29
docs: fix Stokes flow transform comments

wenxiaojie1:
## Summary\n- correct the sine/cosine annotation for the Stokes flow transformation matrix\n- explicitly annotate the zero cosine entries for symmetry, as suggested in the issue\n\n## Checks\n- git diff --check\n\nCloses #1357


fredrikekre:
#1358

PR #1361, opened 2026-06-29
Refactor evaluate_at_grid_nodes for L2Projector

lijas:
I noticed that there were a lot of overlap between the DofHandler and L2Projector's evaluate_at_grid_nodes, so I combined them.

### DIFF
--- a/src/Dofs/DofHandler.jl
+++ b/src/Dofs/DofHandler.jl
@@ -943,7 +943,7 @@ function_value_init(::ScalarInterpolation, ::AbstractVector{T}) where {T} = zero
 function_value_init(::VectorInterpolation{vdim}, ::AbstractVector{T}) where {vdim, T <: Number} = zero(Vec{vdim, T})
 
 # Internal method that have the vtk option to allocate the output differently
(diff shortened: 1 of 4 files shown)


src/Dofs/DofHandler.jl
```diff
@@ -988,13 +952,18 @@ function _evaluate_at_grid_nodes(dh::DofHandler{sdim}, u::AbstractVector{T}, fie
...
+            @assert n_c == 1 # L2Projector always stores scalar field
+            S <: Vec{2} ? 3 : M # Pad 2D Vec to 3D
+        end
         # Float32 is the smallest float type supported by VTK
         TT = promote_type(T, Float32)
         data = fill!(Matrix{TT}(undef, vtk_dim, getnnodes(get_grid(dh))), NaN)
     else
         # Just evaluation at grid nodes
-        RT = typeof(function_value_init(ip, u))
+        RT = S <: Number ? typeof(function_value_init(ip, u)) : S
```
termi-official, line 966: Can we introduce a cleaner implementation for this query? Note that e.g.
```
julia> Complex isa Number
false
```

KnutAM: `Complex isa Type`?

termi-official: Oh, right.

PR #1367, opened 2026-07-05
Generalize grid generator to support sdim > rdim

KnutAM:
```
grid = generate_grid(Quadrilateral, (10, 10), Vec((-1.0, -1.0, 0.0)), Vec((1.0, 1.0, 0.0)));
transform_coordinates!(grid, x -> Vec((x[1], x[2], (x[1]^2 + 2*x[2]^2)/4)));
VTKGridFile("plate", grid) do vtk; end
```
image

This PR introduces a node generator that works by using the interpolations on a given reference shape (= shape of the generated grid global geometry). Is slightly slower, but since this doesn't take much time anyways, I think that is fine (and with this PR a bit faster in most cases since we now always preallocate the nodes). Not used for non-gridded node distribution (pyramid and serendipty-hex).


docs/src/literate-tutorials/linear_shell.jl
```diff
@@ -15,13 +15,9 @@ using ForwardDiff
...
 function main() #wrap everything in a function...
 
-# First we generate a flat rectangular mesh. There is currently no built-in function for generating
-# shell meshes in Ferrite, so we have to create our own simple mesh generator (see the
-# function `generate_shell_grid` further down in this file).
+# First we generate a flat rectangular mesh embedded in 3D.
 #+
-nels = (10,10)
-size = (10.0, 10.0)
-grid = generate_shell_grid(nels, size)
```
termi-official, line 24: This function is now dead code in the tutorial.

KristofferC:
Net negative line diff <3

docs/src/literate-tutorials/linear_shell.jl
KnutAM: Forgot to remove it :)

src/Grid/grid_generators.jl
```diff
@@ -590,3 +535,44 @@ function generate_grid(::Type{Tetrahedron}, cells_per_dim::NTuple{3, Int}, left:
...
+# ====================================================================== #
+function _calculate_coordinate(M::Vector{T}, coords::Vector{Vec{sdim, T}}) where {sdim, T}
+    x = zero(eltype(coords))
+    @inbounds for i in 1:length(M)
+        x += M[i] * coords[i]
+    end
+    return x
+end
+
+function _generate_nodes(ipg::ScalarInterpolation{<:AbstractRefShape{rdim}}, nnodes::NTuple{rdim, Int}, corners::Vector{Vec{sdim, T}}) where {rdim, sdim, T}
```
termi-official, line 550: Maybe something more descriptive?
```suggestion
function _generate_uniform_nodes(ipg::ScalarInterpolation{<:AbstractRefShape{rdim}}, nnodes::NTuple{rdim, Int}, corners::Vector{Vec{sdim, T}}) where {rdim, sdim, T}
```

KnutAM: Currently it is only used with linear interpolations, but :D
```julia
grid = generate_grid(Quadrilateral, (10,10));
nu_nodes = Ferrite._generate_nodes(Lagrange{RefQuadrilateral,2}(), (11, 11), Vec{2}.([[-1.0, -1.0], [1.0, -1.0], [1.0, 1.0],[-1.0, 1.0], [0.0, -1.3], [0.8, 0.0], [0.0, 1.3], [-1.2, 0.0], [0.0, 0.0]]));
VTKGridFile("nonuniform", Grid(grid.cells, nu_nodes)) do vtk; end
```

image

KnutAM: Jokes aside, kept it strictly internal for now. Although it generates uniform nodes on the reference shape, it map to the possibly non-uniform spatial domain, so calling it `uniform` seems a bit misleading? But can be changed later. 

But if that blocks merging - I'm happy to change :)

termi-official: Not really a blocker. Maybe structured/regular is a better fit then? If not, then just proceed as is.

KnutAM: I'll leave it as it is for now then, matches `generate_grid` which does uniform, and then we can change in the future since it is strictly internal

KnutAM:
For future ref, the update was tested with the following script before updating the `generate_grid` function to use the `_generate_nodes` function.
```julia
using Ferrite

function nels_to_nnodes(::Type{<:Union{Line, Triangle, Quadrilateral, Hexahedron, Tetrahedron, Wedge}}, nels)
    return nels .+ 1
end
function nels_to_nnodes(::Type{<:Union{QuadraticLine, QuadraticTriangle, QuadraticQuadrilateral}}, nels)
    return 2 .* nels .+ 1
end

function testgen(::Type{CT}, nels::Tuple, left, right) where {CT <: Ferrite.AbstractCell}
    nnodes = nels_to_nnodes(CT, nels)
    return Ferrite._generate_nodes(Lagrange{Ferrite.RefHypercube{Ferrite.getrefdim(CT)}, 1}(), nnodes, left, right)
end

for CT in (
        Line, Quadrilateral, Triangle, Hexahedron, Tetrahedron, Wedge,
        QuadraticLine, QuadraticTriangle, QuadraticQuadrilateral,)# QuadraticHexahedron, QuadraticTetrahedron)
    rdim = Ferrite.getrefdim(CT)
    left = rand(Vec{rdim})
    right = left + rand(Vec{rdim})
    nels = ntuple(_ -> rand(1:10), rdim)
    grid = generate_grid(CT, nels, left, right)
    nodes = testgen(CT, nels, left, right)
    check = get_node_coordinate.(nodes) ≈ get_node_coordinate.(grid.nodes)
    println(CT, ": passed = ", check)
end
```

src/Grid/grid_generators.jl
termi-official, line 550: @KnutAM Interestingly this change tanks downstream tests which add custom cellsets via `addcellset!` to split the geometry into a left and right half. Nodes which were previously machine precision 0 are now very slightly disturbed next to zero and `addcellset!' missed them, because we check as the default condition whether all *nodes* fulfill the criterion. Maybe we should add a way to make `addcellset!` check whether the cell center fulfills the criterion instead to add cells (or even change this default)? This would probably be also more in line with what users expect, as I think most of them will be surprised that if we add two cellsets, one with norm(x) < 1.0 (e.g. for some pore) and one with norm(x) >= 1 (for the matrix), that they have a band of cells which does not belong to either domain.

PR #1369, opened 2026-07-09
Support mixed reference dimensions in ExclusiveTopology

### LINKED ISSUE #843: Revive mixed dimensional relations in topology

termi-official:
Revive mixed-dimensional relations in the topology (issue #843). The shared entity between two connected cells is classified by the number of shared vertices, so mixed-dimensional connections are stored in the existing vertex/edge/face collections without new data structures.

- Allow construction when the highest cell reference dimension equals the spatial dimension; still reject purely embedded (shell) grids.
- Drop the same-refdim restriction in the neighbor loop.
- Handle :mixed in _getsizehint via the spatial dimension.
- For embedded cells whose shared vertices don't form the expected entity in both cells (e.g. a Line along a Hexahedron's face-diagonal), fall back to the next lower-dimensional entity instead of silently dropping the connection.
- Resolve the facet/edge dimension per cell in getneighborhood (via get_reference_dimension(grid, cellid)) rather than assuming a common reference dimension, so per-cell FacetIndex/EdgeIndex queries work on mixed grids.

The bulk facetskeleton / get_facet_facet_neighborhood remain unsupported for mixed grids since they assume a common facet dimension.


termi-official:
I am good with the PR as-is. Is this also good on your end @koehlerson ?

src/Grid/topology.jl
```diff
@@ -33,10 +33,20 @@ edge, only the face neighborhood is saved. The lower dimensional neighborhood is
...
+`VertexIndex`, `EdgeIndex`, `FaceIndex`, and `FacetIndex` work for such grids (the facet
+dimension is resolved per cell). The bulk operations [`facetskeleton`](@ref) and
+`get_facet_facet_neighborhood` remain unsupported, since they assume a common facet
+dimension across the whole grid.
+
 !!! warning "Limitations"
     The implementation only works with conforming grids, i.e. grids without "hanging nodes". Non-conforming grids will give unexpected results.
-    Grids with embedded cells (different reference dimension compared
-    to the spatial dimension) are not supported, and will error on construction.
+    Purely embedded grids, where the highest cell reference dimension is smaller than the
```
KnutAM, line 49: If logic works if I add a single cell with rdim=sdim, how come it doesn’t support this case?

termi-official: I am pretty sure we still have some open issues with some non-manifold topologies (e.g. a T carrier shell model) not being identified correctly, but I have not looked into this one exhaustively. This definitely also affects mixed-dimensional topologies. We can fix these later. I just need the cables working for now.

src/Grid/topology.jl
```diff
@@ -75,14 +85,21 @@ function ExclusiveTopology(grid::AbstractGrid{sdim}) where {sdim}
...
+            # In that case we fall back to the next lower-dimensional entity.
             if num_shared_vertices == 1
                 _add_single_vertex_neighbor!(vertex_vertex_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id)
-            elseif num_shared_vertices == 2 # Shared edge
-                _add_single_edge_neighbor!(edge_edge_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id)
-            elseif num_shared_vertices >= 3 # Shared face
-                _add_single_face_neighbor!(face_face_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id)
+            elseif num_shared_vertices == 2 # Shared edge (or two separate vertices)
+                _add_single_edge_neighbor!(edge_edge_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id) ||
+                    _add_single_vertex_neighbor!(vertex_vertex_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id)
```
koehlerson, line 100: spacing seems weird but maybe required by runic

src/Grid/topology.jl
```diff
@@ -75,14 +85,21 @@ function ExclusiveTopology(grid::AbstractGrid{sdim}) where {sdim}
...
+            elseif num_shared_vertices >= 3 # Shared face (or lower-dimensional entities)
+                _add_single_face_neighbor!(face_face_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id) ||
+                    _add_single_edge_neighbor!(edge_edge_neighbor_buf, cell, cell_id, neighbor_cell, neighbor_cell_id) ||
```
koehlerson, line 103: same as above

src/Grid/topology.jl
```diff
@@ -33,10 +33,20 @@ edge, only the face neighborhood is saved. The lower dimensional neighborhood is
 - `edge_skeleton::Union{Vector{EdgeIndex}, Nothing}`:     List of unique edges in the grid given as `EdgeIndex`
 - `vertex_skeleton::Union{Vector{VertexIndex}, Nothing}`: List of unique vertices in the grid given as `VertexIndex`
 
+Grids with mixed reference dimensions (e.g. a 3D grid containing both `Hexahedron` and
+`Quadrilateral` cells) are supported: the shared entity is classified by the number of
+shared vertices, so mixed-dimensional connections are stored in `vertex_vertex_neighbor`,
+`edge_edge_neighbor`, or `face_face_neighbor` accordingly. Per-entity queries with
```
koehlerson, line 39: this could need some better explanation I think because it was unclear to me what that means. Maybe make it directly illustrative by a minimal example. Or state that an edge_edge connection can return an edge->cell connection (as in the case in the test files

koehlerson (commented):
I think this could need some helper functions for production code in order to make users clear what kind of connection they are dealing with

```julia
entity_dim(::VertexIndex) = 0
entity_dim(::EdgeIndex)   = 1
entity_dim(::FaceIndex)   = 2
codim(grid, idx) = Ferrite.get_reference_dimension(grid, idx[1]) - entity_dim(idx)
```

and some comment somewhere in the docs like
```julia
for n in getneighborhood(topo, grid, FacetIndex(1, 2))
    if codim(grid, n) == 0      # integrate over the embedded element
    elseif codim(grid, n) == 1  # standard facet-to-facet coupling
    else                        # lower-dimensional contact only
  end
end
# -> neighbor EdgeIndex((2, 1)) is a whole cell -> integrate over the embedded element
```

src/Grid/topology.jl
termi-official: I do not understand. The do string already mentions that `edge_edge` can be mixed.

> shared vertices, so mixed-dimensional connections are stored in `vertex_vertex_neighbor`,

Any chance github does not refresh again?

koehlerson: no that's fine but I think it wasn't clear (to me at least) that mixed-dimensional connections are still `idx::EdgeIndex -> nidx::EdgeIndex` even though the map to the other edge is actually the whole cell. If you read that without knowing how it's implemented you could assume to get something like `idx::EdgeIndex -> nidex::CellIndex`

koehlerson: I think it would benefit if the docs state somewhere close to the topology that for mixed dimensional entities https://ferrite-fem.github.io/Ferrite.jl/stable/topics/reference_shapes/#Entity-naming is taken seriously in that sense that if you connect a 2D cell to a 1D cell you will have an edge connection which is why `EdgeIndex -> EdgeIndex` make sense. However, you have to have code that makes the distinction between what to integrate (which are the type of functions I put in the review comment) because you can't interfere that by the return type

termi-official: Please take a look at the latest commit. Is this what you had in mind?

benchmark/benchmarks-boundary-conditions.jl
```diff
@@ -0,0 +1,42 @@
+#----------------------------------------------------------------------#
```
KristofferC: Does this file has anything to do with the PR?

termi-official: Rebased and the was file marked for deletion. Looks like I fucked up the rebase.

termi-official:
That CI failure is a todo for another day ...

koehlerson:
why does the forest topology choke now? :thinking:

termi-official:
> why does the forest topology choke now? 🤔

See commit [e02a224](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1369/commits/e02a224689d5140838d90dc1fa569862592c574f) . We need to query cell info to distinguish different cases in the mixed-dimensional case, which is not allowed in the forest data structure.

docs/src/topics/grid.md
```diff
@@ -148,6 +148,36 @@ In order to use boundaries, e.g. for Dirichlet constraints in the ConstraintHand
...
+|     | |
++-----+ +
+|     |
+|  3  |
+|     |
++-----+
+```
+Then we can handle the following cases
+```julia
+for n in getneighborhood(topo, grid, FacetIndex(1, 2))
```
koehlerson:
```suggestion
for n in getneighborhood(topo, grid, FacetIndex(1, n))
```
I guess?

koehlerson (approved):

PR #1375, opened 2026-07-14
Fix VTK symmetric tensor export

### LINKED ISSUE #1374: write_cell_data LoadError when writing SymmetricTensor data.
I get the error 
"ERROR: LoadError: MethodError: no method matching length(::Type{SymmetricTensor{2, 3, Float64, 6}})
The function `length` exists, but no method is defined for this combination of argument types."

when calling write_cell_data where the cell data is a Vector{SymmetricTensor}. 

Overloading the length function does make it work as intended, but this should already be the case as Tensors.jl  (https://github.com/Ferrite-FEM/Tensors.jl/blob/83e84dad1e7fa1ac43847261bcf99dba70cbf1f1/src/Tensors.jl#L193)

Here is a small code I made with the issue : [MRE.zip](https://github.com/user-attachments/files/30002571/MRE.zip) 

```julia
using Ferrite

#If we overload this, this works 
#Base.length(::Type{SymmetricTensor{order,dim,T,M}}) where{order,dim,T,M} = M 

grid = generate_grid(Quadrilateral, (1, 1));

dh = DofHandler(grid)
cell_data = [Tensor{2,3}((1.,2.,3.,4.,5.,6.,7.,8.,9.))]
cell_data_symmetric = [SymmetricTensor{2,3}((1.,2.,3.,4.,5.,6.))]

l1 = length(cell_data[1]) #9
l2 = length(cell_data_symmetric[1]) #Returns 9 no matter what, shouldn't it be 6?

VTKGridFile("example_cell_data", dh) do vtk
    write_cell_data(vtk,cell_data,"cell_data") #This works
    write_cell_data(vtk,cell_data_symmetric,"cell_data_symmetric") #This does not
    #ERROR: LoadError: MethodError: no method matching length(::Type{SymmetricTensor{2, 3, Float64, 6}})
    #The function `length` exists, but no method is defined for this combination of argument types.
end

```

termi-official:
Closes #1374


src/Export/VTK.jl
```diff
@@ -241,10 +241,23 @@ end
...
     return vtk
 end
+function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{<:SymmetricTensor{O, D, T, M}}, name) where {O, D, T, M}
+    noutputs = M
+    npoints = length(celldata)
+    out = zeros(T, noutputs, npoints)
+    for i in 1:npoints
+        toparaview!(@view(out[:, i]), celldata[i])
+    end
+    WriteVTK.vtk_cell_data(vtk.vtk, out, name)
```
KnutAM:
```suggestion
    WriteVTK.vtk_cell_data(vtk.vtk, out, name; component_names = component_names(eltype(celldata)))
```

Unrelated, but seems like we are using internal API...
https://github.com/JuliaVTK/WriteVTK.jl/blob/813f021448cc154d77035d628edd49ff0ae6c679/src/write_data.jl#L455-L457

src/Export/VTK.jl
```diff
@@ -241,10 +241,23 @@ end
 
 Write the `celldata` that is ordered by the cells in the grid to the vtk file.
 """
-function write_cell_data(vtk::VTKGridFile, celldata, name)
+function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{T}, name) where {T <: Union{Real, Tensor}}
     WriteVTK.vtk_cell_data(vtk.vtk, celldata, name)
     return vtk
 end
+function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{<:SymmetricTensor{O, D, T, M}}, name) where {O, D, T, M}
+    noutputs = M
```
KnutAM: Technically, this is "breaking" since it will use the "correct" component names for `Tensor`, but I don't think any will complain :D 
Code from: https://github.com/Ferrite-FEM/Ferrite.jl/blob/db867ea7456c0016fcf08a758b9828ca4f955e1e/src/Export/VTK.jl#L168-L170

```suggestion
function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{S}, name) where {O, D, T, M, S <: Union{Tensor{O, D, T, M}, SymmetricTensor{O, D, T, M}}}
    noutputs = S <: Vec{2} ? 3 : M # Pad 2D Vec to 3D
```

src/Export/VTK.jl
```diff
@@ -241,10 +241,23 @@ end
 
 Write the `celldata` that is ordered by the cells in the grid to the vtk file.
 """
-function write_cell_data(vtk::VTKGridFile, celldata, name)
+function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{T}, name) where {T <: Union{Real, Tensor}}
```
KnutAM:
```suggestion
function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{<:Real}, name)
```

KnutAM (commented):
I think we should use `component_names` here to avoid the strange output, which will be even stranger I think for symmetric tensors showing just component 0-5: (11, 21, 31, 22, 23, 33)?

src/Export/VTK.jl
```diff
@@ -241,10 +241,23 @@ end
...
+    out = zeros(T, noutputs, npoints)
+    for i in 1:npoints
+        toparaview!(@view(out[:, i]), celldata[i])
+    end
+    WriteVTK.vtk_cell_data(vtk.vtk, out, name; component_names = component_names(eltype(celldata)))
+    return vtk
+end
+function write_cell_data(vtk::VTKGridFile, celldata::AbstractVector{T}, name) where {T}
+    error("Export of $T is not supported.")
+end
```
KnutAM, line 260: Is this to catch complex numbers etc. ?

If so, we should also add a test, and disallow complex number in the `celldata::Vector{<:AbstractTensor}` case as well? If not we hit the complex number error in  WriteVTK.jl for the tensor-case, but a Ferrite error for the scalar case...

But really good to add this! Perhaps add a test for `MixedTensor2` that we hit this case in addition to a `Complex`? I don't think order>2, e.g. `Tensor{4}`, is supported though, and will hit some downstream error during Voigt conversion I guess?

termi-official: > Is this to catch complex numbers etc. ?

This gives users a meaningful error message instead of the cryptic message they get right now if some internal dispatch is missing, as e.g. in the case which this PR fixes.

For complex WriteVTK will throw a meaningful error
```
julia> VTKGridFile("test", grid) do vtk
           Ferrite.WriteVTK.vtk_cell_data(vtk.vtk, [0.0im for i in 1:27], "u")
       end
ERROR: ArgumentError: data type not supported by VTK: ComplexF64
```

PR #1376, opened 2026-07-15
Fill Ke and fe outside the element routine in the heat equation example

KnutAM:
For more advanced examples when reusing the assembly loop, it is nice to not have the reseting inside. This has the advantage in e.g. #1291 that we can refer to the heat equation that we have exactly the same element routine.

Edit: Added also the change that we pass `K` to assemble global, but not `f`. Now we pass both and modify both. I think this is cleaner and less confusing.


KristofferC:
Mars sense, it seems wrong to zero it inside there.

PR #1378, opened 2026-07-16
[POC] Showcase ImplicitDifferentiation.jl usage with Voce plasticity

### LINKED ISSUE #1231: Tutorial for combining AD for inner iterations (for e.g. plasticity) and AD on the element level
This is not trivial to do, what I have done earlier is:

- Extract the real part from the dual numbers on the element level AD.
- Call the inner solver like normal (with real numbers).
- Get back the value for e.g. the stress.
- Use the implicit function theorem to compute the gradient (d sigma / d_epsilon)
- Propagate back that gradient to the element level AD.

termi-official:
Proof of concept for using ImplicitDifferentiation.jl . Still needs quite a bit of polishing.

~~This should move us closer towards showcasing #1231.~~ Closes #1231 .


docs/src/literate-tutorials/nonlinear-plasticity.jl
```diff
@@ -0,0 +1,462 @@
...
+# Constructor for initializing a material state. Every quantity is set to zero.
+MaterialState() = MaterialState(zero(SymmetricTensor{2,3}), zero(SymmetricTensor{2,3}), 0.0)
+Base.zero(::MaterialState) = MaterialState()
+
+# --- SVector interface for ImplicitDifferentiation ---
+# ImplicitDifferentiation requires AbstractArray inputs/outputs. We use allocation-free
+# SVector{6} for strains and SVector{13} for the full state (both in Mandel convention).
+# Layout: [εᵖ(1:6), σ(7:12), γ(13)] with off-diagonal Mandel factor √2.
+
+@inline function to_svec(A::SymmetricTensor{2,3,T}) where T
```
KristofferC: This is such a common thing to want to do, I wonder if we should somewhere define some helper that easily helps you define a `flatten` and `unflatten` (is that even a word...) on a struct that can contain things like tensors, numbers, etc.

termi-official: Indeed. But TBH I would actually rather prefer to have some stack-allocated array available here, since the tuples become suboptimal for slightly larger problems.

PR #1381, opened 2026-07-21
Add VTKHDF export via VTKHDF.jl package extension

KristofferC:
Putting this up so the API can be reviewed, even though VTKHDF is not registered yet.

Adds VTKHDFGridFile, the VTKHDF (HDF5-based) counterpart of VTKGridFile, implemented in the new FerriteVTKHDF extension for https://github.com/Ferrite-FEM/VTKHDF.jl. Supports the same data functions (write_solution, write_cell_data, write_projection, write_node_data, set writers) including discontinuous grids, plus temporal files via VTKHDF.write_timestep: a whole simulation lands in a single file with the grid stored only once.

The transient heat equation tutorial now exports one temporal .vtkhdf file instead of a .pvd collection with one .vtu file per time step.

### DIFF
--- a/ext/FerriteVTKHDF.jl
+++ b/ext/FerriteVTKHDF.jl
@@ -74,8 +74,8 @@ function VTKHDF.write_timestep(f::Function, vtk::VTKHDFGridFile, t::Real; kwargs
     return vtk
 end
 
-# ---- node data conversion (mirrors Ferrite's WriteVTK path, minus component
-# names, which the VTKHDF format does not support) ----
+# Node data conversion mirrors the WriteVTK path, minus component names,
+# which the VTKHDF format does not support.
 
 function _node_matrix(nodedata::Vector{S}) where {O, D, T, M, S <: Union{Tensor{O, D, T, M}, SymmetricTensor{O, D, T, M}}}
     noutputs = S <: Vec{2} ? 3 : M # pad 2D Vec to 3D
--- a/ext/FerriteVTKHDF.jl
+++ b/ext/FerriteVTKHDF.jl
@@ -1,0 +1,133 @@
+# Implementation of `Ferrite.VTKHDFGridFile` (see src/Export/VTKHDF.jl), the
+# VTKHDF counterpart of `VTKGridFile`, backed by VTKHDF.jl.
+module FerriteVTKHDF
+
+using Ferrite:
+    Ferrite, VTKHDFGridFile, DofHandler, AbstractDofHandler, AbstractGrid,
+    L2Projector, Vec, Tensor, SymmetricTensor,
+    get_grid, getfieldnames, getnnodes, conformity, H1Conformity
+import Ferrite: write_solution, write_projection, write_cell_data, write_node_data
+using VTKHDF: VTKHDF, vtkhdf_grid, VTKPointData, VTKCellData
+
+function Ferrite.VTKHDFGridFile(filename::String, dh::DofHandler; kwargs...)
+    for sdh in dh.subdofhandlers
+        for ip in sdh.field_interpolations
+            if !isa(conformity(ip), H1Conformity)
+                return VTKHDFGridFile(filename, get_grid(dh); write_discontinuous = true, kwargs...)
+            end
+        end
+    end
+    return VTKHDFGridFile(filename, get_grid(dh); kwargs...)
+end
+
+function Ferrite.VTKHDFGridFile(filename::String, grid::AbstractGrid; write_discontinuous = false, kwargs...)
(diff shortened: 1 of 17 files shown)


fredrikekre:
Should we just move to this wholesale? Is there a disadvantage?

KristofferC:
There is a binary dependency with VTKHDF and I think there is less support for reading it in various tools.

ext/FerriteVTKHDF.jl
```diff
@@ -0,0 +1,134 @@
...
+    write_solution(vtk, dh, u)
+end
+```
+"""
+function VTKHDF.write_timestep(f::Function, vtk::VTKHDFGridFile, t::Real; kwargs...)
+    VTKHDF.write_timestep(_ -> f(vtk), hdf(vtk), t; kwargs...)
+    # flush each step so a hard kill (OOM, scheduler kill, segfault) still
+    # leaves a valid file with all completed steps
+    file = hdf(vtk).file
+    file === nothing || Base.flush(file)
```
fredrikekre: Perhaps this should be a `flush = true` kwarg to `VTKHDF.write_timestep`? Perhaps that should even be default?

KristofferC: I feel it should maybe not be the default in VTKHDF itself because that should maybe be considered lower level, but I am not sure...

KristofferC:
I converted more examples to HDF where it made sense

ext/FerriteVTKHDF.jl
KnutAM: I see the argument why not default, but a kwarg seems reasonable and convenient to me. 

Also, what is the case when `file === nothing`?

src/Export/VTKHDF.jl
```diff
@@ -0,0 +1,39 @@
...
+    end
+end
+close(vtkhdf)
+```
+"""
+struct VTKHDFGridFile
+    vtk::Any  # ::VTKHDF.VTKHDFFile (untyped since VTKHDF is a weak dependency)
+    cellnodes::Union{Vector{UnitRange{Int}}, Nothing}
+    node_mapping::Union{Vector{Int}, Nothing}
+end
```
KnutAM: Any reason not to do like this? 

```suggestion
struct VTKHDFGridFile{FT}
    vtk::FT  # ::FT <: VTKHDF.VTKHDFFile (FT unrestricted since VTKHDF is a weak dependency)
    cellnodes::Union{Vector{UnitRange{Int}}, Nothing}
    node_mapping::Union{Vector{Int}, Nothing}
end
```

ext/FerriteVTKHDF.jl
KristofferC: The cost is like sub ms so it didn't feel worth adding a keyword.

KnutAM: You mean that it could just as well be default to always do it then? To me it just feels like there should be some API for this in VTKHDF, since `file = vtk.file` for `vtk::VTKHDF.VTKHDFFile` seems to be using internals in `VTKHDF`? 

Perhaps overload `Base.flush(vtk::VTKHDFFile) = flush(vtk.file)` then to match the `close` overload?

KristofferC: > To me it just feels like there should be some API for this in VTKHDF, since file = vtk.file for vtk::VTKHDF.VTKHDFFile seems to be using internals in VTKHDF?

Yes, this I agree with. Will look at making that better.

KristofferC: https://github.com/Ferrite-FEM/VTKHDF.jl/pull/3 and now using flush

KristofferC:
VTKHDF is now registered. Rebased this.

ext/FerriteVTKHDF.jl
```diff
@@ -0,0 +1,133 @@
...
+function VTKHDF.write_timestep(f::Function, vtk::VTKHDFGridFile, t::Real; kwargs...)
+    VTKHDF.write_timestep(_ -> f(vtk), hdf(vtk), t; kwargs...)
+    # flush each step so a hard kill (OOM, scheduler kill, segfault) still
+    # leaves a valid file with all completed steps
+    flush(hdf(vtk))
+    return vtk
+end
+
+# ---- node data conversion (mirrors Ferrite's WriteVTK path, minus component
+# names, which the VTKHDF format does not support) ----
```
fredrikekre:
```suggestion
# Node data conversion mirrors the WriteVTK path, minus component names,
# which the VTKHDF format does not support.
```

fredrikekre (approved):

PR #1382, opened 2026-07-21
Generate figures from output VTK files from examples

KristofferC:
Also uses animated webps for cases where we time step. The figures are pushed to the gh-pages branch to avoid bloating the branch

PR #1388, opened 2026-07-22
Add Darcy flow tutorial with H(div) Raviart-Thomas elements 

KristofferC:
Made on top of https://github.com/Ferrite-FEM/Ferrite.jl/pull/1382 to get nice figures, etc. easily, so only look at the last commit.

Iterated back and forth with Claude on this today. We didn't have anything using the RT elements, and this felt like the most simple way to show them off.


KristofferC:
Rendered tutorial https://ferrite-fem.github.io/Ferrite.jl/previews/PR1388/tutorials/darcy_flow/

docs/src/literate-tutorials/darcy_flow.jl
```diff
@@ -0,0 +1,529 @@
...
+# where local conservation comes from: since ``\delta p`` may be chosen as the indicator
+# function of any single element ``K`` (the pressure space is discontinuous!), the
+# discrete solution satisfies ``\int_K \mathrm{div}(\boldsymbol{q}_h)\, \mathrm{d}\Omega = 0``
+# for *every element individually*. With a continuous pressure space this argument is not
+# available, and indeed the standard method violates local conservation.
+#
+# ## Implementation
+#
+# We start by loading the packages, and generating a triangulated unit square.
+using Ferrite, SparseArrays, LinearAlgebra
```
KnutAM, line 144:
```suggestion
using Ferrite
```
Doesn't `Ferrite` suffice? At least doesn't look like `SparseArrays` is used...

KnutAM:
Just skimming through, but this seems like a very nice way to show the H(div) functionality. Better option than #1161, but I should also finish that at some point :D

KnutAM:
https://github.com/Ferrite-FEM/Ferrite.jl/discussions/1402#discussioncomment-17796243

I think this fully supersedes https://ferrite-fem.github.io/Ferrite.jl/previews/PR798/tutorials/heat_equation_hdiv/, but perhaps (if not done already), ask Fable to check if some concepts introduced there are missing in this tutorial? I don't think we need this for both heat and flow problems since they are basically equivalent

KristofferC:
> but perhaps (if not done already), ask Fable to check if some concepts introduced there are missing in this tutorial? 

🤖 

Four things were genuinely missing:

1. What an H(div) dof actually is. #798 had an unfinished block trying to explain this. I wrote it out: the dofs are moments of the normal flux over a facet, not point values. This single fact explains both why the normal component is continuous (neighbors share the facet dof) and why ProjectedDirichlet is needed (no nodal support points) â previously those two were stated separately without the underlying reason.
2. The general stable pairing for arbitrary order: RaviartThomas{RefTriangle, k} or BrezziDouglasMarini{RefTriangle, k} with DiscontinuousLagrange{RefTriangle, k-1}.
3. The trade-off. #798 was explicit that the mixed method gives a worse primary field. Our tutorial only argued for the mixed method; it now also says what it costs â bigger, indefinite saddle-point system and an element-wise constant pressure against the primal method's continuous linear one.
4. Gatica2014 as an accessible reference alongside BoffiâBrezziâFortin.

KnutAM:
> Four things were genuinely missing:

Makes me feel genuinely useful (still, until the next model comes out) :D

KristofferC:
I read through this a few times, and I think it is quite good now.

PR #1389, opened 2026-07-22
Add high-order `:polyquad` quadrature rules for tetrahedra

KristofferC:
Add the symmetric quadrature rules of Witherden & Vincent (2015) for `RefTetrahedron`, supporting orders 1 to 10, as quadrature rule type `:polyquad`.

Prompted by the review in #1343: tetrahedral quadrature currently tops out at order 5 (Keast), which is insufficient for the higher-order interpolations that #1343 enables — e.g. exact stiffness integration for `Lagrange{RefTetrahedron, 4}` needs order 6, and mass matrices need order 8.

Some notes:

* This is the same rule family (and citation, [WitVin:2015:isq]) already used for `RefPrism` and `RefPyramid` under the `:polyquad` name. All weights are positive and all points lie strictly inside the reference tetrahedron.
* The point/weight data is taken from the rule files distributed with [PyFR](https://github.com/PyFR/PyFR/tree/master/pyfr/quadrules/tet) (generated with the polyquad tool accompanying the paper, ~38 significant digits), defined on the [-1,1] tetrahedron and transformed to the unit tetrahedron in the same way as the existing prism table.
* The paper provides no distinct degree-4 tet rule, so order 4 uses the 14-point degree-5 rule (deal.II does the same in `QWitherdenVincent`).
* The default rule for `RefTetrahedron` remains `:keast_minimal`; `:polyquad` is opt-in, mirroring how `:gaussjacobi` extends `:dunavant` for triangles. The Keast error message now points to `:polyquad` for orders > 5.

Tests verify polynomial exactness for all monomials `x^a y^b z^c` with `a+b+c <= order` against the analytic integral `a!b!c!/(a+b+c+3)!`, as well as positivity of the weights. I have additionally verified (outside the test suite) that each rule fails one degree above its stated order, i.e. the degrees are genuinely as claimed.


termi-official:
Needs a rebase.

termi-official (approved):
LGTM. Thanks! Also, that polynomial expression on the tet is really helpful to have for future integration rules of higher order.

PR #1391, opened 2026-07-23
Add C1-continuous cubic Hermite interpolation Hermite{RefLine, 3}

KristofferC:
Adds a cubic Hermite interpolation on `RefLine` with two dofs per vertex (value + derivative), making fourth-order problems (Euler-Bernoulli beams) solvable. A follow-up PR (#1392) adds a beam tutorial using it.

**Stacked on #1405**, which contains the library fixes that were originally part of this PR (`evaluate_at_grid_nodes` reinit fix, `apply_analytical!`/nodeset-`Dirichlet`/`PeriodicDirichlet` guards, `function_hessian` export + hessian docs). This PR retargets to `master` automatically when #1405 is merged.

## Design
- The derivative dofs are **physical** derivatives `du/dx`: a new `HermiteMapping` scales the corresponding basis functions by the geometric Jacobian, so the shared vertex dofs are consistent between neighboring cells of different length (the sign of J handles orientation). Restricted to affine 1D geometry, enforced at `FunctionValues` construction.
- `Dirichlet` gets a `kind::Symbol = :value` keyword, backed by a new `dof_kinds(::Interpolation)` trait (all `:value` by default; Hermite tags its derivative dofs `:derivative`). The default constrains only the value dofs; `kind = :derivative` prescribes the derivative dofs, with `f` returning the prescribed `dw/dx` (well-defined since the dofs are physical derivatives). A clamped end is thus two `Dirichlet`s on the same set, and derivative BCs are `update!`-able in time like any other. This replaces the earlier master-less-`AffineConstraint` workaround discussed in the review.
- Unsupported paths error explicitly instead of silently corrupting: vectorization (`Hermite^n`), nodeset `Dirichlet`, `PeriodicDirichlet`, `apply_analytical!` (the guards for the latter three live in #1405 and apply to all non-identity mappings).
- `reinit_needs_cell` is now a `mapping_needs_cell` trait (resolves the TODO in common_values.jl); Hermite does not need the cell, so cell-less `reinit!` and `PointValues` work.

## Tests
`test/test_interpolations_hermite.jl`: generalized Kronecker duality, cubic reproduction, mapping exactness on a nonuniform mesh (catches any missed J-scaling), C1 continuity, dof distribution/sharing, BC behavior, guard errors, Float32/inference, FacetValues/MultiFieldCellValues/PointEvalHandler/VTK integration, and a cantilever solved against the analytic solution (nodally exact) on uniform and nonuniform meshes.

### DIFF
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1,6 +1,6 @@
 # abstract type Constraint end
 """
-    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing)
+    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing; kind::Symbol = :value)
 
 Create a Dirichlet boundary condition on `u` on the `∂Ω` part of
 the boundary. `f` is a function of the form `f(x)` or `f(x, t)`
@@ -8,6 +8,18 @@ where `x` is the spatial coordinate and `t` is the current time,
 and returns the prescribed value. `components` specify the components
 of `u` that are prescribed by this condition. By default all components
 of `u` are prescribed.
+
+`kind` selects which kind of dofs the condition constrains, for interpolations that have
+more than one kind (cf. [`Ferrite.dof_kinds`](@ref)). The default, `:value`, constrains the
+dofs that represent the value of the function on the boundary. For example, for a field
+approximated with [`Hermite`](@ref) interpolation, `kind = :derivative` instead constrains
+the derivative dofs, and `f` then returns the prescribed derivative, e.g. for a clamped
+beam end:
+
+```julia
+add!(ch, Dirichlet(:w, ∂Ω, Returns(0.0)))                     # w = 0
+add!(ch, Dirichlet(:w, ∂Ω, Returns(0.0); kind = :derivative)) # dw/dx = 0
+```
 
 The set, `∂Ω`, can be an `AbstractSet` or `AbstractVector` with elements of
 type [`FacetIndex`](@ref), [`FaceIndex`](@ref), [`EdgeIndex`](@ref), [`VertexIndex`](@ref),
@@ -43,11 +55,12 @@ struct Dirichlet # <: Constraint
     facets::OrderedSet{T} where {T <: Union{Int, FacetIndex, FaceIndex, EdgeIndex, VertexIndex}}
     field_name::Symbol
     components::Vector{Int} # components of the field
+    kind::Symbol # kind of dofs to constrain (cf. dof_kinds), :value by default
     local_facet_dofs::Vector{Int}
     local_facet_dofs_offset::Vector{Int}
 end
-function Dirichlet(field_name::Symbol, facets::AbstractVecOrSet, f::Function, components = nothing)
-    return Dirichlet(f, convert_to_orderedset(facets), field_name, __to_components(components), Int[], Int[])
+function Dirichlet(field_name::Symbol, facets::AbstractVecOrSet, f::Function, components = nothing; kind::Symbol = :value)
+    return Dirichlet(f, convert_to_orderedset(facets), field_name, __to_components(components), kind, Int[], Int[])
 end
 
 # components=nothing is default and means that all components should be constrained
@@ -251,6 +264,7 @@ function Base.show(io::IO, ::MIME"text/plain", ch::ConstraintHandler)
         print(io, "  BCs:")
         for dbc in ch.dbcs
             print(io, "\n    ", "Field: ", dbc.field_name, ", ", "Components: ", dbc.components)
+            dbc.kind === :value || print(io, ", Kind: :", dbc.kind)
         end
     end
     return
@@ -378,7 +392,7 @@ end
 # Dirichlet on (facet|face|edge|vertex)set
 function _add!(ch::ConstraintHandler, dbc::Dirichlet, bcfacets::AbstractVecOrSet{Index}, interpolation::Interpolation, field_dim::Int, offset::Int, bcvalue::BCValues, _) where {Index <: BoundaryIndex}
     local_facet_dofs, local_facet_dofs_offset =
-        _local_facet_dofs_for_bc(interpolation, field_dim, dbc.components, offset, dirichlet_boundarydof_indices(eltype(bcfacets)))
+        _local_facet_dofs_for_bc(interpolation, field_dim, dbc.components, offset, dirichlet_boundarydof_indices(eltype(bcfacets)), dbc.kind)
     copy!(dbc.local_facet_dofs, local_facet_dofs)
     copy!(dbc.local_facet_dofs_offset, local_facet_dofs_offset)
 
@@ -403,14 +417,18 @@ end
 
 # Calculate which local dof index live on each facet:
 # facet `i` have dofs `local_facet_dofs[local_facet_dofs_offset[i]:local_facet_dofs_offset[i+1]-1]
-function _local_facet_dofs_for_bc(interpolation, field_dim, components, offset, boundaryfunc::F = dirichlet_facetdof_indices) where {F}
+function _local_facet_dofs_for_bc(interpolation, field_dim, components, offset, boundaryfunc::F = dirichlet_facetdof_indices, kind::Symbol = :value) where {F}
     @assert issorted(components)
+    kinds = dof_kinds(interpolation)
     local_facet_dofs = Int[]
     local_facet_dofs_offset = Int[1]
-    for (_, facet) in enumerate(boundaryfunc(interpolation))
-        for fdof in facet, d in 1:field_dim
-            if d in components
-                push!(local_facet_dofs, (fdof - 1) * field_dim + d + offset)
+    for facet in boundaryfunc(interpolation)
+        for fdof in facet
+            kinds[fdof] === kind || continue
+            for d in 1:field_dim
+                if d in components
+                    push!(local_facet_dofs, (fdof - 1) * field_dim + d + offset)
+                end
             end
         end
         push!(local_facet_dofs_offset, length(local_facet_dofs) + 1)
@@ -956,6 +974,9 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
             interpolation = interpolation.ip
         end
         getorder(interpolation) == 0 && error("No dof prescribed for order 0 interpolations")
+        if dbc.kind ∉ dof_kinds(interpolation)
+            error("interpolation $(interpolation) for field :$(dbc.field_name) has no dofs of kind :$(dbc.kind) (existing kinds: $(join(unique(dof_kinds(interpolation)), ", ")))")
+        end
         # Set up components to prescribe (empty input means prescribe all components)
         components = isempty(dbc.components) ? collect(Int, 1:n_comp) : dbc.components
         if !all(c -> 0 < c <= n_comp, components)
@@ -974,9 +995,9 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
             EntityType = FacetIndex
         end
         CT = getcelltype(sdh) # Same celltype enforced in SubDofHandler constructor
-        bcvalues = BCValues(interpolation, geometric_interpolation(CT), EntityType)
+        bcvalues = BCValues(interpolation, geometric_interpolation(CT), EntityType, dbc.kind)
         # Recreate the Dirichlet(...) struct with the filtered set and call internal add!
-        filtered_dbc = Dirichlet(dbc.field_name, filtered_set, dbc.f, components)
+        filtered_dbc = Dirichlet(dbc.field_name, filtered_set, dbc.f, components; kind = dbc.kind)
         _add!(
             ch, filtered_dbc, filtered_dbc.facets, interpolation, n_comp,
             field_offset(sdh, field_idx), bcvalues, sdh.cellset,
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -964,9 +964,10 @@ function add!(ch::ConstraintHandler, dbc::Dirichlet)
         # Create BCValues for coordinate evaluation at dof-locations
         EntityType = eltype(dbc.facets) # (Facet|Face|Edge|Vertex)Index
         if EntityType <: Integer
-            if !(mapping_type(interpolation) isa IdentityMapping)
+            if !physical_basis_is_reference_basis(interpolation)
                 # The node idx -> local dof idx assumption in _add! below does not hold
-                # when dofs are not nodal values (e.g. for H(div)/H(curl) interpolations)
+                # when dofs are not nodal values (e.g. for H(div)/H(curl) interpolations,
+                # or for Hermite with its derivative dofs)
                 throw(ArgumentError("Dirichlet conditions on a nodeset are not supported for $(interpolation), use a facetset or vertexset instead."))
             end
             # BCValues are just dummy for nodesets so set to FacetIndex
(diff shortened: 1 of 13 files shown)


fredrikekre:
I thought we had an old PR for this, but I guess that was only https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...hermite

src/Dofs/DofHandler.jl
```diff
@@ -1022,8 +1022,11 @@ function _evaluate_at_grid_nodes!(
         u::AbstractVector{T}, cv::CellValues, drange::UnitRange
     ) where {T}
     ue = zeros(T, length(drange))
+    # For identity mappings the physical shape values alias the reference values, but other
+    # mappings require reinit! to compute them
+    needs_reinit = !isa(mapping_type(function_interpolation(cv)), IdentityMapping)
     for cell in CellIterator(sdh)
-        # Note: We are only using the shape functions: no reinit!(cv, cell) necessary
+        needs_reinit && reinit!(cv, cell)
```
lijas, line 1029: `reinit`! already has "needs_reinit" logic. Should be enough to only call `reinit`!

src/FEValues/FunctionValues.jl
```diff
@@ -142,9 +143,29 @@ end
...
+    if !(sdim == 1 && ip_geo.ip isa Lagrange{RefLine, 1})
+        throw(
+            ArgumentError(
+                "$(ip) requires an affine geometry in 1D (ip_geo = Lagrange{RefLine, 1}()^1), got $(ip_geo). " *
+                    "Embedded (sdim > 1) or curved line elements are not supported."
+            )
+        )
+    end
+    return nothing
+end
```
lijas, line 186: I think it is quite easy to allow for non-linear geometries (Lagrnge{RefLine,2}) if we want. If I recall correctly, that would require computing the jacobian at the end points \xi = 1 and \xi -1 and use that for the scaling in apply_mapping, see https://arxiv.org/abs/1706.09017 .

src/interpolations.jl
```diff
@@ -1570,12 +1573,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+struct Hermite{shape, order} <: ScalarInterpolation{shape, order} end
+
+conformity(::Hermite) = H1Conformity() # C¹ ⊂ C⁰; finer classification is not needed
+mapping_type(::Hermite) = HermiteMapping()
+adjust_dofs_during_distribution(::Hermite) = false
+
+getnbasefunctions(::Hermite{RefLine, 3}) = 4
+vertexdof_indices(::Hermite{RefLine, 3}) = ((1, 2), (3, 4))
+# Only the value dofs (not the derivative dofs) are constrained by Dirichlet conditions
+dirichlet_vertexdof_indices(::Hermite{RefLine, 3}) = ((1,), (3,))
```
lijas: This is a nice "hack" to only lock the function value. But what if one wants to model a clamped beam, how do one constrain the function derivative?

lijas (commented):
Nice. As Fredrik said, I also though we had a Hermite interpolation PR open. Did github delete it :P ?

I left a few comments. Note that this PR overlaps a bit with my hermtian-type Argyris element implementation for triangles: #1270

src/interpolations.jl
KristofferC: Right now you use `AffineConstraint`, see the tutorial in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1392:

```
# The slope dof is instead prescribed directly with a master-less `AffineConstraint`.
# Local dof 2 is the slope at vertex 1 (cf. `vertexdof_indices` above), so the global dof
# number at the clamped end of the first cell is `celldofs(dh, 1)[2]`.
slope_dof = celldofs(dh, 1)[2]
add!(ch, AffineConstraint(slope_dof, Pair{Int, Float64}[], 0.0))
```

but I agree it is not very pretty. Any thoughts how to do this better?

KnutAM: Haven't looked into the details at all, but is it possible to have different fields for the rotation and translations? That would make this easy, but I guess that would violate the interface that each interpolation can be mapped separately?

lijas: Perhaps we can tag each dof,

doftypes(::Hermite{RefLine,3}) = (:value, :derivative, :value, :derivative)

And then as input argument for Dirichlet, you specify the dof type you want to constrain:

Dirichlet(:w, facetset, x -> 0.0, components, :derivative)

KnutAM: Interesting idea, would be interesting to see how that works out!

src/Dofs/DofHandler.jl
KristofferC: I pulled this out into https://github.com/Ferrite-FEM/Ferrite.jl/pull/1405. I do not think the comment is really accurate however, yes the `reinit!` would be cheap but not free for identity mapping args, or?

src/interpolations.jl
KristofferC: I was thinking about this a while ago and I think I had some very similar idea (which is promising) so I will test that and update the tutorial and see how it feels.

KristofferC: Commit up + change to EB-tutorial https://github.com/Ferrite-FEM/Ferrite.jl/pull/1392/changes/19a4601503bfb659492787035cc3dbf5833450f2.

lijas: Nice! Even though we dont go with this approach for applying BCs, it can still be nice to add dof tags for testing the kronecker delta property of the exotic interpolations like HDiv and HCurl (they have :integral_moments on edges I think). Your previous commit added this "@testset "generalized Kronecker property" begin" for the hermite element :)

KristofferC:
I think this looks pretty good. Perhaps one thing to discuss is the `kind` keyword for DirichletBC and corresponding

```
dof_kinds(::Hermite{RefLine, 3}) = (:value, :derivative, :value, :derivative)
```

src/Dofs/ConstraintHandler.jl
```diff
@@ -1,6 +1,6 @@
 # abstract type Constraint end
 """
-    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing)
+    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing; kind::Symbol = :value)
```
lijas, line 3: Should we call the argument dof_kind instead?

src/interpolations.jl
```diff
@@ -483,6 +485,17 @@ dirichlet_boundarydof_indices(::Type{EdgeIndex}) = dirichlet_edgedof_indices
...
+    dof_kinds(ip::Interpolation)
+
+A tuple of length [`getnbasefunctions(::Interpolation)`](@ref) with a `Symbol` classifying
+the kind of each local dof. By default all dofs are `:value` dofs, i.e. the dof value is
+the value of the approximated function at the dof location. Interpolations with other kinds
+of dofs overload this function, e.g. [`Hermite`](@ref) which also has `:derivative` dofs.
+The kind is used by [`Dirichlet`](@ref) to select which dofs a condition constrains.
+"""
+dof_kinds(ip::Interpolation) = ntuple(_ -> :value, getnbasefunctions(ip))
+
```
lijas, line 498: I think using symbols for the dof-kind is clean, but should we make the dof kinds types instead? Like ValueDof(), DerivativeDof(), DerivativeXDof() NormalGradientDof() etc? That would make it possible to dispatch on the DofType, which might be useful in some cases (for example when testing, see my other comment.)

src/interpolations.jl
```diff
@@ -1570,12 +1583,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+[`Dirichlet`](@ref) conditions on facet- or vertex-sets constrain the value dofs by
+default. To also clamp the derivative at a boundary, add a second `Dirichlet` with
+`kind = :derivative`:
+```julia
+add!(ch, Dirichlet(:w, ∂Ω, Returns(0.0)))                     # w = 0
+add!(ch, Dirichlet(:w, ∂Ω, Returns(0.0); kind = :derivative)) # dw/dx = 0
+```
+`Dirichlet` conditions on node-sets and `PeriodicDirichlet` are not supported.
+"""
```
lijas, line 1609: Should we minify this docstring to make it consistent with the other interpolations?

src/interpolations.jl
```diff
@@ -1570,12 +1583,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+    i == 3 && return (1 + x)^2 * (2 - x) / 4
+    i == 4 && return -(1 + x)^2 * (1 - x) / 4
+    throw(ArgumentError("no shape function $i for interpolation $ip"))
+end
+
+# Scaling of basis functions for derivative dofs, applied by the HermiteMapping such that
+# the dof value is the physical derivative du/dx (consistent between cells of different
+# length). The sign of J accounts for reversed cell orientation.
+@inline function hermite_derivative_scaling(::Hermite{RefLine, 3}, shape_nr::Int, J::Tensor{2, 1})
+    return (shape_nr == 2 || shape_nr == 4) ? J[1, 1] : one(J[1, 1])
```
lijas, line 1643: Move this function to FunctionValues.jl where it is used.

src/interpolations.jl
```diff
@@ -1570,12 +1583,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+
 ##################################################
 # VectorizedInterpolation{<:ScalarInterpolation} #
 ##################################################
 struct VectorizedInterpolation{vdim, refshape, order, SI <: ScalarInterpolation{refshape, order}} <: VectorInterpolation{vdim, refshape, order}
     ip::SI
     function VectorizedInterpolation{vdim}(ip::SI) where {vdim, refshape, order, SI <: ScalarInterpolation{refshape, order}}
+        # The vectorized interpolation uses the identity mapping, so interpolations with a
+        # non-trivial mapping (e.g. Hermite) cannot be vectorized.
+        mapping_type(ip) isa IdentityMapping || throw(ArgumentError("$(ip) cannot be vectorized"))
```
lijas: Note that in #1270 i changed the definition of mapping_type(ip::VectorizedInterpolation) to  mapping_type(ip::VectorizedInterpolation) = mapping_type(ip.ip) which i think is correct (why would a vectorizedInterpolation not inherent the mapping type of the scalar interpolation? 

However, Hermite{RefLine,3} is a 1D element, so I don't think it makes much sense to Vectorize it anyways.

test/test_interpolations_hermite.jl
```diff
@@ -0,0 +1,246 @@
+# Imports for parallel (isolated) test execution:
+using Test, LinearAlgebra
+include(joinpath(@__DIR__, "test_utils.jl"))
+include(joinpath(@__DIR__, "interpolation_test_utils.jl"))
+
+using Ferrite: reference_shape_value, reference_shape_gradient, OrderedSet
+
+@testset "Hermite{RefLine, 3}" begin
```
lijas, line 8: I think we should not create a separate test file for the Hermite element, but try to incorporate it in to the existing test suite instead (e.g. test_interpolations.jl). If there are some unique tests that the only the Hermite element needs, they could go in its own @testset in the same file.

Can you try to add Hermite{RefLine.3} to all the other interpolations here? 
https://github.com/Ferrite-FEM/Ferrite.jl/blob/9e6a48561ba730287524cf707815d63a063a6613/test/test_interpolations.jl#L10
From #1270 , I know it will not fit in perfectly since the Hermite dofs are a bit special and breaks some assumptions in the test. For example, the Kronecker delta test will not pass since it currently assumes all the dofs of the interpolations are :value dofs. Here, it could be nice to use the DofType-system i suggested earlier to dispatch on the correct evaluation function when testing the Kronecker delta property.  

Also, we can test for continuity for the Hdiv and Hcurl elements in the test_continuity.jl file. Can you add the hermite interpolation there as well (again you can look at #1270 for inspiration)?

src/interpolations.jl
```diff
@@ -1570,12 +1583,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+`Dirichlet` conditions on node-sets and `PeriodicDirichlet` are not supported.
+"""
+struct Hermite{shape, order} <: ScalarInterpolation{shape, order} end
+
+conformity(::Hermite) = H1Conformity() # C¹ ⊂ C⁰; finer classification is not needed
+mapping_type(::Hermite) = HermiteMapping()
+adjust_dofs_during_distribution(::Hermite) = false
+
+getnbasefunctions(::Hermite{RefLine, 3}) = 4
+vertexdof_indices(::Hermite{RefLine, 3}) = ((1, 2), (3, 4))
```
lijas, line 1618: I think it is amazing that vertexdof_indices could just take multiple dofs per vertex without any other change needed in the code :)

lijas (commented):
I think it looks good. My biggest request is to refactor the tests a bit.

fredrikekre:
> I think this looks pretty good. Perhaps one thing to discuss is the kind keyword for DirichletBC and corresponding

What about just thinking of the two dofs as components and reuse the component functionality for Dirichlet?

lijas:
> What about just thinking of the two dofs as components and reuse the component functionality for Dirichlet?

How do you mean? Should components=[2] lock only derivative dofs? That will not work in general, because in the argyris you have normal gradient dofs on edges, so i dont think there is a good way to number the dofs Kinds. It also possible to vectorize the interpolation, and what should components keyword represent in that case?

src/Dofs/ConstraintHandler.jl
KristofferC: Probably better name, yes

src/interpolations.jl
KristofferC: Maybe, I am just a bit worried about putting multiple different types into a container like that (and then we loop over it and maybe get some type instabilities etc). But maybe it is fine and it will union split ok etc.

src/interpolations.jl
KristofferC: Perhaps, or maybe the other docstrings should be expanded :P. But yeah, some trimming here could be done.

src/interpolations.jl
```diff
@@ -1570,12 +1583,74 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
+dof_kinds(::Hermite{RefLine, 3}) = (:value, :derivative, :value, :derivative)
```
KristofferC, line 1619: I wonder if one could get away with only specifying this as `(:value, :derivative)` and have it be implicit that it repeats for every node sort of.

lijas: No, not it #1270 :)
See https://defelement.org/elements/examples/triangle-argyris-5.html

The three last dofs are on the edge so it does follow a repeated pattern.

termi-official:
Can you run another check if this is sufficient for dim>1 (e.g. for Bogner-Fox-Schmidt elements, xref https://github.com/Ferrite-FEM/FerriteInterpolations.jl/issues/12)?

KristofferC:
Yeah, this need some cleanup etc, might not get to it until in a bit more than a week.

KristofferC:
Here are two elements made on top of this. https://github.com/Ferrite-FEM/FerriteInterpolations.jl/commit/bb3d7a7483eaf647ed285f2ffba49b077faf73ec. However, this PR currently breaks another elements:

> he branch's new `VectorizedInterpolation` constructor guard (`physical_basis_is_reference_basis(ip) || throw`) breaks FerriteInterpolations' existing vector dPc — `DPC^vdim` *deliberately* vectorizes the L2-Piola-mapped scalar with the identity mapping (DefElement declares vector dPc that way), and master has no such guard. The guard can't be worked around downstream (overriding `physical_basis_is_reference_basis` for DPC would corrupt the Nx/Nξ aliasing), so it needs its own opt-out trait before the PR merges.

PR #1393, opened 2026-07-23
Fix ProjectedDirichlet dof lookup for non-first fields and make its projection kernel allocation-free

KristofferC:
Two changes to `ProjectedDirichlet`, split into separate commits:

## Bug fix

`_update_projected_dbc!` looked up global dofs as `celldofs(fc)[shape_nr]`, where `shape_nr` indexes the constrained field's interpolation *without* the field offset in the cell dof vector. For a `DofHandler` where the projected field is not the first field, `update!` therefore resolved dofs belonging to other fields (typically erroring with a `KeyError` in the `dofmapping` lookup). The fix uses the `facet_dofs` views — which are built with `field_offset` at `add!` time and were already passed in, but unused — for the global dof lookup. Includes a regression test with the projected field as the second field, for both H(div) and H(curl).

## Performance

The per-facet projection solve is now allocation-free:

- The small SPD facet systems are solved in closed form for the common 1x1 and 2x2 cases (2x2 via unrolled Cholesky, which unlike Cramer's rule does not form products of matrix entries and is therefore scale-safe; covered by unit tests at entry scales where such products would overflow/underflow), and with in-place Cholesky otherwise, instead of allocating an LU factorization per facet.
- The methods that only pass the boundary-condition function through now force specialization on it (`bc_fun::F ... where F`). Julia does not specialize on `Function`-typed arguments that are never called in the method body, which made the inner kernel call a dynamic dispatch per facet, boxing the isbits `shape_nrs` tuple argument.

Note on failure modes: for degenerate input (near-singular or indefinite `K` from e.g. degenerate geometry), the Cholesky-based solves raise `PosDefException`/`DomainError` where pivoted LU might previously have returned a (garbage) result quietly. Valid SPD Gram matrices are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixes
  - `FacetIterator` now works with `AbstractVector{FacetIndex}` and `AbstractSet{FacetIndex}` inputs as documented, instead of requiring an `OrderedSet` ([#1384])
+ - `ProjectedDirichlet` now updates the correct dofs when the constrained field is not the first
+   field in the `DofHandler`; previously the dof lookup ignored the field offset in the cell dof
+   vector (typically causing a `KeyError` during `update!`) ([#1393])
 
 ## [v1.5.0] - 2026-07-13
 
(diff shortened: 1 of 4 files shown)


KristofferC:
Checked benchmarks locally, regressions are false and improvement is real :P

PR #1394, opened 2026-07-23
Add project_analytical! for H(div) and H(curl) interpolations

KristofferC:
Adds a new function `project_analytical!`, the projection-based counterpart to `apply_analytical!` for interpolations whose dofs are not nodal function values — currently `RaviartThomas`, `BrezziDouglasMarini`, and (2D) `Nedelec`. Following [review feedback](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1394#issuecomment-5219788528), this is a separate function rather than an extension of `apply_analytical!` (analogous to the `Dirichlet` vs `ProjectedDirichlet` split), which keeps `apply_analytical!` with pure nodal-interpolation semantics and leaves room to later support H1 interpolations in `project_analytical!` as well.

Dof values are determined by L2 projection of the analytical function's normal/tangential trace on each facet — the same projection `ProjectedDirichlet` performs, reusing its kernel — followed by a cell-local L2 minimization for interior dofs with the facet dofs fixed. All projections are cell-local; no global system is solved (unlike `L2Projector`).

Properties (all tested):
- Reproduces exactly any function in the global FE space (verified for all supported interpolations on affinely distorted grids).
- Shared facet dofs receive the same value from both neighboring cells (verified on non-affine perturbed grids and curved quadratic geometry, by applying per-cell and comparing).
- On affine facets the facet dofs coincide with the classical moment functionals; no commuting-diagram/canonical-interpolant claims are made beyond that.
- Matches `ProjectedDirichlet` values exactly on boundary facets.

A keyword `qr_order` allows overriding the default quadrature order (`2 * ip_order`). Work arrays respect `eltype(a)` (Float32 tested).

`apply_analytical!` now throws an informative `ArgumentError` pointing to `project_analytical!` for H(div)/H(curl) interpolations, and `project_analytical!` likewise refers back to `apply_analytical!` for nodal interpolations. 3D H(curl) interpolations are not supported, since their edge dofs require line integrals along edges and Ferrite has no `EdgeValues` machinery (the same limitation as `ProjectedDirichlet`); they throw an informative `ArgumentError`.

Performance on an 80k-cell RT2 triangle grid: 74.7 ms / 138 total allocations (initial naive implementation: 114.6 ms / 2.16M allocations).

### DIFF
--- a/src/Dofs/apply_analytical.jl
+++ b/src/Dofs/apply_analytical.jl
@@ -4,76 +4,46 @@ function _geometric_interpolations(dh::DofHandler)
     return ntuple(i -> geometric_interpolation(getcelltype(i)), length(sdhs))
 end
 
-# Error for unsupported interpolations before any dof value has been assigned, so that
-# `a` is never partially modified. See the `apply_analytical!` docstring for which
-# interpolations are supported and how they are handled.
-function _check_apply_analytical_supported(ip_fun::Interpolation{RefShape}) where {dim, RefShape <: AbstractRefShape{dim}}
+# Which interpolations `apply_analytical!` is implemented for; used to error before any
+# dof value has been assigned so that `a` is never partially modified. `apply_analytical!`
+# requires the dof values to be function values at the reference coordinates (so a
+# `reference_coordinates` method must exist); `project_analytical!` covers H(div) and
+# 2D H(curl) interpolations instead.
+function _check_apply_analytical_supported(ip_fun::Interpolation)
     applicable(reference_coordinates, ip_fun) && return nothing
-    cf = conformity(ip_fun)
-    if cf isa HcurlConformity && dim == 3
-        throw(ArgumentError("apply_analytical! is not implemented for 3D H(curl) interpolations ($ip_fun), whose edge dofs require line integrals along edges"))
-    elseif !(cf isa Union{HdivConformity, HcurlConformity})
-        error("apply_analytical! is not implemented for $(ip_fun).")
+    if conformity(ip_fun) isa Union{HdivConformity, HcurlConformity}
+        throw(ArgumentError("apply_analytical! is not applicable for $(ip_fun), whose dofs are not nodal function values. Use project_analytical! instead."))
     end
-    return nothing
+    error("apply_analytical! is not implemented for $(ip_fun).")
 end
 
 """
     apply_analytical!(
         a::AbstractVector, dh::AbstractDofHandler, fieldname::Symbol,
-        f::Function, cellset = 1:getncells(get_grid(dh));
-        qr_order = -1
+        f::Function, cellset = 1:getncells(get_grid(dh))
     )
 
 Apply a solution `f(x)` by modifying the values in the degree of freedom vector `a`
 pertaining to the field `fieldname` for all cells in `cellset`.
-The function `f(x)` is given the spatial coordinate `x`. For scalar fields,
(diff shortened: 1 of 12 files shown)


src/Dofs/apply_analytical.jl
```diff
@@ -88,3 +146,99 @@ function _apply_analytical!(a::AbstractVector, dofs::Vector{Int}, coords::Vector
...
+        celldofs!(c_dofs, dh, cellnr)
+        for facetnr in 1:nfacets(cell)
+            shape_nrs = facetdof_indices[facetnr]
+            isempty(shape_nrs) && continue
+            reinit!(fv, cell, coords, facetnr)
+            solve_projected_dbc!(aᶠ, Kᶠ, fᶠ, g, fv, shape_nrs, coords, 0)
+            for (i, shape_nr) in pairs(shape_nrs)
+                a[c_dofs[celldofinds[shape_nr]]] = aᶠ[i]
+            end
+        end
```
termi-official, line 261: Can we add some reference for this and is it valid for higher order Hdiv/Hcurl ansatz functions, too? I think if you increase the order to at least two you get bubble functions which vanish at the boundary. See e.g. https://link.springer.com/content/pdf/10.1007/978-981-96-0100-4_12.pdf?pdf=inline%20link .

KristofferC: It should be valid, it is already tested (`RaviartThomas{RefTriangle, 2}` and `Nedelec{RefTriangle, 2}`)

termi-official: Just for the public record (as we discussed a bit on slack), I think the tests should catch a potential failure if we add higher order Hdiv/Hcurl . The confusion here comes from the order argument in RT and Nedelec not being the order of the element.

src/Dofs/apply_analytical.jl
```diff
@@ -4,36 +4,91 @@ function _geometric_interpolations(dh::DofHandler)
     return ntuple(i -> geometric_interpolation(getcelltype(i)), length(sdhs))
 end
 
+# Which interpolations `apply_analytical!` is implemented for; used to error before any
+# dof value has been assigned so that `a` is never partially modified. The nodal path
+# requires the dof values to be function values at the reference coordinates (so a
+# `reference_coordinates` method must exist), and H(div) and 2D H(curl) interpolations
+# use the facet-trace projection path instead.
```
termi-official: This comment is now a wild mix of process documentation and outdated information. Needs an update to document the final result or should point to a central doc string to avoid synchronizing docs info for different places.

termi-official (commented):
LGTM!

termi-official (approved):

KnutAM:
Late to the party but shouldn’t this be a different function? `project_analytical`. The same logic could also be applied to H1 interpolations - the same way ProjectedDirichlet can (which is why we decided to not make that part of `Dirichlet`

KristofferC:
I think that is a fair argument. Although it is a bit confusing with our global L2 projection stuff, (`project_analytical` sounds a bit like a global L2 projection).

KristofferC:
I changed it to `project_analytical!`. I wonder if it should be called `projected_apply_analytical!` instead (to be somewhat consistent with `[Projected]Dirichlet`)

KnutAM:
An alternative could be to use apply_analytical with a ”kind” keyword or type that says L2Project (global, codim 0 average), LocalL2Project (this PR) LocalSupport (nodal / precalculated weights for non-nodal) (with the latter default and error if not implemented)

PR #1395, opened 2026-07-24
Add global fields via the GlobalConstant interpolation

KristofferC:
This adds *global fields* to Ferrite: fields whose degrees of freedom are shared between **all** cells of the `SubDofHandler` they are added to, declared with the new interpolation `GlobalConstant{refshape}` (vector valued via e.g. `GlobalConstant{refshape}()^3`, constant shape function `N = 1`). Typical use cases are Lagrange multipliers for integral constraints (e.g. mean value constraints) and macroscopic unknowns in computational homogenization (e.g. prescribed average stress). Prior art: #1009, #596 (and KnutAM's `kam/GlobalDofs` POC).

### Design

- The dofs are created once and appended to `celldofs` of every cell in the `SubDofHandler`, so sparsity patterns (including coupling and constraint elimination), assembly, `dof_range`, `CellValues`, evaluation, and export all work like for any other field. Adding a field with the same name to several `SubDofHandler`s shares the dofs.
- New accessor `global_field_dofs(dh, name)` returns the dof numbers of a global field (kept consistent by `renumber!`, though previously returned vectors are not updated).
- Sparsity pattern rows for global dofs are built with a buffered sort/unique bulk insertion, since per-entry sorted insertion is quadratic in the number of cells for these rows. To support such rows, `PoolAllocator` now serves allocations larger than the fixed 4 MiB page size from dedicated pages instead of erroring.
- The dof-order validation in `close!` classifies every base function index exactly once as vertex, edge, face, volume, or global.

### Guard rails

- `Dirichlet`, `PeriodicDirichlet`, `ProjectedDirichlet`, and `apply_analytical!` reject global fields with descriptive errors (constrain the dofs directly with `AffineConstraint` instead).
- Colored threaded assembly is not safe for cells carrying a global field (no valid coloring exists); this is documented together with a safe task-local reduction recipe, which is also tested.


lijas:
It seems a bit overkill to add the global dof too all cells in the SubDofHandler (i.e. the celldof-array from `celldofs(dh, cellid)`)? I much more prefer Knut's approach in #1012 where the global dofs are just appended at the end of the dof distribution in close!. The drawback is that the user has to handle the sparsity pattern/coupling themselves, but we can add utility functions for that.

termi-official:
I also though that the idea of adding these dofs to the celldofs has been rejected here https://github.com/Ferrite-FEM/Ferrite.jl/pull/1012#discussion_r1663744873 . Mean value constraints can already be added via AffineConstraints. So I guess the really interesting missing case to cover is if we have nonlinear constraints, which typically emerge only on subsets of the elements. I think for a mechanics framework the most pressing missing feature is contact infrastructure. However, it seems to be a bit cumbersome to handle with this approach. I also think that we can handle the case you outline here by simply adding special elements and special interpolations on these elements handling the constraint, right? Not the biggest fan of this approach, but it works.

KristofferC:
> The drawback is that the user has to handle the sparsity pattern/coupling themselves, but we can add utility functions for that.

FWIW, I thought this approach was very elegant in that it is just another field and more or less everything works the same. Not requiring the user to do their own work etc is kind of the point of a FE framework imo.

KnutAM:
I think both approaches could be added, they are orthogonal. 

1) This PR (aka https://github.com/Ferrite-FEM/Ferrite.jl/compare/master...kam/GlobalDofs) 
   Provides ways to connect a dof that has the same value in all cells. This is useful when doing e.g. dissipation constraint, when each cell should be connected to a certain global equation. Following Ferrite's structure, if this is only used for a subset, you should define a subdofhandler for that subset. If used only for the surface, the same way as with all in Ferrite, you then still integrate the full cell with facetvalues, and in which case here you should still have a subdofhandler for those cells.
   These dofs allow you to add equations that have integral contributions.
   
2) #1012: 
   Provides a way to add additional equations and dofs that are not connected to other cells directly, but that can be included in the global equation system via discrete equations (e.g. affine constraints). This is useful for adding e.g. reference-point like equations (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1119 offers an alternative to this), but by not connecting to a physical cell, this doesn't require to change the grid in order to add additional constraints and equations.
   These dofs allow you to add equations that have discrete contributions.

## Regarding naming

1) I don't think `GlobalConstant` is good, as it gives the idea that it could be temporally constant as well. `GlobalInterpolation`, `GlobalDof`, `GlobalCoefficient`, or `GlobalValue` is better IMO.

2) I don't think my name `GlobalDof` fits here, that is better for nr 1. Possibly `AuxillaryDof` is better for this (or `ReferenceDof` to mimic the common `ReferencePoint` used in commercial softwares, but not a huge fan)

KnutAM:
And note that when trying the approach in this PR I observed a significant slowdown in `assemble!`, this should be checked and at least documented if merged.

KristofferC:
Thanks for your comments @KnutAM! I agree that both features are useful (and orthogonal) and they should not really be conflated. I am away tenting ⛺ so cannot really write a full reply to everything here now but will do so later. For the stuff in #1396 the approach here feels really nice (with the exception that we do not have second order tensor fields yet so that part is a bit awkward). I also like that everything just "falls out" of this being "just a field" in how you define the coupling etc. But as I said, will adress specific points tomorrow evening.

lijas:
I think Knut brought up a good distinction between the two approaches. However, I still think #1012 is a better design. My reasoning is:

1. Boundary-only coupling still affects all cells. Suppose the global DOF is only coupled through a boundary integral (e.g. on a subset of faces). With the current approach, every cell in the corresponding SubDofHandler must nevertheless carry that global DOF in its celldofs vector, even though only the boundary cells actually interact with it.
2. Unnecessary enlargement of element operators. If the user also assembles a volume integral over the same SubDofHandler that does not involve the global DOF, the element still has to allocate an extra row and column in its local stiffness matrix (and residual vector), because the DOF is part of celldofs. The additional entries will simply remain zero. This works, but it feels unnecessarily cumbersome, since the global DOF is not part of the volume formulation.

Knut also mentions some performence problem in the assembly loop. I *think* with an approach as in #1012, the performence loss would not be as big, as you would most naturally integrate and assemble the sub matrices on the diagonal(e.g. $`(K^{e}_{u\varepsilon})_{i\beta} = \int \varepsilon[N_i] : \mathbf{E} : \mathbf{E}_{\beta}\,\mathrm{d}\Omega`$ from #1396) in a seperate element-routine function. 

It feels like this PR works well in #1396, but we would need something like #1012 for a more general approach I think. But maybe both approaches can be included in Ferrite :)

KristofferC:
> I also though that the idea of adding these dofs to the celldofs has been rejected here https://github.com/Ferrite-FEM/Ferrite.jl/pull/1012#discussion_r1663744873 .

That's not how I interpreted the discussion there.

> Mean value constraints can already be added via AffineConstraints.

Well, in some cases. You may want to have a mean value on something that is not a dof, or you describe the mean-value on the dual (like in #1396). You may also want to actually get the value of the new dof (the "reaction force"). Also, if you have a volume based mean value you get quadratic fill-in (https://github.com/Ferrite-FEM/Ferrite.jl/blob/c5215f5742e866597f9058c1ef942ba664a569c8/src/Dofs/sparsity_pattern.jl#L568-L578). So it works on Stokes because of boundary mean value but not in general. And as you mention you can have non-linear constraints.

>  I think for a mechanics framework the most pressing missing feature is contact infrastructure. However, it seems to be a bit cumbersome to handle with this approach.

This is not meant for contact, you need more refined things for that.

KristofferC:
> Knut also mentions some performence problem in the assembly loop.

Assemble performance should be fixed I think (will look at it again though).

> since the global DOF is not part of the volume formulation.

Hm, is it not? Or in another question, how is this different from any other field in the same situation (when you have a mixed formulation).


> I much more prefer Knut's approach in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1012 where the global dofs are just appended at the end of the dof distribution in close!. The drawback is that the user has to handle the sparsity pattern/coupling themselves, but we can add utility functions for that.

Regarding #1012, the problem there is that it doesn't integrate at all with Ferrite. There is no cell coupling, no `dof_range`, no `renumber!` etc etc. The utility functions you mention almost feel like they are this PR :P.

----

One reason I want this feature is to reach feature "nirvana" for an old paper (https://onlinelibrary.wiley.com/doi/abs/10.1002/nme.6050) that I am updating to new Ferrite code. My goal is to not require any hacks or ugly stuff to implement the things in there. Part of that is global "constant" fields and tensorial base functions.

KnutAM:
> I think Knut brought up a good distinction between the two approaches. However, I still think https://github.com/Ferrite-FEM/Ferrite.jl/pull/1012 is a better design.
> It feels like this PR works well in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1396, but we would need something like https://github.com/Ferrite-FEM/Ferrite.jl/pull/1012 for a more general approach I think. But maybe both approaches can be included in Ferrite :)

Yes, they are different and orthogonal designs with strengths and weaknesses for different problems - I think we should have both. What this discussions highlights is that we need to be very clear in the documentation about this and what the intended use cases are.

lijas:
I thought of a cool application for global constant fields, and that is to model rigid bodies.

Normally, how I have modelled rigid bodies before is that I would replace all elements of the subdomain (that I want to be rigid), with a "Point Cell" to which I can add rigid body DOFs. The drawback with this is that you lose all geometric meaning of the rigid body, which is annoying when you want to, e.g., export it to VTK or model contact (because the rigid body is only stored as is only a point in the geometric sense, and you would need to store the geometry of the rigid body in some other way).

Instead of replacing the elements with a single point cell, you could instead add a constant displacement field and a constant angular velocity field to the subdomain (as implemented in this PR). In this way, you retain the geometry of the rigid body, which is nice if you want to to contact modelling or integrate traction forces on the surface, and more.

PR #1397, opened 2026-07-24
Rewrite `SparsityPattern` with contiguous buffer storage

fredrikekre:
# Rewrite `SparsityPattern` with contiguous buffer storage

## Summary

This PR rewrites the storage and construction of `SparsityPattern`, making pattern building
2.8–17x faster (depending on the arguments) at roughly half the peak memory, while keeping
the entire public interface unchanged and producing identical patterns. It also lets us
delete two internal components: `FastSparsityPattern` (its algorithm becomes the built-in
fast path) and the `PoolAllocator` module (no longer needed by the new storage).

## Motivation

The old `SparsityPattern` stored one `PoolAllocator.PoolVector` per row. The pool solved the
"millions of tiny allocations" problem, but the per-row metadata still embedded heap
references, and every entry was inserted one at a time through a sorted insert. Meanwhile
`FastSparsityPattern` (#1302) showed that an exact-counting marker algorithm is much faster —
but it was cell-only: no constraints, no coupling, no `add_entry!`, so we had to maintain two
types and route between them in `allocate_matrix`.

Benchmarking several candidate backings (contiguous buffer vs. pool, sorted-insert vs.
append-then-sort vs. count-then-fill, arithmetic vs. geometric growth) showed that one design
matches `FastSparsityPattern` on its own turf while staying fully dynamic. That design is now
the production `SparsityPattern`, and both `FastSparsityPattern` and `PoolAllocator` are
retired.

## Design

The pattern stores all rows in a single contiguous `CollectionsOfViews.ConstructionBuffer`:
one growable `data::Vector{Int}` plus an isbits `AdaptiveRange` per row. Construction builds
on three ideas:

1. **Fast path** — for any fresh pattern, `add_sparsity_entries!` counts the exact number of
   distinct columns per row with a column marker, presizes the buffer in one allocation, and
   fills each row with a marker-dedup append. This is `FastSparsityPattern`'s algorithm,
   generalized: non-full `coupling` (any form) and `keep_constrained = false` are applied as
   identical filters in the count and fill passes. For builds with `topology`, row space for
   the interface entries is reserved up front by also counting the dofs of facet-neighbor
   cells (filtered by `interface_coupling`), and when that enumeration is provably exact —
   single subdofhandler; any interpolation continuity and any mask, thanks to the
   self-sufficient `interface_coupling` semantics from #1432 — the interface entries are
   filled directly by the same pass. Only pre-populated patterns take
   the generic per-entry path (a planned follow-up moves those onto the fast path too);
   interface entries in the remaining cases and constraints layer on afterwards through
   per-entry insertion into reserved space.

2. **Lazy sorting** — the fast fill leaves rows in marker order; rows are sorted only when
   something needs order (`add_entry!`, `eachrow`, CSR instantiation). The
   `SparseMatrixCSC`/`Symmetric` instantiation does not: it reads rows through an internal
   order-free accessor and lets the CSC counting transpose order `rowval`, so the common
   pattern-to-matrix path never sorts at all.

3. **Growable everywhere** — entries added after the fast fill use a sorted dedup insert with
   geometric reservation growth (`CollectionsOfViews.insert_sorted_at_index!`, a new sibling
   of `push_at_index!`). A row that outgrows its reservation relocates to the end of the
   buffer, leaving a transient hole.

## Benchmarks

Pattern build time, master (including #1428 and #1432) vs this PR on the same machine (3D
hexahedron grids; CG = mixed vector/scalar Lagrange, 1.4M dofs; DG = `DiscontinuousLagrange`,
1.0–1.4M dofs; CG/mixed with `topology` at 0.2–0.5M dofs; identical `nnz` verified for every
case; ratios consistent from 80k dofs up):

| case | master | **PR** | speedup | peak pattern memory |
|---|---|---|---|---|
| cell-only | 7.4 s | **1.0 s** | 7.6x | 2.1 → 1.2 GiB |
| field coupling `[t t; f t]` | 5.9 s | **1.1 s** | 5.5x | 1.8 → 0.9 GiB |
| `keep_constrained = false` | 9.1 s | **2.4 s** | 3.8x | 2.1 → 1.1 GiB |
| periodic constraints | 8.5 s | **3.0 s** | 2.8x | 2.1 → 1.7 GiB |
| DG with `topology` | 11.2 s | **0.9 s** | 12.8x | 3.5 → 1.3 GiB |
| CG with `topology` | 9.7 s | **0.6 s** | 16.7x | 1.2 → 0.4 GiB |
| mixed DG, full `interface_coupling` | 17.8 s | **1.6 s** | 11.4x | 4.6 → 2.3 GiB |
| mixed DG, one-field `interface_coupling` | 2.4 s | **0.6 s** | 4.2x | 1.0 → 0.5 GiB |
| mixed DG, asymmetric `interface_coupling` | 14.5 s | **1.3 s** | 11.3x | 3.6 → 1.9 GiB |
| mixed CG/DG, one-field `interface_coupling` | 2.4 s | **0.5 s** | 5.1x | 0.8 → 0.4 GiB |
| mixed CG/DG, full `interface_coupling` | 11.2 s | **0.8 s** | 13.5x | 1.8 → 0.9 GiB |

With the self-sufficient `interface_coupling` semantics (#1432) the direct interface fill
covers every single-`SubDofHandler` discretization, so all `topology` builds above —
including continuous, mixed-continuity, and asymmetric-mask cases — take the same one-pass
fill as pure DG.

Matrix instantiation (`allocate_matrix` from a built pattern) is at parity (233 vs 253 ms for
a 28M-nnz matrix when measured per-repetition; the naive min-of-N numbers in the raw logs
misattribute one-time OS page-fault costs in master's favor, because master's build frees
several GiB that pre-fault the pages the matrix arrays then land on).

## Removals

- **`FastSparsityPattern`** (internal, #1302): `allocate_matrix(MatrixType, dh)` is now a
  single code path through `SparsityPattern` with no performance loss and full kwarg support.
  Its marker machinery survives as the fast-fill helpers.
- **`PoolAllocator`** (internal): unused after the rewrite, removed in a separate commit.

## Behavior

No changes to the public interface or to the resulting matrices: patterns produce identical
`SparseMatrixCSC`/`SparseMatrixCSR`/`Symmetric`/`BlockSparsityPattern` results across
cell-only, constraints, all coupling forms (field/component/template), and
`keep_constrained = false`, verified against the old implementation during development and by
the full test suite. One documentation-level clarification for custom `AbstractSparsityPattern`
implementations: the `eachrow` contract now states that column indices are iterated in sorted
order (all Ferrite patterns, and any reasonable implementation, already comply — the CSR
instantiation has always implicitly relied on it).


KristofferC:
The `sparsity-pattern/matrix-from-pattern/` regressions are AFAIU not "real", they are just moving work from one part to another in the pipeline (where the sorting occurs).

termi-official:
> The sparsity-pattern/matrix-from-pattern/ regressions are AFAIU not "real", they are just moving work from one part to another in the pipeline (where the sorting occurs).

So the baseline benchmarks which I provided as foundation are too fine grained for a sane regression testing pipeline? I am not sure yet if we can really improve the situation significantly tho.

KristofferC:
Well it depends what you want to measure. It could be interesting to measure this part of the pipeline. But you always have to be aware that you can trade performances between steps and what matters is the end result.

fredrikekre:
Updated benchmark table in the OP after the latest commits look promising.

KnutAM:
> So the baseline benchmarks which I provided as foundation are too fine grained for a sane regression testing pipeline? I am not sure yet if we can really improve the situation significantly tho.

Am I missing it, or don't we have `sparsity-pattern/matrix-from-dofhandler/SparseMatrixCSR`, which would validate

> The sparsity-pattern/matrix-from-pattern/ regressions are AFAIU not "real", they are just moving work from one part to another in the pipeline (where the sorting occurs).

KristofferC:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/1425 didn't add a `sparsity-pattern/matrix-from-dofhandler/SparseMatrixCSR`, right?

fredrikekre:
I believe the regressions in the benchmarks as of c1ab39d9a8f6571cd62d00e3c2a715e13643b9a3 are covered by the improvements for a full dofhandler -> matrix pipeline.

CHANGELOG.md
```diff
@@ -29,6 +29,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+   filling passes, and builds with `topology` (interface entries, e.g. for DG) use it for the
+   cell portion, with row space for the interface entries reserved up front (counted from the
+   facet-neighbor cells) so that layering them on top does not grow the buffer. For
+   discretizations with a single `SubDofHandler` (any interpolation continuity, any
+   `interface_coupling` mask) the interface entries are even filled directly by the same fast
+   pass (~25x faster interface builds). Constructing a
+   sparsity pattern and allocating a matrix from it is around 3x faster for typical problems
+   (up to ~10x with coupling/`keep_constrained`), with unchanged behavior across the full
+   interface. ([#1397])
+ - The internal `FastSparsityPattern` (from [#1302]) has been removed: its algorithm is now the
```
KristofferC: I don't think we have to put internal changes in the change log. Or at least put it in internal changes?

CHANGELOG.md
```diff
@@ -29,6 +29,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    assembly](https://ferrite-fem.github.io/Ferrite.jl/stable/howto/threaded_assembly/).
    ([#1417])
 
+### Performance
+ - `SparsityPattern` has been rewritten: all rows are now stored in a single contiguous buffer
```
KristofferC: Getting kind of long...

fredrikekre:
I think this can be reviewed now, just some comments to cleanup perhaps.

To summarize: This basically takes the algorithm from FastSparsityPattern (#1302 ) and moves it to the pattern creation stage and also improves it by adding support for `coupling = `, `interface_coupling = ` and `keep_constrained = false`. The OP is updated and should be current.

src/Dofs/sparsity_pattern.jl
```diff
@@ -787,17 +899,47 @@ function create_celldofs(dh::DofHandler)
         num = ndofs_per_cell(dh, cell_idx)
         num == 0 && continue
         r = n:(n + num - 1)
-        #celldofs!(view(cell_dofs, r), dh, cell_idx), but faster without view:
-        soffs = dh.cell_dofs_offset[cell_idx]
-        copyto!(cell_dofs, n, dh.cell_dofs, soffs, num)
```
KnutAM, line 800: Was there no longer a performance benefit of using `copyto!` instead of `celldofs!(view(...`?

src/Dofs/sparsity_pattern.jl
```diff
@@ -815,97 +960,138 @@ function create_row_to_cells(cell_dofs::ArrayOfVectorViews, sp)
...
+        rowlen::Vector{Int}, data::Union{Nothing, Vector{Int}}, indices::Union{Nothing, Vector{AdaptiveRange}},
+        marker::Vector{Int}, row_to_cells::ArrayOfVectorViews, row_to_localidx::Union{Nothing, ArrayOfVectorViews},
+        cell_dofs::ArrayOfVectorViews, cell_to_sdh::Vector{Int},
+        couplings::Union{Nothing, Vector{Matrix{Bool}}}, isconstrained::Union{Nothing, Vector{Bool}},
+        neighbor_cells::Union{Nothing, ArrayOfVectorViews} = nothing,
+        interface_couplings::Union{Nothing, Vector{Matrix{Bool}}} = nothing,
+    )
+    counting = data === nothing
+    ncols = length(marker) # marker has one slot per column
+    @inbounds for row in 1:length(rowlen)
```
KristofferC, line 1006: gpt is worried this can do an OOB access for oversized patterns (with `keep_constrained=false`) and wants something like:

```diff
-function _isconstrained_by_dof(ch::ConstraintHandler)
-    isconstrained = zeros(Bool, ndofs(ch.dh))
+function _isconstrained_by_dof(ch::ConstraintHandler, nrows::Int)
+    isconstrained = zeros(Bool, nrows)
    for d in keys(ch.dofmapping)
        isconstrained[d] = true
    end
    return isconstrained
end

And at the call site:

-isconstrained = keep_constrained ? nothing : _isconstrained_by_dof(ch)
+isconstrained = keep_constrained ? nothing :
+    _isconstrained_by_dof(ch, getnrows(sp))
```

KnutAM (commented):
Quick comment while looking through a while ago

src/Dofs/sparsity_pattern.jl
fredrikekre: It was slightlyg slower but in total just 0.4% of the total cost which is why I replaced it with the more readable form. This isn't a generic path though so we can use the old code.

PR #1405, opened 2026-07-27
Fixes and guards for interpolations with non-identity mappings

KristofferC:
Pulled out of #1391 since these fixes are independent of the new Hermite interpolation:

- **Fix `evaluate_at_grid_nodes`** (and thus `write_solution`) returning garbage (values from an undef buffer) for interpolations with non-identity mappings (e.g. `Nedelec`, `RaviartThomas`), by calling `reinit!` internally when the mapping requires it. Includes a regression test.
- **`apply_analytical!`** now errors up front for interpolations whose dofs are not nodal function values, instead of silently assigning `f(x)` to non-nodal dofs. The check runs before any mutation of `a`, so the vector is never partially modified.
- **`Dirichlet` on a nodeset** now throws a friendly `ArgumentError` for such interpolations (previously an obscure `MethodError` from `BCValues`/`reference_coordinates`), since the node-index → local-dof assumption does not hold for non-nodal dofs.
- **`PeriodicDirichlet`** now throws for such interpolations, since it would pair only the `dirichlet_facetdof_indices` dofs and lacks `mirror_local_dofs` methods.
- **Docs**: export `function_hessian` (`shape_hessian` already was), fix its docstring formatting, add the missing `shape_hessian` docstring, add both to the FEValues reference page, and add `AffineConstraint` to the boundary conditions reference page.

#1391 will be rebased on top of this.

### DIFF
--- a/src/Dofs/apply_analytical.jl
+++ b/src/Dofs/apply_analytical.jl
@@ -34,6 +34,18 @@ function apply_analytical!(
     fieldname ∉ getfieldnames(dh) && error("The fieldname $fieldname was not found in the dof handler")
     ip_geos = _geometric_interpolations(dh)
 
+    # Check all field interpolations before mutating `a` to avoid partial application
+    for sdh in dh.subdofhandlers
+        isnothing(_find_field(sdh, fieldname)) && continue
+        ip_fun = getfieldinterpolation(sdh, find_field(sdh, fieldname))
+        if !(mapping_type(ip_fun) isa IdentityMapping)
+            error(
+                "apply_analytical! is only supported for interpolations where the dof values " *
+                    "are function values at the nodes (identity mapping), got $(ip_fun)."
+            )
+        end
+    end
+
     for (sdh, ip_geo) in zip(dh.subdofhandlers, ip_geos)
         isnothing(_find_field(sdh, fieldname)) && continue
         field_idx = find_field(sdh, fieldname)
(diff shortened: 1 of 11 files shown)


src/Dofs/apply_analytical.jl
```diff
@@ -34,6 +34,18 @@ function apply_analytical!(
...
     ip_geos = _geometric_interpolations(dh)
 
+    # Check all field interpolations before mutating `a` to avoid partial application
+    for sdh in dh.subdofhandlers
+        isnothing(_find_field(sdh, fieldname)) && continue
+        ip_fun = getfieldinterpolation(sdh, find_field(sdh, fieldname))
+        if !(mapping_type(ip_fun) isa IdentityMapping)
+            error(
+                "apply_analytical! is only supported for interpolations where the dof values " *
+                    "are function values at the nodes (identity mapping), got $(ip_fun)."
```
termi-official, line 44: Good catch.

Should we frame this a bit differently, as I think new users might read this as "not possible"? We could either frame this as not implemented or give users a hint that they need to do a the L2 projection manually.

KristofferC: not implemented sounds good (but see https://github.com/Ferrite-FEM/Ferrite.jl/pull/1394)

termi-official: Oh, I see. Sorry, I am still catching up on my inbox.

Also note that the condition here is faulty. Being nodal or not is not a property of the mapping, but of the interpolation itself. The different mappings help in preserving the correct continuity. To make this clean should have an additional trait on the interpolation declaring whether it is nodal or integral. Probably the easiest way to check this for now in this PR is to check whether the `reference_coordinates` dispatch exists. I.e. we can have something like `is_nodal_interpolation(ip::Interpolation) = applicable(reference_coordinates, ip)`.

PR #1417, opened 2026-07-27
add atomic assembler

### LINKED ISSUE #1331: Atomic assembler

KristofferC:
inline LLVM needed due to https://github.com/JuliaLang/julia/issues/33325

I think there are tricks to make the result deterministic (I have some very vague memory about e.g. deal.ii docs discussing this) but this is a starting point.

Fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/1331.

### DIFF
--- a/src/assembler.jl
+++ b/src/assembler.jl
@@ -152,11 +152,16 @@ end
 
 Assembles the element residual `ge` into the global residual vector `g`.
 """
-@propagate_inbounds function assemble!(g::AbstractVector{T}, dofs::AbstractVector{Int}, ge::AbstractVector{T}) where {T}
+@propagate_inbounds function assemble!(g::AbstractVector{T}, dofs::AbstractVector{Int}, ge::AbstractVector{T}, ::Val{atomic} = Val(false)) where {T, atomic}
     @boundscheck checkbounds(g, dofs)
     @boundscheck checkbounds(ge, keys(dofs))
     @inbounds for (i, dof) in pairs(dofs)
-        addindex!(g, ge[i], dof)
+        if atomic
+            v = ge[i]
+            iszero(v) || _atomic_add!(g, dof, v)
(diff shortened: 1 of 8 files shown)


src/assembler.jl
```diff
@@ -375,6 +422,46 @@ end
...
+        GC.@preserve x begin
+            p = pointer(x, i)
+            Base.llvmcall(
+                $(
+                    """
+                    %p = inttoptr i$(Sys.WORD_SIZE) %0 to $(llvmT)*
+                    %rv = atomicrmw fadd $(llvmT)* %p, $llvmT %1 monotonic
+                    ret void
+                    """
+                ), Cvoid, Tuple{Ptr{$T}, $T}, p, v
```
termi-official, line 447: Is there any major advantage of directly injecting the llvm code over `Atomix.@atomic`? Also, how portable is this IR across architectures?

KristofferC: Atomix does the same so might as well avoid that dependency. It is a target-independent LLVM intrinsic so it works everywhere julia runs.

fredrikekre:
```
Output generated during execution of 'examples/threaded_assembly':
┌ WARNING: llvmcall with integer pointers is deprecated.
│ Use actual pointers instead, replacing i32 or i64 with i8* or ptr
└ in (::var"##examples/threaded_assembly#287".TestMultiThreading.var"#28#29"{scratch, C, b})(Any) at /Users/fredrik/.julia/packages/OhMyThreads/wX0Fh/src/macro_impl.jl
```

KristofferC:
I had to use typed pointers for 1.10, I didn't know they were deprecated (thought it silently upgraded). Will fix.

PR #1418, opened 2026-07-27
add tensorial interpolations

KristofferC:
still need to go through this properly myself but putting it up for consideration anyway

needed for my "feature nirvana" for my paper implementation :P


termi-official:
xref https://github.com/Ferrite-FEM/Ferrite.jl/pull/958 for some prior discussion. We already use these in FerriteViz here https://github.com/Ferrite-FEM/FerriteViz.jl/blob/8e713d9be1e767c982a0999a60d847274eae4567/src/utils.jl#L680-L766 (although not sure how much we can trust this one).

termi-official:
Also, there is https://mfem.org/pdf/seminar/gopalakrishnan.pdf as a reference for some example on where these can be useful.

KnutAM:
This is cool! I guess with a generic `TensorInterpolation{TT<:Type{<:AbstractTensor}}` we could even get rid of the vectorized interpolations, as they would become
```
const VectorizedInterpolation{vdim, refshape, order, SI} = TensorizedInterpolation{Vec{vdim}, refshape, order, SI}
```
?

lijas:
ip^(3, 3)
(or, alternatively,
ip^(3 × 3)
? 😎)

KristofferC:
You also gotta get the `sym` in there :P

KristofferC:
>  we could even get rid of the vectorized interpolations,

the last commit does that, it seems to work pretty well!


```
const VectorizedInterpolation{vdim} = TensorizedInterpolation{Vec{vdim}}
```

beautiful

src/Dofs/ConstraintHandler.jl
```diff
@@ -932,6 +937,30 @@ function meandiag(K::AbstractMatrix)
     return z / size(K, 1)
 end
 
+# Wrapper for functions prescribing tensor-valued fields, allowing the function to return
+# the natural tensor value (e.g. `x -> SymmetricTensor{2, 2}((1.0, 2.0, 3.0))`) from which
+# the prescribed components (in the tensor's data order) are extracted. Returning a
+# collection with one entry per prescribed component (as for vector fields) is also
+# supported and passed through as is.
+function _wrap_tensor_bc_function(f::F, ::TensorInterpolation{TB}, components::Vector{Int}) where {F, TB <: SecondOrderTensor}
```
KristofferC: This whole thing is kind of ugly but comes from the mismatch between length of a symmetric tensor and the number of components it actually contains. We check

```
bc_value = f(x, time)
@assert length(bc_value) == length(components)
```

KnutAM:
Could we have that 
`active_components::AbstractTensor{dim, Bool}`, something like 
```julia
active_components = SymmetricTensor{2,2}((i,j) -> i == j)
dsig = SymmetricTensor{2,2}((1.0, 2.0, 3.0))
dbc = Dirichlet(:sig, left, (x,t) -> t * dsig, active_components)
```
to prescribe only the normal components of `sig` on `left`,
where we require that `f(args...)::Tensors.get_base(typeof(active_components))`

KristofferC:
> Could we have that...

implemented

KristofferC:
I think this is in a somewhat reviewable state now

src/interpolations.jl
```diff
@@ -1570,19 +1579,52 @@ function reference_shape_value(ip::RannacherTurek{RefHexahedron, 1}, ξ::Vec{3,
...
-    function VectorizedInterpolation{vdim}(ip::SI) where {vdim, refshape, order, SI <: ScalarInterpolation{refshape, order}}
-        return new{vdim, refshape, order, SI}(ip)
-    end
 end
-conformity(ip::VectorizedInterpolation) = conformity(ip.ip)
+TensorizedInterpolation{Vec{vdim}}(ip::SI) where {vdim, refshape, order, SI <: ScalarInterpolation{refshape, order}} =
+    TensorizedInterpolation{Vec{vdim}, refshape, order, SI}(ip)
+TensorizedInterpolation{Tensor{2, dim}}(ip::SI) where {dim, refshape, order, SI <: ScalarInterpolation{refshape, order}} =
+    TensorizedInterpolation{Tensor{2, dim}, refshape, order, SI}(ip)
+TensorizedInterpolation{SymmetricTensor{2, dim}}(ip::SI) where {dim, refshape, order, SI <: ScalarInterpolation{refshape, order}} =
```
termi-official, line 1614: Can we add proper error messages for the order 3 and 4, so users understand that is is just not implemented yet? Because the doc string reads like an example, not like a restriction to order<=2.

test/test_tensorized_interpolation.jl
termi-official, line 1: Is this a development artifact (to speed up the development)? If so, should this file be split up and the individual component tests (like CellValues tests) should pair up with the remaining tests on the components?

test/test_interpolations.jl
```diff
@@ -168,6 +168,43 @@ using Ferrite: reference_shape_value, reference_shape_gradient
...
+                        dNdξ = @inferred(reference_shape_gradient(t_interpolation, x, dof))
+                        @test dNdξ isa GradT
+                        # One-hot in data component `comp`, scaled by the scalar base function
+                        Nbase = reference_shape_value(interpolation, x, base_dof)
+                        @test N.data[comp] ≈ Nbase
+                        @test sum(abs, N.data) ≈ abs(Nbase)
+                        # Analytic derivatives match the scalar base ones
+                        dNdξ_base, _ = Ferrite.reference_shape_gradient_and_value(interpolation, x, base_dof)
+                        E = Ferrite._tensorized_basis(TB, comp, one(value_type))
+                        @test dNdξ ≈ otimes(E, dNdξ_base)
```
termi-official, line 198: Just that I understand this correctly, these three tests essentially tell us that we did not transpose our tensor interpolation in the pipeline accidentally -- as long as the modulo logic above to extract the component is not flaky?

PR #1422, opened 2026-07-28
Global dofs 2

lijas:
Based on the implementation in #1012 (adding global DOFs to the `DofHandler`).

Use global DOFs to solve the computational homogenisation tutorial with Neumann boundary conditions.

With this PR, we can more easily compare the alternative implementation of global DOFs in #1395.

Docs: https://ferrite-fem.github.io/Ferrite.jl/previews/PR1422/tutorials/computational_homogenization/


KristofferC:
Just as an example from one of my papers (just look at sigma_bar...):

image

Here we have two pieces of information about sigma_bar from the weak form:

- What field it is: R^(3x3_sym)
- Where the field lives (on the RVE).

With https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 this information is encoded (just like any other field) .The `DofHandler` / `SubDofHandler` encodes the domain, and the "interpolation" says what field it is. Everything works the same as a normal field with coupling, etc. 

With this, you are kind of raw dogging new dofs into your DofHandler. What does that even physically mean? What if your global field only lives on a subdomain? What if you want to swap from a symmetric -> non-symmetric tensor, is that easy? What if you want to allocate a block sparsity pattern?  Here, you have to add a specialized version of standard ferrite functions (`add_cells_to_global_dof_coupling!`) and other "utility" functions to "patch up "the fact that you deviated from the standard pipeline of implementing a weak form in Ferrite.

In https://github.com/Ferrite-FEM/Ferrite.jl/pull/1396/ there is:

```
ip_ε = GlobalConstant{RefTriangle}()^3 # field
dh = DofHandler(grid)
add!(dh, :εbar, ip_ε) # domain
cv_ε = CellValues(qr, ip_ε, Lagrange{RefTriangle, 1}());
reinit!(cv_ε, cell)
shape_value(cv_ε, qp, iε)
function_value(cv_ε, qp, ae, range_ε)
```

You basically "forget" that it is a global field and write your weak form just as it is written on paper and how you would do it in any other problem. I think that has a pedagogical strength to it.

lijas:
Sorry, I am still not fully convinced :-P . I agree that with #1395 the implementation of #1396 becomes quite elegant, but I still see the same big drawbacks. The biggest drawback is that you can't do multithreading with #1395, which, to me, says that it is probably the wrong approach. Do we really want global dofs to be limited to single-threading?

Do you agree that, as a general-purpose global dof implementation, #1395 is not good enough, and that something like this PR would be required? I mean, if you want to enforce a mean value constraint only on the boundary with Lagrange parameters, then #1395 would not suffice (since it has ties global dofs to subdofhandlers). Overall, I think this PR is more flexible. I guess this is what Knut already mentioned, that they try to solve different problems, where #1395 is a bit more tuned towards your RVE problem.

Implementation-wise though, the biggest difference between our two approaches, if one were to implement the problem in #1396, is that with #1395 you have to write one assembly function for the entire system, (K = [K_{uu}, K_{us}; K_{su}, 0]), but with this PR you would have one assembly function for (K_{uu}) and one for (K_{us}). I guess perhaps it comes down to preference, but I don't see any major drawback with my approach. And the fact that it can be threaded makes it feel like the better approach. Also, in this PR, where I extend the computational homogenisation tutorial with Neumann boundary conditions, my global dof approach is quite nice because you can reuse the (K_{uu}) assembly function that is used by the periodic and Dirichlet BCs, and only implement the Kus assembly. 

Edit: Regarding e.g. BlockArrays and renumbering etc.... Yeah, since global fields are their own special thing, one would need to modify those parts of the codes to take globald dofs in to account. Not the prettiest but...

KnutAM:
> I guess this is what Knut already mentioned, that they try to solve different problems, where https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 is a bit more tuned towards your RVE problem.

I would consider that approach for adding e.g. your dissipation solver based method? Then that seems straight forward instead of having to store the results from a nonlinear material model in each quadrature point, and then loop again to apply it in the second assembly loop?

On the other hand, I use the approach herein in my homogenization package, because I only add the couplings to the boundary and not the entire volume - which is why I advocate having both methods - then we don't have to discuss which is better :) 

> And the fact that it can be threaded makes it feel like the better approach.

Actually, as long as the `K[global_dof, global_dof]` is not assembled (often is zero right), then #1395 should be thread safe if the cells are colored in a standard fashion? Of course, easy to get a problem here though. And with #1417, that wouldn't be a problem in any case?

KristofferC:
> Do we really want global dofs to be limited to single-threading?

I mean, the coupling of dofs is intrinsic to the math and the coupling is what makes the coloring not work. Exactly like in the other version, this PR has `doassembly2` with the same problem in that every cell writes into the same global rows. In the other version, you can of course hoist the assembly of that block into a separate routine if you wish.

And it is also a little bit funny to see

```
    #Create the cellvalues for the lagrange basis.
    #TODO: How to introduce this?
```

in the example :P. Isn't that kind of #1395?

src/Dofs/DofHandler.jl
```diff
@@ -413,7 +442,12 @@ function __close!(dh::DofHandler{dim}) where {dim}
             field_ncells,
         )
     end
-    dh.ndofs = nextdof - 1
+    n_dofs = nextdof - 1
+    for (key, dofs) in dh.global_dofs
+        dh.global_dofs[key] = n_dofs .+ (1:length(dofs))
```
KristofferC: I think this is buggy w.r.t dof reordering (which happens after `close!`)

KristofferC:
> as a general-purpose global dof implementation, https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 is not good enough, and that something like this PR would be required? I mean, if you want to enforce a mean value constraint only on the boundary with Lagrange parameters, then https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 would not suffice

How do you define a "global dof". What #1395 introduces is a concept of a spatially constant field that lives on a domain. It doesn't really talk about dofs I think. For introducing fields on boundaries, that is a feature we still need (which would also be useful for contact).

lijas:
> Actually, as long as the K[global_dof, global_dof] is not assembled (often is zero right), then https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 should be thread safe if the cells are colored in a standard fashion? Of course, easy to get a problem here though. And with https://github.com/Ferrite-FEM/Ferrite.jl/pull/1417, that wouldn't be a problem in any case?

I think you are right, the approach in 1395 should also be thread safe in general if Kss block i zero. I was confused about the warning in the coloring algorithm in 1395... (multi threading should work, but is it the coloring algorithm that does not work?).  I guess then my biggest concern about 1395 is gone then if the coloring algorithm for global dofs can be implemented :P 

> Exactly like in the other version, this PR has doassembly2 with the same problem in that every cell writes into the same global rows

Correct me if I am wrong, but it would be possible to multi threading as normal in doassembly2, no? The coloring works as normal, and the cell loop never write to the same entry in the stiffness matrix?

> How do you define a "global dof". What https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395 introduces is a concept of a spatially constant field that lives on a domain. It doesn't really talk about dofs I think.

Good point.. viewing 1395 as a field rather than a global dofs makes it make more sense to me. Did not really notice that distiction earlier. 

> Here we have two pieces of information about sigma_bar from the weak form:
> 
> What field it is: R^(3x3_sym)
> Where the field lives (on the RVE).

Perhaps a bit philosophical, but where does it say that sigma_bar lives on the RVE? the displacement field lives in U_Box, which should we square integrable over the domain Omega, but sigma_bar lives in R^3x3_sym which does not mention any domain?

> And it is also a little bit funny to see
> 
>     #Create the cellvalues for the lagrange basis.
>     #TODO: How to introduce this?
> in the example :P. Isn't that kind of https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395?

Perhaps a bit unclear comment. I meant how to introduce that a introduce that GlobalConstant interpolation can be passed to the cellvalues as normal. But yes, I stole that from 1395 becuase it makes the element assembly of the Kus term very clean.

src/Dofs/DofHandler.jl
lijas: This would be need to refactored to global_dofs and global_dofs_offset style.

lijas:
Unfortunately deal.ii does not have global dofs or fields, so we cant take inspiration from there 😛

lijas:
Regarding the design for global DOFs in this PR (I had chat gpt formulate my thoughts):

Right now, a common workaround is to attach global fields (like \lambda) to dummy cells/nodes. The advantage of this approach is that downstream infrastructure—such as DOF renumbering, block matrix setups, and numbering schemes—works out-of-the-box because the field is formally tied to a topological entity.
While we definitely don't want to secretly inject fake cells or nodes into the actual Grid structure, this perspective provides a useful mental model for our implementation:

Abstraction without Grid Pollution: We can handle global DOFs in DofHandler as if they belong to a "virtual" top-level entity, keeping the Grid strictly spatial and clean.

Seamless Compatibility: Structuring global DOFs this way internally ensures that features like DOF renumbering, block handling, and sparsity pattern generation can interact with global fields seamlessly without requiring special-case hacks across the ecosystem.

KnutAM:
I keep saying - let's just have both :D 

Regarding naming, I discussed with myself and ChatGPT. I would suggest the following (hope it is ok I put all the details here even if half refers to #1395, @KristofferC )

* #1395: `ip = SpatialConstant{TB}()` where `TB` is either `N::Int` or `TB = Tensors.get_base(TT) where {TT<:AbstractTensor}`. Examples
  - `ip = SpatialConstant()`: Scalar spatial constant field (same as `SpatialConstant{1}()`)
  - `ip = SpatialConstant{3}()`: 3 scalar spatial constant fields
  - `ip = SpatialConstant{Vec{3}}()`: 1 vector spatial constant field
  - `ip = SpatialConstant{SymmetricTensor{2,3}}()`: 1 symmetric tensor constant spatial field
  - `ip = SpatialConstant()^dim`: Can be defined as `SpatialConstant{Vec{dim}}()` to follow vectorization.
  
* #1422: `ip = SystemVariable{TB}()` where `TB` has same meaning and convention as above.
  Here, however, we also need to for tensorial variables have the option of getting the dof-number from the tensor-indexing, but that is just one new function. Another difference is that `SpatialConstant` is added to `SubDofHandlers`, whereas `SystemVariable` is always added to the full `DofHandler`. 
  
For both cases, it doesn't make sense to me to use a reference shape however, this will break some assumptions/checks, but I think it should be possible to fix by defining `NoRefShape` and `Dimless` types. We should also have the mapping type being `NoMapping` for these, since they will not require coordinates for mapping their function values, so these should not be calculated if not needed by something else.

While convenient, I had to think a bit about the difference between `ips = SpatialConstant{3}()` and `ipv = SpatialConstant{Vec{3}}()`. And IMO we should have the key difference that

* `reference_shape_value(ips, nr, \xi) = 1`
* `reference_shape_value(ipv, nr, \xi)::Vec{3}(i -> i == nr ? 1 : 0)`

Affecting the result from `shape_value` from `AbstractValues`. The exact same analogy can be implemented for `SystemVariable`, but arguably they don't belong in `CellValues`. However, here I think we can go with the convenience over mathematical stringency and also support that case.

lijas:
Nice suggestions Knut. To aid the design of the implementation of SpatialConstant field, it could be helpful to think how a generalization to SpatialLinear field would look like (not sure if that would ever be useful in a real problem). Like, how would cell values handle the mapping in that case etc.

termi-official:
> if the coloring algorithm for global dofs can be implemented

We just merged the atomic assembly (https://github.com/Ferrite-FEM/Ferrite.jl/pull/1417), your problem has evaporated.

KristofferC:
For what type of problem is the "SystemVariable" needed (that isn't just to hack around other missing features in Ferrite)?

KnutAM:
> For what type of problem is the "SystemVariable" needed (that isn't just to hack around other missing features in Ferrite)?

I’m using this to put affine constrains of the form (e.g. `u = ubar + cdot(epsbar, x)`, where `epsbar::SymmetricTensor` and some of its components might be unknown (conversely if known I can get the stress components as reaction ”forces”), on the boundary.

lijas:
Or for example if we want to enforce a mean value constraint with a Lagrange parameters, $\lambda \int_{\Gamma} p\ \mathrm{d}\Gamma = 0$ where $\lambda \in \mathbb{R}$.

My understanding now is that SystemVariable creates DoFs in spaces like $\mathbb{R}$, $\mathbb{R}^4$, $\mathbb{R}^{3\times3}$, etc., while SpatialConstant creates fields over domains, e.g. mathematically defined by $\lambda \in \mathcal{V}_{\mathrm{const}}$ where Vconst is something like

$\mathcal{V}_{\mathrm{const}} = \{u:\Omega\rightarrow\mathbb{R}^3 \mid u(\mathbf{x})=\mathbf{a},\ \mathbf{a}\in\mathbb{R}^3\}$

Both can be used to solve the same problem in many cases, but SystemVariable is not attached to any domain.
SpatialConstant could be extended/generalized to SpatialyLinear etc if that would ever be needed in the future.

KristofferC:
Even for a `SystemVariable`, can’t we still think in terms of fields? It could be a field living on a "virtual zero-dimensional entity" that belongs to the `DofHandler`, but not to the physical `Grid` and not connected to any cells. Couplings would then be added explicitly with some API. 

For example, what I don't like here:

```
add_global_dofs!(dh, :λ, 3)
```

has a naked 3, it doesn't say what `λ` is (3 scalars, one vector, one symmetric tensor). Also

```
add_cells_to_global_dof_coupling!
```

couples the global dofs to all cell dofs, which is not needed (u-p and you want to couple λ to only p). 

Instead something like

```
add!(dh, :σ̄, SystemVariable{SymmetricTensor{2, 2}}())
```

seems like a stronger API to add this information. And then we have the `SystemVariable` / `GlobalConstant` difference deciding if this field is "everywhere uncoupled" or "everywhere coupled". Thinking about it more, I think this is kind of what Elias just said... So I guess +1 to that :P

KristofferC:
I asked ChatGPT to try formulate my thoughts above a bit clearer. I do not know if it helps but hey:

  > Ferrite generally tries to derive its abstractions from the weak form. From that perspective, the question is not merely where we can
  > append a few extra DOFs, but what trial and test space those DOFs represent.
  >
  > For a SpatialConstant, that space is a constant function space over some physical domain. For a SystemVariable, it may instead be a
  > finite-dimensional space such as (\mathbb{R}^n). We could still implement the latter through the field/entity machinery by viewing it
  > as a field on a virtual zero-dimensional entity owned by the DofHandler, but not belonging to the physical Grid. Its coupling to
  > spatial fields would be specified by the weak form rather than inferred from mesh topology.
  >
  > That would make the distinction variationally meaningful while allowing both cases to reuse numbering, renumbering, block layouts,
  > constraints, and other DOF infrastructure. In contrast, simply appending n anonymous DOFs and patching the required couplings
  > afterwards feels more like a problem-specific escape hatch than a Ferrite-style representation of the variational problem.

KristofferC:
More from talking with ChatGPT:

“Where the field lives” and “how the field varies” are independent properties.

For example, over a domain \(\Omega\), you could have:

- A spatially constant field: `u(x)=a`
- A spatially affine field: `u(x)=a+b\cdot x`
- A higher-order global polynomial field.
- An ordinary FE field whose basis functions have local support.

All are fields defined on \(\Omega\). **The constant field is special only because its approximation space has the single basis function \(N(x)=1\), whose support is the entire domain.**

So the deeper distinction is not really `SystemVariable` versus `SpatialConstant`. It is:

- domain-attached space versus domain-free algebraic space;
- independently, the choice of basis/approximation within that space.


“where the unknown lives” and “how it varies” should be treated as separate questions. `SpatialConstant` combines two choices: it is attached to a physical domain, and its approximation space consists of constant functions. In principle, we could also have a globally supported spatially affine or polynomial field on that domain. A `SystemVariable`, on the other hand, is not attached to a physical domain at all; it belongs to an algebraic space such as \(\mathbb{R}^n\). This suggests that domain attachment and interpolation/basis should be separate parts of the abstraction.

One additional terminology nuance: an ordinary FE field may “live everywhere” in the sense of being defined over all of \(\Omega\), while each of its basis functions has only local support. A spatially affine global field can both live over all of \(\Omega\) and have basis functions with global support.

termi-official:
Let me throw a problem into the mix here with some "semi-global" dofs which I manage right now by hand in my frameworks. I often have some volume control formulations to couple different problems (e.g. as the constraint that some 3D volume = some 0D problem, where the 3D volume integral can be reformulated as a surface integral over some 2D boundary. This way things stay a bit sparser. Any chance we can handle this case too without adding the full global dof on all cells? I would prefer common infrastructure over handling this case separately, as there is considerable overhead.

KristofferC:
Firstly, @lijas, I am starting to come around to the idea that this is perhaps a better approach and sigma_bar shouldn't be looked at as a global field but an algebraic variable, and we can add "utility functions" that make using it as nice as in #1396. Kind of like you said from the beginning :P.

For @termi-official 

> I often have some volume control formulations to couple different problems (e.g. as the constraint that some 3D volume = some 0D problem, where the 3D volume integral can be reformulated as a surface integral over some 2D boundary. 

You would do roughly

```julia
pvar = AlgebraicVariable() # scalar
add!(dh, :p0, pvar)
close!(dh)

coupling = FacetCoupling(dh, boundary; coupling = coupling_matrix)
K = allocate_matrix(dh; couplings = (coupling,))
```

For this we would need to introduce a `FacetCoupling` 	that you can pass to the sparsity pattern constructor.
The algebraic DOF would be numbered and handled by the DofHandler, but would not be inserted into celldofs for every cell.
Assembly would look like:

```julia
# Algebraic values are global, so evaluate them once before the facet loop.
p0 = function_value(dh, a, :p0) # if needed by the element routine

for facet in FacetIterator(dh, boundary)
    # Returns a vector-like local layout containing the cell dofs followed
    # by the global dofs belonging to :p0.
    dofs = local_dofs(facet, dh; append = (:p0,))

    # Ranges in the augmented local matrix/vector.
    range_u = dof_range(dofs, :u)
    range_p = dof_range(dofs, :p0)

    Ke = zeros(length(dofs), length(dofs))
    fe = zeros(length(dofs))

    reinit!(facetvalues, facet)

    assemble_boundary_term!(
        Ke,
        fe,
        facetvalues,
        p0,
        range_u,
        range_p,
    )

    assemble!(assembler, dofs, Ke, fe)
end
```

Something along those lines I think

src/Dofs/DofHandler.jl
```diff
@@ -335,6 +341,29 @@ function add!(dh::DofHandler, name::Symbol, ip::Interpolation)
...
+    dh.global_dofs[name] = (-n):(-1) # Range shifted in close
+end
+
+"""
+    global_dof_range(dh::DofHandler, name::Symbol)
+
+Get a `UnitRange{Int}` describing the global dofs associated with `name`.
+Such dofs can be added with [`add_global_dofs!`](@ref).
+"""
+function global_dof_range(dh::DofHandler, name::Symbol)
```
termi-official: SubDofHandler dispatch missing

KnutAM:
I think this suggestion goes in a very nice direction for improving the use-case of `SystemVariable` / `AlgebraicVariable` (#1422). Still not sure why we wouldn't allow a `SpatialConstant` case too though. 

> dofs = local_dofs(facet, dh; append = (:p0,))

If you have gradient-dependent terms, they will also couple to the other dofs on the cell?  So could be `celldofs`? 

> for facet in FacetIterator(dh, boundary)

Could potentially be `FacetIterator(dh, boundary; append_dofs = (:p))`, which would avoid allocating for each facet.

KristofferC:
> If you have gradient-dependent terms, they will also couple to the other dofs on the cell? So could be celldofs?

Could you elaborate on this?

> Could potentially be FacetIterator(dh, boundary; append_dofs = (:p)), which would avoid allocating for each facet.

Yeah, I haven't really thought about things like optimization, only what information has to be there to somehow get the data you need.

KnutAM:
> > If you have gradient-dependent terms, they will also couple to the other dofs on the cell? So could be celldofs?

> Could you elaborate on this?

```
(4) --- (3)
 |       |
(1) --- (2)
 |---L---|
```
If we have the facet $\Gamma$ for nodes 2-3, and we evaluate 
```math
c = \int_\Gamma \nabla u \mathrm{d}\Gamma
```
we have 
```math
c = \int_\Gamma \nabla N_i \mathrm{d}\Gamma a_i
```
where e.g. $\nabla N_1(x_2) = [-1/L, 0]$, so a non-zero contribution from node 1 which is not on the facet.

lijas:
> Firstly, @lijas, I am starting to come around to the idea that this is perhaps a better approach and sigma_bar shouldn't be looked at as a global field but an algebraic variable, and we can add "utility functions" that make using it as nice as in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1396. Kind of like you said from the beginning :P.

Nice! But I must say I also started to like the idea of global constant fields on subdomains haha (#1395). I thought of a cool application of #1395 for modelling rigid bodies. But I think I will elaborate on that in a different issue/PR :P 

I have updated the code based on your suggestions. I changed the name to SystemVariable and added some utility functions, and made it work with the dof renumbering system. Still a lot to do, but if someone wants to review the main design of system variables in the the DofHandler, please feel free to do so :)

KristofferC:
I think AlgebraicVariable is better than SystemVariable. In some sense, every variable is part of the system. With Algebraic you say that its coupling only comes in from the weak form and not from any physical coupling on the mesh.

KristofferC:
Some things I think almost "have" to be done.

- Adding a new algebraic variable needs to take the type the variable represents, not just the number of dofs:
	```
	add!(dh, :p0, AlgebraicVariable{}())
	add!(dh, :p0, AlgebraicVariable{Vec{3}())
	add!(dh, :p0, AlgebraicVariable{SymmetricTensor{2, 2}())
	...
	```
	Note element type is not part of it, that is decided by e.g. the solution vector.
- `add_system_variable_entries!` should imo use the same `coupling` interface that already exists. The current coupling hard codes:
	  - only cell sets;
	  - one algebraic variable at a time;
	  - always symmetric spatial–algebraic coupling;
	  - no component-level mask;
	  - no algebraic–algebraic coupling;
	  - only diagonal K_λλ entries come from the generic diagonal pass you might want it to be dense;
- 	Some coupling interface
	```
	  cell_coupling = CellCoupling(
	      dh, cells;
	      fields = (:u, :λ),
	      coupling = Bool[
	          0 1
	          1 1
	      ],
	  )
	
	  facet_coupling = FacetCoupling(
	      dh, boundary;
	      fields = (:u, :p0),
	      coupling = Bool[
	          0 1
	          1 1
	      ],
	  )
	```
   where these couplings are "additive".
- There should be a `algebraic_value` or extending `function_value`, something like `p0 = function_value(dh, a, :p0)`
- There needs to be some way to get the local dof values for the algebraic variable inside an element assembly, in  
- To show that the features above are implemented, the example should show that manual block-array assembly is not needed. 

I could make an example implementation just to show how I think it could be done, and maybe it is easier to discuss from there. "Global dofs 3"... 😆

KristofferC:
I did a bunch of back and forth with ChatGPT to try to summarize the discussion we've had so far (what problems we want to solve, what different strategies have been discussed, etc) and make an implementation plan. I also fed it some of my own thoughts from above. It is a bit long, but I have read through it, and I think it is quite readable.

<details><summary>Plan</summary>

# Algebraic Variables and Explicit Entity Couplings in Ferrite.jl

## Status and purpose

- **Status:** Proposed design, ready for implementation review
- **Last updated:** 2026-08-04
- **Primary discussion:** [Ferrite.jl PR #1422](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1422)
- **Related spatial-field proposal:** [Ferrite.jl PR #1395](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1395)

This document makes the proposal reviewable without requiring the full GitHub discussion.
It covers the problem, its relationship to the spatially constant field in PR #1395, the
alternatives considered, and a concrete implementation plan. The scope is limited to:

1. represent typed algebraic unknowns that are not attached to the mesh;
2. state explicitly where those unknowns couple to spatial fields;
3. allocate the correct sparse matrix structure;
4. assemble those terms through Ferrite's ordinary square local-matrix path; and
5. integrate with numbering, blocks, constraints, and atomic assembly.

It does not propose a general weak-form framework or a new finite element abstraction.

## Executive summary

Some variational problems contain unknowns in finite-dimensional spaces such as `R`, `R^n`,
or the space of symmetric tensors. They are not fields over a physical domain, so they
should not appear in every cell's `celldofs`; individual weak-form terms instead couple them
to fields on selected cells or facets.

The proposed representation is an `AlgebraicVariable`. It declares a typed value and,
optionally, which of that value's components are unknown. Only active components receive
DOFs from the `DofHandler`, and the variable has no implicit mesh support. All spatial
coupling is introduced through one of three structural descriptors:

- `CellCoupling`, for terms integrated over selected cells;
- `FacetCoupling`, for terms integrated over selected facets; and
- `AlgebraicCoupling`, for terms involving only algebraic variables.

Descriptors contain only structural metadata: support, participating variables, and allowed
test/trial blocks. User code still supplies the weak-form kernel.

Applications normally store descriptors in a named tuple. Matrix allocation unions their
entries; assembly selects one by name and uses the ordinary `CellIterator` or
`FacetIterator`. A `LocalDofLayout` appends the relevant algebraic DOFs to the ordinary cell
DOFs and provides named ranges, allowing square local systems to use existing assembly paths.

## 1. Problem statement

### 1.1 Ordinary finite element fields

An ordinary Ferrite field has a domain and interpolation. Together they determine:

- where the field's DOFs are distributed;
- which DOFs occur in `celldofs`;
- the local range of the field on a cell;
- the default cell sparsity; and
- the values and gradients that can be evaluated through `CellValues` or `FacetValues`.

This is the correct model for spatial fields such as displacement, temperature, or pressure.

### 1.2 Algebraic unknowns

Other unknowns are domain-free. Examples include:

- a scalar Lagrange multiplier in `R`;
- the state of a coupled 0D model in `R^n`;
- a macroscopic symmetric stress tensor;
- a prescribed-volume multiplier; and
- unknown coefficients appearing in an affine constraint.

The weak form, rather than mesh topology, determines where such an unknown couples: the full
domain, a subdomain, a boundary, or only other algebraic variables.

The infrastructure therefore needs to keep two decisions separate:

1. what algebraic value the DOFs represent; and
2. on which integration entities a particular weak-form term couples that value to spatial
   fields.

### 1.3 Why putting the DOF on every cell is insufficient

Putting an algebraic DOF in every `celldofs` vector gives it implicit global cell support:

- it creates structural entries on cells where no weak-form coupling exists;
- it cannot represent a boundary-only coupling without overallocating the matrix;
- it makes coloring treat all cells as mutually coupled through the shared DOF;
- it conflates a domain-free variable with a spatial field having a globally supported
  basis; and
- it makes semi-global 3D--0D systems unnecessarily dense.

The DOF should therefore be globally numbered but absent from ordinary `celldofs`. It is
added to a local assembly layout only for the terms that need it.

## 2. Representative problems

### 2.1 Computational homogenization

For an RVE problem with Neumann-type boundary conditions, the macroscopic stress
`sigma_bar` can be an unknown symmetric tensor. In two dimensions it has three independent
components. It couples to the displacement field through an integral over the RVE.

It is a finite-dimensional symmetric tensor, not a spatial field. A volume weak-form term
couples it to displacement, and its value and tensor basis directions should be available
without a fictitious `CellValues` object. This requires a typed algebraic variable and a cell
coupling. If only some macroscopic stress components are unknown, an active-component
selection defines that tensor subspace without allocating DOFs for prescribed components.

### 2.2 Boundary volume control coupled to a 0D problem

A 3D volume can sometimes be expressed as a surface integral over a 2D boundary. A scalar
or vector state from a 0D problem then couples only to the displacement DOFs of cells
adjacent to that boundary.

This is the motivating semi-global case:

- adding the 0D DOF to every 3D cell would destroy useful sparsity;
- coupling only to trace-node DOFs is not generally correct because a facet integral may
  contain gradients; and
- the initial implementation must use all DOFs of each adjacent cell, but only on selected
  facets.

This requires a facet coupling.

### 2.3 Mean-value constraint on a boundary

A constraint such as

```math
lambda \int_Gamma p\,dGamma = 0
```

has a scalar algebraic multiplier `lambda` coupled to the pressure field only on `Gamma`.
It should not couple to unrelated fields on the same cells, and it should not couple on
interior cells.

This requires facet support and a field-level test/trial mask.

### 2.4 Multiple algebraic variables and a 0D tangent

A coupled finite element--0D system may have the block form

```text
K_uu  K_up
K_pu  K_pp
```

where `p` contains several algebraic variables. Because `K_pp` may be dense and
cross-coupled, algebraic diagonals alone are insufficient. This requires an algebraic-only
descriptor alongside cell and facet descriptors.

## 3. Requirements

### 3.1 Algebraic-variable requirements

- An algebraic variable is added by name to a `DofHandler` before `close!`.
- Its value structure is explicit: scalar, vector, tensor, or symmetric tensor.
- It may select a nonempty subset of the declared value's canonical components as active;
  all components are active by default.
- Its descriptor does not fix the coefficient type to `Float64`.
- It receives stable global DOF identities and participates in `ndofs(dh)`.
- It receives one DOF per active component, not per component of the full value type.
- Its current global DOF numbers can be queried after arbitrary renumbering.
- Its value can be reconstructed from a solution vector by name.
- Its active-component mapping and corresponding basis directions are available for
  linearization.
- Algebraic and spatial names share one uniqueness namespace.

### 3.2 Coupling requirements

- Coupling is explicit rather than inferred from the mesh.
- Cell, facet, and algebraic-only support are covered.
- A descriptor names the participating variables.
- A Boolean mask follows Ferrite's existing convention: rows are test variables and columns
  are trial variables.
- Asymmetric coupling is allowed.
- Multiple algebraic variables and dense algebraic--algebraic blocks are allowed.
- Multiple descriptors can overlap; sparsity is their union and assembly contributions add.
- The descriptor is reusable by both sparsity generation and local DOF construction.

### 3.3 Assembly requirements

- Existing `CellIterator` and `FacetIterator` remain the assembly iterators.
- The local layout contains ordinary cell DOFs followed by the required algebraic DOFs.
- Facet layouts contain all DOFs of the adjacent cell, not only trace DOFs.
- Named `dof_range` queries work on the augmented layout.
- A square `Ke` and `fe` can be assembled through the existing `assemble!` path.
- The same local representation works with block assemblers and `apply_assemble!`.
- Atomic assembly remains available for threaded assembly into shared algebraic entries.

### 3.4 Infrastructure requirements

- Explicit permutation, field-wise, and component-wise renumbering include algebraic DOFs.
- Public introspection does not expose internal `DofHandler` storage.
- Coupling entries are added before constraint-induced sparsity entries.
- Existing spatial-field behavior and the existing `coupling=` keyword remain compatible.
- Existing APIs that require a spatial field reject algebraic names with descriptive errors,
  while continuing to work normally for spatial fields in a handler that also contains
  algebraic variables.
- The implementation must not cause a material performance regression when a `DofHandler`
  contains no algebraic variables.

## 4. Designs considered

### 4.1 A spatially constant field

`GlobalConstant` or `SpatialConstant` represents a constant function over a physical domain.
It is a meaningful spatial approximation space and can make a globally supported field look
like an ordinary Ferrite field in `CellValues`, `celldofs`, and `dof_range`.

This is useful but solves a different problem:

- it is domain-attached;
- its basis has spatial support over that domain; and
- its natural coupling follows that support.

An algebraic variable is domain-free and receives support only through explicit couplings.
A spatial constant may be implemented separately, but is neither part of nor an internal
substitute for this proposal.

### 4.2 Anonymous global DOFs

Appending anonymous DOFs loses whether `x` represents unrelated scalars, a vector, or a
tensor, leaving reconstruction and basis ordering to user code. `AlgebraicVariable` retains
global numbering while making that structure explicit.

### 4.3 Dummy cells or nodes

Fake mesh entities reuse field machinery but pollute the grid and make coupling an accidental
consequence of nonphysical topology. The `DofHandler` should own algebraic variables without
changing the `Grid`.

### 4.4 Adding every algebraic DOF to every cell

This gives convenient square matrices but cannot represent boundary- or subdomain-only
support sparsely. Instead, each relevant term constructs an augmented local layout.

### 4.5 Rectangular block assembly

It is possible to assemble `K_up` and `K_pu` separately using row and column DOF vectors.
This is useful low-level functionality, but it is not supported uniformly by every assembler
and makes residual assembly, local constraints, and block assembly less natural.

The public workflow uses one augmented DOF vector and a square local system. Existing
rectangular methods remain available as low-level tools.

### 4.6 A dedicated coupling iterator

A `CouplingIterator` could combine entity iteration and augmented DOFs, but different terms
still need different kernels, and Ferrite already has cell and facet iterators. A local layout
operation is sufficient.

### 4.7 Trace-only facet coupling

Restricting a facet term to DOFs geometrically located on the facet is incorrect for terms
that evaluate cell gradients on the facet. For example, the gradient of a shape function
associated with a non-facet node can be nonzero at the facet.

The initial implementation therefore uses the complete adjacent-cell DOF vector. Trace-space
optimization is outside scope.

### 4.8 Coupling descriptors containing assembly callbacks

Attaching kernels to descriptors would create a weak-form registry or form compiler.
Descriptors remain structural metadata and do not own user physics.

## 5. Proposed public model

### 5.1 Algebraic variables

The public declarations are conceptually:

```julia
add!(dh, :p0, AlgebraicVariable())
add!(dh, :z, AlgebraicVariable{Vec{3}}())
add!(dh, :sigma_bar, AlgebraicVariable{SymmetricTensor{2, 2}}())
add!(dh, :sigma_partial, AlgebraicVariable{SymmetricTensor{2, 2}}(
    active_components = ((1, 1), (2, 2)),
))
```

`AlgebraicVariable()` denotes one scalar. Tensor descriptors encode shape but not scalar
coefficient type. The exact internal scalar marker is private. Omitting `active_components`
activates the full value. A vector selection uses natural indices, such as `(1, 3)`, while a
rank-two tensor uses Cartesian indices, such as `((1, 1), (2, 2))`. Higher-rank tensors use
one index tuple per selected component. Users never specify flattened component numbers.

A flattened Boolean mask is intentionally avoided because its positions are unclear for
structured values such as `SymmetricTensor`. A tensor-shaped Boolean array would be clearer
but would duplicate symmetry-equivalent entries. Natural indices are explicit, compact, and
match ordinary indexing of the reconstructed value.

Selections must be nonempty, in bounds, and free of duplicates. Symmetry-equivalent indices,
such as `(1, 2)` and `(2, 1)` for a `SymmetricTensor`, identify the same component and are
normalized before duplicate checking. The descriptor stores selected components in the
value type's canonical order, independent of the order in which they were supplied. A scalar
has only its single active component.

The API omits `AlgebraicVariable{3}()`, which is ambiguous between three scalars and one
three-component value. Users instead declare separate names or use `Vec{3}`. This need not
match a future `SpatialConstant{N}` convention: an algebraic name denotes one typed value.

The active-component selection belongs to `AlgebraicVariable`, not to a coupling descriptor.
It defines the unknown space consistently for numbering, value reconstruction, basis
directions, renumbering, and every coupling in which the variable appears. A coupling mask
answers a different question--which existing test and trial DOFs may interact in one term--
and must not change how many DOFs the variable owns. Separate scalar variables could
represent the same number of unknowns, but would lose their grouping and reconstruction as
one typed tensor.

This MVP supports coordinate subspaces only. It does not represent arbitrary linear
combinations such as a rotated or trace-free basis. Such spaces would require an explicit
basis map rather than a component selection and are outside scope.

Required queries are:

```julia
getalgebraicvariablenames(dh)  # ordered vector of algebraic-variable names
algebraic_variable(dh, :p0)   # the AlgebraicVariable descriptor registered as :p0
algebraic_dofs(dh, :p0)       # current global DOF indices belonging to :p0
active_components(algebraic_variable(dh, :sigma_partial))  # selected natural indices
```

More precisely:

- `getalgebraicvariablenames(dh)` returns a `Vector{Symbol}` in declaration order, analogous
  to `getfieldnames(dh)`.
- `algebraic_variable(dh, name)` returns the registered `AlgebraicVariable` descriptor. The
  descriptor carries the declared shape and active-component selection, not its value or DOF
  numbers.
- `algebraic_dofs(dh, name)` returns global DOF numbers in active-component order. It returns
  an integer vector rather than promising a `UnitRange`, since renumbering may make the
  indices noncontiguous.
- `active_components(variable)` returns the immutable, canonicalized natural component
  indices. For the example above it returns `((1, 1), (2, 2))`.

An unknown name produces a descriptive error. These queries return metadata only;
`algebraic_value(dh, a, name)` is the separate operation that reads the corresponding
coefficients from a global solution vector `a` and reconstructs the typed value.

`getfieldnames(dh)` continues to mean spatial fields. Internal operations may combine both
registries, but public introspection keeps them distinct. Internal code obtains the algebraic
DOF count with
`length(active_components(variable))`; a public `n_components(dh, :p0)` overload is neither
required nor desirable because it is ambiguous between the full value component count and
the active DOF count. The accessors above are exported. The exact canonical ordering for each
supported value type must be documented and shared by selection normalization,
reconstruction, basis evaluation, coupling expansion, and component-wise renumbering.

All additions return `dh`. Spatial and algebraic names share a namespace, checked in both
declaration orders:

- algebraic `add!` scans the names currently stored in every `SubDofHandler`, because
  `dh.field_names` is not populated until `close!`; and
- spatial `add!(sdh, ...)` checks the algebraic registry owned by `sdh.dh`.

Compatible `SubDofHandler`s may still share a spatial name.

### 5.2 Value and basis evaluation

The global value API is:

```julia
p0 = algebraic_value(dh, a, :p0)
sigma_bar = algebraic_value(dh, a, :sigma_bar)
```

For a scalar this returns one coefficient; for a vector or tensor it reconstructs the full
typed value from the active global DOFs and inserts `zero(eltype(a))` in inactive components.
The result is therefore the unknown contribution in the declared coordinate subspace. Its
scalar type comes from `a`, preserving `Float32`, dual numbers, and other compatible types.
A spatial name produces a descriptive error. Unlike `function_value`, this operation has no
values object or evaluation point.

Prescribed inactive components remain problem data rather than descriptor state. User code
adds them explicitly when constructing the physical value:

```julia
sigma = sigma_prescribed + algebraic_value(dh, a, :sigma_partial)
```

This keeps the descriptor independent of load steps, boundary data, and coefficient scalar
type.

Linearization also needs the coordinate direction associated with each active scalar DOF.
These directions play the algebraic role of constant shape functions: a scalar has direction
`1`, while a vector or tensor has one constant direction per active component. They have no
spatial argument, interpolation, or gradient. The minimal operation is:

```julia
E_i = algebraic_basis_value(algebraic_variable(dh, :sigma_bar), i, eltype(a))
```

Here `i` indexes active DOFs, and the result is a full typed value with a `1` in the
corresponding declared component and zeros elsewhere. It is the derivative of the typed
value with respect to coefficient `i`. Inactive components have no basis direction because
they own no DOF. This replaces a `GlobalConstant` `CellValues` object used only to construct
tensor basis directions.

No algebraic `function_gradient`, mapping, quadrature, or `CellValues` integration is added.

### 5.3 Coupling descriptors

The three descriptor constructors are:

```julia
CellCoupling(dh, cells; fields, coupling)
FacetCoupling(dh, facets; fields, coupling)
AlgebraicCoupling(dh; fields, coupling)
```

Each descriptor stores:

- the owning `DofHandler` identity;
- an entity set, except for `AlgebraicCoupling`;
- an ordered tuple of variable names; and
- a Boolean coupling matrix.

The interface exposes `entities(descriptor)` where applicable, `fields(descriptor)`, and
`coupling_matrix(descriptor)`.

Construction requires a closed `DofHandler`; validates names, uniqueness, entities, and
matrix dimensions; and verifies that each selected spatial field exists on every relevant
cell. Errors identify the field and entity.

The field-level matrix is ordered according to `fields`. For example:

```julia
FacetCoupling(
    dh,
    boundary;
    fields = (:u, :p0),
    coupling = Bool[
        0 1  # test u, trial p0
        1 0  # test p0, trial u
    ],
)
```

Rows are test variables and columns are trial variables. The matrix may be asymmetric.

The MVP accepts either:

1. a field-level matrix of size `length(fields) x length(fields)`; or
2. a component-level matrix whose size is the sum of the participating DOF-component counts
   of `fields`: all components for spatial fields and active components for algebraic
   variables.

This follows Ferrite's existing convention and permits component-selective coupling without
a new per-DOF mask language. Inactive algebraic components do not appear in this matrix
because they have no rows or columns in the global system.

Descriptors may also contain only spatial fields. Masks need a new expander keyed to the
descriptor's selected `fields`; the existing expander assumes all handler fields and is
unsafe for subsets. Existing cell and interface coupling APIs otherwise remain unchanged.

### 5.4 Collections of couplings

For a fixed problem, couplings are normally stored in a named tuple:

```julia
couplings = (
    volume_control = FacetCoupling(
        dh,
        controlled_boundary;
        fields = (:u, :p0),
        coupling = Bool[0 1; 1 0],
    ),
    zero_d_problem = AlgebraicCoupling(
        dh;
        fields = (:p0, :z),
        coupling = Bool[1 1; 1 1],
    ),
    average_temperature = CellCoupling(
        dh,
        controlled_cells;
        fields = (:T, :lambda_T),
        coupling = Bool[0 1; 1 0],
    ),
)
```

Named-tuple keys carry application meaning; descriptors need no mutable label.

Matrix construction accepts any iterable of descriptors. Named tuples are recommended so
assembly can use `couplings.volume_control` instead of array positions; dynamic vectors are
also valid.

## 6. Coupling semantics

### 6.1 Couplings are additive

Ordinary cell sparsity and the existing `coupling=` keyword keep their current meaning.
`extra_couplings=` only adds structural relations. Passing descriptors to `coupling=` gives
a targeted error pointing to `extra_couplings=`.

Consequently, a zero in the `u-u` position of an additional descriptor means:

> This descriptor adds no `u-u` entries.

It does not mean that the global `K_uu` block is identically zero. The ordinary cell pass may
already provide that block.

Sparsity unions overlapping descriptors; numerical assembly adds overlapping contributions.

### 6.2 What one descriptor represents

A descriptor represents one structural integration support and one possible block pattern:

> On these entities, these test variables may couple to these trial variables.

It does not represent coefficients, material data, quadrature, or the numerical integrand.

Combine variables that share an entity set and assembly loop. Use separate descriptors for
different supports or kernels, or when separate names clarify the application.

Several weak-form terms may reuse one descriptor if their support and local variable layout
are identical.

## 7. Local DOF layouts

### 7.1 `LocalDofLayout`

`LocalDofLayout` is a read-only vector-like object containing global DOF indices and named
local ranges. It implements the `AbstractVector{Int}` operations needed by `assemble!` and
owns its DOF vector rather than viewing iterator scratch storage, so it remains valid after
the iterator advances.

For a `CellCoupling` or `FacetCoupling`, its ordering is:

1. all ordinary DOFs of the cell, in existing `celldofs` order;
2. the active DOFs of each algebraic variable named by the descriptor, in descriptor field
   order and then active-component order.

Unselected spatial fields remain in the full cell layout, with zero rows and columns for that
term. This avoids a second filtered-cell-DOF convention.

For an `AlgebraicCoupling`, the layout contains only the selected algebraic variables in
field order.

The constructors are conceptually:

```julia
layout = local_dofs(cell, cell_coupling)
layout = local_dofs(facet, facet_coupling)
layout = local_dofs(algebraic_coupling)
```

The entity must belong to the descriptor's set and owning `DofHandler`. Internal accessors
resolve the adjacent cell's `SubDofHandler` and normalize caches created from either handler
type; users do not inspect `cell_to_subdofhandler`.

### 7.2 Named ranges

Ranges are queried from the layout:

```julia
range_u = dof_range(layout, :u)
range_p0 = dof_range(layout, :p0)
```

Existing `dof_range(dh, :u)` continues to mean the ordinary cell-local range of a spatial
field for the unambiguous single-`SubDofHandler` case.

There is no `dof_range(dh, :p0)`: an augmented local range depends on the descriptor and its
ordering, while global DOFs may be noncontiguous after renumbering. The APIs therefore answer
three distinct questions:

- `algebraic_dofs(dh, :p0)` returns current global DOF numbers;
- `algebraic_value(dh, a, :p0)` returns the represented value; and
- `dof_range(layout, :p0)` returns a range in one augmented local system.

`dof_range(::LocalDofLayout, name)` returns one range. This differs from
`dof_range(::InterfaceCache, field)`, which returns one per interface side.

### 7.3 Facet behavior

Facet layouts include all adjacent-cell DOFs, including non-facet nodes, because facet terms
may evaluate cell gradients. Minimal trace extraction is outside scope.

## 8. Sparsity construction

The main allocation entry point is extended conceptually as follows:

```julia
K = allocate_matrix(
    dh,
    ch;
    coupling = standard_cell_mask,
    extra_couplings = couplings,
)
```

Both keywords are optional: `coupling` controls ordinary cell entries and
`extra_couplings` adds descriptor entries.

The required ordering is:

```julia
sp = init_sparsity_pattern(dh)
add_diagonal_entries!(sp)
add_cell_entries!(sp, dh, ch; coupling = standard_cell_mask)
if topology !== nothing
    add_interface_entries!(sp, dh, ch; topology, interface_coupling)
end

for descriptor in values(couplings)
    add_coupling_entries!(sp, descriptor, ch)
end

# Constraint expansion depends on all pre-existing structural entries.
add_constraint_entries!(sp, ch)
K = allocate_matrix(sp)
```

The implementation should reuse `add_sparsity_entries!` and add constraints exactly once,
after every coupling descriptor.

For each descriptor:

- `CellCoupling` visits selected cells;
- `FacetCoupling` visits selected facets and uses the adjacent cell's full field ranges;
- `AlgebraicCoupling` operates once on its algebraic layout;
- the field/component mask is expanded to DOF pairs; and
- allowed row/column pairs are added to any `AbstractSparsityPattern`.

This remains generic over `SparsityPattern` and `BlockSparsityPattern`. Algebraic diagonals
are always present; self and cross masks add any required dense off-diagonal `K_pp` entries.

The low-level mutation entry point is:

```julia
add_coupling_entries!(sp, descriptor, ch = nothing; keep_constrained = true)
```

Prefer the high-level path, which guarantees constraint ordering.

## 9. Assembly model

### 9.1 Matrix allocation may loop over couplings

Sparsity construction can loop uniformly over descriptors because they share a small
structural interface.

### 9.2 User assembly selects couplings explicitly

User assembly normally selects descriptors by name rather than looping generically: each
weak-form term usually has its own kernel, quadrature, and iterator.

For the boundary volume-control term:

```julia
descriptor = couplings.volume_control
p0 = algebraic_value(dh, a, :p0)

# This example has one SubDofHandler and one scalar algebraic variable.
nlocal = ndofs_per_cell(dh) + length(algebraic_dofs(dh, :p0))
Ke = zeros(nlocal, nlocal)
fe = zeros(nlocal)

for facet in FacetIterator(dh, entities(descriptor))
    layout = local_dofs(facet, descriptor)
    range_u = dof_range(layout, :u)
    range_p0 = dof_range(layout, :p0)

    @assert length(layout) == nlocal
    fill!(Ke, 0)
    fill!(fe, 0)

    reinit!(facetvalues, facet)
    assemble_volume_control!(
        Ke,
        fe,
        facetvalues,
        p0,
        range_u,
        range_p0,
    )

    assemble!(assembler, layout, Ke, fe)
end
```

For the algebraic-only 0D tangent:

```julia
descriptor = couplings.zero_d_problem
layout = local_dofs(descriptor)

range_p0 = dof_range(layout, :p0)
range_z = dof_range(layout, :z)

Ke = zeros(length(layout), length(layout))
fe = zeros(length(layout))

assemble_zero_d_problem!(Ke, fe, a, range_p0, range_z)
assemble!(assembler, layout, Ke, fe)
```

The bulk `K_uu` loop may remain separate, or a term may assemble ordinary and algebraic
blocks together in one augmented square matrix.

### 9.3 Threading

Algebraic entries are shared across their coupling set, so ordinary cell coloring cannot make
those writes independent.

Ferrite's atomic assembler is the supported threaded path:

```julia
assembler = start_assemble(K, f; atomic = true)
```

Each task owns its buffers and layout. Existing atomic assembly supports `Float32` and
`Float64`; the BlockArrays assembler has no `atomic` mode. This proposal adds no scheduler,
coloring, reduction scheme, or combined block/atomic path.

## 10. Numbering, renumbering, blocks, and constraints

### 10.1 Initial numbering

`close!(dh)` distributes spatial DOFs as today, then algebraic DOFs in declaration order.
They do not enter `cell_dofs`.

Each algebraic entry stores its current DOF list, not an assumed permanent unit range.

### 10.2 Renumbering

Explicit permutation renumbering updates `cell_dofs` and every algebraic DOF list.

`DofOrder.FieldWise` treats each algebraic variable as one block; `ComponentWise` creates
blocks only for its active components, preserving their declared component identities.
Custom block vectors use a public combined variable ordering. The current component-wise
code gathers only `celldofs`, so its DOF-count assertion and target-block checks must be
updated to include algebraic lists.

Keep the no-algebraic path as close as possible to its current implementation to avoid the
regression observed in an earlier #1422 benchmark.

Metis remains unaware of coupling descriptors and may treat algebraic variables as isolated
while ordering. Coupling-aware graph construction is outside scope and this limitation must
be documented.

### 10.3 Block sparsity and assembly

Field/component numbering and generic sparsity insertion allow blocks containing both
spatial and algebraic variables.

The augmented square `LocalDofLayout` is passed to the standard block assembler. No new
rectangular BlockArrays assembly overload is required by this proposal.

Domain-wide coupling can create dense algebraic rows and columns around a sparse spatial
block. Field-wise blocking exposes that structure for application-level Schur complements;
no solver or preconditioner is added.

### 10.4 Constraints

Coupling sparsity precedes affine-constraint expansion. Because the augmented layout is an
ordinary global DOF vector, existing `apply_local!` and `apply_assemble!` work on it.

No new name-based `Dirichlet` or constraint syntax for algebraic variables is introduced.
Users may refer to `algebraic_dofs` when constructing existing low-level affine constraints.

### 10.5 Guard rails for spatial-only APIs

APIs requiring spatial interpolation reject algebraic names descriptively. This includes
name-based `Dirichlet`, `PeriodicDirichlet`, `ProjectedDirichlet`, `apply_analytical!`, grid
and point evaluation, projection, and spatial/VTK export.

Those APIs continue to work for spatial names in a mixed handler. Whole-solution output may
omit algebraic variables or expose them separately, but not as nodal or cell data; a new
output representation is outside scope.

## 11. Concrete implementation plan

The phases below are ordered so that every phase establishes invariants needed by the next.

### Phase 1: Introduce the algebraic-variable model

1. Add `AlgebraicVariable`, the internal scalar/vector/tensor shape trait, and an optional
   natural-index `active_components` selection that defaults to the full value.
2. Validate component rank, bounds, symmetry-equivalent duplicates, and a nonempty selection;
   canonicalize its order and derive the algebraic DOF count from its length. Do not add an
   ambiguous public `n_components(dh, name)` API.
3. Add `DofHandler` storage for ordered algebraic entries.
4. Make both spatial and algebraic `add!` paths check the shared name namespace: algebraic
   addition scans all pre-close `SubDofHandler` field names, and spatial addition checks the
   owning handler's algebraic registry.
5. Ensure all `add!` methods return the owning handler.
6. Allocate algebraic DOFs at the end of `close!` without modifying `cell_dofs` or
   `ndofs_per_cell`.
7. Add `getalgebraicvariablenames`, `algebraic_variable`, and `algebraic_dofs`.
8. Export and document the algebraic introspection API and active-component accessor.
9. Update `show`, exports, docstrings, and error messages to use "algebraic variable".

Implementation invariants after this phase:

- every variable name is unique;
- `ndofs(dh)` counts spatial and algebraic DOFs;
- `celldofs` remains purely spatial; and
- every algebraic variable has one DOF per active component of its declared shape.

### Phase 2: Add value and algebraic-basis evaluation

1. Implement `algebraic_value(dh, a, name)` for algebraic names and descriptive errors for
   spatial names.
2. Implement scalar, vector, tensor, and symmetric-tensor reconstruction while preserving
   `eltype(a)` and inserting typed zeros in inactive components.
3. Implement `algebraic_basis_value(variable, i, scalar_type)` by mapping active-DOF index
   `i` back to the declared component ordering.
4. Add bounds, unknown-name, incompatible-vector-length, and unsupported-shape errors.
5. Test reconstruction before and after arbitrary renumbering.

Implementation invariant after this phase:

- the active selection, DOF numbering, value reconstruction, and basis directions all use
  one canonical ordering.

### Phase 3: Add coupling descriptors and validation

1. Add an internal `AbstractCoupling` interface.
2. Implement `CellCoupling`, `FacetCoupling`, and `AlgebraicCoupling`.
3. Normalize `fields` to an ordered immutable collection.
4. Validate handler identity, closed state, names, entity membership, and field availability.
5. Validate and classify field-level versus component-level masks, counting only active
   components for algebraic variables.
6. Add `entities`, `fields`, and mask accessors plus useful `show` output.
7. Add a small normalization helper that iterates over named tuples, tuples, and vectors of
   descriptors without exposing collection details to the rest of the code.

Implementation invariant after this phase:

- a valid descriptor contains all information needed to determine allowed global DOF pairs.

### Phase 4: Implement augmented local layouts

1. Add `LocalDofLayout <: AbstractVector{Int}`.
2. Store an owned copy of the global DOF vector and a name-to-range mapping with
   deterministic ordering; never retain a view into iterator-cache scratch storage.
3. Implement `local_dofs(cell, ::CellCoupling)`.
4. Implement `local_dofs(facet, ::FacetCoupling)` using the complete adjacent-cell DOFs.
5. Implement `local_dofs(::AlgebraicCoupling)`.
6. Implement `dof_range(layout, name)` and the minimal read-only vector interface required by
   assemblers.
7. Reject mismatched entities, handlers, and missing fields with actionable errors.
8. Resolve the owning `DofHandler` and per-cell `SubDofHandler` through internal accessors so
   layouts work for iterators constructed from either handler type.

The initial implementation may allocate a layout per call. Cache-based or mutating layout
APIs are deferred until profiling demonstrates a need; they are not required to validate the
design. Immutable name/range metadata that depends only on a descriptor and
`SubDofHandler` should nevertheless be prepared once and reused rather than rebuilt for
every entity. Examples and tutorials hoist fixed-size matrices, vectors, and other reusable
buffers out of entity loops.

### Phase 5: Integrate descriptors with sparsity generation

1. Implement field/component mask expansion against a `LocalDofLayout`.
2. Implement `add_coupling_entries!` generically for `AbstractSparsityPattern`.
3. Add specialized entity traversal for the three descriptor types.
4. Extend `add_sparsity_entries!` and `allocate_matrix` with `extra_couplings=()` and add a
   targeted error when descriptors are passed to the existing `coupling=` keyword.
5. Ensure additional coupling entries are inserted before constraint entries.
6. Preserve existing `coupling=` behavior for ordinary cells.
7. Cover asymmetric masks, overlapping descriptors, cell subsets, facet subsets, dense
   algebraic blocks, and multiple `SubDofHandler`s.

Implementation invariant after this phase:

- every entry that standard square local assembly may touch is structurally allocated, and
  no algebraic--spatial entries are added outside the declared entity support and mask.

### Phase 6: Complete numbering and block integration

1. Update explicit permutation renumbering to mutate algebraic DOF lists.
2. Update field-wise and component-wise ordering through the public combined variable model;
   component-wise ordering includes only active algebraic components while retaining their
   declared component identities.
3. Replace the existing spatial-only gathered-DOF count assertion and target-block length
   checks with checks against that same combined ordered-variable model.
4. Keep the ordinary no-algebraic path allocation- and performance-equivalent to the current
   implementation where possible.
5. Test `BlockSparsityPattern` allocation with algebraic blocks.
6. Test standard square block assembly with `LocalDofLayout`.
7. Add descriptive wrong-kind guard rails to existing spatial-only name-based APIs while
   preserving their behavior for spatial names in mixed spatial/algebraic handlers.
8. Document the Metis limitation without adding coupling-aware graph construction.

### Phase 7: Add the representative stress-driven example

The current master tutorial is strain-driven and does not contain the algebraic formulation.
This phase adds the stress-driven/Neumann formulation developed in #1396/#1422 rather than
claiming to remove workarounds from the existing master tutorial.

1. Declare the macroscopic stress in the stress-driven example as a typed
   `AlgebraicVariable`.
2. Declare its RVE `CellCoupling`.
3. Allocate the matrix through `extra_couplings=`.
4. Implement `K_u_sigma` and `K_sigma_u` through one augmented square local layout rather
   than teaching rectangular block assembly.
5. Use `algebraic_basis_value` directly rather than introducing a `GlobalConstant`
   `CellValues` object as an algebraic-basis workaround.
6. Demonstrate `algebraic_value(dh, solution, :sigma_bar)`.
7. Keep the existing ordinary `K_uu` assembly reusable where that improves the tutorial.
8. Hoist fixed-size `Ke`, `fe`, and other reusable buffers out of entity loops; the example
   must not teach avoidable per-entity allocation even though layout caching is deferred.

### Phase 8: Documentation, cleanup, and verification

1. Add a focused manual section distinguishing algebraic variables from spatial constants.
2. Document named coupling collections and explain why allocation loops over them while user
   assembly normally does not.
3. Document additive mask semantics and the full-adjacent-cell facet rule.
4. Document that domain-wide algebraic coupling creates dense rows/columns and point to
   field-wise blocking and application-level Schur complements without adding solver code.
5. Document descriptive rejection by spatial-only APIs and the atomic/block combination
   limitation.
6. Remove `GlobalConstant` from this change unless it is independently completed and merged
   as the separate spatial-constant feature.
7. Remove prototype-only names such as `SystemVariable` and
   `add_system_variable_entries!` rather than shipping two public APIs.
8. Run formatting, spelling, explicit-import, documentation, unit, downstream, and benchmark
   checks.

## 12. Test plan

### 12.1 Algebraic-variable unit tests

- scalar, vector, tensor, and symmetric-tensor full and active component counts;
- default-all-active and partial natural-index selections;
- wrong-rank, out-of-bounds, duplicate, symmetry-equivalent duplicate, and empty selections;
- name conflicts in both declaration orders;
- adding after `close!`;
- `add!` return value;
- deterministic initial numbering;
- unknown-name and wrong-kind errors;
- `algebraic_value(dh, a, spatial_name)` produces a descriptive wrong-kind error;
- value reconstruction for `Float64` and `Float32` coefficient vectors, including typed
  zeros in inactive components;
- active-DOF-to-declared-component mapping for value reconstruction and basis directions;
- explicit addition of prescribed inactive components outside the descriptor; and
- value reconstruction after permutation renumbering.

### 12.2 Renumbering tests

- arbitrary permutation and inverse round trip;
- field-wise ordering with one and several algebraic variables;
- component-wise ordering of fully and partially active vector/tensor variables;
- custom block targets;
- public variable-name discovery; and
- unchanged behavior for a handler with no algebraic variables.

### 12.3 Cell-coupling sparsity tests

- one spatial field and one scalar algebraic variable;
- a cell subset rather than the full grid;
- asymmetric `K_up` versus `K_pu` masks;
- selected spatial fields on a mixed-field cell;
- component-level coupling sized and expanded against active algebraic components;
- multiple `SubDofHandler`s; and
- clear failure when a selected cell lacks a named field.

### 12.4 Facet-coupling sparsity tests

- only cells adjacent to selected facets receive algebraic coupling entries;
- interior cells remain uncoupled;
- all DOFs of the adjacent selected spatial field are included, including non-trace nodes;
- asymmetric masks;
- multiple boundaries with separate descriptors; and
- duplicate/overlapping facet descriptors produce the structural union.

### 12.5 Algebraic-only sparsity tests

- dense self-coupling between components of one vector variable;
- cross-coupling between two algebraic variables;
- asymmetric algebraic blocks; and
- the expected diagonal storage when no mathematical self-coupling is declared.

### 12.6 Local layout and assembly tests

- deterministic appended ordering for several algebraic variables;
- a layout collected from an iterator remains unchanged after the iterator advances,
  confirming that it owns its DOF vector rather than viewing cache scratch storage;
- correct spatial and algebraic `dof_range` results;
- correctness after global renumbering;
- ordinary CSC square assembly;
- block square assembly;
- residual-vector assembly;
- `apply_assemble!` with a constrained spatial field;
- equivalence to direct global block insertion on a small reference problem;
- an `AffineConstraint` whose master is an algebraic DOF; and
- forgetting `extra_couplings=` produces the existing descriptive missing-sparsity-entry
  error when augmented assembly is attempted.

### 12.7 Threading test

- assemble a small shared algebraic coupling with independent task-local assemblers and
  `atomic=true`;
- exercise both existing CSC and CSR atomic assembly paths;
- compare the result with serial assembly within floating-point tolerance;
- verify that unsupported numeric types and the BlockArrays/atomic combination fail through
  their documented existing limitations; and
- do not introduce a new coloring or scheduling test suite.

### 12.8 Spatial-only API guard-rail tests

- name-based `Dirichlet`, `PeriodicDirichlet`, and `ProjectedDirichlet` reject algebraic
  names descriptively;
- `apply_analytical!`, grid-node evaluation, point evaluation, projection, and spatial/VTK
  export reject algebraic names descriptively; and
- every affected operation continues to work for a spatial name in a `DofHandler` that also
  contains algebraic variables.

### 12.9 End-to-end examples

- the added stress-driven homogenization example based on the #1396/#1422 formulation;
- a minimal boundary scalar-constraint example covering the semi-global facet use case; and
- a small algebraic `K_pp` test rather than a new full 0D application tutorial.

## 13. Acceptance criteria

The implementation is complete for this proposal when all of the following are true:

1. A scalar or tensor algebraic variable can be declared without a mesh domain or fixed
   coefficient scalar type, with all or a selected nonempty subset of components active.
2. Its global DOFs are numbered, queried, and renumbered without appearing in ordinary
   `celldofs`.
3. `algebraic_value(dh, a, name)` reconstructs the full typed value, inserts typed zeros for
   inactive components, and rejects spatial names descriptively.
4. Cell, facet, and algebraic-only descriptors express symmetric or asymmetric coupling;
   component-level masks contain only active algebraic components.
5. Several named descriptors can be passed together to matrix allocation.
6. Constraint-induced sparsity is correct because descriptor entries are processed first.
7. `local_dofs` and `dof_range(layout, name)` support a square local matrix containing the
   ordinary cell fields and selected algebraic variables.
8. The boundary use case couples only cells adjacent to the selected facets while retaining
   all adjacent-cell field DOFs.
9. Multiple algebraic variables can produce dense and cross-coupled `K_pp` blocks.
10. Standard, block, constrained, and atomic assembly use existing Ferrite assembly entry
    points, subject to the documented existing restriction that atomic and BlockArrays
    assembly do not combine.
11. Spatial-only APIs reject algebraic names descriptively and remain unchanged for spatial
    names in a handler containing algebraic variables.
12. The added stress-driven homogenization example uses augmented square assembly and
    algebraic basis values rather than rectangular block assembly or a spatial
    `GlobalConstant` workaround.
13. Forgetting to allocate an explicitly assembled coupling fails with a descriptive
    missing-sparsity-entry error.
14. The full CI suite passes and the no-algebraic-variable path has no material benchmark
    regression.

## 14. Explicit non-goals

To prevent the first implementation from expanding into a broader redesign, the following
items are outside scope:

- implementing `SpatialConstant`, `GlobalConstant`, spatially affine, or other globally
  supported spatial approximation spaces;
- inserting algebraic variables into ordinary `celldofs`;
- trace-only or minimal facet DOF extraction;
- a `CouplingIterator` or new iterator hierarchy;
- automatic assembly by attaching callbacks or weak-form kernels to descriptors;
- a form compiler or general term registry;
- new quadrature, mapping, `CellValues`, gradient, or spatial-coordinate operations for
  algebraic variables;
- arbitrary non-coordinate algebraic bases or prescribed component values stored in the
  variable descriptor;
- a new nonlinear or time-dependent 0D solver;
- new name-based constraint types for algebraic variables;
- a new threading scheduler, coloring algorithm, or reduction framework;
- coupling-aware Metis graph construction;
- replacing the existing ordinary cell `coupling=` API;
- generalized two-cell interface/DG coupling beyond the existing Ferrite interface
  machinery; and
- caching or allocation-elimination APIs before the basic design is implemented and
  profiled.

## 15. Resulting conceptual model

The final model has three independent pieces:

```text
AlgebraicVariable
    what typed value and coordinate subspace the DOFs represent

Coupling descriptor
    on which entities and between which test/trial variables entries may occur

User assembly kernel
    the actual weak-form contribution evaluated on those entities
```

This separation preserves the useful part of PR #1422--domain-free global numbering--while
adding the missing infrastructure required for typed values, boundary-only support, sparse
3D--0D coupling, ordinary square local assembly, and multiple algebraic variables. It also
keeps PR #1395 conceptually available for the distinct case of a spatially constant field
defined over a physical domain.


</details>

lijas:
> I could make an example implementation just to show how I think it could be done, and maybe it is easier to discuss from there. "Global dofs 3"...

Please feel free to take over this PR or open a new one :P I have not really embraced AI yet, and I spent hours yesterday evening writing tests... I think It will be more efficient if we let Claude do that xD 

I like your suggestions. Here are some of my thoughts:

- The point of allowing e.g. AlgebraicVariable{4}() is if someone has to enforce a constraint with four lagrange parameters (that live in R^4). 
- I like the changes to add_system_variable_entries! to make it more similar current API with coupling matrix etc (and that it can be created directly in allocate_matrix(..). But I guess in this case:
```
cell_coupling = CellCoupling(
      dh, cells;
      fields = (:u, :λ),
      coupling = Bool[
          0 1
          1 1
      ],
  )
```
the uu-block will always be zero, because that connectivity is already handled by the standard cell coupling? Perhaps it should not be called CellCoupling in that case? Also, we dont really have to pass `dh` as argument here, since the dofhandler is already an argument in allocation_matrix(dh, ...).
- When it comes to extending function_value with `function_value(dh, a, :p0)` or creating a `algebraic_value`, I think I would prefer the approach you suggested earlier with having a GlobalConstant Interpolation and you can pass to CellValues. I think this matches the current work flow with how element routines are typically written flow much better. Otherwise the user has to start passing the dofhandler to the element routines which feels off.

KristofferC:
> The point of allowing e.g. AlgebraicVariable{4}() is if someone has to enforce a constraint with four lagrange parameters (that live in R^4).

Isn't this just four separate variables? What makes them go "together". Is there an example weak form?

> the uu-block will always be zero

Yes, I am not sure if the block format here is the best or if there should be something like

```
cell_coupling = CellCoupling(
      dh, cells;
      [:u => :λ, :λ => :u, :λ => :λ]
  )
```

> Perhaps it should not be called CellCoupling in that case?

Well, it talkes about how the algebraic variables couple to fields and other alg variables over the cells (as opposed to `FacetCoupling` that talks about how they couple on facets).

> Otherwise the user has to start passing the dofhandler to the element routines which feels off.

I would say you compute the value outside the element routine and pass it in. Having a function that computes the value at all seems quite useful, other wise you have to transform the `a[algebraic_variable_dofs(dh, :sigma_var)` (a `Vector{Float64}`) into a SymmetricTensor yourself.

lijas:
> Isn't this just four separate variables? What makes them go "together". Is there an example weak form?

I am thinking one might have a constraint $g$ with four rows? So you have $\lambda \cdot g$ where $\lambda \in R^4$. But I dont have any concrete example so lets only use tensors for now.

> cell_coupling = CellCoupling(
>       dh, cells;
>       [:u => :λ, :λ => :u, :λ => :λ]
>   )

This makes a bit more sense to me, and removes the ambiguity for the uu-block, but I am fine with both :) 

Regarding function_value for the AlgebraicVariable:
In my mind we still need to access the shape function of the algebraic variable, and that is most easily done via the CellValues. If we have $\lambda \in R^3$, then we want express this in terms of sum of shape functions $\lambda = \sum_i^3 e_i * \lambda_i$ so we can implement the element routine like we typically implement them (like you did with sigma_bar in #1395 and my assemble_kuσ function in this PR). And in this case, it is very cheap to evaluate the function value? Ok, now I see that we have to do it once per element, but IMO I think that fine. But I guess if the user wants to evaluate the function value outside the element routine, it could be nice with a utility function like function_value(dh, a, :p)

```
function assemble_kuL(cv_u, cv_L, udofs, Ldofs)

    #Eval L (lambda) once (any gauss point is fine)
    L = function_value(cv_L, 1, Ldofs)
      ...
      for i in 1:n_L_dofs
           shape_value(cv_L, qp, i) # we need shape value.
```

KristofferC:
> In my mind we still need to access the shape function of the algebraic variable

Well, you don't really have shape functions, right (the variables do not have any spatial domain)? But for derivatives you want someting that is kind of similar to a shape function. In the plan above, this was written as:

----

> Linearization also needs the coordinate direction associated with each active scalar DOF.
These directions play the algebraic role of constant shape functions: a scalar has direction
`1`, while a vector or tensor has one constant direction per active component. They have no
spatial argument, interpolation, or gradient. The minimal operation is:
> ```julia
> E_i = algebraic_basis_value(algebraic_variable(dh, :sigma_bar), i, eltype(a))
> ```
> Here `i` indexes active DOFs, and the result is a full typed value with a `1` in the
corresponding declared component and zeros elsewhere. It is the derivative of the typed
value with respect to coefficient `i`. Inactive components have no basis direction because
they own no DOF. This replaces a `GlobalConstant` `CellValues` object used only to construct
tensor basis directions.
> No algebraic `function_gradient`, mapping, quadrature, or `CellValues` integration is added.

-----------

A side note, if you want to do AD on your element routine it is nicer to calculate the function value (even for an algebraic variable) inside it so there should maybe be some good support for that...

lijas:
Sorry, I did not read that part.
Ok, i kind of like that approach because you dont have to supply a dummy quadrature rule to the CellValues.

My only suggestion would be to have function_value(::AlgebraicVariable{T}, dofvals::Vector) to decouple it from the dof handler a bit.

termi-official:
> `function_value(::AlgebraicVariable{T}, dofvals::Vector) `

I don't understand this part, sorry. So the idea is to store the dof index in the object you pass into the dof handler? I.e.

```
p1var = AlgebraicVariable{Float64}(1) # scalar at dof index 1
p2var = AlgebraicVariable{Float64}(2) # scalar at dof index 1
add!(dh, :p0, p1var)
add!(dh, :p1, p2var)
close!(dh)
u = rand(2)
function_value(p2var, u)
```

KristofferC:
I think `dofvalues` here is already extracted from the "global unknown vector" (so it already only contains the dofs that pertain to the algebraic variable).

lijas:
#1482

PR #1423, opened 2026-07-28
docs: zero local matrices/vectors outside the element routine

KristofferC:
Most tutorials follow the pattern of resetting the local arrays (`Ke`, `fe`, `re`, ...) in the assembly loop and letting the element routine only *accumulate* into them. A few examples did the `fill!`ing inside the element routine instead, which is inconsistent and a bad thing to copy from the docs (it makes the element routine unusable for accumulating multiple contributions into the same local arrays).

Moved the zeroing out to the caller in:

- `reactive_surface.jl` (`assemble_element_mass!`, `assemble_element_diffusion!`)
- `dg_heat_equation.jl` (`assemble_element!`, `assemble_interface!`, `assemble_boundary!`)
- `hyperelasticity.jl` (`assemble_element!`)
- `quasi_incompressible_hyperelasticity.jl` (`assemble_element!`)
- `threaded_assembly.jl` (`assemble_cell!`, including the commented-out `tforeach` variant)

The rest of the tutorials, gallery and how-to examples were checked and already zero in the assembly loop.

All touched examples were run locally and give unchanged results (the reference `norm(u)` checks in `dg_heat_equation` and `hyperelasticity` pass).

### DIFF
--- a/docs/src/literate-gallery/quasi_incompressible_hyperelasticity.jl
+++ b/docs/src/literate-gallery/quasi_incompressible_hyperelasticity.jl
@@ -190,11 +190,9 @@ end;
 # The function to assemble the element stiffness matrix for each element in the mesh now has a block structure like in
 # `incompressible_elasticity`.
 function assemble_element!(Ke, fe, cell, cellvalues, mp, ue, pe)
-    ## Reinitialize cell values, and reset output arrays
+    ## Reinitialize cell values
     ublock, pblock = 1, 2
     reinit!(cellvalues, cell)
-    fill!(Ke, 0.0)
-    fill!(fe, 0.0)
 
     n_basefuncs_u = getnbasefunctions(cellvalues.u)
     n_basefuncs_p = getnbasefunctions(cellvalues.p)
@@ -270,6 +268,8 @@ function assemble_global!(
         @assert size(global_dofs, 1) == nu + np # sanity check
         ue = w[global_dofsu] # displacement dofs for the current cell
         pe = w[global_dofsp] # pressure dofs for the current cell
+        fill!(ke, 0.0)
+        fill!(fe, 0.0)
         assemble_element!(ke, fe, cell, cellvalues, mp, ue, pe)
         assemble!(assembler, global_dofs, ke, fe)
     end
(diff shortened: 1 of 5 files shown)


termi-official:
> it makes the element routine unusable for accumulating multiple contributions into the same local arrays

I think we are missing that explanation in the docs. Should we add it to the heat tutorial? PR is good otherwise.

KristofferC:
I mean, unless you do actually need it, it doesn't matter where you put it. And if you do need it you will put it outside. It's just aesthetically nice to do it the more general way. Not sure it really need docs.

termi-official:
Indeed a design decision, but since we opt for a design where we show this consistently, it might be worth to leave a line to educate new users why this can be beneficial.

KristofferC:
We can have a devdocs tutorial style where we write down some principles we use to make the tutorials aesthetically nice and consistent. But I don't think there has to be user-facing docs for exactly where to zero the global arrays.

termi-official (approved):

docs/src/literate-gallery/quasi_incompressible_hyperelasticity.jl
```diff
@@ -190,11 +190,9 @@ end;
 # The function to assemble the element stiffness matrix for each element in the mesh now has a block structure like in
 # `incompressible_elasticity`.
 function assemble_element!(Ke, fe, cell, cellvalues, mp, ue, pe)
-    ## Reinitialize cell values, and reset output arrays
+    ## Reinitialize cell values
     ublock, pblock = 1, 2
     reinit!(cellvalues, cell)
```
fredrikekre, line 195: We are also inconsistent with this (reinit outside/inside).

fredrikekre:
Isn't it more logical to do it inside? I think I like that better because then the element routine becomes more complete in a way, i.e. you can review the element routinte in isolation for correctness?

KristofferC:
The comment popped up literally a second after I clicked merge..

But I don't think it is more logical on the inside. You are accumulating contributions. Other parts (that you are not aware of) could potentially also accumulate into it, or?

fredrikekre:
Yea I guess

fredrikekre:
Looks like it broke CI though? https://github.com/Ferrite-FEM/Ferrite.jl/actions/runs/30659633867/job/91252481612

fredrikekre:
No filling with 0 before https://github.com/Ferrite-FEM/Ferrite.jl/blob/ea273ca736febbfd18ab41c2cfe1d62bc13afa9b/docs/src/literate-howto/threaded_assembly.jl#L342 I think.

PR #1426, opened 2026-07-29
Correct transpose error in `_condense!`

KnutAM:
Discovered (by 🤖) in a downstream package causing a failure for stiffness matrix.

Fixes there and looks correct, but need to find a clear test to show that this is correct before merging.

(Feel free to use more powerful bots (or humans) to directly add tests to this PR, or make a separate)

### DIFF
--- a/test/test_constraints.jl
+++ b/test/test_constraints.jl
@@ -334,7 +334,7 @@ end
         close!(ch)
 
         C, _ = Ferrite.create_constraint_matrix(ch)
-        K = sparse(reshape(1.0:121.0, 11, 11))
+        K = sparse(reshape(1.0:ndofs(dh)^2, ndofs(dh), ndofs(dh)))
         Kcondensed = C' * K * C
         apply!(K, ch)
         @test K[ch.free_dofs, ch.free_dofs] == Kcondensed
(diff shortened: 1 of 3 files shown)


test/test_constraints.jl
```diff
@@ -327,6 +327,19 @@ end
...
     close!(dh)
 
+    @testset "nonsymmetric matrix condensation" begin
+        ch = ConstraintHandler(dh)
+        add!(ch, AffineConstraint(1, [3 => 2.0], 0.0))
+        add!(ch, AffineConstraint(2, [4 => 3.0], 0.0))
+        close!(ch)
+
+        C, _ = Ferrite.create_constraint_matrix(ch)
+        K = sparse(reshape(1.0:121.0, 11, 11))
```
KnutAM: Nice and easy 👍 
Maybe 
```suggestion
        K = sparse(reshape(1.0:ndofs(dh)^2, ndofs(dh), ndofs(dh)))
```

lijas:
Omg, it feels like this should affect all affine constraints? Strange that it has not been detected before.

termi-official:
@koehlerson AMR fine?

KnutAM:
> Omg, it feels like this should affect all affine constraints? 

only if K is unsymmetric, which is probably why not detected

koehlerson:
> @koehlerson AMR fine?

constraints and tested weak forms all symmetric I guess

PR #1428, opened 2026-07-29
Fix asymmetric interface_coupling in sparsity pattern creation

fredrikekre:
Asymmetric interface_coupling masks now follow the same convention as the
cell coupling: interface_coupling[i, j] = true creates entries in the
blocks (rows of field i) x (columns of field j) for both orientations of
every interface. Previously a single mask entry created entries in both
the (i, j) and the transposed (j, i) block, and the result could depend
on the internal orientation of the interface. Symmetric masks, for which
the two conventions coincide, produce identical patterns as before.

This is fixed by i) emitting only the mask-sanctioned direction for each
orientation of the interface (previously both directions were emitted
from a single mask lookup), and ii) selecting the row/column cells by
the orientation loop index instead of the subdofhandler index, which
never swapped the orientation when both cells belong to the same
subdofhandler (every pair was emitted twice, with the duplicates
silently removed by add_entry! dedup). As a side effect the interface
work is halved for same-subdofhandler meshes.

The mask is now expanded to a rectangular local matrix per
(row cell subdofhandler, column cell subdofhandler) pair, since with
only the mask-sanctioned direction emitted, entries between fields that
do not exist in both subdofhandlers must be gated with the correct local
layout on each side (previously the reverse emission covered these, with
transposed gating).


fredrikekre:
This was found while working on #1397 but extracted out as a separate bug fix.

Issue #1103, opened 2024-11-06
Interface coupling is broken in sparsity patterns for mixed grids


AbdAlazezAhmed:
kinda related to #1102  and #850 . MWE:
```julia
grid = generate_grid(Hexahedron, (1,1,1))

nodes = grid.nodes
cells = grid.cells

push!(nodes, Node((2.0, -1.0, -1.0)))
push!(nodes, Node((2.0, -1.0, 1.0)))

cells = vcat(cells, Wedge((2,9,4,6,10,8)))

grid_mixed = Grid(cells, nodes)

ip1 = DiscontinuousLagrange{RefHexahedron, 1}()
ip2 = DiscontinuousLagrange{RefPrism, 1}()

dh = DofHandler(grid_mixed)

sdh1 = SubDofHandler(dh,  Set([1]))
sdh2 = SubDofHandler(dh,  Set([2]))

add!(sdh1, :u, ip1)
add!(sdh2, :u, ip2)

close!(dh)
```
```julia
julia> allocate_matrix(dh)
14×14 SparseArrays.SparseMatrixCSC{Float64, Int64} with 100 stored entries:
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
 0.0  0.0  0.0  0.0  0.0  0.0   ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅    ⋅ 
  ⋅    ⋅    ⋅    ⋅    ⋅    ⋅   0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
[...]

PR #1432, opened 2026-07-30
Make interface_coupling self-sufficient for interface assembly

fredrikekre:
Declaring interface_coupling[f1, f2] now means that, for every interface,
entries exist for every (test dof of f1, trial dof of f2) pair within the
union of the dofs of the two cells sharing the interface. Follow-up to
the orientation/block-semantics fix in #1428.

Previously the interface pass skipped every pair where either dof is
shared between the two cells (continuous interpolations) and never
considered pairs where both dofs belong to the same cell, so those
entries existed only when the cell coupling provided them. Interface
terms with nonzero values at shared dofs (e.g. flux-type couplings of
continuous fields, or any jump/average product expanding into same-side
terms) could then hit missing sparsity pattern entries when a restricted
cell coupling was used.

The interface pass now loops over all four (row cell, column cell)
combinations of an interface and adds every masked pair, with duplicates
against the cell pass merged by add_entry! as usual. Patterns built with
full cell coupling (the default coupling = nothing) are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    assembly](https://ferrite-fem.github.io/Ferrite.jl/stable/howto/threaded_assembly/).
    ([#1417])
 
+### Changed
+ - `interface_coupling` in `allocate_matrix`/`add_sparsity_entries!`/`add_interface_entries!`
+   is now self-sufficient for interface assembly: `interface_coupling[i, j] = true` creates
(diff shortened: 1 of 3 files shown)


fredrikekre:
Performance wise this more or less exactly undoes #1428 but I think we do things correctly after this. Also, the reason for this change is to get further speedups in #1397

fredrikekre:
Okay, I guess this was a perf win after all. Probably removal of the `to_check` bogus dict where we allocate a bunch of small arrays before.

KristofferC:
Did this fix https://github.com/Ferrite-FEM/Ferrite.jl/issues/1103?

PR #1434, opened 2026-07-30
Fix PointEvalHandler corner cases and extrapolation

### LINKED ISSUE #1238: Add extrapolation option to PointEvalHandler
It sometimes happens that the points to evaluate the solution on are barely outside of the element. For many element formulations we can safely extrapolate the solution to these points.

termi-official:
Fixes #1238 #1181


KristofferC (approved):
I looked commit by commit and it makes sense to me. I did not look at the test and I didn't super carefully varify the distance formula for the various geometric primitives.

termi-official:
The distance formulas are now definitely more painful to read, but from my checks they follow the geometric primitives. I do not know why the bot compatified the formulas tho.

PR #1438, opened 2026-07-31
add back missing zeroing in threaded assembly howto

KristofferC:


KnutAM:
~Add test to catch?~
nvm

fredrikekre:
CI fails on master so there are tests but strange that the original PR didn't fail.

KnutAM:
Just saw - but ok, then something is odd

termi-official:
> CI fails on master so there are tests but strange that the original PR didn't fail.

I also noticed unfortunately that GitHub actions and the web view started to be super unreliable over the last months. E.g. I also saw a partially upgraded p4est PR when doing the "pre-final" review and wasted a bit of time there.

PR #1445, opened 2026-08-02
Set version to 1.6.0

fredrikekre:


KristofferC:
When a new release is made Tachometer runs the benchmarks against the previous release, which in this case is broken.

fredrikekre:
```
ERROR: LoadError: LoadError: ArgumentError: snlu! requires the same sparsity pattern as the analyzed matrix
```
Looks like structural zeroes are broken in SciML somewhere, reverting the doc upgrade for now.

PR #1446, opened 2026-08-02
improve incompressible elasticity tutorial and show checkerboard stress in the result

### LINKED ISSUE #369: incompressible_elasticity.jl example needs improvement
I have been looking through the examples to get a better grasp at some of the basics of Ferrite, and while the incompressible elasticity example seem to demonstrate some strong features in mixed elements, the example lacks in clarity. The main issues seem to me to be:

- It isn't clear which PDE is being solved. Other examples actually explain what PDE they solve, but here there is only a refrence to "Cook's Membrane", which (despite being a named problem?) doesn't appear to be easy to look up. For instance it doesn't seem to have a Wikipedia entry. Maybe I'm just not smart enough, but I haven't been able to figure out what is being solved.
- Lines 133 and 134 appear to serve no function(?)
https://github.com/Ferrite-FEM/Ferrite.jl/blob/75a4dab91a5f6ce4e239db68515732f1a1ac579e/docs/src/literate/incompressible_elasticity.jl#L133
- Unclear why lines 129 deserves its own loop and isn't just part of the subsequent loop over the same indices.
https://github.com/Ferrite-FEM/Ferrite.jl/blob/75a4dab91a5f6ce4e239db68515732f1a1ac579e/docs/src/literate/incompressible_elasticity.jl#L129

There might be other things too, like in line 142 where `j` is iterated from `1` to `n_basefunc_u` instead of `1` to `ì` like every other instance. It would seem to me that the symmetrization process overwrites the extra entries anyway.
https://github.com/Ferrite-FEM/Ferrite.jl/blob/75a4dab91a5f6ce4e239db68515732f1a1ac579e/docs/src/literate/incompressible_elasticity.jl#L142

KristofferC:
Shows

image

in the end (which is kind of the point of the whole tutorial.

Also adds the weak form and some discusson.

Fixes https://github.com/Ferrite-FEM/Ferrite.jl/issues/369.

### DIFF
--- a/docs/download_resources.jl
+++ b/docs/download_resources.jl
@@ -33,6 +33,7 @@ fetch_assets(
         variants(
             [
                 "heat_equation", "linear_elasticity", "incompressible_elasticity",
+                "incompressible_elasticity_pressure",
                 "hyperelasticity", "plasticity", "computational_homogenization",
                 "stokes-flow", "dg_heat_equation", "linear_shell",
                 "linear_elasticity_stress", "elastodynamics_modes", "darcy_flow",
--- a/docs/generate_screenshots.jl
+++ b/docs/generate_screenshots.jl
@@ -76,6 +76,7 @@ const POSTRUN = Dict(
 # belong to a scene when uploading a subset.
 const OUTPUTS = Dict(
     "linear_elasticity" => ["linear_elasticity", "linear_elasticity_stress"],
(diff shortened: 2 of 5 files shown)


termi-official (approved):
Major improvement! Thanks for the update!

KristofferC:
Thanks for review

PR #1452, opened 2026-08-04
Fix atomics wiring for block assembler

termi-official:


ext/FerriteBlockArrays.jl
```diff
@@ -70,15 +70,17 @@ end
 Ferrite.matrix_handle(ba::BlockAssembler) = ba.K
 Ferrite.vector_handle(ba::BlockAssembler) = ba.f
 
-function Ferrite.start_assemble(K::BlockMatrix, f; fillzero::Bool = true)
+function Ferrite.start_assemble(K::BlockMatrix, f; fillzero::Bool = true, atomic = false)
```
KristofferC: This might need a `Base.@constprop :aggressive`

src/arrayutils.jl
KristofferC: Need to update this docstring?

src/arrayutils.jl
```diff
@@ -24,22 +24,65 @@ Fallback: `A[i, j] += v`.
 """
 addindex!
 
-function addindex!(A::AbstractMatrix{T}, v, i::Integer, j::Integer) where {T}
-    return addindex!(A, T(v), Int(i), Int(j))
+function addindex!(A::AbstractMatrix{T}, v, i::Integer, j::Integer, ::Val{atomic} = Val(false)) where {T, atomic}
```
KristofferC, line 28: Personally, I would rather have `addindex!` and `addindex_atomic!` than a `Val` value, but that might be a matter of taste.

test/blockarrays.jl
```diff
@@ -39,7 +39,7 @@ using Ferrite, BlockArrays, SparseArrays, Test
     @test K == KB
 
     # Zeroing out in start_assemble
-    assembler = start_assemble(K, f)
+    assembler = start_assemble(K, f; atomic = false)
```
KristofferC, line 42: Is this meant to be here?

lijas:
While you are at it, can you update block arrays such that it can also assemble rectangular matrices? :) Like you did in #1279

termi-official:
> While you are at it, can you update block arrays such that it can also assemble rectangular matrices? 

Will do as I need it, but in a separate PR. Let me fix the atomics first. Note that constraint handler apply is also not functional for block matrices yet.


I will answer Kristoffer tomorrow. Thanks for taking a detailed look!

src/arrayutils.jl
termi-official: I can change this if you want, but here my reasoning first why I merged them:

For the current design we have the downside that we need to pass the atomic around and have the case distinction only at the innermost calls. However, with the addindex_atomic! variant we need to distinguish every time and we have two duplicate addindex paths which need duplication, so the total code is longer.

KristofferC: Yeah, sure, it's a valid argument

test/blockarrays.jl
termi-official: Yes, I wanted to make sure the new interface is picked up in any test. I can also remove it again. Any idea how we could improve the tests here? Maybe some loop over all combinations of start_assemble with the supported arguments?

KristofferC: I just meant that this is the default so adding the kwarg doesn't do anything, or?

termi-official: Not on this branch, indeed. But this is the failure I get downstream when I the latest release or master (as the kwarg is missing).

KristofferC (approved):

PR #1467, opened 2026-08-06
Add magnetostatics tutorial with H(curl) Nedelec elements

fredrikekre:
This is basically the H(div) tutorial with find and replace `div` with `curl` so not sure it is worth adding but putting it up for archival if nothing else. Perhaps the code gallery.


KristofferC:
Restarting to get CI to run (github was borken yesterday)

KristofferC:
> `ERROR: LoadError: RequestError: HTTP/2 404 while requesting https://raw.githubusercontent.com/Ferrite-FEM/Ferrite.jl/gh-pages/assets/magnetostatics-light.png`

Looks like the new figures haven't been pushed to gh-pages?

PR #1466, opened 2026-08-06
Optimize ExclusiveTopology construction and queries

KristofferC:
Benchmark script:

```julia
using Ferrite, BenchmarkTools

grids = [
    "Quadrilateral 40×40" => generate_grid(Quadrilateral, (40, 40)),
    "Triangle 40×40" => generate_grid(Triangle, (40, 40)),
    "Hexahedron 15×15×15" => generate_grid(Hexahedron, (15, 15, 15)),
    "Tetrahedron 10×10×10" => generate_grid(Tetrahedron, (10, 10, 10)),
]

println("== ExclusiveTopology construction ==")
for (name, grid) in grids
    print(rpad(name, 22))
    @btime ExclusiveTopology($grid)
end

hexgrid = last(grids[3])
top = ExclusiveTopology(hexgrid)

println("== getneighborhood (Hexahedron 15×15×15, interior cell) ==")
print("CellIndex:   "); @btime getneighborhood($top, $hexgrid, CellIndex(1000))
print("VertexIndex: "); @btime getneighborhood($top, $hexgrid, VertexIndex(1000, 1))
print("EdgeIndex:   "); @btime getneighborhood($top, $hexgrid, EdgeIndex(1000, 1))
print("FaceIndex:   "); @btime getneighborhood($top, $hexgrid, FaceIndex(1000, 1))

println("== Bulk operations (Hexahedron 15×15×15) ==")
# facetskeleton caches its result, so reset the cache in setup
print("facetskeleton:        ")
@btime Ferrite.facetskeleton($top, $hexgrid) setup = ($top.facet_skeleton = nothing) evals = 1
print("vertex_star_stencils: ")
@btime vertex_star_stencils($top, $hexgrid)
```

Before

```
== ExclusiveTopology construction ==
Quadrilateral 40×40     268.124 μs (56 allocations: 1.11 MiB)
Triangle 40×40          560.734 μs (61 allocations: 2.65 MiB)
Hexahedron 15×15×15     3.325 ms (61 allocations: 7.04 MiB)
Tetrahedron 10×10×10    8.893 ms (70 allocations: 22.40 MiB)
== getneighborhood (Hexahedron 15×15×15, interior cell) ==
CellIndex:     1.302 ns (0 allocations: 0 bytes)
VertexIndex:   35.899 ns (2 allocations: 192 bytes)
EdgeIndex:     715.496 ns (91 allocations: 7.45 KiB)
FaceIndex:     14.068 ns (0 allocations: 0 bytes)
[...]

PR #1468, opened 2026-08-06
Fix interface_coupling behavior in add_interface_entries

fredrikekre:
The first patch just make sure that a passed `interface_coupling` isn't ignored when
`topology` is missing (bugfix).

The second commit can be discussed: it creates the topology internally when it is needed and
when not passed. You can still pass it for performance reasons since you most likely would
need it for the interface iterations anyway. Note that `InterfaceIterator` already created
one internally.

There is of course a cost of doing this but compared to the full pipeline it doesn't really
matter (see also #1466). Here are some benchmarks from running the DG tutorial for
reference:

```
┌───────────────────────┬──────────────────┬──────────────┬──────────────┬──────────────┐
│         step          │ n=40 (6.4k dofs) │  n=80 (26k)  │ n=160 (102k) │ n=240 (230k) │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ solve K \ f           │ 10.1ms (46%)     │ 39.7ms (45%) │ 608ms (74%)  │ 751ms (55%)  │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ allocate_matrix       │ 5.2ms (24%)      │ 23.1ms (26%) │ 119ms (15%)  │ 288ms (21%)  │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ assembly (interfaces) │ 5.2ms (24%)      │ 20.5ms (23%) │ 83ms (10%)   │ 301ms (22%)  │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ ExclusiveTopology     │ 0.6ms (2.7%)     │ 1.5ms (1.8%) │ 5.7ms (0.7%) │ 18ms (1.3%)  │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ assembly (cells)      │ 0.3ms            │ 1.0ms        │ 4.3ms        │ 9.3ms        │
├───────────────────────┼──────────────────┼──────────────┼──────────────┼──────────────┤
│ close!(dh)            │ 0.09ms           │ 0.2ms        │ 0.7ms        │ 2.2ms        │
└───────────────────────┴──────────────────┴──────────────┴──────────────┴──────────────┘
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixes
  - Atomic assembly support for BlockAssembler. ([#1452])
+ - `add_sparsity_entries!` (and thereby `allocate_matrix`) now guarantees that passing
+   `interface_coupling` adds the requested interface entries since forgetting to pass
+   `topology` now throws an error instead of ignoring the requested `interface_coupling`.
+   Calls that pass both keyword arguments behave exactly as before. ([#1468])
 
 ### Added
  - `ExclusiveTopology` now supports grids with mixed reference dimensions (e.g. a 3D grid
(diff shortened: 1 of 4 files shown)


KristofferC (approved):
Makes sense to me, from my understanding at least. Bugfix + create topology when needed, seems reasonable 👍

PR #1469, opened 2026-08-07
RFC: Merge MultiFieldCellValues into CellValues

KristofferC:
Thinking about it, it didn't really make sense to have a `MultiField` version of `CellValues` when that could just be `CellValues`. And if you have one field, then you just don't have to specify the field.


src/FEValues/CellValues.jl
```diff
@@ -47,65 +50,141 @@ values of nodal functions, gradients and divergences of nodal functions etc. in
...
-    return CellValues(copy(get_fun_values(cv)), copy(get_geo_mapping(cv)), copy(cv.qr), _copy_or_nothing(cv.detJdV))
+# Compat constructors for the MultiFieldCellValues alias
+(::Type{MultiFieldCellValues})(qr::QuadratureRule, ip_funs::NamedTuple, args...; kwargs...) = CellValues(qr, ip_funs, args...; kwargs...)
+(::Type{MultiFieldCellValues})(::Type{T}, qr::QuadratureRule, ip_funs::NamedTuple, args...; kwargs...) where {T} = CellValues(T, qr, ip_funs, args...; kwargs...)
+
+# Given the old and new (e.g. copied) unique fun_values tuples, rebuild the NamedTuple
+# such that keys with aliased values remain aliased in the new tuple.
+_rebuild_fun_values_nt(::Nothing, ::Tuple, ::Tuple) = nothing
+function _rebuild_fun_values_nt(nt::NamedTuple, old_fun_values::Tuple, new_fun_values::Tuple)
+    return map(named_fv -> new_fun_values[findfirst(fv -> fv === named_fv, old_fun_values)], nt)
```
termi-official: Is this really compatible with GPU? (CI unfortunately failed on setup already for unrelated reasons)

termi-official (commented):
I think @KnutAM and me had some discussion about this and decided for splitting the structs for now. But I think this is indeed the same fallacy as with the DofHandler/MixedDofhandler pair we ran into. The argument for the split was primarily an attempt to not complicate the CellValues object further. I mean, nothing in the software is stopping us from increasing the capabilities of the CellValues object further and further. It is just that we might hit some point in the future where where it becomes really difficult to add more features to the feature matrix.

I am not strictly against the merge either. Just note that the split is quite lightweight so far and we don't have much maintainance overhead.

KristofferC:
> The argument for the split was primarily an attempt to not complicate the CellValues object further. I mean, nothing in the software is stopping us from increasing the capabilities of the CellValues object further and further. It is just that we might hit some point in the future where where it becomes really difficult to add more features to the feature matrix.

I must say I don't get the argument... What is the worry here? Why would having `CellValues` as a separate specialization of a one-field `MultiFieldCellValues` make anything better?

src/FEValues/CellValues.jl
KristofferC: Good point, should be fixed (which also means that there is now GPU support for multi field cell values).

termi-official:
My argument is the mental overhead of tracking the feature matrix when trying to read the code. Again, this is not a strong argument to push back on merging the code paths, I am not strictly against it. And probably this argument is becoming less and less relevant today with bots helping devs to understand code paths in larger frameworks.

If I am not mistaken, then one example where things get actually more involved is if we think about elements with more advanced interpolations. For example, we have Argyris + Lagrange+Lagrange (thinking about some hypothetical thermochemomechanical bending problem), then merging the caches would complicate the management of the geometry buffers. In contrast, if we pass in here for the argyris CellValues and the two Lagrange elements MultiFieldCellValues, then we have some slight overhead, but the internal management of the buffers (GeometryMapping) is simpler. Does that clarify my (minor) concern here? But again, both are legitimate design decision in my honest opinion, so I do not have any preference for either path (leave separate vs merge).

KristofferC:
> My argument is the mental overhead of tracking the feature matrix when trying to read the code

There is mental overhead in having two separate implementations that do virtually the same thing (except one is specialized to `n=1`) as well. Whenever you do a change to the multi one, you should do it to the single one as well etc.

src/FEValues/CellValues.jl
```diff
@@ -47,65 +50,134 @@ values of nodal functions, gradients and divergences of nodal functions etc. in
...
+
+function CellValues(
+        ::Type{T}, qr::QuadratureRule, ip_funs::NamedTuple, ip_geo::VectorizedInterpolation,
+        ::ValuesUpdateFlags{FunDiffOrder, GeoDiffOrder, DetJdV}
+    ) where {T, FunDiffOrder, GeoDiffOrder, DetJdV}
+
+    geo_mapping = GeometryMapping{GeoDiffOrder}(T, ip_geo.ip, qr)
+    unique_ips = unique(values(ip_funs)) # Not type-stable, but ok for advanced users this should be constructed outside...
+    fun_values = tuple((FunctionValues{FunDiffOrder}(T, ip_fun, qr, ip_geo) for ip_fun in unique_ips)...)
+    fun_values_nt = NamedTuple((key => Val(findfirst(unique_ip -> ip == unique_ip, unique_ips)) for (key, ip) in pairs(ip_funs)))
```
KnutAM, line 129: Was there a specific reason for this change (`NamedTuple` of `FunctionValues` to `NamedTuple` of `Val`s)? (not saying it is a bad change, just curious. I think I chose the old version to avoid one indirection, but probably the compiler removes it in all cases anyways so not a performance difference I guess?)

KristofferC: Yes, to allow GPU compatibility (IIRC)

KristofferC: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1469/changes/282d5c42d4e1a052d33b8413b274be62d9f8afc7

PR #1474, opened 2026-08-08
Support Float16 and Complex eltypes in atomic assembly

KristofferC:
This extends atomic assembly (`start_assemble(K, f; atomic = true)`) from `Float32`/`Float64` to also cover `Float16` and `Complex` of `Float16`/`Float32`/`Float64`.

- **`Float16`**: the same `atomicrmw fadd` llvmcall is emitted with the `half` type; LLVM legalizes it to a 16-bit compare-exchange loop on targets without a native instruction (with the addition inlined in the loop, unlike `Core.Intrinsics.atomic_pointermodify`).
- **`Complex`**: complex addition is component-wise and assembly only accumulates -- no task reads the values until after the task join -- so the real and imaginary parts can be accumulated with two independent atomic `fadd`s. No 16-byte CAS needed.

The primitive stress-tested clean (no lost updates) with 8 threads on Julia 1.10 and 1.12 for all six eltypes, and generates the expected IR (a single `atomicrmw fadd` for reals, two for complex).

Along the way this consolidates duplication left behind by #1452: there were two copies of the llvmcall primitive and two accumulation wrappers (`addindex!` and `_addindex!`) with flipped argument orders. Now there is a single pointer-based `_atomic_fadd!` primitive and a single `addindex!` wrapper; the CUDA extension method is updated to the unified argument order.

The first commit fixes a bug from #1452 (unreleased) in the generic matrix `addindex!` fallback where the atomic branch was inverted: non-atomic calls errored with "Atomic addindex! not supported for matrices." and atomic calls silently performed a plain (racy) `+=`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    common facet dimension across the whole grid. ([#843])
  - Experimental CUDA GPU support for assembly using type-stable, non-allocating element
    routines. ([#1291])
+ - Atomic assembly (`start_assemble(K, f; atomic = true)`) now supports `Float16` and
+   `Complex` of `Float16`/`Float32`/`Float64` as value types, in addition to `Float32`
+   and `Float64`. ([#1474])
 
 ## [v1.6.0] - 2026-08-02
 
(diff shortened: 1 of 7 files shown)


termi-official (approved):
LGTM. Thanks for taking care of the extension!

Since I don't want to keep anything under the rug, the CUDA extension does not support assembly with Complex yet (xref https://github.com/JuliaGPU/CUDA.jl/issues/1994). But it is looks easy to be added upstream, so I would be against adding a guard on our end.

PR #1476, opened 2026-08-10
Propagate atomics through constraint condensation

### LINKED ISSUE #1465: apply_assemble!/apply_local! does not respect atomics
~~This is some funny interaction here and I am not sure how to resolve it right now.~~ The assemblers atomic flag is not correctly propagated into the inner `addindex!`

KristofferC:
## Summary

- forward the assembler's atomic mode to non-local affine constraint writes
- add atomic `addindex!` support for `SparseMatrixCSR`
- add regression tests for matrix and vector accumulation

Closes #1465.

### DIFF
--- a/test/test_constraints.jl
+++ b/test/test_constraints.jl
@@ -1,6 +1,41 @@
 # Imports for parallel (isolated) test execution:
 using LinearAlgebra, SparseArrays, Logging
 
+# Minimal atomic assembler used to verify that `apply_assemble!` propagates its atomic
+# mode into the non-local constraint condensation before regular assembly.
+struct AtomicApplyAssembler{M, V} <: Ferrite.AbstractAssembler{Float64}
+    K::M
+    f::V
+end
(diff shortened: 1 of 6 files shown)


test/test_constraints.jl
```diff
@@ -1620,6 +1655,27 @@ end # testset
...
+    close!(dh)
+    ch = ConstraintHandler(dh)
+    add!(ch, AffineConstraint(1, [3 => 1.0], 0.0))
+    close!(ch)
+
+    dofs = celldofs(dh, 1)
+    K = AtomicProbeArray(zeros(3, 3))
+    f = AtomicProbeArray(zeros(3))
+    assembler = AtomicApplyAssembler(K, f)
+    apply_assemble!(assembler, ch, dofs, ones(2, 2), ones(2))
```
termi-official, line 1671: Should we add another test for the normal `assemble!` to have a regression check on the normal atomic assembly path?

KristofferC: I looked at it and I think we have good coverage.

PR #1482, opened 2026-08-13
Algebraic variables (Global dofs 3)

KristofferC:
This is my proposal for the "global dofs" (here called algebraic variables) feature. I think the best way to review this is to look at the doc, then look at the tutorials and if that seems good only look at the code later because there isn't much point in reading code unless the API feels good. Summarized, the components this is built on is:

- `AlgebraicVariable`, describes a variable/value that does not live on a domain, it couples to other fields via the weak. The type of the variable is defined on creation and it is added to the dof handler so we get dofs for it.
- `algebraic_coupling` passed to `allocate_matrix`, this describes how and where the algebraic variable couples to other fields (or other algebraic variables). Is used to allocate entries in the sparsity pattern and to get the final set of local dof numbers when assembling.
- `layout = LocalDofLayout()` + `local_dofs!(layout, facet, coupling)` + `range_σ = dof_range(layout, :σ̄)` for dof handling in an element
- `algebraic_value` to compute the value based on dof values
- `AlgebraicValues` to mimic `CellValues` that can be used to calculate the algebraic values inside an element

A full (but intentionally simple) example using this could look like:

```julia
using Ferrite

grid = generate_grid(Quadrilateral, (20, 20))

# Pure-Neumann Poisson: -Δu = x₁ with ∂u/∂n = 0 is singular up to a constant;
# a scalar multiplier λ enforcing ∫u dΩ = 0 makes it unique and olvable.
dh = DofHandler(grid)
add!(dh, :u, ip)
add!(dh, :λ, AlgebraicVariable())
close!(dh)

coupling = CellCoupling(dh, 1:getncells(grid); algebraic_coupling = ((:u, :λ),))
K = allocate_matrix(dh; algebraic_couplings = coupling)
f = zeros(ndofs(dh))

cv = CellValues(QuadratureRule{RefQuadrilateral}(2), ip)
Ke = zeros(ndofs_per_cell(dh) + 1, ndofs_per_cell(dh) + 1)
fe = zeros(ndofs_per_cell(dh) + 1)
layout = LocalDofLayout()
for cell in CellIterator(dh)
    reinit!(cv, cell)
    local_dofs!(layout, cell, coupling)
    range_u = dof_range(layout, :u)
    λdof = only(dof_range(layout, :λ))
    fill!(Ke, 0); fill!(fe, 0)
        dΩ = getdetJdV(cv, qp)
        x = spatial_coordinate(cv, qp, getcoordinates(cell))
        for (i, I) in pairs(range_u)
            Ni = shape_value(cv, qp, i)
            fe[I] += x[1] * Ni * dΩ
            for (j, J) in pairs(range_u)
                Ke[I, J] += (shape_gradient(cv, qp, i) ⋅ shape_gradient(cv, qp, j)) * dΩ
            end
            Ke[I, λdof] += Ni * dΩ  # λ ∫ δu dΩ
            Ke[λdof, I] += Ni * dΩ  # δλ ∫ u dΩ
        end
    end
    assemble!(assembler, layout, Ke, fe)
end

a = K \ f
λ = algebraic_value(dh, a, :λ)
```

which gives

```
julia> λ = algebraic_value(dh, a, :λ)
-2.4640734385680637e-16
```

The tutorial has the stress driven homogenization example as well as Stokes updated to use the formulation here.


docs/src/topics/algebraic_variables.md
```diff
@@ -0,0 +1,202 @@
...
+combined with a [`BlockSparsityPattern`](@ref); the
+[stress-driven homogenization tutorial](@ref tutorial-stress-driven-homogenization) shows
+this.
+
+## Scope
+
+Algebraic variables have no interpolation or spatial evaluation. Operations that require
+a spatial field, such as Dirichlet conditions, point evaluation, projection, and VTK
+export, do not accept algebraic names. To constrain an algebraic dof, use an
+[`AffineConstraint`](@ref) with [`algebraic_dofs`](@ref).
```
termi-official, line 203: Documenter is not happy about this ref.

docs/src/literate-tutorials/stokes-flow.jl
```diff
@@ -538,6 +542,141 @@ main()
 
 # The resulting magnitude of the velocity field is visualized in *Figure 1*.
 
+# ## [Alternative: retaining the Lagrange multiplier](@id stokes-multiplier)
```
termi-official, line 545: I guess the next logical step is to make the block assembly more generic. We typically have some efficient algorithms to invert these matrices when grouping the global dofs and the field dofs into blocks. See e.g. https://github.com/JuliaHealth/Thunderbolt.jl/blob/7c93f732bf8863940762757160e42068ce159753/src/solver/linear/schur.jl . Not for this PR tho, just something to keep in mind for subsequent work. We might want to showcase this here alongside Uzawa iterations (which is just a few lines if we have the block matrices).

docs/src/literate-tutorials/stokes-flow.jl
termi-official, line 1: I guess we should update the tutorial index with more keywords on this to make this more discoverable.

docs/src/literate-tutorials/stokes-flow.jl
```diff
@@ -538,6 +542,141 @@ main()
...
+    add!(dh, :u, ipu)
+    add!(dh, :p, ipp)
+    add!(dh, :λ, AlgebraicVariable())
+    close!(dh)
+    ## All external boundaries
+    Γ = union(
+        getfacetset(grid, "Γ1"), getfacetset(grid, "Γ2"),
+        getfacetset(grid, "Γ3"), getfacetset(grid, "Γ4"),
+    )
+    coupling = FacetCoupling(dh, Γ; algebraic_coupling = ((:p, :λ),))
```
termi-official: Two questions on this. 
1. Does the nested tuple trigger potential recompilation for each new combination and shouldn't this be just Vector{Vector{Symbol}}?
2. Should we allow a variant where we just pass one tuple/Vector?

test/test_algebraic_variables.jl
```diff
@@ -0,0 +1,974 @@
...
+    grid = generate_grid(Quadrilateral, (3, 3))
+    dh = DofHandler(grid)
+    add!(dh, :u, Lagrange{RefQuadrilateral, 1}())
+    add!(dh, :p0, AlgebraicVariable())
+    close!(dh)
+    p0dof = only(algebraic_dofs(dh, :p0))
+    cpl = CellCoupling(dh, 1:getncells(grid); algebraic_coupling = ((:u, :p0), (:p0, :p0)))
+    ch = ConstraintHandler(dh)
+    add!(ch, Dirichlet(:u, getfacetset(grid, "left"), (x, t) -> 0.0))
+    # An AffineConstraint whose master is an algebraic dof
```
termi-official, line 568: I would suggest to add a regression guard on AffineConstraints on non-algebraic dofs too for a case where we have algebraic variables present to show that the column/row for the algebraic dof is really correctly incorporated (in case we e.g. want to optimize the elimination code paths in the future).

docs/src/topics/algebraic_variables.md
```diff
@@ -0,0 +1,202 @@
...
+
+```julia
+assembler = start_assemble(K, f; atomic = true)
+```
+
+A variable coupled to a field over the whole domain gives the matrix a few dense rows and
+columns. This is inherent to global coupling and nothing to fix at the sparsity level.
+If the solver should exploit the structure instead (e.g. eliminating the small algebraic
+block with a Schur complement, or a block preconditioner), renumber with
+[`DofOrder.FieldWise`](@ref) to group the algebraic dofs into their own block, optionally
```
termi-official: Uhm, the examples (stress driven homogenization `@test algebraic_dofs(dh, :εbar) == collect((ndofs(dh) - 2):ndofs(dh)) #src`) suggest that this is given without renumbering already. I also feel like we are missing a demo for how to do this for e.g. stokes, where the first block has :v and :p and the second block the global constraint dof.

termi-official (commented):
First review round on the docs (and a drive-by test thing I noticed)

docs/src/literate-tutorials/stokes-flow.jl
```diff
@@ -538,6 +542,141 @@ main()
...
+# integral, but assembled directly into the matrix blocks instead of being turned into
+# an `AffineConstraint`. (The augmented local matrix also contains (zero) entries
+# coupling `:u` and `:λ`, which is fine: assembling an explicit zero into an entry that
+# is missing from the sparsity pattern is allowed.)
+
+function assemble_multiplier_terms!(assembler, dh, coupling, fvp, Γ)
+    nl = ndofs_per_cell(dh) + 1
+    Ke = zeros(nl, nl)
+    fe = zeros(nl)
+    layout = LocalDofLayout()
```
termi-official: Shouldn't we make this something like `LocalDofLayout(sdh, coupling)` and `local_dofs!` a pure query on this? Our assembly loop here assumes a constant number of dofs per element so far to get good performance. Furthermore, to make this feature GPU compatible we need to get rid of mutable structs in the assembly loop. We also need to make sure that `resize!` calls on the GPU are unreachable and error/throw is behind a function barriere (which I need to override on the GPU).

src/Dofs/algebraic_variables.jl
```diff
@@ -0,0 +1,237 @@
...
+```
+
+See also [`AlgebraicValues`](@ref) and [`algebraic_value`](@ref).
+"""
+struct AlgebraicVariable{V, N, CI}
+    active_components::NTuple{N, CI}
+end
+
+function AlgebraicVariable{V}(; active_components::Union{Nothing, Tuple} = nothing) where {V}
+    components = _canonicalize_active_components(V, active_components)
```
lijas: We could handle active components via the ConstraintHandler instead. That could be more generic since you could set the algebraic variables to specific values also. Do you have any use case for active/inactive components in mind?

src/Dofs/DofHandler.jl
```diff
@@ -289,6 +308,68 @@ function n_components(dh::DofHandler, field_idxs::NTuple{2, Int})
...
+    return idx
+end
+
+"""
+    algebraic_dofs(dh::DofHandler, name::Symbol)
+
+Return a copy of the global dof numbers of the algebraic variable `name`.
+"""
+function algebraic_dofs(dh::DofHandler, name::Symbol)
+    isclosed(dh) || error("the DofHandler must be closed before dofs can be queried")
```
lijas, line 350: add `algebraic_dofs!(dofs, dh, name)` to by symmetric with `celldofs!(dofs, dh, field)`?

src/Dofs/DofHandler.jl
```diff
@@ -289,6 +308,68 @@ function n_components(dh::DofHandler, field_idxs::NTuple{2, Int})
...
+    return copy(dh.algebraic_dofs[_algebraic_variable_index(dh, name)])
+end
+
+"""
+    algebraic_value(dh::DofHandler, a::AbstractVector, name::Symbol)
+
+Reconstruct the typed algebraic value `name` from the global solution vector `a`.
+"""
+function algebraic_value(dh::DofHandler, a::AbstractVector, name::Symbol)
+    isclosed(dh) || error("the DofHandler must be closed before values can be reconstructed")
```
lijas: I guess this could be a nice utility function, but to me this feels a bit like having a function like `function_value(dh, \xi, a, :u) `. I dont think we should include this.

lijas (commented):
Very nice. I downloaded the code an played around with it and it felt solid. 

It seems like we have a bit of a different mental model on how to implement the element routine. You like to to assemble the full element matrix at once e.g: `size(ke) = n_u_dofs + n_lambda_dofs x n_u_dofs + n_lambda_dofs`, while I like to assemble them as block matrices separately, i.e. `ke_ul = zeros(n_u_dofs, n_lambda_dofs)`. But this PR allows for both so that is nice. I updated the computational_homogenization tutorial with Neumann boundary conditions, I think it fits with this PR. 

I added a two small comments. I did not really look at the implementation at this point.

docs/src/literate-tutorials/stokes-flow.jl
KristofferC: Thanks for the comments, this is the part of the feature I am most unsure about how to handle.

KristofferC: I thought about it and decided to just delete the whole thing. https://github.com/Ferrite-FEM/Ferrite.jl/pull/1482/commits/d8588c95e52cde956aa5ac03caf7435c890512e0

termi-official: Yea, I also thought so. A dof layout feature seems to be orthogonal to the algebraic dofs.

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -143,6 +143,8 @@
 # \bar{\boldsymbol{\varepsilon}} \cdot [\boldsymbol{x}^+ - \boldsymbol{x}^-].
 # ```
 #
+# **Neumann boundary conditions**
+#  TODO
```
termi-official: TODO is missing the TODO.

src/Dofs/algebraic_variables.jl
KristofferC: I added it to avoid the user having to think about the "linear" order of the dofs when it comes to eg SymmetricTensor. With this you could use the actual index instead. But I can remove it for now.

docs/src/literate-tutorials/stokes-flow.jl
KristofferC: 1. I did some tweaks so compile time should not he a problem 
2. It is allowed

KristofferC:
I think I addressed all the comments as well as fixing all spelling errors from WIP commit so that CI passes.

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -305,19 +347,18 @@ Ei = 10 * Em;
 # we want to solve the system 3 times, once for each macroscopic strain component, we
 # assemble 3 right-hand-sides.
 
-function doassemble!(cellvalues::CellValues, K::SparseMatrixCSC, dh::DofHandler, εᴹ, Ei, Em)
+function assemble_kuu!(cellvalues::CellValues, K::SparseMatrixCSC, dh::DofHandler, εᴹ)
```
termi-official: Why is this now accessing global variables?

docs/src/literate-tutorials/computational_homogenization.jl
```diff
@@ -344,12 +385,50 @@ function doassemble!(cellvalues::CellValues, K::SparseMatrixCSC, dh::DofHandler,
     return f
 end;
 
+# For the problem statement with Neumann boundary conditions, we must additionally assemble
+# TODO, explain this more.
```
termi-official: Another open TODO

termi-official (commented):
Just missing some polish on the docs from my side. Other than that this new feature looks really awesome!

src/Dofs/DofHandler.jl
KristofferC: We could do that but you can it outside the element routine so there isn't really any need

docs/src/literate-tutorials/computational_homogenization.jl
KristofferC: @lijas what's the plan for this part?

lijas: Delete that commit and I can complete it in another PR :)

KristofferC:
Based on suggestion by @lijas I removed the `dh` argument to the coupling (making that struct pure data) and instead doing the validation when creating the matrix.

src/Dofs/algebraic_variables.jl
```diff
@@ -0,0 +1,194 @@
...
+            "AlgebraicVariable value shapes must not fix the coefficient scalar type: " *
+                "use e.g. `AlgebraicVariable{$(V.name.wrapper){$(join(V.parameters[1:2], ", "))}}()` instead of `AlgebraicVariable{$(V)}()`."
+        )
+    end
+    error(
+        "unsupported value shape `$(V)` for AlgebraicVariable. Supported shapes are " *
+            "scalar (`AlgebraicVariable()`), `Vec{dim}`, `Tensor{2, dim}`, `SymmetricTensor{2, dim}`, " *
+            "`Tensor{4, dim}`, and `SymmetricTensor{4, dim}`."
+    )
+end
```
termi-official, line 92: I am not sure whether I understand this error path here. Why do we exclude the concrete types here?

src/Dofs/algebraic_variables.jl
```diff
@@ -0,0 +1,194 @@
...
+    algebraic_basis_value(av::AlgebraicValues, i::Int)
+
+Return the cached basis direction for the `i`th algebraic dof.
+"""
+function algebraic_basis_value(av::AlgebraicValues, i::Int)
+    if !(1 <= i <= getnbasefunctions(av))
+        error("dof index $i out of bounds for AlgebraicValues with $(getnbasefunctions(av)) dof(s)")
+    end
+    return av.basis[i]
+end
```
termi-official, line 182: Where do we need this (or the basis machinery)? I just could find it in the tests so far.

src/Dofs/algebraic_variables.jl
```diff
@@ -0,0 +1,194 @@
...
+struct AlgebraicVariable{V, N, CI}
+    components::NTuple{N, CI}
+end
+
+function AlgebraicVariable{V}() where {V}
+    components = _canonical_components(V)
+    return AlgebraicVariable{V, length(components), eltype(typeof(components))}(components)
+end
+
+AlgebraicVariable() = AlgebraicVariable{ScalarValueShape}()
```
termi-official, line 38: I do not understand the rationale of this design here (the ScalarValueShape part). Isn't it easier to just pass the number type one wants instead of this proxy type?

termi-official (approved):
Looks awesome! I found nothing critical. Just have a few design decision here which I would like to understand (or reflect on).

src/Dofs/algebraic_variables.jl
KristofferC: The number type is not part of the algebraic variable, that comes from e.g.  the number type of the unknowns / sparse matrix.

src/Dofs/algebraic_variables.jl
KristofferC: Same as https://github.com/Ferrite-FEM/Ferrite.jl/pull/1482#discussion_r3813573386

src/Dofs/algebraic_variables.jl
KristofferC: When you do derivatives.

KristofferC:
Let's try this, we can get some fine-tuning in with some more experience with it.

termi-official:
Thanks for the hard work on this everyone! I think we ended up converging to a really good design here.

KristofferC:
Thanks @lijas, @KnutAM, and @termi-official for many very good suggestions.

PR #1486, opened 2026-08-15
Get rid of `NonConformingGrid` as the intermediate datastructure

### LINKED ISSUE #1413: AMR: use ForestBWG directly as the grid (lazy materialization facade, drop NonConformingGrid)
Today a user of AMR juggles **two** objects: the `ForestBWG` they refine, and the
`NonConformingGrid` they get from `creategrid(forest)` and hand to the `DofHandler`. Every
refinement step invalidates the second one, and keeping them in sync is the user's problem:

```julia
refine!(forest, marked)
balanceforest!(forest)
grid = creategrid(forest)      # must not forget this
```

Nothing prevents assembling against a stale `grid` after a further `refine!`.

**Goal:** one object. A `ForestBWG` is passed to `DofHandler`, `ConstraintHandler`, VTK
export and so on, and always answers grid queries for its *current* refinement state. The
explicit `creategrid` step disappears from user code.

### Proposed mechanism: epoch-guarded lazy cache

- `ForestBWG` carries a mutable cache box holding a materialized snapshot plus an `epoch`.
- Grid accessors materialize on demand and stamp the epoch; `refine!` / `coarsen!` /
  `refine_and_coarsen!` / `balanceforest!` bump the epoch and drop the cache.
- A `DofHandler` records the epoch at `close!` and errors on mismatch, so assembling against
  a stale dof distribution is caught rather than silently wrong. A `reclose!(dh)`
  re-distributes dofs against the new epoch instead of forcing a full rebuild.

The snapshot is essentially today's `NonConformingGrid` minus the `<: AbstractGrid`
supertype and minus the set types that are empty in practice (`nodesets`, `vertexsets`) —
i.e. cells, nodes, `conformity_info`, `facetsets`, `cellsets`. The forest *is* the grid, so
`NonConformingGrid` can then be deleted, with constraint dispatch moving off it.

### Why cache-and-invalidate rather than "true" laziness

Assembly touches every cell on every solve, so arrays are the right hot-loop
representation — recomputing a cell by octree descent per access loses to an array load.
The value of the facade is **lifecycle management** (no user-managed snapshot, no stale
grid) rather than avoiding materialization.

### Known hazard: `forest.cells` holds *trees*

Ferrite's `AbstractGrid` fallbacks are field-based (`getcells(g) = g.cells`), and for a
`ForestBWG` that field holds octrees, not cells. This already leaks today —
`getcelltype(::ForestBWG)` returns `eltype(grid.cells)`, i.e. a tree type. Any accessor not
explicitly overridden will therefore return octree internals instead of erroring, which is
the main correctness risk of this change.

The override set is: `getcells` (all variants), `getncells`, `getcelltype`, `getnodes`,
`getnnodes`, `get_coordinate_type`, `get_coordinate_eltype`, `get_node_coordinate`,
`getcellset(s)`, `getfacetset(s)`. `getnodeset`/`getvertexset` return empty (node/vertex set
reconstruction is a known gap in `creategrid` today too). `ExclusiveTopology(::ForestBWG)`
should **error** with a pointer to `facetskeleton(forest)`: it reads `grid.cells` raw, and
the topology of the materialized grid only knows the macro mesh anyway, which silently gives
wrong facet neighbours.

The existing `getcells(forest)` returns octants; repurposing it to the materialized cells
needs an audit of internal and test callers that rely on the current semantics.

### Encouraging finding

No new `AbstractDofHandler` subtype appears to be needed. `__close!`,
`_close_subdofhandler!`, `CellIterator` and `getcoordinates` reach the grid only through
`get_grid`, `getncells`, `getnnodes`, `getcells(grid, i)`, `get_node_coordinate` and
`vertices`/`edges`/`faces` of materialized cells — so accessor overrides on `ForestBWG`
should suffice and the whole `DofHandler` code path is reused unchanged.

### Possible follow-up (out of scope here)

With the facade in place, dof distribution could be made Lnodes-native: for degree-1
Lagrange the element-node matrix `E` built during materialization already *is* `cell_dofs`
up to the `node_map` permutation and field blocking. That would remove the
`vertexdicts`/`edgedicts`/`facedicts` re-identification in `__close!` — the same dict-based
identification the Lnodes port already removed from `creategrid`. Worth keeping in mind when
shaping the cache, but it should not block the facade itself.

koehlerson:
`ForestBWG` can hold essentially all fields that `NonConformingGrid` has in order to show one materialization step. This eases quite some code parts. One example is that you can hand over `ForestBWG` in each step of the finite element code and even have an invalidation mechanism that tells you that a materialization is needed (which is nice in order to avoid errors). Another example are things that are built on top of adaptivity, e.g. https://github.com/Ferrite-FEM/FerriteViz.jl/pull/168

Another nice part is that `ForestBWG` can actually fulfill in a meaningful way the `AbstractGrid` contract. The `NonConformingGrid` is still there, but I'd like to delete it if everyone agrees thats a better way forward.

### DIFF
--- a/docs/src/devdocs/AMR.md
+++ b/docs/src/devdocs/AMR.md
@@ -252,13 +252,43 @@ Ferrite.AMR.inside
 Ferrite.AMR._maximum_size
 ```
 
-## From a forest to a `NonConformingGrid`
+## From a forest to a grid: the materialization facade
 
 The operations above manipulate the forest of octrees (refine, coarsen, balance, neighbour
-lookups). To actually solve a finite element problem we must turn that forest into a concrete
-grid — this is [`creategrid`](@ref Ferrite.AMR.creategrid), which produces a
-`NonConformingGrid`: an ordinary grid plus the *hanging-node constraints* (`conformity_info`)
-that make a conforming finite element field possible.
+lookups). To actually solve a finite element problem the forest must answer *grid* queries —
+cells, nodes, sets — for its current refinement state. The `ForestBWG` does this itself: it
+subtypes `AbstractGrid` and overrides the accessor interface (`getcells`, `getnodes`,
+`getfacetset`, …) to forward to a lazily materialized snapshot
+(`Ferrite.AMR.ForestSnapshot`) kept in an internal cache box
+(`Ferrite.AMR.MaterializedForest`). Every mutator (`refine!`, `coarsen!`,
+`refine_and_coarsen!`, `refine_all!`, `balanceforest!`) bumps the cache's *epoch*
+([`Ferrite.grid_epoch`](@ref)) and drops the snapshot; the next grid query rebuilds it. A
+`DofHandler` records the epoch at `close!` and every per-loop entry point
+(`CellIterator`/`ConstraintHandler` construction, `ndofs`, VTK export) errors on a mismatch,
+pointing to [`reclose!`](@ref), which re-distributes the dofs against the current state:
+
+```julia
+forest = ForestBWG(grid0, b)
+dh = DofHandler(forest); add!(dh, :u, ip); close!(dh)
+while adapting
+    ch = ConstraintHandler(dh); add!(ch, ConformityConstraint(:u)); add!(ch, dbc); close!(ch)
+    # assemble / solve / estimate (facetskeleton(forest)) / mark
+    refine!(forest, marked); balanceforest!(forest)
+    reclose!(dh)
+end
+```
+
+The snapshot is produced by the [`creategrid`](@ref Ferrite.AMR.creategrid) pipeline
+described below. `creategrid` itself remains as a thin transitional wrapper producing a
+detached `NonConformingGrid` (deprecated): an ordinary grid plus the *hanging-node
+constraints* (`conformity_info`, read it through the accessor
+[`Ferrite.AMR.conformity_info`](@ref)) that make a conforming finite element field possible.
+
+Known facade gaps, shared with `creategrid`: node/vertex sets are not reconstructed onto the
+refined grid (the accessors report them empty), `ExclusiveTopology(forest)` errors (use
+[`facetskeleton`](@ref) instead — the
+fine-grid topology is ill-defined on hanging meshes), and `transform_coordinates!(forest, f)`
+errors (the transform would be lost on rematerialization; transform the base grid instead).
 
 !!! warning "`conformity_info` is subject to change"
     `conformity_info` currently stores hanging *vertices* and their master vertices — exactly
--- a/src/Adaptivity/forest.jl
+++ b/src/Adaptivity/forest.jl
@@ -357,7 +357,6 @@ function coarsen!(forest::ForestBWG, cellids::AbstractVector{<:Integer}; require
     isempty(cellids) && return
     cmarked = issorted(cellids) ? cellids : sort(cellids)
     _apply_refine_coarsen!(forest, cmarked, eltype(cmarked)[], require_all_siblings)
-    _invalidate!(forest)
     return
 end
 
@@ -390,7 +389,6 @@ function refine_and_coarsen!(
         cmarked = issorted(coarsen_ids) ? coarsen_ids : sort(coarsen_ids)
         rmarked = issorted(refine_ids) ? refine_ids : sort(refine_ids)
         _apply_refine_coarsen!(forest, cmarked, rmarked, require_all_siblings)
-        _invalidate!(forest)
     end
     balance && balanceforest!(forest)
     return
@@ -419,6 +417,8 @@ function _apply_refine_coarsen!(forest::ForestBWG, cmarked::AbstractVector{<:Int
         end
         offset += n
     end
+    # one bump for the whole logical mutation, however many trees it rewrote
+    _invalidate!(forest)
     return
 end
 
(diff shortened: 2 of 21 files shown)


src/Adaptivity/forest.jl
```diff
@@ -53,6 +113,67 @@ function ForestBWG(grid::Ferrite.AbstractGrid{dim}, b = DEFAULT_MAXLEVEL[dim]) w
...
+@inline function _materialized(forest::ForestBWG)
+    state = forest.mcache.state
+    if state === nothing
+        state = _materialize_snapshot(forest)
+        forest.mcache.state = state
+    end
+    return state
+end
+
+Ferrite.has_hanging_nodes(::ForestBWG) = true
```
termi-official, line 160: To make this future-proof, I we should call this `is_conforming` or `is_nonconforming`, as we can on paper construct meshes (e.g. coming from p-adaptivity) where we are non-conforming, but don't have hanging nodes.

src/Adaptivity/forest.jl
```diff
@@ -53,6 +113,67 @@ function ForestBWG(grid::Ferrite.AbstractGrid{dim}, b = DEFAULT_MAXLEVEL[dim]) w
...
+dof-distributions, plot data, …) keyed on this value: the cache is valid exactly as long as
+`grid_epoch(forest)` is unchanged. This accessor is internal-but-stable API — it will keep
+working across releases, but is deliberately not exported.
+"""
+Ferrite.grid_epoch(forest::ForestBWG) = forest.mcache.epoch
+
+# Invalidate the facade: bump the epoch and drop the snapshot. Called by every mutator.
+function _invalidate!(forest::ForestBWG)
+    mc = forest.mcache
+    mc.epoch += 1
```
termi-official, line 137: Not convinced yet that this epoch system is the way to go, as I do not understand at all why we have some counter on the refinement/coarsening/balance calls.  Typically the libraries either go with some state/phase system or just a dirty flag that the stuff changed.

I guess with this PR we fix the mesh to be a state machine either way, as the other structs like the Dof Handler are also state machines in disguise. The tricky part is now who becomes responsible for which parts? E.g. who regenerates the mesh? If we reclose the Constraint Handler, do we also reclose the Dof Handler? Who manages the refinement and coarsening mappings?

src/Dofs/DofHandler.jl
```diff
@@ -442,16 +491,77 @@ function __close!(dh::DofHandler{dim}) where {dim}
...
+    end
+    # Re-resolve the (epoch-bound) whole-domain cellset against the current grid state and
+    # reset every ncells-sized structure fixed at construction/close time (shared with the
+    # constructor, so the two lists cannot drift apart).
+    ncells = getncells(get_grid(dh))
+    empty!(sdh.cellset)
+    union!(sdh.cellset, 1:ncells)
+    _reset_dof_state!(dh, ncells)
+    __close!(dh)
+    return dh
```
termi-official, line 562: Now we enter some spicy territory and I already had some fun here when using MFEM's AMR. Refinement is easy, we can just take the the cellset for each subdofhandler and push it forward with the refinement maps. Coarsening is a bit more tricky. Users might have refined the grid at some preprocessing step to resolve some interface and then assign to the leaves different subdomains. We need to catch this somehow, as this is an invalid coarsening -- and if we arrive here in `reclose!`, then I would suggest that we throw a hard error for now, or we add some user-callback which can override which subdofhandler should be assigned.

With the new design in this PR it should be straight forward to add the tracker cache to the forest data structure.

src/Adaptivity/forest.jl
```diff
@@ -36,6 +79,23 @@ struct ForestBWG{dim, C <: OctreeBWG, T <: Real} <: Ferrite.AbstractGrid{dim}
...
+
+# The materialized cell type is a pure function of the dimension.
+_materialized_celltype(::Val{2}) = Quadrilateral
+_materialized_celltype(::Val{3}) = Hexahedron
+
+function ForestBWG(
+        cells::Vector{C}, nodes::Vector{Node{dim, T}}, cellsets, nodesets, facetsets,
+        vertexsets, topology::ExclusiveTopology
+    ) where {dim, C <: OctreeBWG, T}
+    MC = _materialized_celltype(Val(dim))
```
termi-official, line 94: Shouldn't this be rather
```suggestion
    MC = _materialized_celltype(C)
```

src/Adaptivity/forest.jl
```diff
@@ -446,14 +573,59 @@ function Ferrite.getcells(forest::ForestBWG{dim, C}) where {dim, C}
...
+end
+
+# The transform would mutate cached snapshot nodes in place and be silently lost on the next
+# rematerialization; transforming the tree corners instead is future work.
+function Ferrite.transform_coordinates!(::ForestBWG, ::Function)
+    error(
+        "transform_coordinates!(forest::ForestBWG, f) is not supported: the transform " *
+            "would be lost on the next rematerialization. Transform the base grid before " *
+            "constructing the ForestBWG instead."
+    )
```
termi-official, line 627: Maybe I am missing something, but can't we just do this call with `forest.nodes`?

src/Adaptivity/forest.jl
```diff
@@ -2349,10 +2524,30 @@ Pipeline:
...
+end
+
+# The `creategrid` pipeline body, emitting the facade's cache entry. Built fresh on every
+# call (never returns the cached snapshot), so `creategrid`'s result shares no state with
+# the facade.
+function _materialize_snapshot(forest::ForestBWG{dim, C, T}) where {dim, C, T}
     node_map = dim == 2 ? node_map₂ : node_map₃
-    celltype = dim == 2 ? Quadrilateral : Hexahedron
+    celltype = _materialized_celltype(Val(dim))
```
termi-official, line 2550:
```suggestion
    celltype = _materialized_celltype(C)
```

docs/src/devdocs/AMR.md
```diff
@@ -252,13 +252,43 @@ Ferrite.AMR.inside
...
+([`Ferrite.grid_epoch`](@ref)) and drops the snapshot; the next grid query rebuilds it. A
+`DofHandler` records the epoch at `close!` and every per-loop entry point
+(`CellIterator`/`ConstraintHandler` construction, `ndofs`, VTK export) errors on a mismatch,
+pointing to [`reclose!`](@ref), which re-distributes the dofs against the current state:
+
+```julia
+forest = ForestBWG(grid0, b)
+dh = DofHandler(forest); add!(dh, :u, ip); close!(dh)
+while adapting
+    ch = ConstraintHandler(dh); add!(ch, ConformityConstraint(:u)); add!(ch, dbc); close!(ch)
```
termi-official, line 274: A questions I immediately have when reading this code is: Why don't we have a `reclose!(ch)`? We can add a vector to track the added constraints and replay what the user provided in `reclose!`, symmetric to what is proposed in the dof hander.

src/Adaptivity/forest.jl
koehlerson: changed it `C` is the tree type here, but I added a dispatch for each tree type such that the tet stuff can work with this too

PR #1488, opened 2026-08-17
Run Julia with 2 threads in CI

fredrikekre:
Since we are starting to utilize multithreading within the package I think we should run CI
with more than one thread. Also add a single job with 1 thread to make sure that path still
works.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 jobs:
   test:
-    name: Julia ${{ matrix.julia-version }} - ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


fredrikekre:
Seems to make running tests a bit slower? Maybe invert and run just a single job with 2 threads then?

termi-official:
> Seems to make running tests a bit slower? Maybe invert and run just a single job with 2 threads then?

The culprit might be oversubscription (https://github.com/Ferrite-FEM/Ferrite.jl/blob/998b60403c0fb1d831bb9c47ec75b4dd0a9f9b67/test/runtests.jl#L27). We should cap this number anyway, as it does not respect system memory.

fredrikekre:
> The culprit might be oversubscription

I doubt it since it isn't much of the code paths in the tests that are multithreaded?

PR #1489, opened 2026-08-19
Constraint support for block systems

### LINKED ISSUE #1460: Constraint application for block systems
Simply not implemented yet. To be honest I am not sure how difficult this will be, as we might need to restructure the constraint machinery a bit.

termi-official:
Closes #1460 . I added some refactoring of the internal assembly and constraints paths to avoid excessive code duplication on the CSR/CSC/Block CSR/Block CSC paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Atomic assembly support for BlockAssembler. ([#1452])
  - `apply_assemble!` now respects the assembler's atomic mode when non-local affine
    constraints write directly to the global matrix and vector. ([#1465])
+ - Assembling into a `BlockMatrix` is no longer orders of magnitude slower than assembling
+   into an unblocked matrix. The `BlockAssembler` cached the cell dofs as `BlockIndex{1}`,
+   which is not a concrete type, so the scatter loop was type unstable and allocated ~8
+   times per matrix entry. It now splits the cell dofs into their blocks and scatters each
+   block through `Ferrite._assemble_inner!`, i.e. the same routine the CSC and CSR
+   assemblers use, and assembles without allocating. ([#1489])
  - `add_interface_entries!` (and thereby `interface_coupling` builds) no longer errors for
    grids where some cells are not covered by any `SubDofHandler`: an uncovered cell has no
    dofs and contributes no interface entries, but the covered side of such an interface
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,9 +47,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    explicit constraint data and index offsets, so the same methods serve a matrix on its own and
    as a block of a blocked matrix, and the BlockArrays extension contains no format specific
    code. As part of this, `apply!` and `apply_zero!` dispatch on `AbstractMatrix` rather than
-   `AbstractSparseMatrix`. ([#1486])
+   `AbstractSparseMatrix`. ([#1489])
  - `apply!` on a `SparseMatrixCSR` now supports affine constraints, which previously threw
-   `"condensation of ::SparseMatrixCSR{...} matrix not supported"`. ([#1486])
+   `"condensation of ::SparseMatrixCSR{...} matrix not supported"`. ([#1489])
 
 ### Performance
  - `create_coloring` is significantly faster: the incidence matrix construction and the
@@ -1448,4 +1448,4 @@ poking into Ferrite internals:
 [#1475]: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1475
 [#1481]: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1481
 [#1490]: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1490
-[#1486]: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1486
+[#1489]: https://github.com/Ferrite-FEM/Ferrite.jl/issues/1489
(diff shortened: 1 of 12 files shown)


termi-official:
I tested for a few days downstream and could not find any issues so far. So this should be good to go from my end @KristofferC .

termi-official:
I added a latent performance bug in [c3898af](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1489/commits/c3898af034546e1ba1b7ec9b72a39d8f9aac2bbb) . I can make a separate PR out of it if you want.

```
┌────────────────────────────────────────────────────────┬───────────────────────────┬────────────────┬───────┐
│                       benchmark                        │          pre-fix          │    post-fix    │ ratio │
├────────────────────────────────────────────────────────┼───────────────────────────┼────────────────┼───────┤
│ BlockMatrix of CSC (Quadrilateral 40×40, u-p)          │ 468 ms / 6 268 613 allocs │ 6.65 ms / 9    │ 70×   │
├────────────────────────────────────────────────────────┼───────────────────────────┼────────────────┼───────┤
│ BlockMatrix of CSR (Quadrilateral 40×40, u-p)          │ 489 ms / 6 268 613 allocs │ 6.75 ms / 9    │ 72×   │
├────────────────────────────────────────────────────────┼───────────────────────────┼────────────────┼───────┤
│ Stokes …×Lagrange{1} BlockMatrix (Quadrilateral 20×20) │ 123 ms / 1 591 640 allocs │ 5.79 ms / 1228 │ 21×   │
└────────────────────────────────────────────────────────┴───────────────────────────┴────────────────┴───────┘
```

termi-official:
The splatting in the benchmarks is not properly optimized away

```
I had reported "reinit! on MultiFieldCellValues allocates ~1.3 allocations per cell." Two errors: @allocated returns bytes, not a count, and I never checked plain CellValues as a control. Measured properly, reinit! is innocent:

iterate only (400 cells)             8 allocs
reinit loop, CellValues              8 allocs
reinit loop, MultiFieldCellValues    8 allocs
stokes_kernel! x400, no reinit       0 allocs
reinit + stokes_kernel! loop         8 allocs

What was actually happening

The 1228 allocations came from assemble_global! in benchmark/helper.jl:

function assemble_global!(K, f, dh, cv, kernel!::F, args...) where {F}
    for cell in CellIterator(dh)
        reinit!(cv, cell)
        kernel!(Ke, cv, args...)   # <-- here
        ...

Splatting a Vararg parameter of two or more elements inside the loop is not optimized away — the call stays a runtime Core._apply_iterate and allocates 3× per cell. Bisected to that single variable, with everything else held fixed:

┌─────────────────────────────────┬────────────────────┐
│            loop body            │ allocs / 400 cells │
├─────────────────────────────────┼────────────────────┤
│ kernel hardcoded, explicit args │ 8                  │
├─────────────────────────────────┼────────────────────┤
│ kernel as ::F, explicit args    │ 8                  │
├─────────────────────────────────┼────────────────────┤
│ kernel as ::F, splatted args... │ 1212               │
├─────────────────────────────────┼────────────────────┤
│ same, but 1 splatted arg        │ 8                  │
└─────────────────────────────────┴────────────────────┘

Independent of the kernel (a no-op kernel behaves identically) and of the values type (plain CellValues too). Inference is not at fault — code_warntype types args::Tuple{UnitRange{Int64}, UnitRange{Int64}} concretely in both cases; it's the optimizer failing to lower the apply. A Ferrite-free minimal reproducer of the same shape is clean, so it needs the surrounding loop's complexity to trigger.

Of the fixes I tried — let-binding the tuple, closing over the args before the loop, @inline at the call site — only a function barrier taking args as an ordinary Tuple works. That's what I committed.

Why it had never surfaced

Every pre-existing kernel takes at most one extra argument (Laplace 0, elasticity 1), which optimizes fine. The Stokes kernel is the first with two (dof_range(dh, :u) and dof_range(dh, :p)), so the new benchmark exposed it. element_sweep! has the same loop shape and no caller with two extra args yet; I gave it the same treatment rather than leave a trap.

Scope

I grepped src/ for the pattern. Every args... there is a deprecation shim, a constructor, or a single-call forwarder — none splat a Vararg inside a loop, so no user-facing Ferrite code is affected. This was purely a benchmark-helper artifact inflating that one entry's allocation count; its timing moved only ~2%, within noise.
```

Is this expected or a performance bug which should be filed to Julia?

Fix is not splatting:

```
diff --git a/benchmark/helper.jl b/benchmark/helper.jl
index 47ee8917c..a9fe82856 100644
--- a/benchmark/helper.jl
+++ b/benchmark/helper.jl
@@ -313,6 +313,10 @@ end
 # Element kernel over a batch of cells, including `reinit!`, i.e. the per-cell work of an
 # assembly loop without the iterator and scatter parts.
 function element_sweep!(Ke, cv, coord_batch, kernel!::F, args...) where {F}
+    # `args` is passed on as a plain tuple, see `assemble_global!` below.
+    return _element_sweep!(Ke, cv, coord_batch, kernel!, args)
+end
+function _element_sweep!(Ke, cv, coord_batch, kernel!::F, args::Tuple) where {F}
     for x in coord_batch
         reinit!(cv, x)
         kernel!(Ke, cv, args...)
@@ -325,7 +329,13 @@ end
 #----------------------------------------------------------------------#
 
 # Full matrix + vector assembly: CellIterator, reinit!, element kernel and scatter.
+# The kernel's extra arguments are handed to the loop as a plain tuple rather than splatted
+# from a `Vararg` parameter: splatting a `Vararg` of two or more elements inside the loop is
+# not optimized away, and the resulting dynamic call allocates three times per cell.
 function assemble_global!(K, f, dh, cv, kernel!::F, args...) where {F}
+    return _assemble_global!(K, f, dh, cv, kernel!, args)
+end
+function _assemble_global!(K, f, dh, cv, kernel!::F, args::Tuple) where {F}
     Ke = zeros(ndofs_per_cell(dh), ndofs_per_cell(dh))
     fe = zeros(ndofs_per_cell(dh))
```

termi-official:
This should go after https://github.com/Ferrite-FEM/Ferrite.jl/pull/1476

KristofferC (approved):

PR #1492, opened 2026-08-21
Improved interface assembly

fredrikekre:


interface-assembly-design.md
termi-official, line 1: Did we get this one as a bonus to the PR? 😆 

I would like to throw another constraint here into the mix, since this PR goes might cause some trouble in its current form: Can we make the interface assembly work on GPU and with AMR? The DH heat could be a starting point. If the interface iterator becomes a problem, then an alternative would be to have a cell iterator carries neighborhood information directly (matrix of max number of neighbors per face \times num faces).

PR #1493, opened 2026-08-21
Add dof functionals information for shape functions

KristofferC:
Adds `Ferrite.dof_functionals(ip)` so boundary conditions and dof distribution can distinguish point values, derivatives, and integral moments. This supports interpolations with several kinds of dofs, including the Hermite elements discussed in [#1391](https://github.com/Ferrite-FEM/Ferrite.jl/pull/1391#discussion_r3647764861).

### Dof descriptions

The query returns one descriptor per local dof, in shape-function order:

- `PointValue()` for function values at a point.
- `PointDerivative(α)` for derivatives in physical coordinates, with multi-index `α` and total derivative order stored in the type.
- `NormalMoment()`, `TangentialMoment()`, and `InteriorMoment()` for integral moments.

For vectorized interpolations, each scalar descriptor is wrapped in `VectorizedFunctional(functional, direction)`. Coordinates and owning entities remain available through the existing coordinate and dof-index queries; vectorized interpolations continue to store coordinates once per scalar dof. Per-entity functional queries are also provided, including `facetdof_functionals`.

### Boundary conditions

`Dirichlet` gains a `functional` keyword. Its default, `PointValue()`, preserves the usual nodal boundary conditions. For an interpolation with derivative dofs, for example:

```julia
Dirichlet(:u, boundary, f; functional = PointDerivative{1})
```

This prescribes all first derivatives. `PointDerivative` selects every derivative, `PointDerivative((1, 0))` selects one derivative, and a tuple of selectors selects their union. The function `f` supplies the prescribed derivative value directly. For vectorized fields, `components` continues to select the vector components.

Selecting integral moments with `Dirichlet` reports an error directing users to `ProjectedDirichlet`. Nodeset constraints and `apply_analytical!` require interpolations whose dofs are all point values.

### Sharing dofs between subdomains

When a field is added to multiple `SubDofHandler`s, the dofs on shared vertices, edges, and faces must have compatible functional layouts and point locations. Only entities carrying dofs on both sides are compared; differences on unshared entities are allowed. When every possible entity pairing is compatible, validation skips the mesh scan.

This changes some previously accepted combinations into errors. For example, sharing edge dofs between quadratic and cubic Lagrange interpolations now errors instead of warning and assigning overlapping global dof numbers. Equal dof counts with different functionals or support points are also rejected.

Moment descriptors do not encode weights or normalization, so matching descriptors alone cannot establish equivalence between different moment bases. Entity-relative derivatives, such as the normal derivatives needed by Argyris elements (#1270), will need a separate descriptor.

### Validation

The interpolation, H(curl)/H(div) moment-duality, dof-distribution, constraint, analytical-assignment, mixed-dof, facet-value, and export tests pass. Regression tests cover mismatches at actual shared vertices, compatible differences at unshared vertices, and scalar/vector derivative boundary conditions. Runic and repository hygiene checks pass.

### DIFF
--- a/docs/src/reference/interpolations.md
+++ b/docs/src/reference/interpolations.md
@@ -23,3 +23,22 @@ BubbleEnrichedLagrange
 CrouzeixRaviart
 RannacherTurek
 ```
+
+## [Dof functionals](@id dof-functionals)
+
+Each local dof of an interpolation has a *dof functional* describing its physical
+meaning. Most interpolations have only [`PointValue`](@ref) dofs, but e.g. H(div)/H(curl)
+interpolations have integral moment dofs, and Hermite-type interpolations have derivative
+dofs. The functional can be used to select which dofs a [`Dirichlet`](@ref) condition
+constrains.
+
+```@docs
+Ferrite.DofFunctional
+PointValue
+PointDerivative
+IntegralMoment
+NormalMoment
+TangentialMoment
+InteriorMoment
+Ferrite.dof_functionals
+```
--- a/docs/src/reference/interpolations.md
+++ b/docs/src/reference/interpolations.md
@@ -40,5 +40,6 @@ IntegralMoment
 NormalMoment
 TangentialMoment
 InteriorMoment
+VectorizedFunctional
 Ferrite.dof_functionals
 ```
--- a/docs/src/reference/interpolations.md
+++ b/docs/src/reference/interpolations.md
@@ -26,8 +26,8 @@ RannacherTurek
 
 ## [Dof functionals](@id dof-functionals)
 
-Each local dof of an interpolation has a *dof functional* describing its physical
-meaning. Most interpolations have only [`PointValue`](@ref) dofs, but e.g. H(div)/H(curl)
+Each local dof of an interpolation has a *dof functional* describing how it evaluates a
+function. Most interpolations have only [`PointValue`](@ref) dofs, but e.g. H(div)/H(curl)
 interpolations have integral moment dofs, and Hermite-type interpolations have derivative
 dofs. The functional can be used to select which dofs a [`Dirichlet`](@ref) condition
 constrains.
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -14,8 +14,9 @@ represent more than one kind of [`Ferrite.DofFunctional`](@ref) (cf.
 [`Ferrite.dof_functionals`](@ref)). The default, `PointValue()`, constrains the
 dofs whose value is the function value at the dof location. Omitted properties act as
 wildcards, e.g. `functional = PointDerivative()` selects all derivative dofs of a
-Hermite-type interpolation while `functional = PointDerivative((1, 0))` selects one
-derivative. For vectorized interpolations `functional` selects the kind of dof (the
+Hermite-type interpolation, `functional = PointDerivative(order = 1)` its first
+derivatives, and `functional = PointDerivative((1, 0))` one derivative. For vectorized
+interpolations `functional` selects the kind of dof (the
 scalar functional) and `components` selects the direction.
 
 The set, `∂Ω`, can be an `AbstractSet` or `AbstractVector` with elements of
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -12,12 +12,12 @@ of `u` are prescribed.
 `functional` selects which dofs the condition constrains, for interpolations whose dofs
 represent more than one kind of [`Ferrite.DofFunctional`](@ref) (cf.
 [`Ferrite.dof_functionals`](@ref)). The default, `PointValue()`, constrains the
-dofs whose value is the function value at the dof location. Omitted properties act as
-wildcards, e.g. `functional = PointDerivative()` selects all derivative dofs of a
-Hermite-type interpolation, `functional = PointDerivative(order = 1)` its first
-derivatives, and `functional = PointDerivative((1, 0))` one derivative. For vectorized
-interpolations `functional` selects the kind of dof (the
-scalar functional) and `components` selects the direction.
+dofs whose value is the function value at the dof location. A functional type acts as a
+wildcard, e.g. `functional = PointDerivative` selects all derivative dofs of a
+Hermite-type interpolation and `functional = PointDerivative{1}` its first derivatives,
+while `functional = PointDerivative((1, 0))` selects a single derivative and a tuple of
+selectors the union of its elements. For vectorized interpolations `functional` selects
+the kind of dof (the scalar functional) and `components` selects the direction.
 
 The set, `∂Ω`, can be an `AbstractSet` or `AbstractVector` with elements of
 type [`FacetIndex`](@ref), [`FaceIndex`](@ref), [`EdgeIndex`](@ref), [`VertexIndex`](@ref),
@@ -53,11 +53,13 @@ struct Dirichlet # <: Constraint
     facets::OrderedSet{T} where {T <: Union{Int, FacetIndex, FaceIndex, EdgeIndex, VertexIndex}}
     field_name::Symbol
     components::Vector{Int} # components of the field
-    functional::DofFunctional # which dofs to constrain (cf. dof_functionals)
+    functional::DofFunctionalSelector # which dofs to constrain (cf. dof_functionals)
     local_facet_dofs::Vector{Int}
     local_facet_dofs_offset::Vector{Int}
 end
-function Dirichlet(field_name::Symbol, facets::AbstractVecOrSet, f::Function, components = nothing; functional::DofFunctional = PointValue())
+function Dirichlet(field_name::Symbol, facets::AbstractVecOrSet, f::Function, components = nothing; functional::DofFunctionalSelector = PointValue())
+    all(_is_selector, _selector_tuple(functional)) ||
+        throw(ArgumentError("`functional` must be a `DofFunctional` instance, type, or tuple of those (got $functional)"))
     return Dirichlet(f, convert_to_orderedset(facets), field_name, __to_components(components), functional, Int[], Int[])
 end
 
@@ -1232,7 +1234,7 @@ function add!(ch::ConstraintHandler{<:Any, Tv, Ti}, dbc::Dirichlet) where {Tv, T
         EntityType = eltype(dbc.facets) # (Facet|Face|Edge|Vertex)Index
         # The selector is matched against the scalar functionals of the base interpolation;
         # `components` selects the direction of a vectorized interpolation
-        if dbc.functional isa VectorizedFunctional
+        if any(s -> s isa VectorizedFunctional || (s isa Type && s <: VectorizedFunctional), _selector_tuple(dbc.functional))
             throw(ArgumentError("`functional` selects the scalar functional of a vectorized interpolation (got $(dbc.functional)); select the direction with `components` instead"))
         end
         fs = dof_functionals(interpolation)
--- a/src/Dofs/ConstraintHandler.jl
+++ b/src/Dofs/ConstraintHandler.jl
@@ -1,6 +1,6 @@
 # abstract type Constraint end
 """
-    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing)
+    Dirichlet(u::Symbol, ∂Ω::AbstractVecOrSet, f::Function, components = nothing; functional = PointValue)
 
 Create a Dirichlet boundary condition on `u` on the `∂Ω` part of
 the boundary. `f` is a function of the form `f(x)` or `f(x, t)`
@@ -8,6 +8,18 @@ where `x` is the spatial coordinate and `t` is the current time,
 and returns the prescribed value. `components` specify the components
 of `u` that are prescribed by this condition. By default all components
 of `u` are prescribed.
+
+`functional` selects which dofs the condition constrains, for interpolations whose dofs
+represent more than one kind of [`Ferrite.DofFunctional`](@ref) (cf.
+[`Ferrite.dof_functionals`](@ref)). The default, [`PointValue`](@ref), constrains the
+dofs whose value is the function value at the dof location. Pass a `DofFunctional`
+subtype to constrain all dofs of that family, or an instance to constrain exactly
+matching dofs, e.g. `functional = PointDerivative` for all derivative dofs of a
+Hermite-type interpolation. Note that all matched dofs at a location are prescribed the
+same value `f(x, t)`, so a family selector matching several distinct dofs (e.g. both
+``\\partial_x`` and ``\\partial_y`` derivative dofs at a vertex) is typically only
+meaningful for homogeneous conditions; prescribe distinct values with one condition per
+exact functional, e.g. `functional = PointDerivative((1, 0))`.
 
 The set, `∂Ω`, can be an `AbstractSet` or `AbstractVector` with elements of
 type [`FacetIndex`](@ref), [`FaceIndex`](@ref), [`EdgeIndex`](@ref), [`VertexIndex`](@ref),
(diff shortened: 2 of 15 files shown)


src/interpolations.jl
```diff
@@ -81,6 +81,115 @@ or `H1Conformity()`, for the interpolation.
...
+
+# Per-entity dof functional signatures of (the base of) an interpolation: a tuple over
+# entities, each element the tuple of DofFunctionals in local entity-dof order.
+function _entity_functionals(ip::Interpolation, entity_indices::Tuple)
+    fs = dof_functionals(ip)
+    return map(dofs -> map(i -> fs[i], dofs), entity_indices)
+end
+vertexdof_functionals(ip::Interpolation) = _entity_functionals(ip, vertexdof_indices(ip))
+edgedof_functionals(ip::Interpolation) = _entity_functionals(ip, edgedof_interior_indices(ip))
+facedof_functionals(ip::Interpolation) = _entity_functionals(ip, facedof_interior_indices(ip))
```
lijas, line 229: I think facetdof_functionals  is missing.

src/interpolations.jl
```diff
@@ -1698,6 +1809,8 @@ hypercubes. For details see the original paper [RanTur:1992:snq](@cite).
 """
 struct RannacherTurek{shape, order} <: ScalarInterpolation{shape, order} end
 conformity(::RannacherTurek) = L2Conformity()
+# Nodal at facet midpoints like CrouzeixRaviart, so the PointValue fallback of
+# dof_functionals applies.
```
lijas: Remove this comment?

src/interpolations.jl
```diff
@@ -1632,6 +1741,8 @@ struct CrouzeixRaviart{shape, order} <: ScalarInterpolation{shape, order}
     CrouzeixRaviart{RefTetrahedron, 1}() = new{RefTetrahedron, 1}()
 end
 conformity(::CrouzeixRaviart) = L2Conformity()
+# The implemented basis is nodal at the facet midpoints, so the PointValue fallback of
+# dof_functionals applies (the classical facet-mean dofs coincide with midpoint values).
```
lijas: Remove this comment?

src/interpolations.jl
```diff
@@ -1773,6 +1886,7 @@ struct VectorizedInterpolation{vdim, refshape, order, SI <: ScalarInterpolation{
     end
 end
 conformity(ip::VectorizedInterpolation) = conformity(ip.ip)
+dof_functionals(ip::VectorizedInterpolation) = dof_functionals(ip.ip) # indexed by base dof number
```
lijas: According to DefElements, each dof has its own functional, so this suggests that length(dof_functionals(ip)) should be equal to getnbasefunctions(ip)?  https://defelement.org/elements/examples/triangle-vector-lagrange-1.html  

Mathematically, I think the vector Lagrange functional are not the same as scalar Lagrange `PointValue`-functionals... they also have a direction associated with them. So I think the PointValue struct should also have a "direction" variable. 

But on closer thought, I am not sure how would this work out with Dirichlet boundary conditions. `Dirichlet(:u, set, f, [1,2,3], functional = PointValue(1))`, ... Now there are two ways to prescribe a component of the vectorized interpolation.

src/Dofs/ConstraintHandler.jl
```diff
@@ -43,11 +55,12 @@ struct Dirichlet # <: Constraint
     facets::OrderedSet{T} where {T <: Union{Int, FacetIndex, FaceIndex, EdgeIndex, VertexIndex}}
     field_name::Symbol
     components::Vector{Int} # components of the field
+    functional::Union{Type{<:DofFunctional}, DofFunctional} # which dofs to constrain (cf. dof_functionals)
```
lijas: Maybe always use the value of the dof functional (PointValue()), and not the type (PointValue) ?

test/interpolation_test_utils.jl
```diff
@@ -68,6 +68,26 @@ function test_interpolation_properties(ip::Interpolation{RefShape, FunOrder}) wh
...
         Ferrite.mapping_type(ip) # Dry-run just to catch if it isn't defined
         @test Ferrite.conformity(ip) isa Union{Ferrite.L2Conformity, Ferrite.HdivConformity, Ferrite.HcurlConformity, Ferrite.H1Conformity}
+
+        # Test F: dof functionals
+        fs = Ferrite.dof_functionals(ip)
+        @test fs isa NTuple{getnbasefunctions(ip), DofFunctional}
+        if all(f -> f isa PointValue, fs)
+            # Point-value dofs must satisfy the Kronecker delta property at their
+            # reference coordinates
+            coords = Ferrite.reference_coordinates(ip)
```
lijas, line 78: I think it would be cool the to also test the moment-type Hcurl and Hdiv elements for the Kronecker delta property. For integral moments you would need to use numerical quadrature on the edge/face the functional is associated with. But in the way this is coded now, this would require extending the definition of the Ferrite.reference_coordinate function in some to return the edge/face of the functional... I dont have a good solution atm.

docs/src/reference/interpolations.md
```diff
@@ -23,3 +23,22 @@ BubbleEnrichedLagrange
 CrouzeixRaviart
 RannacherTurek
 ```
+
+## [Dof functionals](@id dof-functionals)
+
+Each local dof of an interpolation has a *dof functional* describing its physical
+meaning. Most interpolations have only [`PointValue`](@ref) dofs, but e.g. H(div)/H(curl)
```
lijas: Is it correct to say "its physical meaning" here? There is no physics associated with the Dof at this point. The physics comes later when setting up the problem.

lijas (commented):
I like this a lot 😄  

I commented a bit on the definition on the VectorizedInterpolations, I am not 100% sure if it is implemented correctly as it is now.

src/interpolations.jl
KristofferC: Yeah, this makes sense to me. Will implement it and see how it feels. 

It is not too surprising there might be some overlap in previous functionalities, we could support both or depreciate the earlier one.

docs/src/reference/interpolations.md
KristofferC: I guess, just using the *functional* terminology might be enough.

test/interpolation_test_utils.jl
KnutAM: Isn’t that what we do here: https://github.com/Ferrite-FEM/Ferrite.jl/blob/30a02c1576d120d72e56c37861e4986e3066d07e/test/test_interpolations_hcurl_hdiv.jl#L110

src/interpolations.jl
KristofferC: fixed

src/Dofs/ConstraintHandler.jl
KristofferC: But the type applies to all instances.

src/interpolations.jl
KristofferC: Implemented

KristofferC:
Did some updates here beased on the review comments

src/Dofs/ConstraintHandler.jl
```diff
@@ -43,11 +52,12 @@ struct Dirichlet # <: Constraint
     facets::OrderedSet{T} where {T <: Union{Int, FacetIndex, FaceIndex, EdgeIndex, VertexIndex}}
     field_name::Symbol
     components::Vector{Int} # components of the field
+    functional::DofFunctional # which dofs to constrain (cf. dof_functionals)
```
lijas: Thinking out loud a bit:

Lets say we want a Hermitian triangle `(u1, dud1ξ, du1dη, u2, dud2ξ, du2dη, u3, du3dξ, du3dη)`, and we make this Vectorized, then we have the dofs 

```
(u1_x, u1_y, du1dξ_x, du1dξ_y, du1dη_x, du1dη_y, 
 u2_x, u2_y, du2dξ_x, du2dξ_y, du2dη_x, du2dη_y, 
 u3_x, u3_y, du3dξ_x, du3dξ_y, du3dη_x, du3dη_y)
```

Then 

```
Dirichlet(:u, set, f, [1,], functional = PointValue()) #Picks all u_x
Dirichlet(:u, set, f, functional = PointValue(2)) #components defaults to [1,2] and  picks all u_[x,y], so there is a conflict with `functional = PointValue(2)`?
Dirichlet(:u, set, f, [1,], functional = PointValue(2)) #Conflict here aswell?

Dirichlet(:u, set, f, [1,2], functional = PointDerivative(2)) #Picks all dudη_[x,y] (no conflict)
Dirichlet(:u, set, f, [2], functional = PointDerivative(2)) #Picks all dudη_y (no conflict)
```

So it seems to be that when we have Vectorized PointValues, there can be conflicts between `components `and `functional` in many cases, There seems to exist a tests for this, so all good. Ignore this comment :)

src/interpolations.jl
```diff
@@ -81,6 +81,180 @@ or `H1Conformity()`, for the interpolation.
...
+
+Dof functional for a point evaluation: ℓ(f) = f(xᵢ), where xᵢ is given by
+`reference_coordinates`. `direction` is the component selected from a vector-valued
+function and is omitted for scalar-valued interpolations. `PointValue()` and
+`PointValue(direction)` omit the point and can be used as selectors matching every point
+with the requested direction.
+"""
+struct PointValue{X, D} <: DofFunctional
+    x::X
+    direction::D
```
lijas: With my suggested VectorizedInterpolation, the direction can be removed.

src/interpolations.jl
```diff
@@ -81,6 +81,180 @@ or `H1Conformity()`, for the interpolation.
...
+for ∂²u/∂x∂y. The point `x` is a reference coordinate, but the derivative is with respect
+to physical coordinates so that the dof is cell-invariant on shared entities. `direction`
+has the same meaning as for [`PointValue`](@ref).
+`PointDerivative()` is a selector matching every point-derivative functional, while
+`PointDerivative(α)` matches the derivative `α` at every point and in every direction.
+"""
+struct PointDerivative{X, A, D} <: DofFunctional
+    x::X
+    α::A
+    direction::D
```
lijas: Direction can also be removed here.

src/interpolations.jl
```diff
@@ -81,6 +81,180 @@ or `H1Conformity()`, for the interpolation.
...
+
+"""
+    NormalMoment([nr::Int])
+
+Moment nr `nr` of the normal trace, ℓ(v) = ∫ (v ⋅ n) q dS, over the entity owning the dof
+(H(div) interpolations).
+`NormalMoment()` is a selector matching every normal moment.
+"""
+struct NormalMoment{N} <: IntegralMoment
+    nr::N
```
lijas: I don't understand what the what nr::N is? Is it an int pointing to the edge index?

src/interpolations.jl
```diff
@@ -81,6 +81,180 @@ or `H1Conformity()`, for the interpolation.
...
+    PointValue(x [, direction])
+
+Dof functional for a point evaluation: ℓ(f) = f(xᵢ), where xᵢ is given by
+`reference_coordinates`. `direction` is the component selected from a vector-valued
+function and is omitted for scalar-valued interpolations. `PointValue()` and
+`PointValue(direction)` omit the point and can be used as selectors matching every point
+with the requested direction.
+"""
+struct PointValue{X, D} <: DofFunctional
+    x::X
```
lijas: I think it could be more clean to instead store the index to the vertex in the reference shape, instead of storing the acutal coordinate. I think this would also match how the IntegralMoment reference their "entity" (with an index to the edges in the refshape). Becomes more clean without a bunch of type parameters :)

src/interpolations.jl
```diff
@@ -81,6 +81,180 @@ or `H1Conformity()`, for the interpolation.
...
+multi-index `α`, e.g. `PointDerivative((1,))` for du/dx in 1D or `PointDerivative((1, 1))`
+for ∂²u/∂x∂y. The point `x` is a reference coordinate, but the derivative is with respect
+to physical coordinates so that the dof is cell-invariant on shared entities. `direction`
+has the same meaning as for [`PointValue`](@ref).
+`PointDerivative()` is a selector matching every point-derivative functional, while
+`PointDerivative(α)` matches the derivative `α` at every point and in every direction.
+"""
+struct PointDerivative{X, A, D} <: DofFunctional
+    x::X
+    α::A
```
lijas: #1270 has second order derivatives, so we need to distinguish this some way for when applying BCs. Either we introduce PointSecondDerivative or add a type parameter PointDerivative{difforder}. I can't imagine anyone defining an element that goes beyond difforder=2 though :P

lijas (commented):
I am pretty sure we should introduce something like `VectorizedFunctional` which can wrap the PointValue-functional. It becomes symmetric with how we handle VectorizedInterpolations, and I think this would solve/improve a few things:

There would no longer be any conflicts or ambigouties when defining BCs. As it is now:

```
# For a vector field :u
Dirichlet(:u, set, x, [1,], PointValue(2)) #Conflict, components specify x component and PointValue specify the y-component
```

But if introduce VectorizedFunctional then we can say that it is always the components-arg that should specify the component, and the functional-kwarg should specify the dof-type (i.e. PointValue() without any direction).  I guess it could also be possible to allow for 
```
Dirichlet(:u, set, x, VectorizedFunctonal{PointValue}(dir=2))
```
if we want (but i prefer not).

`VectroizedFunctional `would also extend cleanly to the PointDerivative-functional.

src/interpolations.jl
KristofferC: It could be removed

src/interpolations.jl
KristofferC: I think https://github.com/Ferrite-FEM/Ferrite.jl/pull/1270 will need a dedicated `NormalDerivative`

lijas: Yes, that aswell!

KristofferC: Added an order field in https://github.com/Ferrite-FEM/Ferrite.jl/pull/1493/changes/4aa5e1ef7e238c377f37e3d704cd6fef38e04f7a

src/interpolations.jl
KristofferC: This could be completely removed and instead the index of the functional can be used directly to do the look up in the reference shape.

src/interpolations.jl
```diff
@@ -81,6 +81,154 @@ or `H1Conformity()`, for the interpolation.
...
+multi-index `α`, e.g. `PointDerivative((1,))` for du/dx in 1D or `PointDerivative((1, 1))`
+for ∂²u/∂x∂y. The point xᵢ is given by `reference_coordinates`, but the derivative is with
+respect to physical coordinates so that the dof is cell-invariant on shared entities.
+The first type parameter is the derivative order |α|, so the types `PointDerivative` and
+`PointDerivative{1}` can be used as selectors for every derivative and every first
+derivative, respectively (cf. [`Ferrite.matches_functional`](@ref)).
+"""
+struct PointDerivative{order, N} <: DofFunctional
+    α::NTuple{N, Int}
+    PointDerivative(α::NTuple{N, Int}) where {N} = new{sum(α), N}(α)
```
lijas: I think this is struct is pretty clean now. But perhaps change N to dim or rdim (reference dim), since that it was it represents?

src/Dofs/DofHandler.jl
```diff
@@ -408,6 +408,144 @@ function add!(sdh::SubDofHandler, name::Symbol, ip::Interpolation)
...
+    return _same_point_locations(first(locations), first(_locations))
+end
+
+# Entity-local coordinates of the point-supported dofs (`nothing` for other dofs) of a
+# non-vectorized interpolation.
+function _entity_point_locations(ip::Interpolation, entity_dim::Int)
+    functionals = entity_dim == 0 ? vertexdof_functionals(ip) :
+        entity_dim == 1 ? edgedof_functionals(ip) : facedof_functionals(ip)
+    all(all(f -> !(f isa Union{PointValue, PointDerivative}), fs) for fs in functionals) &&
+        return map(fs -> ntuple(_ -> nothing, length(fs)), functionals)
```
lijas: Perhaps just me, but I feel like AI sometimes generate code that is hard to read, like this for example 😬

src/Dofs/DofHandler.jl
```diff
@@ -408,6 +408,144 @@ function add!(sdh::SubDofHandler, name::Symbol, ip::Interpolation)
...
+
+    dof_indices = entity_dim == 0 ? vertexdof_indices(ip) :
+        entity_dim == 1 ? edgedof_interior_indices(ip) : facedof_interior_indices(ip)
+    dof_coordinates = reference_coordinates(ip)
+    vertex_coordinates = reference_coordinates(Lagrange{getrefshape(ip), 1}())
```
lijas, line 465: Hard coding Lagrange elements feels wrong here...but maybe it works

lijas (commented):
I think this PR looks pretty good now :) The dof matching checks in the DofHandler is nice, and I think we need something like this if we are going to support Hermitian type elements, so I think we can merge this, but perhaps someone else can take a look. 

In the future we could add a entity field to the dof functional structs if we want to tie the funcional to an entity. e.g. `struct TangentialMoment <: IntegralMoment; edgeid::Int; end`, but it seems like we don't really need this atm.

src/Dofs/DofHandler.jl
KristofferC: Added a comment

KristofferC:
Regarding

> In the future we could add a entity field to the dof functional structs if we want to tie the funcional to an entity. e.g. struct TangentialMoment <: IntegralMoment; edgeid::Int; end, but it seems like we don't really need this atm.

We already have this information right, for example:

```
edgedof_interior_indices(ip) == ((1,), (2,), (3,))
dof_functionals(ip) == (
      TangentialMoment(),
      TangentialMoment(),
      TangentialMoment(),
  )
```

Storing that information again doesn't seem useful?

lijas:
Good point, I did not think of that! And the coordinates for PointValues and PointDerivatives are stored in `reference_coordinates`, but this requires that PointValues and PointDerivatives are always the first functionals in an element so they can be indexed correctly (which is fine).

PR #1495, opened 2026-08-22
Conflict-aware grid coloring for DG assembly and constraints

fredrikekre:
I haven't reviewed this yet so this is just the robots output.

---

Extend create_coloring beyond node-sharing conflicts so that colored
multithreaded assembly is safe for more problem types:

- Interface (DG) assembly via the interface_coupling keyword argument.
  Three conflict graphs, each exact for its field configuration and
  selected automatically from the interpolations by the new
  create_coloring(dh, [ch]; ...) method: A_f ∪ A_f² for purely
  discontinuous discretizations, node graph ∪ A_f² when continuous
  fields exist but only discontinuous ones cross interfaces, and the
  closure product (I ∪ A_f)(I ∪ A_n)(I ∪ A_f) when a continuous field
  crosses (this contains node-distance-3 pairs; node-distance-2 is
  insufficient). The grid-only method always uses the conservative
  product graph since it has no field information. Facet neighbors are
  classified by shared-vertex counting, or taken from an optionally
  passed topology via the same helper as the sparsity pattern.
- Constraint condensation during assembly (apply_assemble!/apply_local!)
  via create_coloring(dh, ch): cells whose condensation writes to the
  same master dofs conflict, even when mesh-distant (e.g. opposite sides
  with PeriodicDirichlet). Only affine constraints add conflicts; plain
  Dirichlet performs no global writes during condensation.
- User-provided conflicts via the extra_conflicts keyword argument
  (Dict or cliques form; symmetrized and restricted to the cellset).

The DofHandler method mirrors the add_sparsity_entries! keyword
arguments so the arguments used for matrix allocation can be forwarded
as-is (coupling and keep_constrained are validated no-ops; nonempty
algebraic_couplings is refused with a pointer to extra_conflicts).

The coloring algorithms are untouched: all conflict edges are unioned
into the incidence matrix before the workstream BFS partitioning, whose
odd/even zone merge remains correct for any edge present during the
BFS. The default code path is unchanged (byte-identical gather loop,
no measurable timing difference) and colorings remain deterministic
and independent of the thread count.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    both keyword arguments behave exactly as before. ([#1468])
 
 ### Added
+ - `create_coloring` can now account for conflicts beyond node-sharing, making colored
+   multithreaded assembly safe for more problem types: interface (DG) assembly via the
+   `interface_coupling` keyword argument, constraint condensation during assembly (e.g.
+   `PeriodicDirichlet`/`AffineConstraint` with `apply_assemble!`, where mesh-distant
+   cells conflict through the master dofs) via a new `create_coloring(dh, ch)` method,
+   and arbitrary user-known conflicts via the `extra_conflicts` keyword argument. The
+   `DofHandler` method mirrors the keyword arguments of `add_sparsity_entries!` so the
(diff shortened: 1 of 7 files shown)


fredrikekre:
Maybe this is the wrong approach, we want to define a coloring for the interfaces instead since that is how we do the looping.

PR #1505, opened 2026-09-06
🤖 Fix assembly, constraints, and projection; improve interface and L2 performance

KristofferC:
New model is out = new bugfixes

Fixes incorrect assembly and constraint results, broken constructors and complex projection, and two performance bottlenecks.

Each fix is a separate commit:

- cc97d3af — Reject incompatible row/column dof lists in symmetric assembly before mutation; accept equal lists.
- 6b0f93e0 — Exclude prescribed masters from the free-dof constraint matrix; retain their contribution in the inhomogeneity.
- c675aa02 — Reject unsupported lower-triangle CSC storage in symmetric assembly and constraint application; preserve CSR support.
- aa94d697 — Accumulate repeated interface dofs correctly in CSC/CSR assembly, including shared diagonal entries.
- ae95a833 — Fix `InterfaceValues(fv)` and verify that its two sides have independent mutable state.
- 52319b9c — Preserve complex coefficients in L2 projection, including tensor data and hanging-node constraints.
- 95acb1b3 — Validate collection view offsets and handle empty collections safely.
- cf557492 — Cache interface topology lookups, avoiding quadratic traversal with abstract cell storage; add benchmarks.
- d978caee — Replace L2 RHS row-slice copies with scalar accumulation; add benchmarks.
- 7ab697be — Add the required changelog entries.

Validation: targeted assembly, constraint, collection, interface, projection, sparsity, and grid/VTK tests passed on Julia 1.12.7. Benchmark fixtures, Runic, and pre-commit checks also passed. The full test suite was not run.

### DIFF
--- a/ext/FerriteSparseMatrixCSR.jl
+++ b/ext/FerriteSparseMatrixCSR.jl
@@ -18,6 +18,10 @@ end
         coldofs::AbstractVector, sortedcoldofs::AbstractVector, colpermutation::AbstractVector,
         sym::Bool, atomic::Val = Val(false), rowoffset::Int = 0, coloffset::Int = 0
     )
+    if Ferrite._has_repeated_dofs(sortedcoldofs)
+        return Ferrite._assemble_repeated!(K, Ke, sortedrowdofs, rowpermutation, sortedcoldofs, colpermutation, sym, atomic, rowoffset, coloffset)
+    end
+
     current_row = 1
     ld = length(coldofs)
     ncols = size(K, 2)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixes
+
+ - Symmetric CSC assembly now rejects incompatible row/column dof lists and
+   lower-triangle storage before modifying the system. Constraint application also
+   rejects unsupported lower-triangle CSC storage.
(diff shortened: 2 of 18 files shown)


ext/FerriteSparseMatrixCSR.jl
```diff
@@ -18,6 +18,10 @@ end
         coldofs::AbstractVector, sortedcoldofs::AbstractVector, colpermutation::AbstractVector,
         sym::Bool, atomic::Val = Val(false), rowoffset::Int = 0, coloffset::Int = 0
     )
+    if Ferrite._has_repeated_dofs(sortedcoldofs)
+        return Ferrite._assemble_repeated!(K, Ke, sortedrowdofs, rowpermutation, sortedcoldofs, colpermutation, sym, atomic, rowoffset, coloffset)
```
fredrikekre, line 22: https://github.com/Ferrite-FEM/Ferrite.jl/pull/1492 was designed on the premise that we *don't* want to support repeated indices. Not sure what is best.

KristofferC: Ok, the robot says:

-----

When assembling interface terms for **continuous fields**. `InterfaceCache` concatenates both cells’ dofs, so two neighboring Q1 cells can produce:

```julia
[1, 2, 3, 4, 2, 5, 6, 3]
```

The shared dofs `2` and `3` appear twice. Contributions must accumulate correctly—for example, when assembling a penalty on gradient jumps across the interface.

Ordinary cell assembly and purely discontinuous DG interface assembly generally have unique dofs. The existing DG tutorial therefore doesn’t encounter this bug.

---

but #1492 obsoletes that we can drop this commit.

KristofferC:
https://github.com/Ferrite-FEM/Ferrite.jl/pull/1492 can decide what to do with repeated dofs, at least for now they work properly.

I don't think anything here is very controversial so will just merge.

PR #1509, opened 2026-09-08
🤖 Tet AMR support

termi-official:


koehlerson:
nice that it works with so less code

I'll review after I covered https://github.com/Ferrite-FEM/Ferrite.jl/pull/1486 because I think it makes sense to first address https://github.com/Ferrite-FEM/Ferrite.jl/pull/1486 and make this PR on top

termi-official:
Sure. I just had Fable tokens left for the week and tried to convert them into something useful. :)

PR #1511, opened 2026-09-11
Fix link to deal.II step-7 documentation

### LINKED ISSUE #1477: deal link is not working in Helmholtz equation
I found that the link for [deal.II step-7](https://www.dealii.org/8.4.1/doxygen/deal.II/step_7.html) on Helmholtz equation gives an error as

Image

Am I right ?  Is that my PC or for all is the same ?

Maybe it is not a big deal.

misosora:
Fixes #1477


KristofferC:
Thank you!

